PR #11, opened 2025-09-12
Fixes to enable producing a proper crates.io version

martin-kolinek:
Align the state of the repository with experiments to publish crates to crates.io. Mainly a version bump to get to a nice minor version and small changes to .gitignore to avoid issues with the internal crates.io publishing process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,18 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
 For more information see the [Code of Conduct FAQ](https://opensource.microsoft.com/codeofconduct/faq/)
 or contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additional questions or comments.
 
+## Releases
+
+Releases to crates.io are handled by an internal process. To release a new version of any of the crates,
+bump the version in accordance to semver. You should use cargo set-version for this to ensure all appropriate
+files are updated. For example:
+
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -11,14 +11,14 @@ members = [
...
-license-file = "LICENSE.txt"
+license-file = "LICENSE"
 homepage = "https://github.com/microsoft/oxidizer"
 repository = "https://github.com/microsoft/oxidizer"
 
 # all dependencies are listed with default-features = false, individual crates add the features they need
 [workspace.dependencies]
 # local dependencies
-data_privacy_macros = { path = "crates/data_privacy_macros", default-features = false  }
+data_privacy_macros = { path = "crates/data_privacy_macros", default-features = false, version = "0.3.0" }
```
geeknoid, line 21: Would it work to have a wildcard here for the version, so we don't need to update this file whenever a crate's version changes?

geeknoid (approved):

Cargo.toml
martin-kolinek: That would mean that we publish the crate without a specified version for its dependencies. `cargo set-version` handles this for you, so hopefully this won't be an issue. I added a section to readme on how to publish new versions.

ralfbiedert (approved):

PR #18, opened 2025-09-29
feat: introduce the recoverable crate

martintmk:
Add a new `recoverable` crate that provides standardized types for classifying 
error conditions as recoverable or non-recoverable, enabling consistent retry 
behavior across different error types and resilience middleware.

Core features:
- `RecoveryInfo` type for classifying errors with recovery metadata
- `Recoverable` trait for types that can determine their recoverability  
- `RecoveryKind` enum distinguishing between retry, outage, never, and unknown
- Support for explicit retry delays via `delay()` method
- Service outage detection with optional recovery hints


crates/recoverable/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "recoverable"
+description = "recovery metadata and classification for resilience patterns"
```
ralfbiedert: Can we capitalize this?

crates/recoverable/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+version = "0.1.0"
+readme = "README.md"
+keywords = ["oxidizer", "resilience", "metadata", "classification", "seatbelt"]
+categories = ["data-structures"]
```
ralfbiedert, line 10: Is this really `data-structures`?

crates/recoverable/README.md
```diff
@@ -0,0 +1,64 @@
...
+## Core Types
+
+- [`Recovery`]: Classifies errors as recoverable (transient) or non-recoverable (permanent).
+- [`Recover`]: A trait for types that can determine their recoverability.
+- [`RecoveryKind`]: An enum representing the kind of recovery that can be attempted.
+
+## Examples
+
+```rust
+use seatbelt_recovery::{Recover, Recovery, RecoveryKind};
```
ralfbiedert: What about these paths?

README.md
```diff
@@ -11,17 +11,19 @@
 
 This repository contains a set of crates that help you build robust highly scalable services in Rust.
 
-- [Crates](#crates)
-- [Repo Guidelines](#repo-guidelines)
-- [Releasing Crate Versions](#releasing-crate-versions)
-- [Trademarks](#trademarks)
+- [The Oxidizer Project](#the-oxidizer-project)
```
geeknoid: I don't think we want this change.

crates/recoverable/Cargo.toml
geeknoid:
```suggestion
description = "Recovery metadata and classification for resilience patterns"
```

crates/recoverable/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "recoverable"
+description = "recovery metadata and classification for resilience patterns"
+version = "0.1.0"
+readme = "README.md"
+keywords = ["oxidizer", "resilience", "metadata", "classification", "seatbelt"]
```
geeknoid:
```suggestion
keywords = ["oxidizer", "resilience", "metadata", "classification", "error"]
```

crates/recoverable/README.md
```diff
@@ -0,0 +1,64 @@
+<div align="center">
+ <img src="./logo.png" alt="Data Privacy Logo" width="128">
```
geeknoid:
```suggestion
 <img src="./logo.png" alt="Recoverable Logo" width="128">
```

crates/recoverable/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+# Data Privacy
```
geeknoid:
```suggestion
# Recoverable
```

crates/recoverable/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+[![crate.io](https://img.shields.io/crates/v/data_privacy.svg)](https://crates.io/crates/recoverable)
```
geeknoid:
```suggestion
[![crate.io](https://img.shields.io/crates/v/recoverable.svg)](https://crates.io/crates/recoverable)
```

crates/recoverable/README.md
```diff
@@ -0,0 +1,64 @@
...
+[![docs.rs](https://docs.rs/data_privacy/badge.svg)](https://docs.rs/recoverable)
```
geeknoid:
```suggestion
[![docs.rs](https://docs.rs/recoverable/badge.svg)](https://docs.rs/recoverable)
```

crates/recoverable/README.md
```diff
@@ -0,0 +1,64 @@
...
+[![CI](https://github.com/microsoft/oxidizer/workflows/main/badge.svg)](https://github.com/microsoft/oxidizer/actions)
+[![Coverage](https://codecov.io/gh/microsoft/oxidizer/graph/badge.svg?token=FCUG0EL5TI)](https://codecov.io/gh/microsoft/oxidizer)
+[![License](https://img.shields.io/badge/license-MIT-blue.svg)](../../LICENSE)
+
+</div>
+
+- [Data Privacy](#data-privacy)
```
geeknoid: Delete this line.

crates/recoverable/README.md
geeknoid: And note that you'll need to create a logo for the crate :-) You can give the Ferris crab to an AI as a baseline and try different ideas for what represents "recoverable".

crates/recoverable/README.md
geeknoid:
```suggestion
use recoverable::{Recover, Recovery, RecoveryKind};
```

README.md
martintmk: reverted

crates/recoverable/README.md
martintmk: used the template now

crates/recoverable/README.md
martintmk: uploaded the logo, wdyt? ;)

shield (protection) with a tag

crates/recoverable/README.md
```diff
@@ -0,0 +1,74 @@
+<div style="text-align: center">
```
geeknoid: I've got a PR outstanding that switches this in the template to

```
<div align="center">
```

In the template. Turns out that GitHub strips away the CSS attribute, so stuff wasn't getting centered.

crates/recoverable/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+[![crate.io](https://img.shields.io/crates/v/recoverable.svg)](https://crates.io/crates/recoverable)
+[![docs.rs](https://docs.rs/recoverable/badge.svg)](https://docs.rs/recoverable)
+[![CI](https://github.com/microsoft/oxidizer/workflows/main/badge.svg)](https://github.com/microsoft/oxidizer/actions)
+[![Coverage](https://codecov.io/gh/microsoft/oxidizer/graph/badge.svg?token=FCUG0EL5TI)](https://codecov.io/gh/microsoft/oxidizer)
+[![License](https://img.shields.io/badge/license-MIT-blue.svg)](../LICENSE)
+
+</div>
+
+- [Recoverable](#recoverable)
```
geeknoid: You don't need the first link.
And you need to add links for all the other headers in the file.

crates/recoverable/src/lib.rs
```diff
@@ -0,0 +1,546 @@
+// Copyright (c) Microsoft Corporation.
+
```
geeknoid, line 3: Need to add a license line here.

crates/recoverable/src/lib.rs
```diff
@@ -0,0 +1,546 @@
...
+/// use recoverable::{Recovery, RecoveryKind};
+///
+/// let recovery = Recovery::retry();
+/// assert_eq!(recovery.kind(), RecoveryKind::Retry);
+/// ```
+#[derive(Debug, PartialEq, Clone)]
+#[non_exhaustive]
+pub struct Recovery {
+    kind: RecoveryKind,
+    delay: Option<Duration>,
```
geeknoid, line 107: Hmmm. Wouldn't this be better as a parameter to the RecoveryKind::Retry enum? Either that, or add RecoveryKind::RetryAfter

geeknoid: Or
```
RecoveryKind::Retry(Option<Duration>)
```

crates/recoverable/src/lib.rs
```diff
@@ -0,0 +1,546 @@
...
+/// # Examples
+///
+/// ```rust
+/// use recoverable::{Recovery, RecoveryKind};
+///
+/// let recovery = Recovery::unknown();
+/// assert_eq!(recovery.kind(), RecoveryKind::Unknown);
+/// ```
+#[derive(Debug, PartialEq, Clone, Eq, Copy, Hash)]
+#[non_exhaustive]
```
geeknoid, line 143: Can we make this exhaustive instead? When it's non-exhaustive, you're forcing users to figure out what to do when there is an unknown recovery type, which can be hard to handle. Since you already have Unknown in there, is that sufficient?

martintmk: we can do that before 1.0.0, but I not sure about whether these are all the states we will have, `non_exhaustive` gives us some leverage for the future

crates/recoverable/src/lib.rs
martintmk: I had this like that, the thing is that it was difficult to enhance the metadata post-creation. My idea is that metadata are dumb, the component that evaluates them interprets the values.

Typical scenario:

- User provides retry recovery metadata but does ignore the retry-after header
- Our resilience middleware detects that retry was requested but no delay was provided
- We look at retry-after header and enrich recovery metadata with just calling `existing_metadata.delay(...)`
- Rest of the code just looks-up the retry delay from recovery metadata or generates the delay from backoff

crates/recoverable/README.md
martintmk: fixed

crates/recoverable/src/lib.rs
```diff
@@ -0,0 +1,538 @@
...
+        match self {
+            Ok(res) => res.recovery(),
+            Err(err) => err.recovery(),
+        }
+    }
+}
+
+impl Display for Recovery {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        Display::fmt(&self.kind, f)
```
geeknoid, line 467: Shouldn't the delay be displayed too?

crates/recoverable/src/lib.rs
```diff
@@ -0,0 +1,538 @@
...
+//! let success_result: Result<(), DatabaseError> = Ok(());
+//! // If we had a wrapper type for success, it would also return Recovery::never()
+//! ```
+
+use std::fmt::{Display, Formatter};
+use std::time::Duration;
+
+/// Represents the recoverability metadata associated with an operation or condition.
+///
+/// This type describes how the operation can be recovered from, if at all. It provides
```
geeknoid:
```suggestion
/// This type describes how an operation can be recovered from, if at all. It provides
```

crates/recoverable/README.md
geeknoid: Looks good.

geeknoid:
I think the names of the types aren't quite right English-wise. I would recommend the following:

RecoveryMetadata - type for classifying errors with recovery metadata
Recoverable - trait for types that can determine their recoverability
RecoveryKind enum distinguishing between retry, outage, never, and unknown

As a first step. And then, I'm still not sure why the retry delay can't be incorporated directly into the RecoveryKind enum which would eliminate a separate type. You could add a delay function for this enum which would return an updated RecoveryKind with the given delay, so it would match the semantics that the current delay function has. So what scenario does this complicate?

crates/recoverable/src/lib.rs
martintmk: done

martintmk:
What about:

Recover -> Recoverable (declares capability)
Recovery -> RecoveryInfo (cleaner, as a bag of properties, feels nicer than `RecoveryMetadata`)

Cargo.toml
```diff
@@ -20,20 +18,22 @@ repository = "https://github.com/microsoft/oxidizer"
...
+proc-macro2 = { version = "1.0.101", default-features = true }
+proc-macro-crate = { version = "3.3.0", default-features = false }
+quote = { version = "1.0.38", default-features = false }
 serde = { version = "1.0.217", default-features = false }
-serde_json = { version = "1.0.135", default-features = false  }
+serde_json = { version = "1.0.135", default-features = false }
 syn = { version = "2.0.106", default-features = false }
-xxhash-rust = { version = "0.8.15", default-features = false  }
+xxhash-rust = { version = "0.8.15", default-features = false }
+static_assertions = { version = "1.1.0", default-features = false }
```
geeknoid, line 80: Please add in alphabetical order.

ralfbiedert:
> "Recover" is a verb, it's not appropriate as a trait name.

That's not true, there are several lenses to look at trait naming, compare [API Guidelines](https://github.com/rust-lang/api-guidelines/discussions/28):

Linguistically:

> Imperatives like `io::Write`, `fmt::Debug`, `clone::Clone`
> Agent nouns like `iter::Iterator`, `hash::Hasher`
> Nouns like `fmt::Binary`, `ops::Fn`
> Adjectives like `marker::Sized`, `panic::UnwindSafe`
> Preposition like `convert::Into`, `borrow::ToOwned`

Functionally:

> - If the trait has a single self-explanatory method (or a set of nearly identical methods), name it after the method: Clone, Hash, Default, Into, Write, ToOwned, AsRef, Extend.
> - If the trait has no methods, name it after what ability (Send, Sync, Copy) or property (UnwindSafe) its implementors have.
> - If the trait has a broader set of methods or a single method that is not self-explanatory, the name should describe either what its implementors are (Iterator, Hasher, Fn, Error, Future, Termination) or what ability/property they have (AsciiExt, fmt::Debug, fmt::Binary).

However, as a universal convention trait names are generally short, unless compound `IntoIter`, `UnwindSafe`. Many of our trait names have unfortunately deviated from that and more leaned into C# interfaces naming style , and as we move here I'd like to revert that. 

With that said I agree `Recover`, while being imperative, does not match the "single method" convention, so given the method is called `recovery`, I'd like the trait to be named `Recovery` as well. since that matches the "Noun + single method" lens (and removes the `-able` that doesn't feel idiomatic. 

About `RecoveryInfo` I don't have strong feelings.

crates/recoverable/examples/recoverable_error.rs
```diff
@@ -0,0 +1,75 @@
...
+enum NetworkError {
+    /// DNS resolution failed - might be transient
+    DnsResolutionFailed,
+    /// Invalid URL format - permanent error
+    InvalidUrl,
+    /// Service is unavailable, for example circuit breaker is open
+    ServiceUnavailable { retry_after: Option<Duration> },
+}
+
+impl Recoverable for NetworkError {
```
ralfbiedert: Should be `Recovery`.

crates/recoverable/Cargo.toml
martintmk: I followed the same thing as `data_privacy`, do you have any other alternatives?

crates/recoverable/README.md
martintmk: seems to be old iteration of readme

crates/recoverable/examples/recoverable_error.rs
martintmk: @geeknoid Would you agree with the rename?

crates/recoverable/Cargo.toml
geeknoid: The set of supported categories is actually surprisingly silly. There isn't anything more appropriate that I can find. See https://crates.io/categories for the supported set.

crates/recoverable/examples/recoverable_error.rs
geeknoid: No, I don't think it's right.

This doesn't do anything about recovery, it merely says that an error can be recovered from. When I see "recovery", I think "action", this IS the recovery to the problem.

crates/recoverable/README.md
geeknoid: You can use cargo-rdme to regen the README. But you should be getting a CI validation error if the README doesn't match what's in lib.rs. Did you not get one of these CI errors?

crates/recoverable/Cargo.toml
martintmk: correct, I don't think there is a better category for this.

crates/recoverable/examples/recoverable_error.rs
martintmk: This trait only tells that particular error or result is recoverable but does not nor it is able to do the recovery action. 

It's up to the caller to do the action itself based on the conditions.

For example, caller might decide ahead of time that particular action can be retried and he needs to put aside all information/parameters required to do that recovery action. This trait only gives the information that such recovery is possible.

Our names should reflect that. With this context in mind, I find the `Recoverable` more appropriate than `Recovery`.
(capability vs action)

crates/recoverable/src/lib.rs
```diff
@@ -0,0 +1,585 @@
...
+//! # Why
+//!
+//! This crate provides types for classifying conditions based on their **recoverability state**,
+//! enabling consistent recovery behavior across different error types and resilience middleware.
+//!
+//! # Recovery Information
+//!
+//! The recovery information describes whether recovering from an operation might help, not whether
+//! the operation succeeded or failed. Both successful operations and permanent failures
+//! should use [`RecoveryInfo::never`][RecoveryInfo::never] since recovery won't change the outcome.
```
geeknoid:
```suggestion
//! should use [`RecoveryInfo::never`] since recovery is not necessary or desirable.
```

crates/recoverable/src/lib.rs
```diff
@@ -0,0 +1,585 @@
...
+use std::time::Duration;
+
+// Naming Convention for Get/Set:
+//
+// This type uses an unconventional naming pattern where setters use plain names (e.g., `delay()`)
+// and getters use the `get_` prefix (e.g., `get_delay()`). This deviates from standard Rust
+// conventions because setters are used much more frequently than getters in typical usage patterns.
+// The `get_` prefix on getters helps distinguish them from their corresponding setters.
+
+/// Represents the recovery information associated with an operation or condition.
```
geeknoid:
```suggestion
/// The recovery information associated with an operation or condition.
```

crates/recoverable/src/lib.rs
```diff
@@ -0,0 +1,585 @@
...
+///
+/// # Examples
+///
+/// ```rust
+/// use recoverable::{RecoveryInfo, RecoveryKind};
+///
+/// let recovery = RecoveryInfo::retry();
+/// assert_eq!(recovery.kind(), RecoveryKind::Retry);
+/// ```
+#[derive(Debug, PartialEq, Clone, Eq)]
```
geeknoid: Would you ever want a hash table of HashMap<RecoveryInfo, xxx>? (Hash trait)

crates/recoverable/src/lib.rs
```diff
@@ -0,0 +1,585 @@
...
+pub struct RecoveryInfo {
+    kind: RecoveryKind,
+    delay: Option<Duration>,
+}
+
+/// Represents the kind of recovery that can be attempted.
+///
+/// To retrieve the recovery kind from a `RecoveryInfo` instance, use the [`RecoveryInfo::kind`] method.
+///
+/// # Examples
```
geeknoid, line 134: I think we should have a recommendation of what code is expected to do in general when it receives these different kinds. In particular, what should code do/assume when it gets Unknown? And what should it do if it gets a kind that not currently in the enum (given the non_exhaustive attribute).

crates/recoverable/src/lib.rs
```diff
@@ -0,0 +1,585 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#![cfg_attr(coverage_nightly, feature(coverage_attribute))]
+#![cfg_attr(docsrs, feature(doc_auto_cfg))]
+#![doc(html_logo_url = "https://media.githubusercontent.com/media/microsoft/oxidizer/refs/heads/main/crates/recoverable/logo.png")]
+#![doc(html_favicon_url = "https://media.githubusercontent.com/media/microsoft/oxidizer/refs/heads/main/crates/recoverable/favicon.ico")]
+
+//! Recovery information and classification for resilience patterns.
```
geeknoid, line 9: There are some instances where a service fails and provides a hint to an alternate endpoint to connect to instead. Just like delay(), would it make sense to have a fallback(url) hint?

crates/recoverable/src/lib.rs
```diff
@@ -0,0 +1,585 @@
...
+/// # Examples
+///
+/// ```rust
+/// use recoverable::{RecoveryInfo, RecoveryKind};
+///
+/// let recovery = RecoveryInfo::retry();
+/// assert_eq!(recovery.kind(), RecoveryKind::Retry);
+/// ```
+#[derive(Debug, PartialEq, Clone, Eq)]
+#[non_exhaustive]
```
geeknoid: Why do we need non_exhaustive here? It has no effect given the private fields.

crates/recoverable/src/lib.rs
martintmk: sure, could be a valid use-case

crates/recoverable/src/lib.rs
martintmk: yes, that's the reason I did not make the `RecoveryInfo` implement `Copy` trait. I expect we will add something like this at some point. Not right now though.

crates/recoverable/src/lib.rs
martintmk: true

crates/recoverable/src/lib.rs
martintmk: enhanced the docs

crates/recoverable/examples/recoverable_error.rs
martintmk: I think we can go ahead with `Recovery`, it's shorter that `Recoverable` and it's aligned with the `recovery` method that returns `RecoveryInfo`.

crates/recoverable/src/lib.rs
```diff
@@ -0,0 +1,595 @@
...
+    /// }
+    ///
+    /// let error = MyError;
+    /// assert_eq!(error.recovery().kind(), RecoveryKind::Retry);
+    /// ```
+    fn recovery(&self) -> RecoveryInfo;
+}
+
+impl Recovery for RecoveryInfo {
+    fn recovery(&self) -> RecoveryInfo {
```
geeknoid: Why is this impl useful?

crates/recoverable/src/lib.rs
geeknoid: I still think we should remove this. It's making it too hard for consumers to handle failures in the right way.

Or at least, document the expected pattern: If you receive a recovery kind value that you don't recognize, then always treat it like the Unknown case:

match kind {
    _ | Unknown: 
}

crates/recoverable/src/lib.rs
martintmk: I forgot why I added this to `RecoveryInfo`. Dropping.

crates/recoverable/src/lib.rs
martintmk: added a docs section

ralfbiedert (approved):

PR #30, opened 2025-10-02
Move over fundle, fix warnings in tests, fix macro warning.

ralfbiedert:


geeknoid:
To get coverage for the macros, check out what I did in data_privacy_macros. Basically, all error handling is tested via the insta crate within the data_privacy_macros folder, while successful use of the macros is tested in data_privacy. Seems to work pretty well that way.

PR #54, opened 2025-11-18
ci: Improve CI experience by always running all static analysis steps even if one fails

geeknoid:
- Update various tool dependencies too.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,8 +12,9 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
-  MSRV: "1.88"        # used for testing, ensures the MSRV promise is kept
-  RUST_LATEST: "1.90" # used for static analysis, mutation testing, and coverage
+  RUST_MSRV: "1.88"                   # used for testing (ensures the MSRV promise is kept)
+  RUST_LATEST: "1.91"                 # used for static analysis, mutation testing
+  RUST_NIGHTLY: "nightly-2025-11-15"  # used for coverage, extended analysis, and external type exposure
 
 jobs:
   testing:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -32,15 +32,15 @@ jobs:
       - name: Start sscache
         uses: mozilla-actions/sccache-action@v0.0.9
         with:
-          version: "v0.10.0"
+          version: "v0.12.0"
       - name: Install Rust
-        uses: actions-rust-lang/setup-rust-toolchain@v1
+        uses: actions-rust-lang/setup-rust-toolchain@v1.15.2
```
sandersaares, line 38: Does it not automatically pick the latest "1.x" if we just put "1" or what is the deal here? I rather hope we do not need to be manually incrementing these numbers on a regular basis - that is horribly busywork. What is the motivation for expanding these version numbers here - am I missing the point? Can't we just specify a major version and forget about it?

.github/workflows/main.yml
```diff
@@ -245,11 +269,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v5
       - name: Install Rust Nightly
-        uses: actions-rust-lang/setup-rust-toolchain@v1
+        uses: actions-rust-lang/setup-rust-toolchain@v1.15.2
         with:
-          toolchain: nightly-2025-08-06-x86_64-unknown-linux-gnu
+          toolchain: nightly-2025-11-06-x86_64-unknown-linux-gnu
```
sandersaares: This looks off already from the old version - we should be installing the latest version of nightly if we really want it to be a "nightly" build. This is just going to be "some random old version from nightly channel" which nobody will expect. Can we remove the extra suffix and just use "nightly"?

.github/workflows/main.yml
geeknoid: Having floating versions is completely undesirable in a CI pipeline. Not selecting a fully-specific version was an oversight when I first set this up.

So this is not busywork, this is making sure we have robust repeatable builds that don't break because of some random version changing on the Internet.

.github/workflows/main.yml
geeknoid: It doesn't matter which nightly build we're using. We just need something stable over time and recent enough relative to the current tip.

PR #57, opened 2025-11-20
feat: Improve builder UX.

ralfbiedert:


crates/fundle/src/lib.rs
```diff
@@ -124,3 +125,19 @@ pub struct Set;
...
+
+#[doc(hidden)]
+#[derive(Debug)]
+pub struct Read;
+
+#[doc(hidden)]
+#[derive(Debug)]
+pub struct Write;
+
+pub trait Reader {
```
martintmk, line 138: is there any reason to not hide these traits?

ralfbiedert: Negligence :)

PR #59, opened 2025-11-21
ci: Use explicit tool versions and action versions.

geeknoid:
- As is best practice, specify explicit versions
  for tools and GitHub Actions to ensure consistent
  behavior across environments and over time.

- Update to latest versions of tools and actions.

- Increase mutation testing timeout to avoid
  intermittent test failures due to timeouts.

- Fix the use of cargo-check-external-types to
  include all features.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,8 +12,23 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
-  MSRV: "1.88"        # used for testing, ensures the MSRV promise is kept
-  RUST_LATEST: "1.90" # used for static analysis, mutation testing, and coverage
+
+  # Rust toolchain versions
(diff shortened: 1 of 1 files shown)


sandersaares (commented):
I disagree with the premise that we want to pin to specific patch versions - this just creates unnecessary maintenance burden. Pinning to specific major version is fine - incrementing across API contract version requiring manual action feels acceptable. Perhaps even minor version for certain tools (e.g. cargo-mutants because minor versions can introduce new mutations). However, pinning to functionally interchangeable versions seems excessive - the cost-benefit balance is not there in my eyes. We simply have better things to do with our time than increment version numbers. Yes, it sacrifices some amount of reproducibility but that is a sacrifice we can and should make.

ralfbiedert (approved):

martin-kolinek (approved):

PR #67, opened 2025-11-23
chore: Enable unwrap_used and panic clippy lints

geeknoid:


crates/ohno_macros/src/derive_error/constructors.rs
```diff
@@ -151,6 +151,7 @@ fn generate_simple_named_constructors(
 }
 
 #[expect(clippy::too_many_arguments, reason = "C'est la vie")]
+#[expect(clippy::unwrap_used, reason = "Field names are guaranteed to be present here")]
```
Vaiz, line 154: the main reason why panics are fine here - these are compile time panics, it won't fire in production

in hindsight, we probably should replace them with proper compile error instead

Vaiz (approved):

crates/data_privacy/examples/employees/main.rs
```diff
@@ -92,20 +91,32 @@ fn app_loop() {
...
-    // For each value, you can control which trait is used to format the value into a string:
-    //   `name` - formats the value with the `Display` trait.
-    //   `name:?` - formats the value with the `Debug` trait.
-    //   `mame:@` - formats the value with the `Display` trait and redacts it.
-    log!(event:? = "Employee created",
-         name:@ = employee.name,
-         address:@ = employee.address,
-         employee_id:@ = employee.id,
-         age = employee.age);
+                Err(e) => {
```
Vaiz: I see this pattern a lot when instead of giving a proper name to error object, it's renamed in tracing macro
instead of renaming error object later, we can 
```suggestion
        Err(error) => {
            log!(event = "Employee database read error", error);
```

crates/data_privacy/examples/employees/main.rs
```diff
@@ -92,20 +91,32 @@ fn app_loop() {
 
     employees.push(employee.clone());
 
-    let file = OpenOptions::new().write(true).create(true).truncate(true).open(json_path).unwrap();
-    serde_json::to_writer_pretty(file, &employees).unwrap();
-    println!("Employee added.\n");
+    match OpenOptions::new().write(true).create(true).truncate(true).open(json_path) {
```
Vaiz: error handling can be simplified by returning an error from `app_loop` function and logging it there. This would allow to remove match statements here making code more readable

Vaiz (approved):

crates/ohno_macros/src/derive_error/constructors.rs
geeknoid: Yes, it would be cleaner to not panic in general, even in macros.

crates/data_privacy/examples/employees/main.rs
geeknoid: Actually, with my little experimental macro, line 113 needs to be error = error.

crates/data_privacy/examples/employees/main.rs
geeknoid: Somewhere along the way, I lost the "loop" part of app_loop. I've reintroduced that now, which makes the code make more sense.

clippy.toml
```diff
@@ -1,11 +1,8 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-# This file contains fine-tuning settings that cannot be specified in cargo.toml (which only supports "error level").
-
-# Absolute paths of length 3 can be useful to emphasize where a particular symbol is coming from,
-# e.g. by using "std::sync::mutex" versus "tokio::sync::mutex".  Anything beyond 3 segments seems
-# excessively verbose, so we limit it to 3 - import or alias to shorten longer symbol paths.
 absolute-paths-max-segments = 3
+allow-panic-in-tests = true
```
ralfbiedert, line 5: Ah cool, wasn't aware this exists!

PR #69, opened 2025-11-24
chore: remove recursive dev dependency in ohno crate

Vaiz:
ohno crate included itself as a dependency to enable `test-util` feature for testing, that negatively affects some tools like rust-analyzer.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -218,10 +218,8 @@ jobs:
       # execute
       - name: Generate Coverage (all-features)
         run: cargo +${{ env.RUST_NIGHTLY }} llvm-cov --all-features --workspace --lcov --output-path lcov-all.info
-      # Note: test-util feature enables additional testing functionality.
-      # Without this feature, coverage would be lower.
       - name: Generate Coverage (no-default-features)
-        run: cargo +${{ env.RUST_NIGHTLY }} llvm-cov --no-default-features --features="test-util" --workspace --lcov --output-path lcov-no-def.info
+        run: cargo +${{ env.RUST_NIGHTLY }} llvm-cov --no-default-features --workspace --lcov --output-path lcov-no-def.info
       - name: Upload Coverage to Codecov
         uses: codecov/codecov-action@v5.5.1
         with:


crates/ohno/tests/caused_by_temp_str.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+#![cfg(feature = "test-util")]
```
psandana, line 4: Is this intending the entire test to be behind the `cfg` attribute?

Vaiz: yes, tests are not supposed to be run without this feature

psandana: I don't think this does a global feature gate. AFAIK, you need to add a scope.

.github/workflows/main.yml
```diff
@@ -213,8 +213,10 @@ jobs:
       # execute
       - name: Generate Coverage (all-features)
         run: cargo +${{ env.RUST_NIGHTLY }} llvm-cov --all-features --workspace --lcov --output-path lcov-all.info
+      # Note: test-util feature enables additional testing functionality.
+      # Without this feature, coverage would be lower.
       - name: Generate Coverage (no-default-features)
-        run: cargo +${{ env.RUST_NIGHTLY }} llvm-cov --no-default-features --workspace --lcov --output-path lcov-no-def.info
+        run: cargo +${{ env.RUST_NIGHTLY }} llvm-cov --no-default-features --features="test-util" --workspace --lcov --output-path lcov-no-def.info
```
geeknoid: This is not needed. We run coverage twice. Once above with all features on, and once here with no features on. The two coverage reports are merged to produce the final coverage data.

crates/ohno/tests/caused_by_temp_str.rs
Vaiz: it's works as expected, if feature is not enabled, the whole module is compiled out

here is a running test with `--all-features` flag
https://github.com/microsoft/oxidizer/actions/runs/19629283941/job/56204944041?pr=69#step:7:607

.github/workflows/main.yml
Vaiz: rolled that change back

PR #82, opened 2025-11-27
ci: reduce number of mutations jobs to 1

Vaiz:
github actions cancels our mutant job due to high resource usage. It works only when number of jobs is set to 1

https://github.com/microsoft/oxidizer/actions/runs/19732684189/job/56537246259

### DIFF
--- a/.github/license-check/config.json
+++ b/.github/license-check/config.json
@@ -16,8 +16,7 @@
     {
         "include": [
             "**/*.toml",
-            "**/*.ps1",
-            "justfile"
+            "**/*.ps1"
         ],
         "exclude": [
             "target"
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -63,7 +63,7 @@ jobs:
           components: clippy, rustfmt
       - name: Installing Wild Linker
         uses: davidlattimore/wild-action@0.7.0
-      - name: Install Tools
+      - name: Install Cargo Tools
         uses: taiki-e/install-action@v2.62.53
         with:
           tool: cargo-mutants@${{ env.CARGO_MUTANTS_VERSION }}


scripts/mutants.ps1
```diff
@@ -0,0 +1,70 @@
...
+    $packages_arg = $group | ForEach-Object { "--package=$_"}
+
+    $args = @(
+        "--no-shuffle",
+        # "--baseline=skip",
+        "--colors=never",
+        "--jobs=$jobs",
+        "--build-timeout=$build_timeout_sec",
+        "--timeout=$timeout_sec",
+        "--minimum-test-timeout=$minimum_test_timeout_sec",
```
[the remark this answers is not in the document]
Vaiz: it was removed intentionally to reduce the number of tests running for each mutation

.github/workflows/nightly.yml
```diff
@@ -63,19 +64,22 @@ jobs:
...
       # execute
       - name: Mutate All
         id: mutate
-        continue-on-error: true
+        shell: pwsh
         env:
           DISABLE_T_REX: 1
-        timeout-minutes: 180
-        run: cargo mutants --no-shuffle --baseline=skip --test-workspace=true --colors=never --jobs=4 --build-timeout=600 --timeout=300 -vV
+        run: just mutants
```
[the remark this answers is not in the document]
Vaiz: restored `continue-on-error: true`

.github/workflows/nightly.yml
```diff
@@ -63,19 +64,22 @@ jobs:
...
         uses: davidlattimore/wild-action@0.7.0
-      - name: Install Cargo Tools
+      - name: Install Cargo Mutants
         uses: taiki-e/install-action@v2.62.53
         with:
           tool: cargo-mutants@${{ env.CARGO_MUTANTS_VERSION }}
+      - name: Install Just
+        uses: taiki-e/install-action@v2.62.53
+        with:
+          tool: just
```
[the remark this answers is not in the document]
Vaiz: applied

scripts/mutants.ps1
```diff
@@ -0,0 +1,71 @@
...
+function mutate_group {
+    param([array]$group)
+    $crates = $group -join ","
+    Write-Host "Mutating group: $crates"
+
+    $packages_arg = $group | ForEach-Object { "--package=$_"}
+
+    $args = @(
+        "--no-shuffle",
+        # "--baseline=skip",
```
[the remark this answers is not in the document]
Vaiz: applied

.github/workflows/nightly.yml
```diff
@@ -63,19 +64,22 @@ jobs:
...
 
       # execute
       - name: Mutate All
         id: mutate
-        continue-on-error: true
+        shell: pwsh
```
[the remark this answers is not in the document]
Vaiz: we don't need it yet

scripts/mutants.ps1
```diff
@@ -0,0 +1,71 @@
...
+    )
+
+    $mutate_command = "cargo mutants " + ($args -join " ") + " " + ($packages_arg -join " ")
+    Write-Host "Running command: $mutate_command"
+
+    Invoke-Expression $mutate_command
+}
+
+foreach ($group in $test_groups) {
+    mutate_group $group
```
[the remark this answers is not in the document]
Vaiz: it's intentional

scripts/mutants.ps1
```diff
@@ -0,0 +1,71 @@
...
+        "--build-timeout=$build_timeout_sec",
+        "--timeout=$timeout_sec",
+        "--minimum-test-timeout=$minimum_test_timeout_sec",
+        "-vV"
+    )
+
+    $mutate_command = "cargo mutants " + ($args -join " ") + " " + ($packages_arg -join " ")
+    Write-Host "Running command: $mutate_command"
+
+    Invoke-Expression $mutate_command
```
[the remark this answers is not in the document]
Vaiz: it should be covered by those lines
```
$ErrorActionPreference = "Stop"
$PSNativeCommandUseErrorActionPreference = $True
```

.github/workflows/nightly.yml
```diff
@@ -63,7 +63,7 @@ jobs:
           components: clippy, rustfmt
       - name: Installing Wild Linker
         uses: davidlattimore/wild-action@0.7.0
-      - name: Install Cargo Tools
+      - name: Install Tools
```
geeknoid: Why change this? It's the wording used in all the other CI jobs right now.

.github/license-check/config.json
```diff
@@ -16,7 +16,8 @@
     {
         "include": [
             "**/*.toml",
-            "**/*.ps1"
+            "**/*.ps1",
+            "justfile"
```
geeknoid: Not needed.

scripts/mutants.ps1
```diff
@@ -0,0 +1,74 @@
...
+$PSNativeCommandUseErrorActionPreference = $True
+
+$jobs = 1
+$build_timeout_sec = 600
+$timeout_sec = 300
+$minimum_test_timeout_sec = 60
+
+
+# Define test groups for mutants testing
+$test_groups = @(
```
geeknoid, line 14: So the point here is to only run on a subset of crates at a time? Is this really needed? Given that for a given run of cargo-mutants, it's doing things sequentially, it's not clear how giving it smaller crate sets actually reduces memory consumption.

What's the behavior when new crates are added to the repo and they aren't listed here? I gather they would just be ignored?

Vaiz: The main point is time reduction by limiting the number of crates tested for each mutation. Originally we were running all the tests in the repo per mutation until at least one of them fails which is wasteful - there is no point in running tests on `fundle` for a mutation in `ohno` crate.

All crates, that are not on the list, are tested, but testing suite includes only crate's own tests.

.github/workflows/nightly.yml
Vaiz: it's just an artifact after multiple edits
fixed it

scripts/mutants.ps1
Vaiz:
```
foreach ($group in $test_groups) {
    mutate_group $group
}
foreach ($crate in $not_included) {
    mutate_group @($crate)
}
```

Vaiz: the difference is 16min vs 23min with current set of crates, it will become even bigger as we add more crates

[this branch](https://github.com/microsoft/oxidizer/actions/runs/19738401401)
[original version with 1 job](https://github.com/microsoft/oxidizer/actions/runs/19740916098)

geeknoid: OK, that's a nice improvement. So what about the case when a new crate is added to the repo? Can we generate an error in that case so we don't forget to register the crate here?

Or potentially, we could improve the scripts/add-crate.ps1 script to automatically put an entry in this file when a crate is added.

Vaiz: creates that are not explicitly added already tested in groups of one. It means there might be more false failures when mutatated code is tested in another crate

Let's merge that PR as is. I can do a follow up with some improvements

PR #85, opened 2025-11-28
feat: Add unknown MemoryAffinity

Darksecond:
To facilitate API changes elsewhere

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,6 +23,7 @@ env:
   CARGO_AUDIT_VERSION: "0.22.0"
   CARGO_CHECK_EXTERNAL_TYPES_VERSION: "0.3.0"
   CARGO_ENSURE_NO_DEFAULT_FEATURES_VERSION: "0.2.0"
+  CARGO_ENSURE_NO_CYCLIC_DEPS_VERSION: "0.1.0"
   CARGO_HACK_VERSION: "0.6.39"
   CARGO_LLVM_COV_VERSION: "0.6.21"
   CARGO_MUTANTS_VERSION: "25.3.1"
@@ -113,7 +114,7 @@ jobs:
       - name: Install Cargo Tools
         uses: taiki-e/install-action@v2.62.53
         with:
-          tool: cargo-audit@${{ env.CARGO_AUDIT_VERSION }}, cargo-rdme@${{ env.CARGO_RDME_VERSION }}, cargo-workspaces@${{ env.CARGO_WORKSPACES_VERSION }}, cargo-ensure-no-default-features@${{ env.CARGO_ENSURE_NO_DEFAULT_FEATURES_VERSION }}
+          tool: cargo-audit@${{ env.CARGO_AUDIT_VERSION }}, cargo-rdme@${{ env.CARGO_RDME_VERSION }}, cargo-workspaces@${{ env.CARGO_WORKSPACES_VERSION }}, cargo-ensure-no-default-features@${{ env.CARGO_ENSURE_NO_DEFAULT_FEATURES_VERSION }}, cargo-ensure-no-cyclic-deps@${{ env.CARGO_ENSURE_NO_CYCLIC_DEPS_VERSION }}
 
       # execute
       - name: Clippy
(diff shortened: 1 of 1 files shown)


crates/thread_aware/src/cell.rs
```diff
@@ -456,21 +459,25 @@ impl<T, S: Strategy> ThreadAware for Trc<T, S> {
...
-                .replace(destination, Arc::<T>::clone(&value));
+            if let MemoryAffinity::Pinned(destination) = destination {
+                let old_data = self
+                    .storage
+                    .write()
+                    .expect("Failed to acquire write lock")
+                    .replace(destination, Arc::<T>::clone(&value));
 
-            assert!(old_data.is_none(), "Data already exists for the destination affinity");
+                assert!(old_data.is_none(), "Data already exists for the destination affinity");
```
martin-kolinek: This issue had been there before, but this is a race condition - we'll panic in case two clones of this Trc try to be relocated to the same destination at the same time.

crates/thread_aware/src/cell.rs
```diff
@@ -422,7 +422,10 @@ impl<T, S: Strategy> Trc<T, S> {
 
 impl<T, S: Strategy> ThreadAware for Trc<T, S> {
     fn relocated(self, source: MemoryAffinity, destination: MemoryAffinity) -> Self {
-        let value = self.storage.read().expect("Failed to acquire read lock").get_clone(destination);
+        let value = match destination {
```
martin-kolinek: This implies that relocating to an unknown destination creates a new value. I don't think that's what we want? That will result in potentially unbounded number of connection pools for HTTP client for example.

Do we really need to have unknown affinity for the destination as well? Couldn't we get away with just having it for source?

More generally, I suspect the unknown affinity will hurt in other places as well, which is why I'd prefer not to introduce it and rather embrace the "you only get one Builtins outside of the runtime" approach.

crates/thread_aware/src/affinity.rs
```diff
@@ -0,0 +1,111 @@
...
+/// An `MemoryAffinity` can be thought of as a placement in a system.
+///
+/// It is used to represent a specific context or environment where data can be processed.
+/// For example a NUMA node, a thread, a specific CPU core, or a specific memory region.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub enum MemoryAffinity {
+    /// An unknown affinity represents no specific binding, like an unpinned thread.
+    Unknown,
+    /// A pinned affinity represents a specific binding to a processor and memory region.
+    Pinned(PinnedAffinity),
```
martin-kolinek, line 13: KnownAffinity? Pinned feels a bit wrong.

crates/thread_aware/src/cell.rs
Darksecond: I agree with the changing destination to never be a unknown affinity; Until we make the final decision on the runtime design I'm going to keep the concept of MemoryAffinity::Unknown for the source however.

ralfbiedert: Yeah, one "option" we discussed was having an `Option<Affinity>` for the source instead. For source we need it, but I agree destination probably not. 

> This implies that relocating to an unknown destination creates a new value. I don't think that's what we want? That will result in potentially unbounded number of connection pools for HTTP client for example.

What I don't get is, why would that imply it, we could just implement "outside" as always a single, shared affinity?

crates/thread_aware/src/affinity.rs
```diff
@@ -0,0 +1,111 @@
...
+// Licensed under the MIT License.
+
+/// An `MemoryAffinity` can be thought of as a placement in a system.
+///
+/// It is used to represent a specific context or environment where data can be processed.
+/// For example a NUMA node, a thread, a specific CPU core, or a specific memory region.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub enum MemoryAffinity {
+    /// An unknown affinity represents no specific binding, like an unpinned thread.
+    Unknown,
```
kate-shine, line 11: Unspecified might be clearer (At least to my understanding). Unknown feels like "It has some affinity set, but we don't know what it is"

crates/thread_aware/src/cell.rs
martin-kolinek: > What I don't get is, why would that imply it, we could just implement "outside" as always a single, shared affinity?

My comment was pointed at the current implementation at the time.

On a more general note, even the "single outside affinity" has issues - how do you relocate e.g. IoDriver to it? IoDriver is tied to a thread that drives the loop, so relocating it outside would be problematic.

martin-kolinek (approved):

PR #86, opened 2025-11-28
feat!: rename ohno::error_trace into enrich_err

Vaiz:


crates/ohno/src/trace_info.rs
```diff
@@ -32,41 +32,23 @@ impl fmt::Display for Location {
 pub struct TraceInfo {
```
martintmk, line 32: nit: more appropriate name alligned with `enrich_err`?

`EnrichInfo`, `EnrichMsg`, `EnrichData`?

Vaiz: it's hard to come up with a good name here...
good news is that it's not supposed to be named directly in user's code

martintmk (approved):

PR #87, opened 2025-11-28
ci: rewrite mutants script in rust

Vaiz:
this is a follow up for #82

### DIFF
--- a/scripts/check-external-types.rs
+++ b/scripts/check-external-types.rs
@@ -9,28 +9,28 @@ edition = "2024"
 [dependencies]
 automation = { path = "../crates/automation" }
 anyhow = "1.0"
+argh = "0.1"
 ---
 
-use std::env;
 use std::path::Path;
 
-use anyhow::{Context, Result};
+use anyhow::Result;
+use argh::FromArgs;
 
 /// Check external types in all workspace library crates.
-///
-/// Usage: check-external-types.rs <TOOLCHAIN>
-///
-/// Arguments:
-///   <TOOLCHAIN>    The Rust toolchain to use (e.g., 'nightly-2025-08-06')
 ///
 /// This script iterates through all workspace packages and runs
 /// `cargo check-external-types` on library crates to verify that
 /// public APIs don't expose types from private dependencies.
+#[derive(FromArgs)]
+struct Args {
+    /// the Rust toolchain to use (e.g., 'nightly-2025-08-06')
+    #[argh(positional)]
+    toolchain: String,
+}
+
 fn main() -> Result<()> {
-    let args: Vec<String> = env::args().collect();
-    let toolchain = args
-        .get(1)
-        .context("Missing required toolchain argument (e.g., 'nightly-2025-08-06')")?;
+    let args: Args = argh::from_env();
 
     let workspace_root = Path::new(env!("CARGO_MANIFEST_DIR")).parent().unwrap();
     let all_packages = automation::list_packages(workspace_root)?;
@@ -43,7 +43,7 @@ fn main() -> Result<()> {
     println!();
     println!("=== External Type Exposure Check ===");
     println!();
-    println!("Toolchain: {}", toolchain);
+    println!("Toolchain: {}", args.toolchain);
     println!("Checking {} crate(s)", filtered_packages.len());
     println!("Skipped internal crates: {}", automation::INTERNAL_CRATES.join(", "));
     println!();
@@ -57,7 +57,7 @@ fn main() -> Result<()> {
 
         if has_lib {
             println!("Checking external types in {}", pkg.name);
-            check_external_types(&pkg.manifest_path, toolchain)?;
+            check_external_types(&pkg.manifest_path, &args.toolchain)?;
             println!("✓ Passed: {}", pkg.name);
             checked += 1;
         } else {


.github/workflows/main.yml
```diff
@@ -210,7 +210,7 @@ jobs:
         env:
           DISABLE_T_REX: 1
         timeout-minutes: 45
-        run: cargo mutants --in-place --no-shuffle --baseline=skip --test-workspace=true --colors=never --build-timeout=600 --in-diff diff.txt --timeout=300
+        run: cargo +nightly -Zscript scripts/mutants.rs --in-place --in-diff diff.txt
```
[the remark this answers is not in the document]
Vaiz: this is intended

crates/ci_aids/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+[package]
+name = "ci_aids"
```
sandersaares: This reads like CI has a serious illness. Let's not follow the unfortunate naming choice from `testing_aids`, please. Also, "Shared code for writing Rust scripts" from the description is not really the same as "ci" - if we are using these scripts locally, they are not really "CI". Also also, CI is a specific methodology, just having build automation is not the same as "CI".

Suggested name: "automation"

scripts/mutants.rs
```diff
@@ -0,0 +1,136 @@
...
+    println!("  Min timeout: {MINIMUM_TEST_TIMEOUT_SEC}s");
+    println!("  In-place: {}", args.in_place);
+    if let Some(ref diff) = args.in_diff {
+        println!("  Diff file: {}", diff.display());
+    }
+    println!("Test groups ({} total):", test_groups.len());
+    for (i, group) in test_groups.iter().enumerate() {
+        println!("  {}: [{}]", i + 1, group.join(", "));
+    }
+    println!("\nSkipped: {}", ci_aids::INTERNAL_CRATES.join(", "));
```
sandersaares: Claude really loves this `\n` at the start of println but it makes the string rather difficult to read. I have not found a way to consistently forbid it, so I always have to manually remind it no tidy up and just "println!()" and empty line instead of this ugly escaping-prefix.

I do not know if you used Claude here or if you just have similar patterns as Claude but regardless, let's please keep our strings readable without escape characters when possible - a separate println to emit a separate blank line would be preferred by me.

crates/ci_aids/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+            output.status.code(),
+            stdout,
+            stderr
+        );
+    }
+
+    Ok(())
+}
+
+#[cfg(all(test, not(miri)))]
```
sandersaares: I think elsewhere in this repo we use `#[cfg_attr(miri, ignore)]` to just ignore tests under miri but still compile them. This reduces dead code warnings and other such side-effects that conditional compilation brings. Sounds like a potentially preferable pattern?

crates/ci_aids/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+rust-version.workspace = true
+authors.workspace = true
+license-file.workspace = true
+homepage.workspace = true
+repository.workspace = true
+publish = false
+
+[dependencies]
+anyhow = { workspace = true, features = ["std"] }
+duct.workspace = true
```
sandersaares: My soul is pained by the difference in notation between this and its neighbors.

scripts/check-external-types.rs
```diff
@@ -0,0 +1,73 @@
...
+---
+
+use std::env;
+use std::path::Path;
+
+use anyhow::{Context, Result};
+
+fn main() -> Result<()> {
+    let args: Vec<String> = env::args().collect();
+    let toolchain = args.get(1)
```
sandersaares: Where/how are we documenting the arguments and their expected contents/formats? Is there a `--help` (I guess not) or some written comment somewhere? We should not force the user or the next maintainer to reverse-engineer arguments.

scripts/check-external-types.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    let mut checked = 0;
+    let mut skipped = 0;
+
+    for pkg in &filtered_packages {
+        // Check if this is a library crate by looking at the targets
+        let has_lib = pkg.targets.iter().any(|t| t.kind.contains(&"lib".to_string()));
+
+        if has_lib {
+            println!("✓ Checking external types in {}", pkg.name);
```
sandersaares: A checkmark before the check is done seems the wrong way around... a checkmark is what you get when the check has passed!

sandersaares (changes requested):
I like it, just a few minor requests for improvement.

[Though I agree with Ralf that we should not be repeating this logic - but I imagine factoring this out to a reusable core with minor repo-specific layer on top can be a next step]

crates/ci_aids/Cargo.toml
Vaiz: renamed to automation
another thought I had, maybe testing_aids and this new crate should be merged

scripts/mutants.rs
Vaiz: It's my personal preference. I don't see any problems with readability, and it allows to make scripts shorter, which is the main benefit for me

Vaiz: but I can live with empty `println!()` statements as well

crates/ci_aids/src/lib.rs
Vaiz: ok

crates/ci_aids/Cargo.toml
Vaiz: I've heard different opinions on that topic. Would be nice to have a decision the whole team agrees on

scripts/check-external-types.rs
Vaiz: switched to argh

PR #89, opened 2025-11-28
feat: Improve data_privacy UX

ralfbiedert:
Most notably, data_privacy now works with compound types, and properly handles deeply nested fields when doing redacted printing.

```rust
#[classified(Taxonomy::PII)]
struct EMailAddress(String);

#[derive(RedactedDisplay, RedactedDebug)]
struct Contact {
    name: String,
    email: EMailAddress,
}

log!("{:@}", contact);
```

This might, for example, print 

```
Contact { name: "Alice", email: <redacted/pii> }
```

The crate also does not leak implementation details of `classified` types anymore. In addition, various UX issues were fixed, e.g., `Taxonomy::PII` can now be used directly instead of `Taxonomy::PII.data_class()`, and the general patterns have been harmonized with the rest of this repo and our Rust Guidelines.

In addition:

- There is now an IntoDataClass trait and the #[taxonomy] macro generates an implementation of it for every taxonomy.
- RedactionEngineBuilder::new is replaced with RedactionEngine::builder
- Redactors have been moved into the simple_redactor and xxh3_redactor modules.
- ClassifiedWrapper<T> was renamed to Sensitive<T>
- DataClass now implements AsRef<Self>

### DIFF
--- a/crates/thread_aware/src/closure.rs
+++ b/crates/thread_aware/src/closure.rs
@@ -1,8 +1,6 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-#![expect(missing_docs, reason = "TODO")]
-
 pub mod erased;
 
 pub use erased::ErasedClosureOnce;
@@ -13,21 +11,27 @@ use crate::ThreadAware;
 ///
 /// Use [`relocate_once`] function to construct these.
 pub trait RelocateFnOnce<T: ?Sized>: ThreadAware {
+    /// Calls the closure, consuming it in the process.
     fn call_once(self) -> T;
 }
 
 /// A trait for callable types that can be called multiple times.
 /// This trait is used to define closures that can be called multiple times, without consuming the closure.
 pub trait RelocateFn<T>: ThreadAware {
+    /// Calls the closure, returning the result.
     fn call(&self) -> T;
 }
 
 /// A trait for callable types that can be called mutably.
 /// This trait is used to define closures that can be called mutably, allowing the closure to modify its internal state.
 pub trait RelocateFnMut<T>: ThreadAware {
+    /// Calls the closure mutably, returning the result.
     fn call_mut(&mut self) -> T;
 }
 
+/// A common implementation of [`RelocateFn`]
+///
+/// Construct this using the [`relocate`] function.
 #[derive(Debug, Copy, Hash)]
 pub struct Closure<T, D> {
     data: D,
@@ -123,6 +127,9 @@ where
     }
 }
 
+/// A common implementation of [`RelocateFnMut`]]
+///
+/// Construct this using the [`relocate_mut`] function.
 #[derive(Debug, Copy, Hash)]
 pub struct ClosureMut<T, D> {
     data: D,
@@ -169,6 +176,11 @@ where
     }
 }
 
+/// Construct a [`RelocateFn`] - a closure-like object where the captured data implement [`ThreadAware`].
+///
+/// Create a closure-like object by explicitly providing closed-over
+/// value and a function pointer to operate on that value, essentially simulating a
+/// parameterless closure that ensures that captured data implements [`ThreadAware`].
 pub fn relocate<T, D>(data: D, f: fn(&D) -> T) -> Closure<T, D>
 where
     D: ThreadAware,
@@ -176,6 +188,11 @@ where
     Closure { data, f }
 }
 
+/// Construct a [`RelocateFnMut`] - a closure-like object where the captured data implement [`ThreadAware`].
+///
+/// Create a closure-like object by explicitly providing closed-over
+/// value and a function pointer to operate on that value, essentially simulating a
+/// parameterless closure that ensures that captured data implements [`ThreadAware`].
 pub fn relocate_mut<T, D>(data: D, f: fn(&mut D) -> T) -> ClosureMut<T, D>
 where
     D: ThreadAware,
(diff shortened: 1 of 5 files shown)


crates/data_privacy/src/redactors/mod.rs
```diff
@@ -1,9 +1,27 @@
...
+    ///
+    /// # Errors
+    ///
+    /// This function should return [`Err`] if, and only if, the provided [`Formatter`](std::fmt::Formatter) returns [`Err`]. String redaction is considered an infallible operation;
+    /// this function only returns a [`std::fmt::Result`] because writing to the underlying stream might fail and it must provide a way to propagate the fact that an error
+    /// has occurred back up the stack.
+    fn redact(&self, data_class: &DataClass, value: &str, output: &mut dyn Write) -> std::fmt::Result;
+}
 
 pub struct Redactors {
```
geeknoid: This type should not be public outside the crate.

crates/data_privacy/src/redacted.rs
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use crate::RedactionEngine;
+use core::fmt::{Formatter, Result};
+
+pub trait RedactedDebug {
```
geeknoid, line 14: Lost the documentation for these traits.

crates/data_privacy/README.md
```diff
@@ -126,8 +126,8 @@ How this all works:
 * The application uses the classified container types to wrap sensitive data throughout the application. This ensures the
   sensitive data is not accidentally exposed through telemetry or other means.
 
-* On startup, the application initializes a [`RedactionEngine`](https://docs.rs/data_privacy/latest/data_privacy/redaction_engine/struct.RedactionEngine.html) using the [`RedactionEngineBuilder`](https://docs.rs/data_privacy/latest/data_privacy/redaction_engine_builder/struct.RedactionEngineBuilder.html)
-  type. The engine is configured with redactors for each data class in the taxonomy. The redactors define how to handle sensitive data for that class.
+* On startup, the application initializes a `RedactionEngine` via [`RedactionEngine::builder()`]. The engine is configured
```
geeknoid: This README file needs to be generated with cargo-rdme, not by hand. Otherwise, you lose links like these.

geeknoid: This is not failing CI right now because for some reason, I have the check configured to only generate a warning instead of failing. I'll fix CI.

crates/data_privacy/src/data_class.rs
```diff
@@ -43,73 +49,12 @@ impl Display for DataClass {
...
-        let hash2 = hasher2.finish();
-
-        let mut hasher3 = DefaultHasher::new();
-        data_class3.hash(&mut hasher3);
-        let hash3 = hasher3.finish();
-
-        assert_eq!(hash1, hash2);
-        assert_ne!(hash1, hash3);
-    }
+pub trait IntoDataClass {
```
geeknoid, line 53: Needs documentation.

crates/data_privacy/src/std.rs
```diff
@@ -0,0 +1,265 @@
...
+
+        impl $crate::RedactedDebug for $ty {
+            fn fmt(&self, _: &$crate::RedactionEngine, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
+                <Self as ::std::fmt::Debug>::fmt(self, f)
+            }
+        }
+    };
+}
+
+/// Implements Classified, `RedactedDebug`, and `RedactedDisplay` for non-generic std types.
```
geeknoid: I think this is fundamentally not a good idea.

This now results in types claiming they participate in redaction when they actually do not. I can now hold some data in a string and since the String type support RedactedDisplay it will lead me to think this is being redacted even though it does nothing.

This is breaking the model in a core way and opens the door to misunderstanding and misuse.

geeknoid:
We probably should add a feature for the simple redactor, like there is one for the xxh3 redactor. Production code would typically never use the simple redactor, so no need to impose compilation costs on everyone.

geeknoid:
Looking through the fog of the massive refactoring of this source code, I see the following relatively small concrete functional changes:

* There is now an IntoDataClass trait and the #[taxonomy] macro generates an implementation of it for every taxonomy.

* RedactionEngineBuilder::new is replaced with RedactionEngine::builder

* Redactors have been moved into the simple_redactor and xxh3_redactor modules.

* ClassifiedWrapper<T> was renamed to Sensitive<T>

* DataClass now implements AsRef<Self>

* Several of the traits have been implemented on primitive types.

Did I miss anything?

As a best practice, I think we need to separate refactoring from functional changes. Given the relatively small and simple changes in the PR, the review was made considerably more lengthy because of the refactoring.

crates/data_privacy/src/std.rs
```diff
@@ -0,0 +1,265 @@
...
+macro_rules! impl_std_traits {
+    ($ty:ty, $data_class:expr) => {
+        impl_std_traits_debug_only!($ty, $data_class);
+
+        impl $crate::RedactedDisplay for $ty {
+            fn fmt(&self, _: &$crate::RedactionEngine, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
+                <Self as ::std::fmt::Display>::fmt(self, f)
```
geeknoid:
```suggestion
                ::std::fmt::Display::fmt(self, f)
```

This is more readable and succinct.

PR #105, opened 2025-12-06
ci: Add license check for dependencies

geeknoid:
- Make sure all our dependencies have acceptable licenses.

- Add missing 'license` field to our Cargo.toml. Turns out having `license-file` defined isn't enough.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -1,0 +1,19 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+[licenses]
+allow = [
+    # You can add new entries to the list only from "Permissive OSS licenses" on this Microsoft-internal page: https://docs.opensource.microsoft.com/legal/resources/oss-licenses-by-type/.
+    "MIT",
+    "Apache-2.0",
+    "ISC",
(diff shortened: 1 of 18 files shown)


ralfbiedert (approved):

deny.toml
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+[licenses]
+allow = [
+    "MIT",
```
ralfbiedert, line 7: When we have BSD or so we'll extend that list?

geeknoid: Yeah, there's a company site that lists the allowed licenses, so we can extend as needed. It's mostly about avoiding GPL in the end.

sandersaares: Let's add instructions here on how the content here is defined and how to identify valid options to insert into the list (e.g. what we have in the private repo). Ralf is not the only one who is going to come up with this question, after all, so let's not make the next reader reverse-engineer the truth of things.

geeknoid: Done.

PR #106, opened 2025-12-06
feat: introduce tick crate

martintmk:
This PR introduces a `tick` crate that contains APIs to retrieve and manipulate the flow of time.

### DIFF
--- a/crates/tick/src/lib.rs
+++ b/crates/tick/src/lib.rs
@@ -7,7 +7,7 @@
     test,
     allow(
         clippy::arithmetic_side_effects,
-        clippy::unchecked_time_subtraction,
+        clippy::unchecked_duration_subtraction,
         reason = "allow these lints in tests to improve the readability of the tests"
     )
 )]


martintmk:
@copilot review this PR

crates/tick/Cargo.toml
```diff
@@ -0,0 +1,96 @@
...
+name = "tick"
+description = "Provides primitives to interact with and manipulate machine time."
+version = "0.1.0"
+readme = "README.md"
+keywords = ["time", "clock", "tick", "timestamp", "runtime", "stopwatch", "oxidizer"]
+categories = ["data-structures"]
+edition.workspace = true
+rust-version.workspace = true
+authors.workspace = true
+license-file.workspace = true
```
geeknoid: Also add "license.workspace = true"

crates/tick/Cargo.toml
```diff
@@ -0,0 +1,96 @@
...
+homepage.workspace = true
+repository.workspace = true
+
+[lib]
```
geeknoid: [lib] is not needed, that's the default.

crates/tick/Cargo.toml
```diff
@@ -0,0 +1,96 @@
...
+# Enables additional test-specific APIs
+test-util = []
+# Serialization support for serde
+serde = ["dep:serde_core"]
+# Integration with tokio runtime
+tokio = ["dep:tokio"]
+# Exposes timestamp APIs
+timestamp = ["dep:jiff"]
+
+[dependencies]
```
geeknoid, line 40: No need for 'default-features = false' in any of these, that's done at the workspace level.

crates/tick/README.md
```diff
@@ -0,0 +1,154 @@
...
+[![CI](https://github.com/microsoft/oxidizer/workflows/main/badge.svg)](https://github.com/microsoft/oxidizer/actions)
+[![Coverage](https://codecov.io/gh/microsoft/oxidizer/graph/badge.svg?token=FCUG0EL5TI)](https://codecov.io/gh/microsoft/oxidizer)
+[![License](https://img.shields.io/badge/license-MIT-blue.svg)](../LICENSE)
+
+</div>
+
+* [Summary](#summary)
+
+## Summary
+
```
geeknoid, line 24: Generate a TOC here for all the headers below. I wish this were automated...

crates/tick/Cargo.toml
martintmk: done

crates/tick/src/clock.rs
```diff
@@ -0,0 +1,610 @@
...
+
+use super::TimerKey;
+#[cfg(any(feature = "test-util", test))]
+use super::clock_control::ClockControl;
+use crate::state::ClockState;
+
+/// Provides an abstraction for time-related operations.
+///
+/// Working with time is notoriously difficult to test and control. The clock enables time control in tests
+/// while providing near-zero overhead in production. When running with the `test-util` feature enabled, the clock
```
geeknoid: What makes it near-zero instead of truly zero cost?

crates/tick/src/clock.rs
```diff
@@ -0,0 +1,610 @@
...
+/// provides additional functionality to control the flow of time. This makes tests faster and more reliable.
+/// See the [Testing](#testing) section for more information.
+///
+/// The clock is used for:
+///
+/// - Retrieving the current absolute time in UTC.
+/// - Creation of [`Stopwatch`][super::Stopwatch] that
+///   simplifies time measurements and can be used as a relative unit of time.
+/// - Creating [`PeriodicTimer`][super::PeriodicTimer] and [`Delay`][super::Delay] instances.
```
geeknoid:
```suggestion
/// - Creating [`Stopwatch`][super::Stopwatch] instances that
///   simplify time measurements and can be used as relative units of time.
/// - Creating [`PeriodicTimer`][super::PeriodicTimer] and [`Delay`][super::Delay] instances.
```

crates/tick/src/clock.rs
```diff
@@ -0,0 +1,609 @@
...
+///
+/// # Testing
+///
+/// When working with time, it's challenging to isolate time-related operations in tests. A typical example is the sleep
+/// operation, which is hard to test and slows down tests. What you want to do is have complete control over the flow
+/// of time that allows you to jump forward in time. This is where the clock comes into play.
+///
+/// The ability to jump forward in time makes tests faster, more reliable and gives you complete control over the flow of time.
+/// By default, the clock does not allow you to control the flow of time. However, when the `test-util` feature is enabled,
+/// this crate provides a `ClockControl` type that can be used to control time.
```
geeknoid:
```suggestion
/// this crate provides a [`ClockControl`] type that can be used to control time.
```

crates/tick/src/clock.rs
```diff
@@ -0,0 +1,609 @@
...
+impl Clock {
+    /// Creates a new clock driven by the Tokio runtime.
+    ///
+    /// # Panics
+    ///
+    /// Panics if called outside of a Tokio runtime context.
+    #[cfg(any(feature = "tokio", test))]
+    #[cfg_attr(docsrs, doc(cfg(feature = "tokio")))]
+    #[must_use]
+    pub fn new_tokio() -> Self {
```
geeknoid, line 157: Wouldn't with_tokio be a better name more in line with things like with_capacity?

crates/tick/src/clock.rs
```diff
@@ -0,0 +1,609 @@
...
+
+    #[cfg(any(feature = "test-util", test))]
+    #[must_use]
+    pub(crate) fn with_control(clock_control: &ClockControl) -> Self {
+        Self::with_state(ClockState::ClockControl(clock_control.clone()))
+    }
+
+    /// Creates a new frozen clock.
+    ///
+    /// This is a convenience method for creating a clock by calling `ClockControl::new().to_clock()`.
```
geeknoid:
```suggestion
    /// This is a convenience function for creating a clock that's equivalent to calling `ClockControl::new().to_clock()`.
```

crates/tick/src/clock.rs
```diff
@@ -0,0 +1,609 @@
...
+    /// ```
+    #[cfg(any(feature = "test-util", test))]
+    #[must_use]
+    pub fn new_frozen() -> Self {
+        crate::ClockControl::new().to_clock()
+    }
+
+    /// Creates a new frozen clock at the specified timestamp.
+    ///
+    /// This is a convenience method for creating a clock by calling `ClockControl::new_at(time).to_clock()`.
```
geeknoid:
```suggestion
    /// This is a convenience function for creating a clock that's equivalent to calling `ClockControl::new_at(time).to_clock()`.
```

crates/tick/src/clock.rs
```diff
@@ -0,0 +1,609 @@
...
+    ///
+    /// This method provides the standard library's representation of absolute time. Use this when:
+    /// - You need basic absolute time functionality
+    /// - You need to interoperate with other crates using `SystemTime`
+    /// - You don't need formatting, parsing, or serialization capabilities
+    ///
+    /// For enhanced absolute time capabilities including formatting, parsing, and serialization support,
+    /// use [`timestamp()`][Self::timestamp] instead (requires the `timestamp` feature).
+    ///
+    /// **Note**: The system time is not monotonic and can be affected by system clock changes.
```
geeknoid: Put the various note blocks you have in your docs in a block quote, which is the recommended way to highlight notes in Rust docs:

For general notes that don't fit the semantic sections above, the community standard is to use a standard Markdown blockquote combined with a bold label. This renders cleanly in IDE hovers (VS Code), docs.rs, and local HTML.

Example:

Rust

/// Configures the network timeout.
///
/// > **Note:** This setting is ignored if the `async-std` feature is disabled.
pub fn set_timeout(millis: u64) { ... }

crates/tick/src/clock.rs
```diff
@@ -0,0 +1,609 @@
...
+            ClockState::System(timers) => timers.with_timers(|t| t.unregister(key)),
+        }
+    }
+
+    pub(crate) fn clock_state(&self) -> &ClockState {
+        self.0.local_state()
+    }
+}
+
+impl AsRef<Self> for Clock {
```
geeknoid, line 429: Why is this needed?

crates/tick/src/clock_control.rs
```diff
@@ -0,0 +1,794 @@
...
+// Licensed under the MIT License.
+
+use std::sync::{Arc, Mutex};
+use std::task::Waker;
+use std::time::{Duration, Instant, SystemTime};
+
+use super::{TimerKey, Timers};
+use crate::{Clock, ClockTimestamp};
+
+/// Controls the flow of time in tests.
```
geeknoid:
```suggestion
/// Controls the passage of time in tests.
```

Throughout the crate, I think 'passage' is a better choice in English than 'flow'

crates/tick/src/clock_timestamp.rs
```diff
@@ -0,0 +1,37 @@
...
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum ClockTimestamp {
+    /// Represents system time.
+    System(SystemTime),
+    /// Represents a specific timestamp.
+    #[cfg(any(feature = "timestamp", test))]
+    #[cfg_attr(docsrs, doc(cfg(feature = "timestamp")))]
+    Timestamp(crate::Timestamp),
+    /// Represents an offset duration from the UNIX epoch.
```
geeknoid: This seems strange to me. Aren't SystemTime and Timestamp both offsets from the UNIX epoch? I don't think we should conflate Duration which is intended to measure time delta values with these absolute time things, that's inviting programmer error.

crates/tick/src/clock_timestamp.rs
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use std::time::{Duration, SystemTime};
+
+/// Represents a type of timestamp used by [`Clock`][crate::Clock] and [`ClockControl`][crate::ClockControl].
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum ClockTimestamp {
```
geeknoid: Shouldn't this be a trait?

trait TimestampProvider {
    fn timestamp() -> Timestamp;
}

crates/tick/src/clock_control.rs
```diff
@@ -0,0 +1,794 @@
...
+    /// let clock = control.to_clock();
+    ///
+    /// assert!(SystemTime::now() >= clock.system_time());
+    /// ```
+    #[must_use]
+    pub fn now() -> Self {
+        Self::new_at(SystemTime::now())
+    }
+
+    /// Converts the `ClockControl` to a `Clock` instance.
```
geeknoid: If I call to_clock N times, are those clocks linked or not? In other words, are these equivalent:

let c1 = clock_control.to_clock();
let c2 = clock_control.to_clock();

vs.

let c1 = clock_control.to_clock();
let c2 = clock.clone();

crates/tick/src/clock_control.rs
```diff
@@ -0,0 +1,794 @@
...
+        self
+    }
+
+    /// Sets a limit on the total auto-advance duration.
+    ///
+    /// When auto-advance is enabled via [`Self::auto_advance`], this method limits the total
+    /// amount of time that can be auto-advanced. Once the limit is reached, further calls to
+    /// access the current time will no longer auto-advance the clock.
+    ///
+    /// **Note:** This method only has an effect if [`Self::auto_advance`] has been called
```
geeknoid: Could this constraint be removed in order to eliminate a potential source of friction/error when using these APIs?

If this can't be removed, could the logic panic when auto_advance_limit is called at the wrong time? Right now, the call is just silently ignored which is an anti-pattern.

crates/tick/src/clock_control.rs
```diff
@@ -0,0 +1,794 @@
...
+        });
+
+        self
+    }
+
+    /// Determines whether the clock control should automatically auto-advance all upcoming timers.
+    ///
+    /// Note that when [`Self::auto_advance_limit`] is used, the maximum total auto-advance duration is respected.
+    /// This means that when the total of all auto-advances exceeds the maximum, the auto-advance will stop and such timers won't be fired.
+    #[must_use]
```
geeknoid, line 234: Does this call also have the constraint of needing to be called after auto_advance, otherwise it is ignored?

crates/tick/src/clock_control.rs
```diff
@@ -0,0 +1,794 @@
...
+    /// # Ok::<(), Box<dyn std::error::Error>>(())
+    /// ```
+    pub fn advance(&self, duration: Duration) {
+        self.with_state(|v| v.advance(duration, TimeFlow::Forward));
+    }
+
+    /// Advances the clock to the specified system time.
+    ///
+    /// The clock can be advanced to the future or to the past. Advancing the clock forward
+    /// fires all timers that are scheduled to be fired before or at the target time.
```
geeknoid: Can you explain the behavior of time changes regarding periodic timers? Presumably, they should not be affected, right?

crates/tick/src/delay.rs
```diff
@@ -0,0 +1,237 @@
...
+// Licensed under the MIT License.
+
+use std::pin::Pin;
+use std::task::{Context, Poll, Waker};
+use std::time::Duration;
+
+use super::Clock;
+use super::timers::TimerKey;
+
+/// Asynchronously delays for the specified duration.
```
geeknoid, line 11: Add a note about the fact Delay is not affected by changes in the system clock.

crates/tick/src/clock.rs
```diff
@@ -0,0 +1,609 @@
...
+use std::sync::Arc;
+use std::task::Waker;
+use std::time::{Instant, SystemTime};
+
+use super::TimerKey;
+#[cfg(any(feature = "test-util", test))]
+use super::clock_control::ClockControl;
+use crate::state::ClockState;
+
+/// Provides an abstraction for time-related operations.
```
geeknoid, line 11: Could you describe the cost of cloning a clock? I think it's really cheap and doesn't allocate, so we should call that out.

crates/tick/src/fmt.rs
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Module that contains primitives for parsing, formatting, and serializing the [`Timestamp`][`crate::Timestamp`].
```
geeknoid:
```suggestion
//! Module that contains primitives for parsing, formatting, and serializing [`Timestamp`][`crate::Timestamp`].
```

crates/tick/src/fmt/iso_8601_timestamp.rs
```diff
@@ -0,0 +1,228 @@
...
+/// let iso = "2024-08-06T23:30:00+02:00".parse::<Iso8601Timestamp>()?;
+/// assert_eq!(iso.to_string(), "2024-08-06T21:30:00Z"); // Note that the UTC offset is applied
+///
+/// let timestamp: Timestamp = iso.into();
+/// assert_eq!(Iso8601Timestamp::from(timestamp), iso);
+///
+/// # Ok::<(), Box<dyn std::error::Error>>(())
+/// ```
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct Iso8601Timestamp(Timestamp);
```
geeknoid: Would it make sense to implement Deref on these formatting types so that they can be used like normal timestamps?

crates/tick/src/fmt/unix_seconds_timestamp.rs
```diff
@@ -0,0 +1,262 @@
...
+///
+/// let unix_seconds = "9999".parse::<UnixSecondsTimestamp>()?;
+/// assert_eq!(unix_seconds.to_string(), "9999");
+///
+/// let timestamp: Timestamp = unix_seconds.into();
+/// assert_eq!(timestamp.to_string(), "1970-01-01T02:46:39Z");
+///
+/// # Ok::<(), Box<dyn std::error::Error>>(())
+/// ```
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
```
geeknoid: How about implementing Ord and PartialOrd for the timestamp formatting types? Sorting those seems like a usable feature.

crates/tick/src/fmt/unix_seconds_timestamp.rs
```diff
@@ -0,0 +1,262 @@
...
+    }
+}
+
+impl Display for UnixSecondsTimestamp {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        write!(f, "{}", self.to_secs())
+    }
+}
+
+impl From<UnixSecondsTimestamp> for Timestamp {
```
geeknoid: You could consider providing interop with the other formatting timestamp types too.

crates/tick/src/future_ext.rs
```diff
@@ -0,0 +1,109 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use std::time::Duration;
+
+use super::{Clock, Delay, Timeout};
+
+/// Extensions for the [`Future`] trait.
+pub trait FutureExt: Future {
```
geeknoid, line 9: Would it make sense to have a less generic name here? Like FutureTimeout

crates/tick/src/future_ext.rs
```diff
@@ -0,0 +1,109 @@
...
+    /// // Apply a timeout to the future and await it
+    /// let timeout_error = future
+    ///     .timeout(Duration::from_millis(200), &clock)
+    ///     .await
+    ///     .unwrap_err();
+    ///
+    /// assert_eq!(timeout_error.to_string(), "future timed out");
+    /// # }
+    /// ```
+    fn timeout(self, timeout: Duration, clock: &Clock) -> Timeout<Self, Delay>
```
geeknoid: Shouldn't the clock parameter come first, like in other functions like Delay::new?

crates/tick/src/periodic_timer.rs
```diff
@@ -0,0 +1,274 @@
...
+/// A periodic timer can be created using the [`PeriodicTimer::new`] constructor,
+/// which requires a reference to a [`Clock`].
+///
+/// # Precision
+///
+/// The timer uses the current thread's scheduler to schedule its ticks. The precision
+/// of the timer is affected by the load on this thread. There are no guarantees about the
+/// precision of the timer other than that it will eventually tick. When the thread is healthy,
+/// the timer's period should be close to the specified one.
+///
```
geeknoid, line 26: Add a note that these times aren't affected by system clock adjustments.

crates/tick/src/periodic_timer.rs
```diff
@@ -0,0 +1,274 @@
...
+    clock: Clock,
+    // Currently scheduled timer. This value is not initialized until
+    // the first call to the `Stream::poll_next` method.
+    current_timer: Option<TimerKey>,
+}
+
+impl PeriodicTimer {
+    /// Creates a timer that fires periodically.
+    ///
+    /// This constructor automatically adjusts the provided period to the minimum allowed value.
```
geeknoid: Can we remove this constraint and simply state that we support 1ms precision. If we want to offer more precision in the future, we can add a new_with_more_precision function.

crates/tick/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+//!     })
+//!     .await;
+//! # }
+//! ```
+//!
+//! # Additional Examples
+//!
+//! The [time examples](https://github.com/microsoft/oxidizer/tree/main/crates/tick/examples)
+//! contain additional examples of how to use the time primitives.
+
```
geeknoid, line 212: Add this magic:

#![doc(html_logo_url = "https://media.githubusercontent.com/media/microsoft/oxidizer/refs/heads/main/crates/tick/logo.png")]
#![doc(html_favicon_url = "https://media.githubusercontent.com/media/microsoft/oxidizer/refs/heads/main/crates/tick/favicon.ico")]

crates/tick/src/clock.rs
martintmk: It would be more aligned wtih Rust APIs true.

What about others though?

- `new_frozen` -> `with_frozen` ?
- `new_frozen_at` -> `with_frozen_at` ?
- `ClockControl::new_at` -> `ClockControl::at` or `ClockControl::with_at` (meh)

crates/tick/Cargo.toml
```diff
@@ -0,0 +1,105 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+[package]
+name = "tick"
+description = "Provides primitives to interact with and manipulate machine time."
+version = "0.1.0"
+readme = "README.md"
+keywords = ["time", "clock", "tick", "timestamp", "runtime", "stopwatch", "oxidizer"]
```
ralfbiedert: This is too long, crates.io only supports 5 IIRC.

crates/tick/src/clock.rs
ralfbiedert: This is mainly for interop with `fundle` so you can pass a a `clock` where `AsRef<Clock>` is expected. `AsRef<Self>` is [reasonably common](https://doc.rust-lang.org/std/convert/trait.AsRef.html).

crates/tick/src/clock_timestamp.rs
ralfbiedert: Yes, I think I'd also prefer that `Timestamp` (the struct) becomes `tick::SystemTime`, since this seems to be the spiritual sibling of `std::time::SystemTime`?

crates/tick/src/future_ext.rs
ralfbiedert: I think the `...Ext` naming is pretty standard for extension traits, which this one here is.

crates/tick/src/periodic_timer.rs
ralfbiedert: >If we want to offer more precision in the future, we can add a new_with_more_precision function.

I wouldn't want adding a 2nd method to that later just to support a slightly larger value range of the same type. I think these adjustments aren't unheard of, but we could also panic on small values for now, since that would be an API contract violation and is in line how some `std` functions handle invalid programming-time values.

crates/tick/src/clock_timestamp.rs
geeknoid: Is Timestamp a no-regrets substitute for SystemTime? If so, if there is no downside to using Timestamp, then I do think calling it tick:SystemTime would be a plus.

crates/tick/src/periodic_timer.rs
geeknoid: I just don't want programmers to put in delicate code to deal with the maybe perhaps sometimes changes to the precision which may never happen. If adding  a new API (which is really really cheap with no risk of breaking anyone) is not worth it to support higher precision, then I would say that supporting higher precision itself is not worth it.

Let's make the API stable and predictable instead of baking in uncertainty from day 1.

crates/tick/src/clock.rs
geeknoid: Actually, I don't understand "frozen" as a name. This implies some sort of immutability, which isn't the case. Any reason not to use "fake"?

crates/tick/src/clock_timestamp.rs
ralfbiedert: I agree this enum feels like it shouldn't exist, at least as user-visible. The the boundary it's only used via `Into` conversion, I think the actual issue is replacing `ClockControl::new_at` which treats all 3 branches differently. If that could be unified, possibly this could be replaced with a trait (maybe something shorter though), or an `Into` into an already existing type (e.g., tbd. `SystemTime`).

crates/tick/src/fmt/iso_8601_timestamp.rs
martintmk: I would rather keep it without deref. The intention for these was to convert them to `Timestamp` not work with them directly

crates/tick/src/fmt/unix_seconds_timestamp.rs
martintmk: Sure, why not.

crates/tick/src/fmt/unix_seconds_timestamp.rs
martintmk: true, will add these

crates/tick/src/clock_timestamp.rs
martintmk: Lot of ecosystem is using std-friendly `SystemTime`, the tick gives you the option to make the work deterministic. 

If your scenario requires to recieve and store "SystemTime", this is the use-case for timestamps wich enable additional formatting options. I would rather have a extra `Timestamp` type to avoid a confusion. 

Additionally, timestamp is guarded behind timestamp feature which brings a `jiff` dependency.

crates/tick/src/clock_control.rs
martintmk: this is explicit auto_advance call which always advances the clock

crates/tick/src/clock_control.rs
martintmk: If periodic timers is scheduled to be fired in 2 seconds and we adnvace the time by 3, it will be fired too.

crates/tick/src/clock_control.rs
martintmk: yes, if clock is produced from clock control and you clone that clock, you still have a link to the original clock control

crates/tick/src/clock.rs
martintmk: The frozen implies that the clock is frozen in time, the passage of time is stopped. FYI, the `tick` does not mention the term "fake" anywhere.

crates/tick/src/clock.rs
martintmk: done.

crates/tick/src/clock_timestamp.rs
martintmk: dropped and replaced with sealed `IntoSystemTime` trait

crates/tick/src/clock_control.rs
martintmk: Originally I though about introducing `AutoAdvance` type with two variants that can describe the behavior, but if felt like overkill. 

> If this can't be removed, could the logic panic when auto_advance_limit is called at the wrong time?
The order is arbitrary, you can call `auto_advance_limit` then `auto_advance` or other way around. Only if you call `auto_advance_limit` without enabling `auto_advance` then it has no effect.

Could we leave this as it is and iterate later?

crates/tick/src/clock.rs
martintmk: done

crates/tick/src/periodic_timer.rs
martintmk: added

crates/tick/src/periodic_timer.rs
martintmk: I have updated the docs. I agree with Martin saying that this should not cause panics.

crates/tick/Cargo.toml
martintmk: updated

crates/tick/src/clock.rs
martintmk: updated the wording

crates/tick/src/clock.rs
geeknoid: Maybe "manual" would be a better choice then? It's not really frozen, the time can change. It just needs to be done by hand, whereas real time changes automatically.

crates/tick/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+    allow(
+        clippy::arithmetic_side_effects,
+        clippy::unchecked_duration_subtraction,
+        reason = "allow these lints in tests to improve the readability of the tests"
+    )
+)]
+#![cfg_attr(coverage_nightly, feature(coverage_attribute))]
+#![cfg_attr(docsrs, feature(doc_cfg))]
+
+//! Provides primitives to interact with and manipulate machine time.
```
ralfbiedert: I would add a half-sentence that the key benefit is the testability of time, maybe:

> Primitives to obtain, interact with and mock system time, for faster and more robust testing.

crates/tick/src/clock.rs
martintmk: > It's not really frozen, the time can change. 

Not really, this method is convenience for calling `ClockControl::new().to_clock()` where one discards the clock control. This way, the time is indeed frozen and won't move forward. We use this one actually quite a bit for APIs that require clock, but we are not testing the time related functionality.

crates/tick/src/clock_timestamp.rs
geeknoid: In my code, if I just systematically change SystemTime with Timestamp and do nothing else than just change that symbol, will my code keep working? In other words, is Timestamp a pure compatible superset of SystemTime?

If it is a pure superset, then why not call it SystemTime?

crates/tick/src/clock_control.rs
geeknoid: That's not what I asked :-)

Are the two sequences I show equivalent? If I create two clocks from a single clock control, are they both connected in the same way as if I clone two clocks? 

Whatever the answer is, let's put it in the docs.

crates/tick/src/clock_control.rs
geeknoid: Is adding a panic! difficult in this case? I just don't like silent API failures

crates/tick/src/clock_timestamp.rs
martintmk: SystemTime exposes some static functions such as SystemTime::now() which we won't have. With the first iterations, we actually didn't have our own type and we just re-used existing `SystemTime` and exposed extensions for it.

The static functions on `SystemTime` actually convinced us to introduce a new type. As for naming, I would follow the jiff naming, which also calls this type `Timestamp`. Since we are using internally I would prefer to keep the name. 
https://docs.rs/jiff/latest/jiff/struct.Timestamp.html


(imho I find the `Timestamp` more suitable than `SystemTime`, because the latter implies the time is related to the environment/system where it's being retrieved)

crates/tick/src/clock_control.rs
geeknoid: There's a difference between "the passage of time" and "adjusting the clock", which I think we might be handling differently in the real clock vs. the fake one.

The timers operate on a relative time. In a real system, if you say "wake me up in 30 seconds" then wait 15 seconds and adjust the clock backwards by 15 seconds, the timer will still wake up in 15 seconds, having waited a total real time of 30 seconds. So once the timer is created, it is not bound to the clock anymore and just waits its allotted time independent of what the system clock says.

So basically, the timer isn't configured to "wake up at <current time> + 30 seconds", it is configured with "30 seconds". 

I think our fake behavior needs to be reconciled with the real-world behavior.

crates/tick/src/clock_control.rs
martintmk: The panic is not appropriate, if we really want to be explicit what about introducing `AutoAdvance` type that encapsulates this behavior?

`AutoAdvance::step(Duration::from_millis(100))`
`AutoAdvance::step_with_limit(Duration::from_millis(100), Duration::from_secs(10))`

But this is something I wanted to advance because of additional APIs.

(or just have `ClockControl::auto_advance_with_limit` method)

crates/tick/src/clock_control.rs
martintmk: Currently, the timers actually do not react to us rewinding time. This is just implementation detail though, because this behavior requires some extra complexity to handle. 

Practically,  I have not encountered the need to handle this use-case because we have not have a need to rewind the clock yet. (albeit it's silently supported if you attempt to jump at particular timestamp which might be behind the current time).

Would you treat this behavior as an must have in an initial version?

crates/tick/src/clock_control.rs
martintmk: done

crates/tick/src/lib.rs
martintmk: done

crates/tick/src/error.rs
```diff
@@ -0,0 +1,135 @@
...
+/// ```
+/// use std::time::Duration;
+///
+/// use tick::fmt::Iso8601Timestamp;
+/// use tick::{Clock, Error, Timestamp};
+///
+/// "invalid date".parse::<Iso8601Timestamp>().unwrap_err();
+/// ```
+#[derive(Debug)]
+pub struct Error(ErrorKind);
```
ralfbiedert, line 31: Should this use `ohno` for backtrace support?

crates/tick/src/clock.rs
```diff
@@ -0,0 +1,618 @@
...
+/// let mut timer = PeriodicTimer::new(&clock, Duration::from_millis(10));
+///
+/// while let Some(()) = timer.next().await {
+///     // do something
+///         # break;
+/// }
+/// # }
+/// ```
+#[derive(Debug, Clone)]
+pub struct Clock(ClockInner);
```
ralfbiedert: And what about `ThreadAware` for this one?

crates/tick/src/into_system_time.rs
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use std::time::{Duration, SystemTime};
+
+/// A trait for types that can be converted into a [`SystemTime`].
+///
+/// This trait is used by [`ClockControl`][crate::ClockControl] to accept various
+/// timestamp representations when setting the clock's timestamp.
```
geeknoid:
```suggestion
/// timestamp representations when setting the clock's time.
```

crates/tick/src/into_system_time.rs
```diff
@@ -0,0 +1,33 @@
...
+    fn into_system_time(self) -> SystemTime;
+}
+
+impl IntoSystemTime for SystemTime {
+    fn into_system_time(self) -> SystemTime {
+        self
+    }
+}
+
+impl IntoSystemTime for Duration {
```
geeknoid: As I mentioned elsewhere, I think this is a bad idea. A duration is completely unrelated to the system's time, and I think it's mislading to provide this conversion.

crates/tick/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+//! async fn main() {
+//!     // Create a clock using the Tokio runtime
+//!     let clock = Clock::new_tokio();
+//!
+//!     // Get the current timestamp
+//!     let now = clock.timestamp();
+//!     println!("Current time: {now}");
+//!
+//!     // Delay execution
+//!     Delay::new(&clock, Duration::from_secs(1)).await;
```
geeknoid: Have you consider this syntax instead:

```rust
clock.delay(Duration::from_secs(1)).await;
```

A bit shorter and perhaps more discoverable?

martintmk: I would love this. It's actually something we had a while back, but the constructor approach was deemed more idiomatic. I would like to bring it back for easier APIs.

crates/tick/src/clock_timestamp.rs
martintmk: I was thinking about this a bit. The timestamp as we have it here is indeed mostly a duplication of `SystemTime` with extra formatting logic on top to allow serialization of system time/ timestamp into common formats such as ISO8601/RFC/UnixSeconds.

What if we only kept these formats and allow serialization of system time to these? Basically, service owners could work with native `std::time::SystemTime` and for storing system time or sending it outside of process boundaries they would use serialization/formatting types defined by this crate? 

For anything more complex, such as working with civil times, jiff can be used.

So I would propose the following changes:

- `Timestamp`: dropped
- `tick::fmt::Iso8601Timestamp`: renamed to `tick::fmt::Iso8601` and allows conversion from and into `SystemTime`
- `tick::fmt::Rfc2822Timestamp`: renamed to `tick::fmt::Rfc2822` and allows conversion from and into `SystemTime`
- `tick::fmt::UnixSecondsTimestamp`: renamed to `tick::fmt::UnixSeconds` and allows conversion from and into `SystemTime`

This would make the `tick` crate much leaner, no ambiguity while still supporting common operations required by services.

@geeknoid @ralfbiedert

crates/tick/src/into_system_time.rs
martintmk: I think you are right, I'll drop this trait and just use the `Into<SystemTime>`

crates/tick/src/clock_timestamp.rs
geeknoid: I like it. Removes all ambiguity.

crates/tick/src/into_system_time.rs
martintmk: done

crates/tick/src/clock_control.rs
geeknoid: Delivering a consistent experience between the real world and the testing world is super important. If the tests behave in one way, but the real system behaves in a different way, we will be inviting developers to create broken code.

So I think consistency is super important.

Now, changing the fake clock to move the time backwards seems like a "can come later" concern. You could constrain it such that the fake clock can only move forward for now, which would simplify some of the edge cases in the implementation.

crates/tick/src/lib.rs
geeknoid: Hmmm. In terms of being idiomatic, to me this lines up with the builder pattern somewhat.

When you have Foo and FooBuilder, we don't expose Foo::new and FooBuilder::new, but instead we expose Foo::builder.

So to me, Clock::delay is like Foo::builder...

ralfbiedert: I'm not very opposed to it. 

The main argument against it, as a general pattern, is that it muddles responsibilities. Like why might `clock.delay()` exist and not `clock.something_else()` for an arbitrary `SomethingElse` that happens to want a clock? We therefore wanted to be conservative adding these, unless there was a clear UX benefit.

That said, I've found myself typing a few times `Sleep`?... `Timeout`? ... `Delay`; and `clock.delay()` might be one of these cases.

crates/tick/src/error.rs
martintmk: this being such a low-level crate, I would rather avoid extra dependency.

crates/tick/src/lib.rs
martintmk: I have added `Clock.delay(duration)` method as a convenience. This is indeed used quite often so having it here makes sense.

crates/tick/examples/interop_jiff.rs
```diff
@@ -0,0 +1,51 @@
...
+    println!(
+        "Current time ({}): {}",
+        TimeZone::system().iana_name().context("failed to get time zone name")?,
+        zoned.strftime(JIFF_DISPLAY_FORMAT)
+    );
+
+    // Temporal is a new pending standard that preserves time zone information.
+    // https://tc39.es/proposal-temporal/docs/index.html
+    let mut buff = String::new();
+    DateTimePrinter::new().print_zoned(&zoned, &mut buff)?;
```
BurntSushi: You can just do `zoned.to_string()` here. It's a [documented guarantee](https://docs.rs/jiff/latest/jiff/struct.Zoned.html#impl-Display-for-Zoned).

Note that the specific output format is [RFC 9557](https://datatracker.ietf.org/doc/rfc9557/), which is not pending. It's a thing now. :-)

martintmk: thanks! I'll simplify this :)

crates/tick/src/clock_timestamp.rs
martintmk: We also agreed on the `Timestamp` removal with Ralf, I have started the work but more changes are needed.

crates/tick/src/clock.rs
martintmk: will add that in follow-up, requires more changes

crates/tick/src/fmt/iso_8601.rs
```diff
@@ -0,0 +1,241 @@
...
+///
+/// The ISO 8601 standard is used worldwide in various applications, ranging
+/// from software and digital formats to international communication, ensuring
+/// consistency across different systems and regions.
+///
+/// Examples:
+/// - `2024-08-06T21:30:00Z` (UTC)
+/// - `2024-08-06T14:30:00-07:00` (UTC offset)
+///
+/// The [`Iso8601`] format is defined in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601).
```
geeknoid: Maybe this is a better link? The actual standard doc:

https://www.iso.org/obp/ui/#iso:std:iso:8601:-1:ed-1:v1:en

crates/tick/src/fmt/iso_8601.rs
```diff
@@ -0,0 +1,241 @@
...
+///
+/// # UTC and time zones
+///
+/// While ISO 8601 can include a UTC offset, the resulting [`Timestamp`] is always represented in the
+/// UTC time zone with an offset of `Z`.
+///
+/// # Serialization and deserialization
+///
+/// `Iso8601` implements the `Serialize` and `Deserialize` traits from the `serde` crate.
```
geeknoid: Add a note that the 'serde' feature needs to be enabled for this.

crates/tick/src/fmt/iso_8601.rs
```diff
@@ -0,0 +1,241 @@
...
+///
+/// let iso = "1990-12-31T23:59:60Z".parse::<Iso8601>()?;
+/// assert_eq!(iso.to_string(), "1990-12-31T23:59:59Z");
+///
+/// # Ok::<(), Box<dyn std::error::Error>>(())
+/// ```
+///
+/// # Examples
+///
+/// ### Formatting and parsing - UTC
```
geeknoid:
```suggestion
/// ## Formatting and parsing - UTC
```

crates/tick/src/fmt/iso_8601.rs
```diff
@@ -0,0 +1,241 @@
...
+/// let iso = "2024-08-06T21:30:00Z".parse::<Iso8601>()?;
+/// assert_eq!(iso.to_string(), "2024-08-06T21:30:00Z");
+///
+/// let timestamp: Timestamp = iso.into();
+/// assert_eq!(Iso8601::from(timestamp), iso);
+///
+/// # Ok::<(), Box<dyn std::error::Error>>(())
+/// ```
+///
+/// ### Formatting and parsing - With UTC offset
```
geeknoid, line 70:
```suggestion
/// ## Formatting and parsing - With UTC offset
```

crates/tick/src/fmt/iso_8601.rs
```diff
@@ -0,0 +1,241 @@
...
+use super::utils::from_jiff;
+use crate::fmt::utils::to_jiff;
+use crate::fmt::{Rfc2822, UnixSeconds};
+use crate::{Error, Timestamp};
+
+/// Parser and formatter for ISO 8601 timestamps.
+///
+/// The ISO 8601 standard is used worldwide in various applications, ranging
+/// from software and digital formats to international communication, ensuring
+/// consistency across different systems and regions.
```
BurntSushi, line 17: It might be worth saying that RFC 3339 is also supported here.

And that the output is always compatible with _both_ RFC 3339 and ISO 8601.

martintmk: done

crates/tick/src/fmt/iso_8601.rs
martintmk: done

crates/tick/src/clock_timestamp.rs
martintmk: done

Darksecond (approved):

crates/tick/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+    allow(
+        clippy::arithmetic_side_effects,
+        clippy::unchecked_duration_subtraction,
+        reason = "allow these lints in tests to improve the readability of the tests"
+    )
+)]
+#![cfg_attr(coverage_nightly, feature(coverage_attribute))]
+#![cfg_attr(docsrs, feature(doc_cfg))]
+
+//! Runtime-agnostic primitives for obtaining, working with, and mocking system
```
geeknoid: The crate has a dependency on tokio, so perhaps remove the "Runtime-agnostic" part.

Also, make sure this first line is repeated in Cargo.toml's description field and in the top-level README.md file.

crates/tick/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+//!         assert_eq!(produce_value(&clock).await, 123);
+//!     }
+//! }
+//! ```
+//!
+//! # Why?
+//!
+//! This crate provides a unified API for working with time that:
+//!
+//! - **Abstracts async runtimes** - Works across Tokio, async-std, etc. without tight coupling
```
geeknoid: Again, qualified by the fact the code does provide tokio integration.

crates/tick/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+//! [jiff]: https://crates.io/crates/jiff
+//! [chrono]: https://crates.io/crates/chrono
+//! [time]: https://crates.io/crates/time
+//!
+//! # Testing
+//!
+//! This crate provides a way to control the passage of time in tests via the `ClockControl`
+//! type, which is exposed when the `test-util` feature is enabled.
+//!
+//! **Important:** Never enable the `test-util` feature for production code. Only use it in your `dev-dependencies`.
```
geeknoid: Use a block quote:


```suggestion
//! > **Important:** Never enable the `test-util` feature for production code. Only use it in your `dev-dependencies`.
```

crates/tick/src/clock.rs
```diff
@@ -0,0 +1,662 @@
...
+///   enabled, you can format `SystemTime` into different formats.
+///
+/// Absolute time is not monotonic and can be affected by system clock changes.
+///
+/// When possible, always prefer [`Stopwatch`][super::Stopwatch] over absolute time due to its monotonic properties.
+/// For scenarios where you need absolute time, use [`system_time()`][Self::system_time].
+///
+/// # Clock construction
+///
+/// The clock requires a runtime to drive the registered timers. For this reason, clock construction is non-trivial
```
geeknoid: Contradicting the "Runtime-agnostic" claims...

crates/tick/src/lib.rs
geeknoid: Is there support for async-std?

crates/tick/src/clock.rs
```diff
@@ -0,0 +1,662 @@
...
+/// and clock access is provided by the runtime.
+///
+/// In production, the clock is typically obtained from your async runtime. Different runtimes
+/// provide different mechanisms for clock access. Consult your runtime's documentation for details.
+///
+/// In tests, the clock can be constructed directly using `new_frozen()` or
```
geeknoid: Can you make new_frozen and new_frozen_at links?

crates/tick/src/clock.rs
```diff
@@ -0,0 +1,662 @@
...
+/// this crate provides a [`ClockControl`] type that can be used to control time.
+///
+/// # Cloning
+///
+/// Cloning a clock is inexpensive (just an `Arc` clone) and will not cause performance bottlenecks.
+/// Cloned clocks share the same underlying state, including registered timers and, in tests, the
+/// controlled passage of time.
+///
+/// # State sharing between clocks
+///
```
geeknoid: These two sections contain redundant text, maybe consolidate?

crates/tick/src/clock.rs
```diff
@@ -0,0 +1,662 @@
...
+/// // Cloned clocks; all these instances are linked
+/// // together and share the same state.
+/// let clock_clone1 = clock.clone();
+/// let clock_clone2 = clock.clone();
+/// # }
+/// ```
+///
+/// # Examples
+///
+/// ### Retrieve absolute time
```
geeknoid:
```suggestion
/// ## Retrieve absolute time
```

crates/tick/src/clock.rs
```diff
@@ -0,0 +1,662 @@
...
+/// # fn retrieve_absolute_time(clock: &Clock) {
+/// // Using SystemTime for basic absolute time needs
+/// let time1: SystemTime = clock.system_time();
+/// let time2: SystemTime = clock.system_time();
+///
+/// assert!(time2 >= time1);
+/// # }
+/// ```
+///
+/// ### Measure elapsed time
```
geeknoid:
```suggestion
/// ## Measure elapsed time
```

crates/tick/src/clock.rs
```diff
@@ -0,0 +1,662 @@
...
+/// use tick::Clock;
+///
+/// # fn measure(clock: &Clock) {
+/// let stopwatch = clock.stopwatch();
+/// // Perform some operation...
+/// let elapsed: Duration = stopwatch.elapsed();
+/// # }
+/// ```
+///
+/// ### Delay operations
```
geeknoid:
```suggestion
/// ## Delay operations
```

crates/tick/src/clock.rs
```diff
@@ -0,0 +1,662 @@
...
+/// let stopwatch = clock.stopwatch();
+///
+/// // Delay for 10 milliseconds
+/// clock.delay(Duration::from_millis(10)).await;
+///
+/// assert!(stopwatch.elapsed() >= Duration::from_millis(10));
+/// # }
+/// ```
+///
+/// ### Create periodic timers
```
geeknoid:
```suggestion
/// ### Create periodic timers
```

crates/tick/src/clock.rs
```diff
@@ -0,0 +1,662 @@
...
+    ///
+    /// # fn retrieve_system_time(clock: &Clock) {
+    /// let time1 = clock.system_time();
+    /// let time2 = clock.system_time();
+    ///
+    /// assert!(time2 >= time1);
+    /// # }
+    /// ```
+    #[must_use]
+    pub fn system_time(&self) -> SystemTime {
```
geeknoid, line 270: Any reason not to call this 'now'?

crates/tick/src/delay.rs
```diff
@@ -0,0 +1,241 @@
...
+    /// let stopwatch = clock.stopwatch();
+    ///
+    /// // Delay for 10 milliseconds
+    /// Delay::new(clock, Duration::from_millis(10)).await;
+    ///
+    /// assert!(stopwatch.elapsed() >= Duration::from_millis(10));
+    /// # }
+    /// ```
+    #[must_use]
+    pub fn new(clock: &Clock, duration: Duration) -> Self {
```
geeknoid, line 72: Given Clock::delay, do we need to make this public?

crates/tick/src/clock_control.rs
```diff
@@ -0,0 +1,819 @@
...
+    ///
+    /// use tick::ClockControl;
+    ///
+    /// let control = ClockControl::now();
+    /// let clock = control.to_clock();
+    ///
+    /// assert!(SystemTime::now() >= clock.system_time());
+    /// ```
+    #[must_use]
+    pub fn now() -> Self {
```
geeknoid: What's the use case for this? Feels like it is inviting users to mix "real-time" with "fake-time".

crates/tick/src/clock_control.rs
```diff
@@ -0,0 +1,816 @@
...
+    /// assert_eq!(clock.system_time(), target);
+    /// ```
+    #[expect(
+        clippy::missing_panics_doc,
+        reason = "we are handling cases where the timestamp is either in future or past and the resulting duration is always positive"
+    )]
+    pub fn set_time(&self, timestamp: impl Into<SystemTime>) {
+        let now = self.system_time();
+        let timestamp = timestamp.into();
+
```
geeknoid, line 320: As discussed, rename this to set_time or something like it. Add to the comment explaining that changing the time does not affect any outstanding timers, delays, or stopwatches.

crates/tick/src/error.rs
```diff
@@ -0,0 +1,158 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use std::{fmt, time::SystemTimeError};
+
+/// The result type for fallible operations that use the [`Error`] type in the `time` module.
+pub type Result<T> = std::result::Result<T, Error>;
+
+/// An error that can occur in the `time` module.
```
geeknoid:
```suggestion
/// An error that can occur in the `tick` crate.
```

crates/tick/src/delay.rs
```diff
@@ -0,0 +1,241 @@
...
+///
+/// // Delay for 10 milliseconds
+/// clock.delay(Duration::from_millis(10)).await;
+///
+/// assert!(stopwatch.elapsed() >= Duration::from_millis(10));
+/// # }
+/// ```
+#[derive(Debug)]
+pub struct Delay {
+    // Currently scheduled timer. This value is not initialized before
```
geeknoid, line 40: Does this mean that the delay only starts counting the first time the application calls await on something?

geeknoid: And same applies to the timeout function, the timeout only starts counting relative to the next time the app awaits.

crates/tick/src/periodic_timer.rs
```diff
@@ -0,0 +1,277 @@
...
+        }
+
+        match this.current_timer {
+            Some(key) if key.tick() <= this.clock.instant() => {
+                // Reset the timer. It will be registered again on the next poll.
+                this.current_timer = None;
+
+                // Unregister the timer, just in case this call was explicit and not due to
+                // timers advancing.
+                this.clock.unregister_timer(key);
```
geeknoid, line 133: So if I have a timer set to tick every millisecond and I have some compute that takes 10ms and then calls await, the timer will then fire in another ms. Right?

crates/tick/src/stopwatch.rs
```diff
@@ -0,0 +1,156 @@
...
+    /// // Perform some operation...
+    /// stopwatch.elapsed()
+    /// # }
+    /// ```
+    #[cfg_attr(
+        not(any(feature = "test-util", test)),
+        expect(unused_variables, reason = "intentionally not using self-references if test-util is disabled")
+    )]
+    #[must_use]
+    pub fn new(clock: &Clock) -> Self {
```
geeknoid, line 60: Just like Delay::new, do we need to expose this given the funciton is exposed on Clock directly?

geeknoid (commented):
A suggestion for the future: maybe it would be nice for timers to have names, to help with debugging?

crates/tick/src/clock.rs
martintmk: there is also an `Clock::instant` which could also be called `now`. We didn't want to have `instant_now` and `system_time_now` so shorter version was used.

crates/tick/src/delay.rs
martintmk: I discussed with Ralf and `Clock::delay` is just a convenience, with constructor being the primary pattern (in case someone want's to expose new primitives built on Clock, we don't want to pollute clock API with too many methods)

crates/tick/src/clock_control.rs
martintmk: done.

crates/tick/src/stopwatch.rs
martintmk: same argument as for Delay, this convenience, while the constructor form is the primary form

crates/tick/src/clock_control.rs
martintmk: dropped, it's indeed not required

crates/tick/src/clock.rs
martintmk: done

crates/tick/src/clock.rs
martintmk: done

crates/tick/src/clock.rs
martintmk: refined this section, worded as "easy integration with async runtimes"

crates/tick/src/lib.rs
martintmk: refined and dropped mention of async-std

crates/tick/src/lib.rs
martintmk: re-worded this section

crates/tick/src/delay.rs
martintmk: yes, we do prefer scalability over accuracy. Basically if async runtime lags behind and is not able to poll the futures, excessive timers won't be created either.

crates/tick/src/periodic_timer.rs
martintmk: Yes it will fire. These are separate, timer will be fired by the runtime that owns the clock drivers.

crates/tick/src/clock_control.rs
martintmk: Created an issue for this #136

PR #109, opened 2025-12-08
ci: Always pass some tests when there isn't enough memory available

martin-kolinek:
Our release pipeline runs on relatively small machines and these tests are flaky there. Getting bigger machines for these releases is non-trivial and we're running these tests as part of our normal build pipelines as well, so it should be ok to skip them on smaller systems.

Code coverage should catch situations where the test is unintentionally skipped.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -71,7 +71,7 @@ checksum = "b35204fbdc0b3f4446b89fc1ac2cf84a8a68971995d0bf2e925ec7cd960f9cb3"
 
 [[package]]
 name = "bytesbuf"
-version = "0.1.0"
+version = "0.1.1"
 dependencies = [
  "alloc_tracker",
  "bytes",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ repository = "https://github.com/microsoft/oxidizer"
 [workspace.dependencies]
 
 # local dependencies
-bytesbuf = { path = "crates/bytesbuf", default-features = false, version = "0.1.0" }
+bytesbuf = { path = "crates/bytesbuf", default-features = false, version = "0.1.1" }
 data_privacy = { path = "crates/data_privacy", default-features = false, version = "0.7.0" }
 data_privacy_macros = { path = "crates/data_privacy_macros", default-features = false, version = "0.7.0" }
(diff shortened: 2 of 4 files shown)


crates/bytesbuf/src/transparent.rs
```diff
@@ -115,6 +116,14 @@ mod tests {
 
     #[test]
     fn giant_allocation() {
+        // This test requires at least 5 GB of memory to run. The publishing pipeline runs on a system
+        // where this may not be available, so we skip this test in that environment.
+        #[cfg(all(not(miri), any(target_os = "linux", target_os = "windows")))]
```
Vaiz, line 121: don't we have the same memory restrictions when we run miri? it might be better to disable the whole test under miri

crates/bytesbuf/src/bytes.rs
```diff
@@ -249,6 +250,14 @@ mod tests {
 
     #[test]
     fn test_giant_bytes_to_sequence() {
+        // This test requires at least 5 GB of memory to run. The publishing pipeline runs on a system
```
psandana, line 253: Is it possible to feature gate it, and not activate the feature when publishing?

I feel the decision here too arbitrary that may hinder problems with this test.

crates/bytesbuf/src/transparent.rs
martin-kolinek: No, because MIRI runs on a machine with more memory. The only thing that runs on memory-constrained machines is the release, where we only need to run the tests normally, without miri. The test actually has quite a bit of value being executed under miri as it exercises some unsafe code paths that are not exercised otherwise.

crates/bytesbuf/src/bytes.rs
martin-kolinek: It doesn't look like building with all features except a given one is possible: https://github.com/rust-lang/cargo/issues/3126. That would mean we need to somehow make sure all the features except this one get applied during release build. There was another option proposed which was to explicitly skip tests in the release pipeline, but that also has drawbacks as it decouples the tests from the place where it's decided which tests are executed.

Given that code coverage will catch situations where this test is not executed when it's meant to be, this sounds like a reasonable option.

PR #114, opened 2025-12-09
feat: Add a redactor that uses RapidHash

geeknoid:
Since RapidHash is faster than xxH3, we should support it.

Also, remove some non-sensical trait implementations for the xxH3Redactor and the SimpleRedactor.
Redactors should not be Ord, PartialOrd, or Hash.

### DIFF
--- a/crates/data_privacy/src/redactors/mod.rs
+++ b/crates/data_privacy/src/redactors/mod.rs
@@ -8,8 +8,12 @@ use std::collections::HashMap;
 use std::fmt::Write;
 
 pub mod simple_redactor;
+
 #[cfg(feature = "xxh3")]
 pub mod xxh3_redactor;
+
+#[cfg(feature = "rapidhash")]
+pub mod rapidhash_redactor;
 
 /// Represents types that can redact data.
 pub trait Redactor {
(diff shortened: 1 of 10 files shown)


crates/data_privacy/src/redactors/rapidhash_redactor.rs
```diff
@@ -0,0 +1,75 @@
...
+        let hash = rapidhash_v3_seeded(value.as_bytes(), &self.secrets);
+        let buffer = u64_to_hex_array(hash);
+
+        // SAFETY: The buffer is guaranteed to be valid UTF-8 because it only contains hex digits.
+        write!(output, "{}", unsafe { core::str::from_utf8_unchecked(&buffer) })
+    }
+}
+
+#[inline]
+fn u64_to_hex_array(mut value: u64) -> [u8; 16] {
```
ralfbiedert: This here could probably be a `u64_to_hex_array<const N: usize>() -> [u8; N]`, be moved to the parent `mod.rs`, and then be called from all redactors.

geeknoid: Done!

crates/data_privacy/src/redactors/mod.rs
```diff
@@ -79,11 +83,41 @@ impl Debug for Redactors {
     }
 }
 
+#[cfg(any(feature = "xxh3", feature = "rapidhash"))]
+#[inline]
+pub fn u64_to_hex_array<const N: usize>(mut value: u64) -> [u8; N] {
```
Vaiz, line 88: this technique should be faster, thought it has some limitations
https://johnnylee-sde.github.io/Fast-unsigned-integer-to-hex-string/

geeknoid: I don't think so :-)

His benchmarks show that a table-based approach is fastest. And I think my table-based approach is in fact quite a bit faster than his as I have 1/2 the number of operators, without the 1K table (which takes up more room and is less cache-friendly)

PR #115, opened 2025-12-09
feat: Thread aware cache

ZacWalk:
This PR introduces `thread_aware_cache`, a high-performance in-memory cache designed to maximize read throughput and minimize latency on **multi-socket NUMA architectures**.

* **Thread-Affinity Sharding**
    Unlike traditional caches that shard by key hash, this implementation shards by thread affinity. This ensures data remains physically local to the executing thread, significantly reducing cross-socket interconnect traffic.

* **"Read-Through" Replication**
    This implementation utilizes a unique replication strategy that automatically promotes hot data to the local shard upon access. This is supported by a **lock-free Bloom filter** to efficiently short-circuit negative lookups across shards.

* **SIEVE Eviction Algorithm**
    Capacity management is handled by the **SIEVE** algorithm. This provides scan-resistance and high concurrency by eliminating the need for heavy write locks during cache reads.

### DIFF
--- a/crates/thread_aware_cache/CHANGELOG.md
+++ b/crates/thread_aware_cache/CHANGELOG.md
@@ -1,0 +1,14 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+## [0.1.0] - 2024-12-09
+
+### Added
+
+- Initial release of `thread_aware_cache`.
--- a/crates/thread_aware_cache/benches/cache.rs
+++ b/crates/thread_aware_cache/benches/cache.rs
@@ -1,0 +1,383 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Benchmarks for the NUMA cache.
+
+#![expect(missing_docs, reason = "Benchmark code does not require documentation")]
+
+use std::hint::black_box;
+use std::sync::Arc;
+use std::thread;
+
+use criterion::{BatchSize, BenchmarkId, Criterion, Throughput, criterion_group, criterion_main};
+use rand::Rng;
+use rand::SeedableRng;
+use rand::rngs::StdRng;
+use thread_aware::create_manual_pinned_affinities;
+use thread_aware_cache::NumaCache;
+
+criterion_group!(benches, bench_basic, bench_concurrent, bench_cross_shard, bench_bloom_filter);
+criterion_main!(benches);
+
+const CACHE_CAPACITY: usize = 10_000;
+const NUM_SHARDS: usize = 4;
+const TOTAL_CAPACITY: usize = CACHE_CAPACITY * NUM_SHARDS;
+
+fn bench_basic(c: &mut Criterion) {
+    let mut group = c.benchmark_group("NumaCache");
+
+    // Create affinities for benchmarks
+    let affinities = create_manual_pinned_affinities(&[1, 1, 1, 1]);
+
+    // Benchmark cache creation
+    group.bench_function("new", |b| {
+        b.iter(|| {
+            black_box(
+                NumaCache::<String, i32>::builder()
+                    .affinities(&affinities)
+                    .capacity_per_shard(CACHE_CAPACITY)
+                    .build(),
+            )
+        });
+    });
+
+    // Benchmark single-threaded insert
+    group.throughput(Throughput::Elements(1));
+    group.bench_function("insert_single", |b| {
+        let cache = NumaCache::<usize, usize>::builder()
+            .affinities(&affinities)
+            .capacity_per_shard(CACHE_CAPACITY)
+            .build();
+        let affinity = affinities[0];
+
+        let mut key = 0usize;
+        b.iter(|| {
+            cache.insert(affinity, key, key);
+            key = (key + 1) % TOTAL_CAPACITY;
+        });
+    });
+
+    // Benchmark single-threaded get (local hit)
+    group.bench_function("get_local_hit", |b| {
+        let cache = NumaCache::<usize, usize>::builder()
+            .affinities(&affinities)
+            .capacity_per_shard(CACHE_CAPACITY)
+            .build();
+        let affinity = affinities[0];
+
+        // Pre-populate the cache on the same affinity
+        for i in 0..CACHE_CAPACITY {
+            cache.insert(affinity, i, i);
+        }
+
+        let mut rng = StdRng::seed_from_u64(42);
+        b.iter(|| {
+            let key = rng.random_range(0..CACHE_CAPACITY);
+            black_box(cache.get(affinity, &key));
+        });
+    });
+
+    // Benchmark single-threaded get (miss)
+    group.bench_function("get_miss", |b| {
+        let cache = NumaCache::<usize, usize>::builder()
+            .affinities(&affinities)
+            .capacity_per_shard(CACHE_CAPACITY)
+            .build();
+        let affinity = affinities[0];
+
+        let mut key = TOTAL_CAPACITY; // Start outside the range
+        b.iter(|| {
+            key += 1;
+            black_box(cache.get(affinity, &key));
+        });
+    });
+
+    // Benchmark eviction
+    group.bench_function("insert_with_eviction", |b| {
+        let single_affinity = create_manual_pinned_affinities(&[1]);
+        b.iter_batched(
+            || {
+                let cache =
+                    NumaCache::<usize, usize>::builder().affinities(&single_affinity).capacity_per_shard(100).build();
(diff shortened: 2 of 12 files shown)


ralfbiedert:
What is the relation between this and the cache Schuyler is working on, do they serve different use cases? Also, I don't quite understand how the cache does thread migration, it doesn't seem to implement the `ThreadAware` trait?

geeknoid:
Please see the top-level README.md files for the steps needed to add a new crate. You need to register it in the top-level README and the top level CHANGELOG, you need to add the requisite logo and favicon, and you need the voodoo lines in the crate-level documentation to pull in the logo and icon into docs.rs. See the script scripts\add-crate.ps1 for everythig.

crates/thread_aware_cache/CHANGELOG.md
```diff
@@ -0,0 +1,14 @@
+# Changelog
```
geeknoid, line 1: We auto-generate the changelog, so you can leave this blank for starters and it'll be filled in when releasing the crate.

crates/thread_aware_cache/benches/cache.rs
```diff
@@ -0,0 +1,383 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Benchmarks for the NUMA cache.
```
geeknoid, line 4: Can you add comparative benchmarks for existing cache crates out there?

crates/thread_aware_cache/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+//!
+//! The cache uses a shared lock-free Bloom filter to optimize cross-shard lookups. Before
+//! searching all shards for a key, the cache queries the Bloom filter:
+//!
+//! - **Negative result:** Key definitely doesn't exist → skip cross-shard search (O(1))
+//! - **Positive result:** Key might exist → proceed with cross-shard search
+//!
+//! The Bloom filter is sized for ~1% false positive rate and uses atomic operations for
+//! lock-free concurrent access. Note that removals don't clear bits from the filter, so
+//! stale positives may occur (safe, just slower), but false negatives never occur.
```
geeknoid, line 114: Does the bloom filter ever get regenerated over time?

crates/thread_aware_cache/src/bloom.rs
```diff
@@ -0,0 +1,283 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Lock-free Bloom filter for fast negative lookups.
```
geeknoid, line 4: Unless this code has some magic in it, can we use an existing crate for this like fastbloom?

If thia code does have some magic, then perhaps it should be in a separate crate?

crates/thread_aware_cache/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+//! 5. **Bloom Filter:** Kirsch-Mitzenmacher optimization for deriving multiple hash functions.
+
+mod bloom;
+mod cache;
+mod shard;
+mod sieve;
+
+pub use cache::{NumaCache, NumaCacheBuilder};
+
+// Re-export thread_aware types for convenience
```
geeknoid, line 132: @ralfbiedert Do we have a policy on reexports?

crates/thread_aware_cache/src/lib.rs
```diff
@@ -0,0 +1,136 @@
+// Copyright (c) Microsoft Corporation.
```
geeknoid, line 1: Couple high-level comments:

- You use the word 'shard' to describe the different partitions. Given this is explicitly a NUMA cache, when would you expect the shards to be different than the NUMA nodes? If they're also going to be 1:1, then can we just use the words 'NUMA node' instead of shard in order to eliminate one concept?

- You call this thread_aware_cache, but it's really a NUMA-aware cache, right? There is no general affordance for thread-specific shards. You can imagine having per-thread L0 caches to completely eliminate contention for common lookups. So should we call this crate numa_aware_cache?

- Would it be possible to support the Borrow<T> pattern like HashMap does? This enables mix-and-match between &str and String which is mighty handy.

- Caches should report some state about their efficacy. Hit & miss rates in particular. Can we add this? Otherwise, it's hard for app to tune the cache size.

- I worry that this won't scale well within a shard given the very coarse lock that's taken.

- I also worry that a contended lock can have catastophic effects on overall systemic perf. If N threads are in the middle of reading from a shard and a writer comes along, that writer will block and in a thread-per-core world the thread's core will go completely idle. Should we be using async mutexes and make cache access be async methods? Is the overhead of async warranted or would that potentially still also suffer from contention inside the async mutex logic?

- I didn't dig enough in the code, but what's the lock behavior when a cache miss occur and probing occurs in other shards? If the shard's lock is held across the whole operation, this could lead to substantial contention. If the lock is not held, then we have a racy situation where a probe could be occurring and then a local thread can put the sought-after data into the shard "behind the scenes". Or 10 threads could potentially be probing for the same state at the same time.

crates/thread_aware_cache/src/cache.rs
```diff
@@ -0,0 +1,691 @@
...
+use crate::bloom::BloomFilter;
+use crate::shard::NumaShard;
+
+/// A high-performance, NUMA-aware in-memory cache with SIEVE eviction.
+///
+/// The cache partitions data across multiple shards, with routing based on
+/// thread affinity to minimize cross-NUMA traffic. Each shard is explicitly
+/// associated with a [`thread_aware::PinnedAffinity`].
+///
+/// A shared Bloom filter optimizes cross-shard lookups by quickly identifying
```
geeknoid, line 23: Seems like an implementation detail that's not needed in the user docs.

crates/thread_aware_cache/src/cache.rs
```diff
@@ -0,0 +1,691 @@
...
+/// keys that definitely don't exist in any shard, avoiding expensive O(n) searches.
+///
+/// # Type Parameters
+///
+/// * `K` - The key type, must implement `Eq + Hash + Clone`.
+/// * `V` - The value type, must implement `Clone`.
+/// * `S` - The hash builder type, defaults to `DefaultHashBuilder`.
```
geeknoid, line 30: The default hasher is very slow. We might want to default to something else, like RapidHash.

crates/thread_aware_cache/src/shard.rs
```diff
@@ -0,0 +1,266 @@
...
+use std::hash::{BuildHasher, Hash};
+
+use hashbrown::DefaultHashBuilder;
+use hashbrown::HashMap;
+use parking_lot::RwLock;
+
+use crate::sieve::{NodeIndex, SieveList};
+
+/// Cache line size for alignment to prevent false sharing.
+const CACHE_LINE_SIZE: usize = 64;
```
geeknoid, line 18: Nobody is making a 128 byte cache line? There's no system-level constant for this we could leverage?

crates/thread_aware_cache/src/shard.rs
```diff
@@ -0,0 +1,266 @@
...
+
+/// A single cache shard containing data and eviction metadata.
+///
+/// Aligned to the CPU cache line (64 bytes) to prevent cache-line bouncing between locks.
+#[repr(align(64))]
```
geeknoid, line 23: Can you use CACHE_LINE_SIZE here, or are you not allowed to use constants in these attribute expressions?

ZacWalk:
> What is the relation between this and the cache Schuyler is working on, do they serve different use cases? Also, I don't quite understand how the cache does thread migration, it doesn't seem to implement the `ThreadAware` trait?

Schuyler told me they did not have an in-memory part to their cache yet. I wanted to contribute this one we wrote last year but was parked. This code has been evolved to use thread_aware. Let me set to draft while we discuss if that is a good idea or not.

crates/thread_aware_cache/src/lib.rs
ZacWalk: Handling removals in the bloom filter is tricky because items are independently removed from the shards. Think of it like SIEVE eviction per NUMA node. To handle this well we would need a bit per affinity. My thinking was to have this decided at compile:
 - Cache has a bit per affinity bloom filter because the cache will have large number of deletes
 - Single bit but does not track removals and may become saturated. This would be right for mostly read only caches.

crates/thread_aware_cache/src/lib.rs
ZacWalk: Good comments. My intention was to start the ball rolling again with previous work. Setting to draft while we mature the ideas here and have some discussions.

crates/thread_aware_cache/src/lib.rs
geeknoid: I guess I was thinking that you could keep track of the number of false positive over time. As the ratio gets over a certain threshold, you'd throw the filter away and rebuild it. I don't know if rebuilding it is possible though.

I don't know if there is such a thing as a mostly read-only cache. A cache with TTL will eventually recycle every item, so the bloom filter will become ineffective.

PR #117, opened 2025-12-10
chore: publish bytesbuf 0.1.2

Vaiz:
The PR fixes `testing_aids` dependency by specifying its path instead of version

```
  Publishing bytesbuf
    Updating `CratesIoMirror` index
   Packaging bytesbuf v0.1.1 (D:\a\_work\1\s\crates\bytesbuf)
    Updating `CratesIoMirror` index
error: failed to prepare local package for uploading

Caused by:
  no matching package named `testing_aids` found
  location searched: `CratesIoMirror` index (which is replacing registry `crates-io`)
  required by package `bytesbuf v0.1.1 (D:\a\_work\1\s\crates\bytesbuf)`
```


sandersaares (approved):
I would prefer just removing `version = ..` from the root `cargo.toml` as there is no purpose to having a version on `testing_aids` - if it is never published, it is never resolved by version.

PR #118, opened 2025-12-10
feat: Add uniflight crate for duplicate request coalescing

schgoo:
Adds `uniflight`, a crate for coalescing duplicate async tasks into a single execution. When multiple tasks request the same work (identified by a key), only the leader performs the work while followers wait and receive a clone of the result.

* Handles cancellation and panic gracefully (followers become the new leader)
* Includes comprehensive module documentation
* Includes example demonstrating cache thundering herd prevention

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -83,15 +83,17 @@ runs:
           echo "cargo_tools=$expanded" >> "$GITHUB_OUTPUT"
         fi
 
+    # Note: sccache is skipped on Windows ARM64 due to missing binaries
+    # See: https://github.com/Mozilla-Actions/sccache-action/issues/189
     - name: Configure Sccache
-      if: inputs.enable-sccache == 'true'
+      if: inputs.enable-sccache == 'true' && !(runner.os == 'Windows' && runner.arch == 'ARM64')
       shell: bash
       run: |
         echo "SCCACHE_GHA_ENABLED=true" >> "$GITHUB_ENV"
         echo "RUSTC_WRAPPER=sccache" >> "$GITHUB_ENV"
 
     - name: Start Sccache
-      if: inputs.enable-sccache == 'true'
+      if: inputs.enable-sccache == 'true' && !(runner.os == 'Windows' && runner.arch == 'ARM64')
       uses: mozilla-actions/sccache-action@v0.0.9
       with:
         version: ${{ env.SCCACHE_VERSION }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-latest, windows-latest, ubuntu-24.04-arm, windows-11-arm]
     steps:
       # prep
       - name: Checkout
@@ -50,7 +50,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-latest, windows-latest, ubuntu-24.04-arm, windows-11-arm]
     steps:
       # prep
       - name: Checkout
@@ -157,13 +157,22 @@ jobs:
           delete: true
 
   extended-analysis:
+    name: extended-analysis (${{ matrix.os }})
     if: github.event_name == 'pull_request'
     needs: [testing, static-analysis]
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        include:
+          - os: ubuntu-latest
+            target: x86_64-unknown-linux-gnu
+          - os: windows-latest
+            target: x86_64-pc-windows-msvc
+          - os: ubuntu-24.04-arm
+            target: aarch64-unknown-linux-gnu
+          - os: windows-11-arm
+            target: aarch64-pc-windows-msvc
     steps:
       # prep
       - name: Checkout
@@ -181,7 +190,7 @@ jobs:
         run: cargo +${{ env.RUST_NIGHTLY }} udeps --all-features --workspace --color always
       - name: Careful
         if: success() || failure()
-        run: cargo +${{ env.RUST_NIGHTLY }} careful nextest run --all-features --workspace --color always --target ${{ runner.os == 'Linux' && 'x86_64-unknown-linux-gnu' || 'x86_64-pc-windows-msvc' }}
+        run: cargo +${{ env.RUST_NIGHTLY }} careful nextest run --all-features --workspace --color always --target ${{ matrix.target }}
       - name: Miri
         if: success() || failure()
         run: cargo +${{ env.RUST_NIGHTLY }} miri test --all-features --workspace
@@ -216,7 +225,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        # Note: windows-11-arm excluded due to LLVM coverage instrumentation bugs
+        # causing "malformed instrumentation profile data: symbol name is empty" errors
+        os: [ubuntu-latest, windows-latest, ubuntu-24.04-arm]
     steps:
       # prep
       - name: Checkout


geeknoid (commented):
The logo.png file isn't rendering for me. Is this a valid png file? I tried even going to your repo & branch, and GitHub won't show the image to me.

schgoo:
@geeknoid I hadn't uploaded the logo yet. I just added it

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,404 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// Based on singleflight-async by ihciah
+// Original: https://github.com/ihciah/singleflight-async
+// Licensed under MIT/Apache-2.0
+
+//! Coalesces duplicate async tasks into a single execution.
```
ralfbiedert, line 4: How about 

>  Deduplicates async tasks into a single execution.

crates/uniflight/examples/cache_population.rs
```diff
@@ -0,0 +1,67 @@
...
+    },
+    time::Duration,
+};
+
+use uniflight::UniFlight;
+
+#[tokio::main]
+async fn main() {
+    // Create a shared UniFlight instance for cache operations
+    let cache_group = Arc::new(UniFlight::<String, String>::new());
```
ralfbiedert: This example is a bit weird (and / or the `UniFlight` definition)

I assume `UniFlight` is meant to be used by multiple tasks, so the type itself should probably contain an inner `Arc`. Looking inside of that type, there is an Arc already, so why wouldn't that be sufficient to get the cloning / sharing of `cache_group` going?

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+            let (call, shared) = BroadcastOnce::new();
+            mapping.insert(key.clone(), call);
+            let waiter = BroadcastOnce::waiter(shared, func, key, owned_mapping);
+            waiter.wait()
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
ralfbiedert, line 488: Can you move most tests out into a `tests/` folder, except the ones which strictly need to access implementation details?

crates/uniflight/examples/cache_population.rs
```diff
@@ -0,0 +1,67 @@
...
+                .await;
+
+            let elapsed = start.elapsed();
+            println!("  [Request {i}] Got result in {elapsed:?}: {result}");
+        });
+
+        handles.push(handle);
+
+        // Stagger the requests slightly to see the deduplication in action
+        tokio::time::sleep(Duration::from_millis(10)).await;
```
martintmk: Maybe you could use the `tick` crate for delays once it's available?

https://github.com/microsoft/oxidizer/pull/106

schgoo: I'm happy to make that change once it gets checked in

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+
+use parking_lot::Mutex as SyncMutex;
+use xutex::AsyncMutex;
+
+type SharedMapping<K, T> = Arc<SyncMutex<HashMap<K, BroadcastOnce<T>>>>;
+
+/// Represents a class of work and creates a space in which units of work
+/// can be executed with duplicate suppression.
+#[derive(Debug)]
+pub struct UniFlight<K, T> {
```
ralfbiedert: I'd love for this to impl `ThreadAware` and investigate how the resulting item should be emitted / shared. 

Naively one could argue the resulting item should be "unique" and shared between threads, but I think in some situations we still want "per thread" semantics being honored, but my feeling is also this isn't quite as obvious for `UniFlight` to implement, and might tell us something about the design of `ThreadAware` itself.

crates/uniflight/examples/cache_population.rs
ralfbiedert: On second thought, maybe this is a simple as using `thread_aware::Arc` and letting your worker / key pick a `Strategy`.

crates/uniflight/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+homepage.workspace = true
+repository.workspace = true
+
+[dependencies]
+parking_lot = { workspace = true, default-features = false }
+xutex = { workspace = true, default-features = false }
+
+[dev-dependencies]
+futures-util = { workspace = true, default-features = false, features = ["alloc", "std"] }
+tokio = { workspace = true, features = ["macros", "rt", "time", "rt-multi-thread"] }
```
geeknoid:
```suggestion
parking_lot.workspace = true
xutex.workspace = true

[dev-dependencies]
futures-util = { workspace = true, features = ["alloc", "std"] }
tokio = { workspace = true, features = ["macros", "rt", "time", "rt-multi-thread"] }
```

crates/uniflight/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+[package]
+name = "uniflight"
+description = "Coalesces multiple ongoing tasks into a leader which does the work, and follower tasks that wait on the result, to prevent duplicate I/O or other downstream overhead."
+version = "0.1.0"
+readme = "README.md"
+keywords = ["oxidizer", "singleflight"]
+categories = ["Concurrency"]
```
geeknoid:
```suggestion
keywords = ["oxidizer", "coalescing", "stempede", "singleflight", "deduplication"]
categories = ["concurrency"]
```

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,428 @@
...
+
+        self.mapping.lock().remove(&self.key);
+
+        value
+    }
+}
+
+impl<K, T> UniFlight<K, T>
+where
+    K: Hash + Eq + Clone,
```
geeknoid: Why does the key need to implement Clone?

In general, you want to minimize the type bounds on methods as much as you can, since it makes composition better on your consumers. As such, I'd move this type constraint to the work function instead, so as to remove the constraint from the new function.

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,428 @@
...
+    /// operation is in-flight at a given moment. If a duplicate call comes in, that caller will
+    /// wait until the original call completes and return the same value.
+    pub fn work<F, Fut>(&self, key: K, func: F) -> impl Future<Output = T>
+    where
+        F: FnOnce() -> Fut,
+        Fut: Future<Output = T>,
+        T: Clone,
+    {
+        let owned_mapping = Arc::clone(&self.mapping);
+        let mut mapping = self.mapping.lock();
```
geeknoid: Right now, this is a global lock which will cause contention as the # of cores increases. You can reduce contention by sharding the lock. For example, you could use a modulo of the hash value for the key to pick one of N locks. If you make the modulo equal to the number of processors, then you get more and more locks the more processors there are, such that contention should remain generally flat as the system scales up.

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,428 @@
...
+    #[inline]
+    #[must_use]
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Execute and return the value for a given function, making sure that only one
+    /// operation is in-flight at a given moment. If a duplicate call comes in, that caller will
+    /// wait until the original call completes and return the same value.
+    pub fn work<F, Fut>(&self, key: K, func: F) -> impl Future<Output = T>
```
geeknoid: I think the key should be &K here, you don't want the key to be consumed by this API.

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,428 @@
...
+    where
+        F: FnOnce() -> Fut,
```
geeknoid: I worry that this logic is susceptible to unexpected side effects. If the closure depends on external state, it could lead to surprising/incorrect behavior. For example:

```rust
let thread_id = get_thread_id();
let result = group.work("user:123", || async {
     thread_id.to_string()
 }).await;
```

This is a racy pattern. Depending on which threads becomes the leader, the result of the closure will be different. I'm making this obvious in this example, but it's a general problem. It's basically inviting concurrency issues to surface.

To get this pattern working reliably, all state going into the closure needs to be evaluated as part of deciding "is this the same work". If you only look at the key and ignore all the other state, you can have false sharing. So I think we can't allow any state to enter the closure "on the side", everything needs to be explicitly passed to the closure and evaluated for "sameness".

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,428 @@
...
+                Err(fm) => fm,
+            };
+            let (new_call, shared) = BroadcastOnce::new();
+            *call = new_call;
+            let waiter = BroadcastOnce::waiter(shared, func, key, owned_mapping);
+            waiter.wait()
+        } else {
+            let (call, shared) = BroadcastOnce::new();
+            mapping.insert(key.clone(), call);
+            let waiter = BroadcastOnce::waiter(shared, func, key, owned_mapping);
```
geeknoid: Have the if and else branches return the waiter object, then call waiter.wait() outside the if construct. This avoids a bit of code duplication.

crates/uniflight/src/lib.rs
schgoo: I've been wondering about that, because it's kind of a strange user experience with intellisense to try and find the `work` function, but it doesn't appear because your bounds are incorrect. I don't have a strong opinion on it, though.

crates/uniflight/src/lib.rs
schgoo: Isn't that user error though? Is it the responsibility of this crate to prevent them from causing themselves this issue?

crates/uniflight/src/lib.rs
schgoo: Especially given the fact that `UniFlight` is completely useless if `work` can't be called - does it really make sense to separate the bounds here? That said, the reason it needs clone is that HashMap takes ownership of it, but it still needs to be passed into the BroadcastOnceWaiter, so the waiter can remove it from the HashMap when the task completes.

crates/uniflight/src/lib.rs
geeknoid: Yes it is a user error. However, we don't want to expose a programming model which puts people in the pit of failure by default.

Rust is great in that it's eliminate race conditions by construction. The pattern that this code exposes encourage the creation of races. So how can we tighten things up to either completely prevent issues, or at least make it unlikely to hit issues.

Not allowing state to be captured and forcing all state to be basically built into the key is one way to achieve this protection. I think this would translate in saying that this code doesn't take a closure, but it takes a raw function pointer. So if you want to pass something into this code, it will need to be in the key. And its in the key, then it'll need to participate in the Hash and Eq trait implementations. And this will then ensure that coalescing only happens when it really should.

Maybe there are other approaches too...

crates/uniflight/src/lib.rs
geeknoid: It's really not great that the key is passed by value and is internally cloned. That means the caller will clone the key before calling the function, and then internally it will get cloned again. If the parameter is passed by reference, at least the caller won't need to clone before calling you.

Regarding the bounds, you're right that you can't create a struct where the key is not hashable, so what's the benefit of having new not having bounds. Why put the bounds on the work function instead? It simply composes better at a higher-level. It can avoid your users needed to explicitly specify types when interacting with your stuff. So if you look around at the ecosystem, you can see that well designed APIs tend to minimize the type bounds on methods in general. You can have different impl blocks with different constraints and then bunch several methoss with the same constraints in those impl blocks. Or you can apply the constraints directly on the API.

Basically, if you don't need the constraint to support your implementation of the method, you shouldn't require that constraint on your callers.

You know, the same argument can be made to apply the constraints on the struct definition itself. You could declare that the Uniflight's K parameter always be hashable. But since you don't need that to define your struct, you don't apply the constraint there and only on the methods.

crates/uniflight/src/lib.rs
ralfbiedert: > Not allowing state to be captured and forcing all state to be basically built into the key is one way to achieve this protection. 

I don't think that's possible in the strictest sense, as you can still (accidentally) bypass that with statics and thread locals, like 

```rust
let result = group.work("user:123", || async { // 'pure' fn-ptr
     logger::last_message()
 }).await;
```

might set a different value depending on which thread executed that.

crates/uniflight/src/lib.rs
schgoo: You really hate that word, huh? 😆

crates/uniflight/src/lib.rs
geeknoid: Yeah, I realize it's not air-tight, but I feel it is substantially reducing the risk of accidental mismatch. Capturing state in a closure is the most likely source of these problems.

This is very similar to classic makefiles and dependencies. If you don't capture all the dependencies in a makefile properly, then your incremental builds will mostly work. Until one day, they don't work and you have to do a "make clean" and start from scratch.

crates/uniflight/src/lib.rs
geeknoid: And use the same tagline in Cargo.toml's 'description' field, and in the top-level README's list of crates.

crates/uniflight/src/lib.rs
schgoo: Then we'll _definitely_ need K to be Clone

crates/uniflight/examples/cache_population.rs
```diff
@@ -0,0 +1,69 @@
...
+use std::{
+    sync::{
+        Arc,
+        atomic::{AtomicUsize, Ordering},
+    },
+    time::Duration,
+};
+
+use tick::Clock;
+use uniflight::UniFlight;
```
sandersaares: I like the name `uniflight` for the crate but for the actual type it might be better to have more "down to business" names so reading the code it becomes clear what it is. Maybe `uniflight::Coalesced` to represent "a coalesced operation"?

crates/uniflight/examples/cache_population.rs
```diff
@@ -0,0 +1,69 @@
...
+
+#[tokio::main]
+async fn main() {
```
sandersaares, line 21: This whole thing seems like [`Once`](https://doc.rust-lang.org/std/sync/struct.Once.html) [or OnceLock/OnceCell] with an added layer of keying, and async support. Is this right?

Perhaps this suggests that the design could be decomposed into something like a thread-aware `AsyncOnce` and a suitably ergonomic thread-aware map?

```rust
let cache_groups = ConcurrentMap::<String, AsyncOnce<String>>::new();

let value = cache_groups
    .entry("key123")
    .or_insert(|| AsyncOnce::new((http_client), |(http_client)| load_item(http_client)))
    .await?;
```

crates/uniflight/examples/cache_population.rs
schgoo: Unfortunately, @ralfbiedert is _very_ resistant to the word "coalesce", but I agree with the essence of what you're saying here.

crates/uniflight/examples/cache_population.rs
schgoo: Interestingly enough, I was working on an update that removed the per-key async mutex and replaced it with OnceLock and EventListener. I'll explore the idea of this decomposition, but I don't believe we yet have anything like a thread-aware concurrent map, right? I'm also not sure about the value of implementing a thread-aware `AsyncOnce`. What would be the desired behavior upon relocation?

crates/uniflight/src/lib.rs
schgoo: So, I've been thinking about this a while. I think I agree that important context should be in the key, and then we can pass that into `work` like

```Rust
pub fn work<Fut>(&self, key: K, f: fn(K) -> Fut) -> impl Future<Output = T>
where
    Fut: Future<Output = T>,
{ ... }
```

The problem is that if they're using this, they probably need some downstream client that needs to do I/O (for instance, a `logger`, as @ralfbiedert pointed out). In that case, we would need to capture that client in the key. Is it reasonable to require that their client/logger/etc. implements Hash+Eq? That seems a bit strange to me.

Maybe we could capture some additional context when `Uniflight` is initialized, and then pass that into the function pointer, as well, like:

```Rust
  impl<K, T, C> UniFlight<K, T, C> {
      pub fn with_context<C2>(self, ctx: C2) -> UniFlight<K, T, C2> { ... }

      pub fn work<Fut>(
          &self,
          key: K,
          f: fn(&C, K) -> Fut,  // Receives context AND key
      ) -> impl Future<Output = T>
      where
          Fut: Future<Output = T>,
      { ... }
  }
```

Would we need to box the future to maintain the reference to C?

**edit**
I did some prototyping, and yes, we would need to box the future because a function pointer needs a concrete return type, it can't return `impl Future`. So that would look something like:

```Rust
async fn work(&self, key: K, f: fn(&C, K) -> BoxFuture<'_, T>) -> T {
    f(&self.context, key).await
}
```

Is the cost to allocate a `BoxFuture` for every leader worth it for this design? Users would also have to pin their futures within the function pointer, which is a bit strange:

```Rust
fn fetch_user(client: &Client, user_id: UserId) -> BoxFuture<'_, String> {
    Box::pin(async move {
        format!("Fetched user {} from {}", user_id.0, client.base_url)
    })
}
```

crates/uniflight/examples/cache_population.rs
schgoo: This got me thinking, though, maybe I should just use an existing `AsyncOnce`? I wasn't aware of this pattern before, but there are some widely used components out there already (for example, [async-once-cell](https://crates.io/crates/async-once-cell))

sandersaares: > I don't believe we yet have anything like a thread-aware concurrent map, right

It has been on the wishlist for a long time but indeed, we do not have it today. Perhaps that just means the opportunity is ripe!

> I'm also not sure about the value of implementing a thread-aware AsyncOnce. What would be the desired behavior upon relocation?

The way I was thinking about this was that this closure given to `or_insert()` only ever gets executed once (approximately - eventual consistency in the collection might conceivably result in temporary copies that get dropped). Therefore there is only one `AsyncOnce` created and only one `http_client` that ever gets used per key. The leader may change but the closure and the HTTP client remain the same.

So if the leader polls it, the `http_client` (and any other state) need to be on the thread of the leader for efficient processing. This suggests to me that when a leader is (re)assigned, a relocation to the leader's thread needs to happen.

I do not have a thorough design mapped out in my head, so take this with a grain of salt.

crates/uniflight/examples/cache_population.rs
martintmk: https://docs.rs/tick is now published :)

crates/uniflight/examples/cache_population.rs
schgoo: > This suggests to me that when a leader is (re)assigned, a relocation to the leader's thread needs to happen.

A relocation of the `AsyncOnce` to the leader's thread? What about the `http_client`? I'm struggling to wrap my brain around thread-awareness.

crates/uniflight/src/lib.rs
geeknoid: Ah, I think this is going off the rails :-) Let's just yolo this and move on with the original design.

crates/uniflight/examples/cache_population.rs
geeknoid: Converged, combined, joined, single_instanced

crates/uniflight/examples/cache_population.rs
geeknoid: If you assume each thread is homed to a NUMA node (it's stack is allocated in that NUMA node and allocating memory from that thread allocates on that NUMA node), then thread-awareness is essentially about "keeping the state that a thread uses in that thread's NUMA node", in order to avoid cross-node access.

So if there's some state that will be accessed repeatedly over time, moving it to the thread's NUMA node can have a considerable performance benefit.

schgoo: Ok, I've decided to use `async_once_cell` and `dashmap` here to improve simplicity and performance. For thread awareness, I just wrap the dashmap in a thread aware Arc. The idea is that a relocated merger will just create a new, empty dashmap. It might be worth some more effort in the future to move leader task results over with it, but I'm not sure how much value that offers, in practice.

crates/uniflight/src/lib.rs
schgoo: I switched it to borrow semantics on `work`

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,495 @@
...
+//!
+//! - **Cache population**: Prevent thundering herd when a cache entry expires
+//! - **API calls**: Deduplicate concurrent requests to the same endpoint
+//! - **Database queries**: Coalesce identical queries issued simultaneously
+//! - **File I/O**: Avoid reading the same file multiple times concurrently
+//!
+//! # Example
+//!
+//! ```
+//! use uniflight::Merger;
```
geeknoid, line 24: Hmmm. You're calling this Merger now. Have you hear of M&A? That's Merger-and-Acquisition. Maybe that would be the crate name? m_and_a? Nah....

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,495 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// Based on singleflight-async by ihciah
+// Original: https://github.com/ihciah/singleflight-async
```
geeknoid: How close are we to this design at this point? Is there still significant lineage there? In other words, is this attribution still meaningful?

crates/uniflight/src/lib.rs
geeknoid: So is this really what is happening? That is "coalesce duplicate async tasks"? Do we even need to talk about tasks at all?

A simple framework designed to avoid redundant asynchronous work within a process.

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,495 @@
...
+//!
+//! ```
+//! use uniflight::Merger;
+//!
+//! # async fn example() {
+//! let group: Merger<String, String> = Merger::new();
+//!
+//! // Multiple concurrent calls with the same key will share a single execution.
+//! // Note: you can pass &str directly when the key type is String.
+//! let result = group.work("user:123", || async {
```
geeknoid: What would you think about calling this function "execute", and the variable "merger", giving you the lovely:

let result = merger.execute(...)

crates/uniflight/src/lib.rs
geeknoid: I guess I'm saying I don't like "uniflight" as a name. "flight" has the wrong overly constrained connotations I think. Given that your main theme is "merging" then maybe a merge-centric name would be better.

merge-o-matic

crates/uniflight/src/lib.rs
schgoo: Hmm, yeah probably not

crates/uniflight/src/lib.rs
schgoo: lol that would be funny, but we're not acquiring anything, unfortunately

crates/uniflight/src/lib.rs
schgoo: fine with me

README.md
```diff
@@ -11,8 +11,9 @@
 
 This repository contains a set of crates that help you build robust highly scalable services in Rust.
 
-- [Crates](#crates)
-- [About this Repo](#about-this-repo)
+- [The Oxidizer Project](#the-oxidizer-project)
```
geeknoid: Can we undo this change?

crates/uniflight/benches/comparison.rs
ralfbiedert, line 1: I'm all for benchmarks, when they track a long-term performance goal or guard against regressions. I have the suspicion though these here are mostly artifacts that came from development? Also, they seem very boilerplatey.

schgoo: I wrote these because I needed a basis for comparison for the performance of this crate (in this case, singleflight-async). I can see why that could be considered a development artifact. Is the logic that, now that we've established our own baseline, we shouldn't need the comparison any longer?

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,491 @@
...
+use thread_aware::{
+    Arc as TaArc, ThreadAware,
+    affinity::{MemoryAffinity, PinnedAffinity},
+    storage::Strategy,
+};
+
+// Re-export strategies for convenience
+pub use thread_aware::{PerCore, PerNuma, PerProcess};
+
+/// Represents a class of work and creates a space in which units of work
```
ralfbiedert: Just reading this, I don't get what it wants to say, it feels a bit convoluted. It also causes `M-FIRST-DOC-SENTENCE` issues.

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,491 @@
...
+    Entry::{Occupied, Vacant},
+};
+use thread_aware::{
+    Arc as TaArc, ThreadAware,
+    affinity::{MemoryAffinity, PinnedAffinity},
+    storage::Strategy,
+};
+
+// Re-export strategies for convenience
+pub use thread_aware::{PerCore, PerNuma, PerProcess};
```
ralfbiedert: I'm not sure these should be re-exported. Having `Merger` use a default `PerProcess` strategy is ok, but I'm inclined to say if people want to change that they should `use thread_aware`.

crates/uniflight/src/lib.rs
schgoo: How about, "Suppresses duplicate execution of hash key-associated work"

schgoo: Going with "Suppresses duplicate async operations identified by a key"

crates/uniflight/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+edition.workspace = true
+rust-version.workspace = true
+authors.workspace = true
+license.workspace = true
+homepage.workspace = true
+repository.workspace = true
+
+[package.metadata.cargo_check_external_types]
+allowed_external_types = [
+    "thread_aware::cell::builtin::PerProcess",
```
geeknoid, line 21: Why do we need to reexport these?

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,483 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Coalesces duplicate async tasks into a single execution.
```
geeknoid, line 4: I suggest using this line as the description in Cargo.toml and in the repo-level README file.

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,483 @@
...
+//! let merger: Merger<String, String, PerNuma> = Merger::new_per_numa();
+//! # }
+//! ```
+//!
+//! # Cancellation and Panic Safety
+//!
+//! `Merger` handles task cancellation and panics gracefully:
+//!
+//! - If the leader task is cancelled or dropped, a follower becomes the new leader
+//! - If the leader task panics, a follower becomes the new leader and executes its work
```
geeknoid: This feels wrong. Shouldn't the panic propagate to all followers?

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,483 @@
...
+//! - Followers that join before the leader completes receive the cached result
```
geeknoid: "receive the cached result" confuses me. Can we just say "receive the value the leader returns"?

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,483 @@
...
+//!
+//! # Memory Management
+//!
+//! Completed entries are automatically removed from the internal map when the last caller
```
geeknoid, line 101: Hmmm. Seems like having a TTL here might be desirable? Keep an entry around for N msec in case someone asks for the result? This can turn into a simple short-lived cache.

The cargo-rank app I'm working on issues sometimes thousands of requests over the course of a few minutes as it runs. I am planning to adopt this crate once it's available. But if I want to truly be efficient, I'm going to need to implement a cache in front of uniflight to avoid issuing the same request multiple times over the course of those minutes. if I could tell uniflight to hold on to the keys & values for a bit, then it could totally avoid me doing the work N times.

But this leads to Hmmm #2. Is uniflight just a cache with an oddball TTL? Instead of holding on to items for a specific time period, it throws out cached data is nobody is actively waiting for it. Why should I use uniflight instead of a cache of tasks? I can achieve the same semantics by setting a TTL of 0, and get more flexible semantics if I increase the TTL.

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,483 @@
...
+
+impl<K, T, S> Default for Merger<K, T, S>
+where
+    K: Hash + Eq + Send + Sync + 'static,
+    T: Clone + Send + Sync + 'static,
+    S: Strategy + Send + Sync,
+{
+    fn default() -> Self {
+        Self {
+            inner: TaArc::new(DashMap::new),
```
geeknoid: You probably should override the hasher, or allow the caller to override the hasher. The default hasher used by DashMap is the default system hasher which is really slow. Consider using ahash instead, or one of the other many fast hashers.

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,483 @@
...
+    fn clone(&self) -> Self {
+        Self { inner: self.inner.clone() }
+    }
+}
+
+impl<K, T, S> Default for Merger<K, T, S>
+where
+    K: Hash + Eq + Send + Sync + 'static,
+    T: Clone + Send + Sync + 'static,
+    S: Strategy + Send + Sync,
```
geeknoid:
```suggestion
    S: Strategy,
```

geeknoid: Probably applies in other places too.

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,483 @@
...
+impl<K, T, S: Strategy> Clone for Merger<K, T, S> {
+    fn clone(&self) -> Self {
+        Self { inner: self.inner.clone() }
+    }
+}
+
+impl<K, T, S> Default for Merger<K, T, S>
+where
+    K: Hash + Eq + Send + Sync + 'static,
+    T: Clone + Send + Sync + 'static,
```
geeknoid:
```suggestion
    T: Send + Sync + 'static,
```

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,483 @@
...
+        Self {
+            inner: TaArc::new(DashMap::new),
+        }
+    }
+}
+
+impl<K, T, S> Merger<K, T, S>
+where
+    K: Hash + Eq + Send + Sync + 'static,
+    T: Clone + Send + Sync + 'static,
```
geeknoid:
```suggestion
    T: Send + Sync + 'static,
```

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,483 @@
...
+    S: Strategy + Send + Sync,
```
geeknoid:
```suggestion
    S: Send + Sync,
```

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,483 @@
...
+    #[must_use]
+    pub fn new() -> Self {
+        Self::default()
+    }
+}
+
+impl<K, T> Merger<K, T, PerProcess>
+where
+    K: Hash + Eq + Send + Sync + 'static,
+    T: Clone + Send + Sync + 'static,
```
geeknoid:
```suggestion
    T: Send + Sync + 'static,
```

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,483 @@
...
+    #[cfg_attr(test, mutants::skip)] // Equivalent mutant: delegates to Default
+    pub fn new_per_process() -> Self {
+        Self::default()
+    }
+}
+
+impl<K, T> Merger<K, T, PerNuma>
+where
+    K: Hash + Eq + Send + Sync + 'static,
+    T: Clone + Send + Sync + 'static,
```
geeknoid:
```suggestion
    T: Send + Sync + 'static,
```

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,483 @@
...
+    #[cfg_attr(test, mutants::skip)] // Equivalent mutant: delegates to Default
+    pub fn new_per_numa() -> Self {
+        Self::default()
+    }
+}
+
+impl<K, T> Merger<K, T, PerCore>
+where
+    K: Hash + Eq + Send + Sync + 'static,
+    T: Clone + Send + Sync + 'static,
```
geeknoid:
```suggestion
    T: Send + Sync + 'static,
```

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,483 @@
...
+    #[must_use]
+    #[cfg_attr(test, mutants::skip)] // Equivalent mutant: delegates to Default
+    pub fn new_per_core() -> Self {
+        Self::default()
+    }
+}
+
+impl<K, T, S: Strategy> Merger<K, T, S>
+where
+    K: Hash + Eq,
```
geeknoid, line 295: Remove this trait bound, it's not needed.

crates/uniflight/src/lib.rs
```diff
@@ -0,0 +1,483 @@
...
+        Self {
+            inner: self.inner.relocated(source, destination),
+        }
+    }
+}
+
+impl<K, T, S> Merger<K, T, S>
+where
+    K: Hash + Eq + Send + Sync,
+    T: Clone + Send + Sync,
```
geeknoid: Remove Clone from here, and put it as a bound on the execute method instead.

crates/uniflight/Cargo.toml
schgoo: I'm not re-exporting them, they're part of the public API. `Strategy` is a generic parameter on `Merger`, `PerProcess` is the default.

crates/uniflight/src/lib.rs
schgoo: I guess it depends on whether the panic is transient? Maybe we make this configurable?

crates/uniflight/src/lib.rs
schgoo: Well, since this is already going to be part of the stampede protection implementation for the cache crate, why don't you just use the cache crate to store your result. There's no reason to keep the task itself around, you just want the result.

crates/uniflight/src/lib.rs
schgoo: Ok I'll look into it

crates/uniflight/src/lib.rs
geeknoid: Hmmm. Maybe we need to figure out how this crate works alongside seatbelt.

If you assume that all network communication is fallible, and you use this crate to coalesce multiple calls, then what should happen when the call fails? I think the right model is to have the leader be the one running the resilience stuff (so doing the retries). So by the time the followers get involved, the request has either succeeded or failed. It's no longer a retriable operation at this point, it's already been retried and deemed kaput.

But this particular case here is about panic. If the leader crashes, how should the followers react? Should they treat the panic as "operation failed" or "operation failed and I should retry" or "I'm panicking because my leader panicked".

I'm generally inclined towards a fail-fast approach to things. If the leader panicked, then the app probably should restart, and hence I'd like the panic to propagate.

@ralfbiedert What do you think our model should be? I think we want the panic propagation story to be consistent across abstractions (like with task spawning & awaiting).

crates/uniflight/src/lib.rs
geeknoid: OK, let's play with this. If I have a cache around this thing, how do I compose the stuff together?

```
result = check_cache();
if (not in cache) {
    result = issue_coalesced_request();
    put_result_in_cache();
}
```

So there's a window there where the result has been collected but not yet added to the cache. Can we close this window and prevent concurrent requests?

schgoo: Well, the way it would work with the cache, is the cache defines the method that is passed into the `Merger`. So basically:

``` rust
pub struct Cache {
  pub async fn get(&self, key: &K) -> Option<CacheEntry<V>>
  {
    // some other stuff
    return merger.execute(key, || async {
      // call get on storage
      // promote to cache, if configured
    }).await;
  }
}
```

We are in control of the item being added to cache within the leader task. In this case, we could get duplicate work if the leader collected the result but failed to promote the result to the cache.

crates/uniflight/src/lib.rs
geeknoid: Thinking about it, it's became more clear in my mind.

The async operation has three possible outcomes:

* Leader succeeded. Yey.
* Leader failed. Too bad, so sad.
* Leader crashed/panicked. Ouch.

All three of these states need to be reflected to the followers. When you spawn tasks with tokio, you get a result at the top-level that indicates whether or not the task panicked. If it didn't panic, then you get to the result of the async operation.

So I think we want a similar pattern here.

If the callers want the semantic that a follower takes over when a leader crashes, they can do that on their own by invoking the async operation again once they detect a leader has crashed.

crates/uniflight/src/lib.rs
geeknoid: All righty, works for me.

Maybe worth adding as an example?

crates/uniflight/src/lib.rs
schgoo: So something like this?

``` Rust
  #[derive(Debug, Clone)]
  pub struct LeaderPanicked {                                                                                                            
      pub message: Option<String>,       
  }

  impl<K, T, S> Merger<K, T, S> {
      pub fn execute<Q, F, Fut>(
          &self,
          key: &Q,
          func: F
      ) -> impl Future<Output = Result<T, LeaderPanicked>>
      where
          // ... existing bounds ...
  }
```

So, we'd have:
* Leader succeeded: Leader and all followers return Ok(T)
* Leader cancelled or dropped: Follower promoted to leader, leader is dropped so no return value
* Leader failed (i.e. T has type Result<X, SomeError>): Leader and all followers return Result<Err(SomeError)>
* Leader crashed/panicked: Leader and all followers return Err(LeaderPanicked)

Would we ever want follower promotion other than leader cancellation/drop?

crates/uniflight/src/lib.rs
schgoo: I assume here you meant remove Send + Sync

crates/uniflight/src/lib.rs
schgoo: These are needed here because the methods here call `DashMap::len` and `DashMap::is_empty`, and `DashMap` requires `K: Hash + Eq`

crates/uniflight/src/lib.rs
geeknoid: I think that feels right. No lies, no silently demoted failure mode.

In terms of other cases where promotion makes sense, I can't think of any in the real world. In theory, if the different participants have some sort of different priority, then you would probably want to promote the highest priority participant into being a leader such that you don't suffer priority inversion. But that's getting too complicated IMO.

crates/uniflight/src/lib.rs
geeknoid: I'm pretty sure all the places I suggested you can remove trait bounds are things I explicitly tried myself and built locally...

schgoo: Huh, I don't know, it didn't build when I tried to remove these ones.

geeknoid (approved):

PR #132, opened 2025-12-15
feat: continued bytesbuf tidying

sandersaares:
Removes the requirement to go through the foreign `bytes::Buf` and `bytes::BufMut` traits when performing reads/writes on `BytesBuf`/`BytesView`. These traits are still implemented but now there are also equivalent inherent methods, sometimes with better signatures, as we are no longer trying to maintain parity with cumbersome `bytes` API signatures.

The inherent read/write methods are separated out to `buf_put.rs` and `view_get.rs` to create some structure as these code files are getting cumbersome to navigate. The idea is that the "elementary" operations are in `buf.rs` and `view.rs` and higher-level helpers (that do not generally need to access type internals) are in separate files.

Adds benchmarks for a selection of the new inherent methods used to read/write bytes. Results indicate equivalent performance compared to the methods in `bytes`. Additional benchmarks planned in the future: https://github.com/microsoft/oxidizer/issues/151

In a later iteration, the `bytes` compatibility traits will move behind a feature flag - not fully rooted out yet from all the places.

Removes usage of `bytes::Buf` and `bytes::BufMut` from examples, relying on inherent `bytesbuf` methods only.

Improved panic documentation and assertion logic with regard to some obscure cases like exceeding `usize::MAX`.

Updates API documentation with new and improved terminology. One step of many - further iterations for API docs improvement will follow.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -155,11 +155,8 @@ jobs:
 
   semver:
     needs: constants
-    runs-on: ${{ matrix.os }}
-    strategy:
-      fail-fast: false
-      matrix:
-        os: [ubuntu-latest, windows-latest]
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
     env:
       RUST_MSRV: ${{ needs.constants.outputs.RUST_MSRV }}
       RUST_LATEST: ${{ needs.constants.outputs.RUST_LATEST }}
@@ -174,7 +171,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v6.0.0
         with:
-          fetch-depth: 2
+          fetch-depth: 0
       - name: Cache Cargo Dependencies
         uses: actions/cache@v4.3.0
         with:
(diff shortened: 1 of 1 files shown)


crates/bytesbuf/src/view_get.rs
```diff
@@ -0,0 +1,519 @@
...
+    ///
+    /// Panics if the view does not cover enough bytes of data.
+    #[inline]
+    #[must_use]
+    pub fn get_byte(&mut self) -> u8 {
+        assert!(!self.is_empty());
+
+        // SAFETY: We asserted above that the view covers at least one byte,
+        // so the first slice must also have at least one byte.
+        let byte = unsafe { *self.first_slice().get_unchecked(0) };
```
Vaiz: looks the same as 
`let byte = self.first_slice().first().unwrap()`

but without unsafe block

crates/bytesbuf/src/view_get.rs
```diff
@@ -0,0 +1,519 @@
...
+        byte
+    }
+
+    /// Transfers bytes into an initialized slice.
+    ///
+    /// The bytes are removed from the front of the view, shrinking it.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the destination is larger than the view.
```
Vaiz, line 35: why does it panic if the destination is larger than the view?

crates/bytesbuf/src/view_get.rs
sandersaares: Almost. The code as written results in slightly less logic to execute since there is no need to do the `unwrap()` as we rely on the assertion above already making `None` impossible. (For performance)

crates/bytesbuf/src/view_get.rs
sandersaares: Ideally, we would want to enforce 1:1 match - caller has to slice both source and destination equally to ensure right range. This is what `slice::copy_to_slice()` does.

However, because the view is meant to be progressively consumed, this would lead to silly code:

```rust
let view = get_my_data();
let mut four_bytes = [0u8; 4];

view.slice(..4).copy_to_slice(&mut four_bytes);
```

What happened? Well, `.slice(..4)` returned a new temporary `BytesView`. Even if we consume 4 bytes off the start of that, the original `view` is not affected. This would be an easy landmine to step on.

So we need to relax the requirement and not require slicing the source to match the right size. In most cases, the source will be some large byte sequence and the destination is just some small subset being extracted, so we should accept `source > destination`.

We still preserve the invariant that you can be sure your entire destination slice was filled, which is useful.

This matches what `bytes` crate APIs do.

crates/bytesbuf/src/view_get.rs
Vaiz: but this is what I mean, it's better to remove assert and UNSAFE block, if there is a safe alternative

sandersaares: I do not agree that the existence of a safe alternative makes it automatically preferable. Reading/writing to/from byte sequences is performance-critical code, every nanosecond counts here.

There is a balance to aim for, of course, but the correctness of this code here seems trivial to verify by review (and by Miri), so the tradeoff feels very much in favor of efficiency in my mind.

Am I thinking wrong?

Vaiz: from my understanding, any unsafe code that claims to be faster should be tested against safe code

and because we already have an assert, I don't think this code in any way is faster

sandersaares: OK, I can buy that. Will add benchmark.

sandersaares: Benchmark proved you are correct - the performance is the same. I have applied your suggested change and removed unsafe code from this function.

geeknoid:
You know, the word "sequence" appears 254 in the bytesbuf source code...

sandersaares:
Yes, `BytesBuf`, `BytesView` and their sidekicks are types for working with byte sequences.

PR #133, opened 2025-12-16
feat: Better serialization and perf.

geeknoid:
- Improve serialization of the DataClass type. It's now serialized as a simple string rather than a clumsy map.

- Use serde_core instead of serde for improved build times.

- Change the Classified trait's data_class method to return a &DataClass rather than a DataClass.
This avoids the need to clone the DataClass in many places, improving performance (since DataClass has two Cow to clone).

- Use FxHashMap to hold the redactors to get much better hash lookup perf. We don't need DoS protection since the hash keys are entirely internal.

### DIFF
--- a/crates/data_privacy/src/data_class.rs
+++ b/crates/data_privacy/src/data_class.rs
@@ -4,26 +4,35 @@
 use core::fmt::Display;
 use std::borrow::Cow;
 
+#[cfg(feature = "serde")]
+use serde_core::{Deserialize, Deserializer, Serialize, Serializer, de};
+
 /// The identity of a well-known data class.
 ///
 /// Each data class has a name, which is unique in the context of a specific named taxonomy.
+///
+/// # Serialization
+///
+/// Serializing a `DataClass` produces a string in the format `taxonomy/name`.
 #[derive(Debug, Clone, Hash, Eq, PartialEq, Ord, PartialOrd)]
-#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 pub struct DataClass {
     taxonomy: Cow<'static, str>,
     name: Cow<'static, str>,
 }
 
-impl AsRef<Self> for DataClass {
-    fn as_ref(&self) -> &Self {
-        self
-    }
-}
-
 impl DataClass {
     /// Creates a new data class instance.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `taxonomy` or `name` are not valid ASCII identifiers. Valid identifiers must
+    /// start with `_` or an ASCII letter, followed by zero or more `_`, ASCII letters, or ASCII
+    /// digits (e.g., `foo`, `_bar`, `Baz123`)
     #[must_use]
     pub const fn new(taxonomy: &'static str, name: &'static str) -> Self {
+        assert!(is_valid_identifier(taxonomy), "taxonomy is not a valid identifier");
+        assert!(is_valid_identifier(name), "name is not a valid identifier");
+
         Self {
             taxonomy: Cow::Borrowed(taxonomy),
             name: Cow::Borrowed(name),
@@ -49,6 +58,12 @@ impl Display for DataClass {
     }
 }
 
+impl AsRef<Self> for DataClass {
+    fn as_ref(&self) -> &Self {
+        self
+    }
+}
+
 /// Helper for converting a type into a [`DataClass`].
 pub trait IntoDataClass {
     /// Converts `self` into a [`DataClass`].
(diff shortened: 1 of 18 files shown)


crates/data_privacy/src/data_class.rs
```diff
@@ -60,3 +62,97 @@ impl IntoDataClass for DataClass {
...
+    }
+}
+
+#[cfg(all(test, feature = "serde"))]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_serialize() {
+        let dc = DataClass::new("contoso", "customer_identifier");
```
martintmk, line 171:
```suggestion
        let dc = DataClass::new("contoso", "customer_identifier/ABC");
```

what happens in this case? Shall we restrict allowed characters?

geeknoid: Restrict allowed characters to what?

crates/data_privacy/src/redactors/mod.rs
```diff
@@ -4,7 +4,7 @@
 use crate::simple_redactor::{SimpleRedactor, SimpleRedactorMode};
 use crate::{DataClass, IntoDataClass};
 use core::fmt::Debug;
-use std::collections::HashMap;
+use rustc_hash::FxHashMap;
```
sandersaares, line 7: I seem to recall hearing good things about foldhash, being even better than fxhash supposedly. Why fxhash here?

crates/data_privacy/src/data_class.rs
```diff
@@ -60,3 +62,97 @@ impl IntoDataClass for DataClass {
...
+#[cfg(feature = "serde")]
+impl<'de> Deserialize<'de> for DataClass {
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        struct DataClassVisitor;
+
+        impl de::Visitor<'_> for DataClassVisitor {
+            type Value = DataClass;
+
+            fn expecting(&self, formatter: &mut core::fmt::Formatter) -> core::fmt::Result {
+                formatter.write_str("a string in taxonomy/name format")
```
sandersaares, line 95: Is the format part of the API contract? If so, should be documented.

crates/data_privacy/src/redactors/mod.rs
geeknoid: > I seem to recall hearing good things about foldhash, being even better than fxhash supposedly. Why fxhash here?

Cause the AI told me so :-)

crates/data_privacy/src/data_class.rs
martintmk: so far, you are only treating the `/` as a special one. So we can either restrict this one or just only allow alphanumeric characters for these idenitifiers (plus some special ones such as `.`, `_`, `-`.

The Otel naming for attributes might be a good reference:
https://opentelemetry.io/docs/specs/semconv/general/naming/

geeknoid: Done, there is now full validation.

crates/data_privacy/src/data_class.rs
geeknoid: Now documented.

crates/data_privacy/src/redactors/mod.rs
sandersaares: So no reason, okay :) 

I suggest we add some benchmark so we can verify the impact of this and verify that the choice of hash algorithm is optimal.

crates/data_privacy/src/redactors/simple_redactor.rs
```diff
@@ -67,7 +67,8 @@ impl SimpleRedactor {
 impl Redactor for SimpleRedactor {
     #[cfg_attr(test, mutants::skip)]
     fn redact(&self, data_class: &DataClass, value: &str, output: &mut dyn Write) -> core::fmt::Result {
-        static ASTERISKS: &str = "********************************";
+        const ASTERISKS: &str =
```
sandersaares, line 70: I would like to see some benchmarks about the performance impact of this.

crates/data_privacy/src/data_class.rs
```diff
@@ -60,3 +75,228 @@ impl IntoDataClass for DataClass {
...
+            }
+        }
+
+        deserializer.deserialize_str(DataClassVisitor)
+    }
+}
+
+/// Checks if a byte is a valid ASCII start character for a Rust identifier.
+const fn is_valid_ascii_ident_start(byte: u8) -> bool {
+    matches!(byte, b'a'..=b'z' | b'A'..=b'Z' | b'_')
```
martintmk: Could we use `is_ascii_alphanumeric` here?

crates/data_privacy/src/data_class.rs
```diff
@@ -60,3 +75,228 @@ impl IntoDataClass for DataClass {
...
+const fn is_valid_ascii_ident_continue(byte: u8) -> bool {
+    matches!(byte, b'a'..=b'z' | b'A'..=b'Z' | b'0'..=b'9' | b'_')
+}
+
+/// Validates that a string is a valid Rust identifier (ASCII only).
+///
+/// This supports standard ASCII identifiers: `foo`, `_bar`, `Baz123`
+///
+/// Valid identifiers must:
+/// - Start with `_` or an ASCII letter (a-z, A-Z)
```
martintmk, line 137: why do we allow starting with `_`?

ralfbiedert (approved):

crates/data_privacy/src/redactors/mod.rs
geeknoid: I don't believe this is significant enough to warrant a benchmark.

crates/data_privacy/src/redactors/simple_redactor.rs
geeknoid: This is not significant enough. The simple redactor is not meant for use in production, it's mostly for tests and examples.

crates/data_privacy/src/data_class.rs
geeknoid: I wanted Rust-compatible identifiers and _ is valid Rust. I opted not to support the full Unicode set of allowed identifier characters however, since that pulls in a ton of Unicode-related crates. So I support the ASCII subset.

crates/data_privacy/src/data_class.rs
geeknoid: Done.

PR #137, opened 2025-12-16
chore: release a new version of tick crate

martintmk:


crates/tick/CHANGELOG.md
```diff
@@ -1 +1,8 @@
 # Changelog
+
+## [0.2.0] - 2025-12-16
+
+- ✨ Features
+
+  - introduce tick crate ([#106](https://github.com/microsoft/oxidizer/pull/106))
```
psandana, line 7: is it a feature?

psandana (approved):

crates/tick/Cargo.toml
```diff
@@ -4,7 +4,7 @@
 [package]
 name = "tick"
 description = "Provides primitives to interact with and manipulate machine time."
-version = "0.1.0"
+version = "0.2.0"
```
psandana, line 7: Why do we need to bump significant version instead of patch?

crates/tick/CHANGELOG.md
sandersaares: Told ya keeping changelogs is a fool's errand 💃

psandana: If we have a template, this will not happen :D

crates/tick/Cargo.toml
martintmk: discussed offline, our internal pipelines need this to push new version (doesn't matter, it's our initial release)

ralfbiedert (approved):

crates/tick/Cargo.toml
psandana: Let's try first the manual CI

psandana (changes requested):
Let's first try to run manual CI

Vaiz (approved):

crates/tick/CHANGELOG.md
geeknoid: The changelog is auto-generated, so there's nothing to do to fix this.

psandana:
> Let's first try to run manual CI

https://crates.io/crates/tick is released. I think we can close this ticket

crates/tick/CHANGELOG.md
psandana: Although this is weird. Previous PR was for 0.1, and this was for 0.2. I guess this happened because it didn't have 0.1 changelog in previous PR.

PR #147, opened 2025-12-16
misc: Add rust-toolchain.toml and reusable constants.toml

sandersaares:
This reduces the chances of toolchain versions going out of sync between local/cloud and different workflows.

In future versions, `constants.toml` can also be used by repo automation/tooling, not only GitHub workflows (e.g. to install same versions of tools locally).

[It is mind-blowing how convoluted this is in GitHub, there being no "include another file" logic]

### DIFF
--- a/.cargo-rdme.toml
+++ b/.cargo-rdme.toml
@@ -2,3 +2,9 @@
 # README.  If this is not set the crate’s documentation will be inserted with its sections
 # belonging to the README section where the insertion happens.
 heading-base-level = 1
+
+[intralinks]
+# The cargo-rdme link parsing functionality is extremely limited and it complains about
+# entirely legitimate links. As it anyway misses a ton of intralinks, let's just remove
+# the links. If someone wants to read the proper linked-up API docs, they can go to docs.rs.
+strip-links = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1449,6 +1449,7 @@ version = "0.6.0"
 dependencies = [
  "many_cpus",
  "mutants",
+ "static_assertions",
  "thread_aware_macros",
 ]
 
--- a/crates/data_privacy/README.md
+++ b/crates/data_privacy/README.md
@@ -64,7 +64,7 @@ to tell over time that an operation is attributed to the same piece of state wit
 
 This crate is built around two primary traits:
 
-* The [`Classified`](https://docs.rs/data_privacy/latest/data_privacy/classified/trait.Classified.html) trait is used to mark types that hold sensitive data. The trait exposes
+* The `Classified` trait is used to mark types that hold sensitive data. The trait exposes
   explicit mechanisms to access the data in a safe and auditable way.
 
 * The [`Redactor`] trait defines the logic needed by an individual redactor. This crate provides a
@@ -85,7 +85,7 @@ This crate also exposes additional traits which are usually, but not necessarily
 A taxonomy is defined using the [`taxonomy`] attribute macro. The macro is applied to an enum
 declaration. Each variant of the enum represents a data class within the taxonomy.
 
-[`DataClass`](https://docs.rs/data_privacy/latest/data_privacy/data_class/struct.DataClass.html) is a struct that represents a single data class within a taxonomy. The struct
+`DataClass` is a struct that represents a single data class within a taxonomy. The struct
 contains the name of the taxonomy and the name of the data class. You can get a `DataClass` instance for a given data class
 by calling the associated `data_class` method on the taxonomy enum.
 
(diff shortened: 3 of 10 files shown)


constants.toml
```diff
@@ -0,0 +1,26 @@
...
+# Shared constants for GitHub workflows and repository automation.
+
+# Rust toolchain versions
+RUST_MSRV = "1.88"                                  # used for testing (ensures the MSRV promise is kept)
+RUST_LATEST = "1.92"                                # used for static analysis & mutation testing; must match rust-toolchain.toml
+RUST_NIGHTLY = "nightly-2025-11-20"                 # used for coverage and extended analysis
+RUST_NIGHTLY_EXTERNAL_TYPES = "nightly-2025-08-06"  # used for external type exposure checks
+
+# Tool versions
+CARGO_AUDIT_VERSION = "0.22.0"
```
geeknoid, line 13: Will dependabot know how to recognize these values and upgrade them automatically when needed, or does this defeat it?

sandersaares: Ah, great question! I have minimal understanding of how Dependabot works and struggle to find useful evidence from its logs but from what documentation and AI insights suggest, Dependabot relies on having literal version strings and will not process anything with variables or other types of indirection.

In context of this PR, I think nothing changes - no new variables were created. All these things were indirectly specified before, all that changed was that the variables became reusable. From what my brief research tells me, these variables were not Dependabot-compatible even before this PR.

geeknoid: Indeed:

Environment Variables: It generally does not support updating versions specified in env blocks or repository secrets. If you do something like:

```yaml

env:
  NODE_VERSION: 14
steps:
  - uses: actions/setup-node@v2
    with:
      node-version: ${{ env.NODE_VERSION }}
```

Dependabot will not know how to update NODE_VERSION. It only looks for the tag after the @ symbol in the uses step.

.github/workflows/main.yml
```diff
@@ -312,7 +387,27 @@ jobs:
...
+      CARGO_ENSURE_NO_DEFAULT_FEATURES_VERSION: ${{ needs.constants.outputs.CARGO_ENSURE_NO_DEFAULT_FEATURES_VERSION }}
+      CARGO_HACK_VERSION: ${{ needs.constants.outputs.CARGO_HACK_VERSION }}
+      CARGO_LLVM_COV_VERSION: ${{ needs.constants.outputs.CARGO_LLVM_COV_VERSION }}
+      CARGO_MUTANTS_VERSION: ${{ needs.constants.outputs.CARGO_MUTANTS_VERSION }}
+      CARGO_RDME_VERSION: ${{ needs.constants.outputs.CARGO_RDME_VERSION }}
+      CARGO_SEMVER_CHECKS_VERSION: ${{ needs.constants.outputs.CARGO_SEMVER_CHECKS_VERSION }}
+      CARGO_SORT_VERSION: ${{ needs.constants.outputs.CARGO_SORT_VERSION }}
+      CARGO_UDEPS_VERSION: ${{ needs.constants.outputs.CARGO_UDEPS_VERSION }}
+      CARGO_WORKSPACES_VERSION: ${{ needs.constants.outputs.CARGO_WORKSPACES_VERSION }}
+      SCCACHE_VERSION: ${{ needs.constants.outputs.SCCACHE_VERSION }}
```
Vaiz, line 410: what is the purpose of keeping unused variables here?

```
      CARGO_SEMVER_CHECKS_VERSION: ${{ needs.constants.outputs.CARGO_SEMVER_CHECKS_VERSION }}
      CARGO_SORT_VERSION: ${{ needs.constants.outputs.CARGO_SORT_VERSION }}
      CARGO_UDEPS_VERSION: ${{ needs.constants.outputs.CARGO_UDEPS_VERSION }}
      CARGO_WORKSPACES_VERSION: ${{ needs.constants.outputs.CARGO_WORKSPACES_VERSION }}
Comment view
      SCCACHE_VERSION: ${{ needs.constants.outputs.SCCACHE_VERSION }}
```

Vaiz:
@sandersaares , have you considered adding those variables as repo config?
https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-variables#creating-configuration-variables-for-a-repository

.github/workflows/main.yml
sandersaares: I was ambivalent about which way to go here:

1. On one hand, we could cherry-pick the variables for each job. Requires a certain amount of diligence but keeps the list small.
2. On the other hand, we could just blanket-forward all variables to all jobs, meaning less need for bookkeeping (at cost of more lines of repeated code spent).

It is not clear to me which way is preferable, so I went with the 2nd option but it was an arbitrary pick.

sandersaares:
> have you considered adding those variables as repo config?

I considered it but rejected it because:

1. It is hidden non-versioned data.
2. It is not actually checked out as part of repo so cannot be used by non-GitHub-Actions build tooling.

.github/workflows/main.yml
Vaiz: is it possible to reassign the whole output to something shorter and just use that instead?

e.g. `consts = needs.constants.outputs`

sandersaares: This did not prove workable because some of the constants are used during "workflow compilation", inserted into action parameters. As far as I can tell, this would only be possible it they were used "at runtime" by the actions but not in the YML itself.

Granted, my knowledge of GitHub Actions is less than of ADO pipelines but neither I nor Copilot could see a way to do that. It is worth trying harder on this in the future, though, as it would be a really nice simplification.

sandersaares:
I agree it is annoying but we do need to have a way to share the constants, though. Do you have a better alternative?

PR #148, opened 2025-12-16
chore: Replace cargo-rdme by cargo-doc2readme

geeknoid:
- Solves issue with cargo-rdme not handling links reliably. The README files produced now have full links.


crates/bytesbuf/README.md
```diff
@@ -538,31 +525,63 @@ For testing purposes, this crate exposes some special-purpose memory providers t
...
-<!-- cargo-rdme end -->
 
 <div style="font-size: 75%" ><hr/>
 
 This crate was developed as part of [The Oxidizer Project](https://github.com/microsoft/oxidizer).
 
 </div>
+
+ [__cargo_doc2readme_dependencies_info]: ggGkYW0CYXSEG2USkr5n8dTvG47r3ve7pFPEG49HUSPBhjRZGzuZSC3wY-xNYXKEG-7Puz6IAGIQG5CqzO13hImhG44P1uV1jkmfGw3omWWkSiU3YWSHgmhCeXRlc0J1ZvaCaUJ5dGVzVmlld_aCdEZpeGVkQmxvY2tUZXN0TWVtb3J59oJqR2xvYmFsUG9vbPaCaUhhc01lbW9yefaCZk1lbW9yefaCdVRyYW5zcGFyZW50VGVzdE1lbW9yefY
+ [__link0]: https://crates.io/crates/BytesView
```
psandana: These links are wrong. They are pointing to a `BytesView` crate, not this crate and struct `BytesView`

geeknoid: Hmm. Turns out the existing docs we have published to crates.io also don't look correct. So there's something in the lib.rs file that's making the readme generation wonky. I'll investigate.

geeknoid: Fixed.

crates/bytesbuf/src/lib.rs
```diff
@@ -614,23 +610,23 @@ mod vec;
...
-pub use memory::*;
-pub use memory_guard::*;
-pub use memory_shared::*;
-pub use opaque_memory::*;
-pub(crate) use span::*;
-pub(crate) use span_builder::*;
-pub use transparent::*;
-pub use view::*;
-pub(crate) use write_adapter::*;
+pub use block::{Block, BlockSize};
```
sandersaares, line 613: This sort of change is out of scope of this PR and not desirable in general, please revert. The wildcards are intentional and need to stay - manually curating some list of types re-exported from a layer down is not a good use of anyone's time.

sandersaares (changes requested):

martin-kolinek (approved):

crates/ohno_macros/README.md
```diff
@@ -1,34 +1,29 @@
 <div align="center">
- <img src="./logo.png" alt="Ohno Macros Logo" width="128">
+ <img src="./logo.png" alt="Ohno_Macros Logo" width="128">
 
-# Ohno Macros
+# Ohno_Macros
```
Vaiz: this looks like a regression

crates/bytesbuf/src/lib.rs
ralfbiedert: I agree this is an unrelated change, but our guidelines say pub re-exports can't be globs, so these should be addressed.

sandersaares: Yeah, I consider that a defect in the guidelines: https://github.com/microsoft/rust-guidelines/issues/4. These globs here are a good thing - only re-exports of already-public things deserve manual curation.

ralfbiedert: It is not a defect in our guidelines, they prevent unintended re-exports.

sandersaares: I acknowledge that we disagree about it being a defect. This does not change my assessment of it being a defect.

crates/ohno_macros/README.md
geeknoid: Fixed.

crates/bytesbuf/src/lib.rs
geeknoid: Is it in scope for this PR, otherwise the generated readme for this crate is completely broken.

martin-kolinek (approved):

PR #150, opened 2025-12-16
ci: Reintroduce semver check, but as a separate optional job

geeknoid:


ralfbiedert (approved):

ralfbiedert:
> What is the story - why was it disabled, how is it meant to be used, what does it mean to have it be optional?

Was originally enabled to prevent semver issues.

Had to be disabled because we currently publish from `master`, so with this on as well you could not "break semver" anymore in a regular commit without being forced to also change to a new major version, which in turn mandated a release.

Is now back on with "warning" to at least signal you might be doing something naughty, without interrupting workflows, until we have better solution.

sandersaares (approved):

PR #153, opened 2025-12-17
docs: Normalize feature handling for docs.rs

geeknoid:
- All crates now have the same docs.rs metadata so we get consistent behavior in our sources.

- Remove all `#[cfg_attr(docsrs, doc(cfg(...)))]` attributes from the codebase as they are no longer needed given I've turned on doc_auto_cfg which has the same effect, without needing manual annotations.

- Rename the scripts/repo-template folder to scripts/crate-template and update the internal references accordingly.

- Update the crate-template README.md and lib.rs to reflect needed modern defaults.


crates/bytesbuf/Cargo.toml
```diff
@@ -24,6 +24,10 @@ allowed_external_types = [
     "bytes::buf::buf_mut::BufMut",
 ]
 
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
```
sandersaares: I believe this cfg flag is already there by default and does not need to be specified explicitly.

Vaiz (approved):

crates/bytesbuf/Cargo.toml
geeknoid: @sanderseares Do you have a reference for this, cause what I find tells me differently:

----

The rustdoc-args line is specifically there so you can use the doc_auto_cfg or doc_cfg features. When docs.rs builds your crate, it passes that flag, allowing you to write code like this in your lib.rs:

```rust
#![cfg_attr(docsrs, feature(doc_auto_cfg))]

/// This function only exists if the "magic" feature is enabled.
/// Because of the 'docsrs' flag, it will have a visual badge in the docs!
#[cfg(feature = "magic")]
pub fn perform_magic() {}
```

What happens if you remove it?
If you remove the rustdoc-args line:

Your documentation will still build (because of all-features).

However, users won't be able to tell which functions require which features. Every function will look like it's part of the standard crate, which can be very confusing for someone trying to use your library without enabling every single feature.

# Recommendation

Keep both. It is currently the "gold standard" for Rust crate documentation. It ensures your docs are complete and that the UI clearly communicates feature requirements to your users.

Would you like me to show you how to set up the doc_auto_cfg attribute in your lib.rs to make those feature badges appear automatically?

sandersaares: Listening to AI "wisdom" is living life dangerously!

https://docs.rs/about/builds#detecting-docsrs

geeknoid: All right, I removed those.

sandersaares (approved):

PR #158, opened 2025-12-19
ci: Add spell checker

geeknoid:


spellcheck.dic
```diff
@@ -0,0 +1,248 @@
+247
```
sandersaares: We already have a `.spelling` file - please do not introduce a second dictionary.

spellcheck.toml
```diff
@@ -0,0 +1,12 @@
+dev_comments = false
```
sandersaares, line 6: I would like to see inline comments documenting the justification for these configuration customizations.

spellcheck.dic
geeknoid: I know, this PR is a work in progress. I spent several hours to get this simple freaking spell checking logic working, and I still have work to do.

.github/workflows/main.yml
```diff
@@ -153,6 +153,73 @@ jobs:
...
+          toolchain: ${{ env.RUST_LATEST }}
+      - name: Install Cargo Tools
+        uses: taiki-e/install-action@v2.62.62
+        with:
+          tool: cargo-spellcheck@${{ env.CARGO_SPELLCHECK_VERSION }}
+
+      # execute
+      - name: Validate Dictionary
+        run: |
+          FILE=".spelling"
```
psandana, line 106: I would rather this validation to exists inside the spellcheck. If a specific format is expected, I rather to have it IN the tool, rather than outside.

.spelling
```diff
@@ -34,7 +34,9 @@ branch_name
 btree_map
 buildable
 bytesbuf
-C #-Rust
+callee
+Cargo.toml
+Cargo.toml.
```
psandana: Cargo.toml. should be invalid, right?

crates/bytesbuf/src/span_builder.rs
```diff
@@ -167,7 +167,7 @@ impl SpanBuilder {
         // This cannot overflow - guarded by assertion above.
         self.filled_bytes = self.filled_bytes.wrapping_sub(len.get());
 
-        // SAFETY: We only seeked over filled bytes, so we must still be in-bounds.
+        // SAFETY: We only skipped over filled bytes, so we must still be in-bounds.
```
psandana:
```suggestion
        // SAFETY: We only seek over filled bytes, so we must still be in-bounds.
```

.github/workflows/main.yml
geeknoid: I don't know what you mean, could you explain further?

.spelling
geeknoid: This string was already present in the .spelling file. I can try to remove it and see if anything fires.

.github/workflows/main.yml
geeknoid: Oh, you mean that cargo-spellcheck should be able to handle input that's not sorted? Yeah, no kidding. But it does require this, so we have this code.

Note that I took inspiration from tokio's GitHub actions, where they do the exact same thing.

.spelling
geeknoid: Removed.

crates/bytesbuf/src/span_builder.rs
geeknoid: Done.

PR #165, opened 2025-12-21
ci: CI improvements

geeknoid:
- Save a bunch of lines of mind-numbing GitHub actions YAML.

- Update to latest tool and action versions.

- Stop using the cargo dependency cache, it's redundant with the other caches we have.

- Enable Sccache on two jobs that should've been using it, which should speed those up.

- Switch to using cargo nextest for running our tests which delivers a considerable perf improvement.

- Run cargo-careful as additional prevention against introducing undefined behaviors.

- Stop using cargo-audit since it is completely redundant with cargo-deny which we already run and does more.


.github/workflows/main.yml
```diff
@@ -218,162 +128,99 @@ jobs:
 
       - name: Post Semver Failure Comment
         if: steps.semver.outcome == 'failure'
-        uses: marocchino/sticky-pull-request-comment@v2
+        uses: marocchino/sticky-pull-request-comment@v2.9.4
```
psandana, line 131: why full version?

geeknoid: In general, we pin everything to specific versions so that the build is completely stable over time. If you don't specify complete versions, you get random upgrades at any time which can break your build. Now, we only upgrade when we want to. And it turns out Claude is perfectly capable to upgrade all the tool and action versions automatically, and it gives a nice report of what's being changed and why.

psandana: I understand the point on deterministic builds. I'm a bit worried if we should have upgraded an action, for instance, due security issues. How can we check for upgrades and apply them?

Also, they should be following SemVer, hence they should not introduce breaking changes.

geeknoid: "should" follow and "do follow" don't always line up.

The core requirement for a hermetic reproducible builds is to have fully pinned down inputs. Consistent tools, consistent source, consistent environment.

dependabot scans our tree and will report semver issues for us for actions. Beyond that, I think the right pattern for us is that whenever a new Rust toolchain is released, we update to that latest version and sync all tools and actions at the same time.

psandana: (Forgot we are on GH, where we have dependabot to look after action versions :D)

PR #169, opened 2025-12-29
feat(tick): implement display extension for `SystemTime`

martintmk:
I found multiple cases when the consumer wants to display `SystemTime` but currently there is no way to do it easily. One can construct `Iso8601` from SystemTime manually that implements the `Display` but it's too verbose. 

This PR adds an `display` extension to simplify the scenario. To fulfill the format requirements we do not fail of `SystemTime` is out of range, but rather display values saturated at boundaries `-009999-01-02T01:59:59Z` and `9999-12-30T22:00:00.999999999Z`.

This is because we are limited by the jiff MIN/MAX values.

Before:

```rust
    println!("Current time: {}", Iso8601::try_from(clock.system_time()).unwrap_or(Iso8601::MAX));
```

After:
```rust
    println!("Current time: {}", clock.system_time().display_iso_8601());
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,7 +25,7 @@ jobs:
         uses: ./.github/actions/setup
         with:
           enable-sccache: true
-          rust-toolchain: RUST_MSRV
+          rust-toolchain: RUST_MSRV, RUST_NIGHTLY
           cargo-tools: CARGO_NEXTEST_VERSION, CARGO_HACK_VERSION
 
       # execute
@@ -37,8 +37,8 @@ jobs:
       - name: Docs
         if: success() || failure()
         env:
-          RUSTDOCFLAGS: "-D warnings"
-        run: cargo doc --verbose --workspace --all-features
+          RUSTDOCFLAGS: "--cfg docsrs -D warnings"
+        run: cargo +${{ env.RUST_NIGHTLY }} doc --no-deps --verbose --workspace --all-features
       - name: Doc Tests
         if: success() || failure()
         run: cargo test --doc --verbose --workspace --all-features
@@ -70,9 +70,6 @@ jobs:
       - name: Cargo.toml Format
         if: success() || failure()
         run: cargo sort --check --grouped --workspace
-      - name: Docs
-        if: success() || failure()
-        run: cargo doc --no-deps --workspace --all-features
       - name: Readmes
         if: success() || failure()
         run: cargo workspaces exec --ignore-private cargo doc2readme --check --lib --template ../README.j2
--- a/crates/bytesbuf/src/lib.rs
+++ b/crates/bytesbuf/src/lib.rs
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 
 #![cfg_attr(coverage_nightly, feature(coverage_attribute))]
-#![cfg_attr(docsrs, feature(doc_auto_cfg))]
+#![cfg_attr(docsrs, feature(doc_cfg))]
 
 //! Types for creating and manipulating byte sequences.
 //!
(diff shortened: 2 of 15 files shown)


crates/tick/src/fmt/mod.rs
```diff
@@ -15,6 +15,12 @@
 //! - [`UnixSeconds`]: Parsing and formatting of system time that is represented as the number of whole seconds since Unix epoch.
 //!   For example, `0` represents `Thu, 1 Jan 1970 00:00:00 -0000`.
 //!
+//! # Extension Traits
+//!
+//! - [`SystemTimeExt`]: Extension trait for [`SystemTime`][`std::time::SystemTime`] that provides convenient
```
geeknoid: Shouldn't we generalize this a bit? That is, instead of .display() maybe it should be display_as_8601 and then we can have similar methods for all the formats?

sandersaares:
> One can construct Iso8601 from SystemTime manually that implements the Display but it's too verbose.

Can you bring an example of the too-verbose alternative?

crates/tick/src/fmt/mod.rs
martintmk: True, I renamed this to `display_iso_8601`

crates/tick/src/future_ext.rs
```diff
@@ -6,7 +6,7 @@ use std::time::Duration;
 use super::{Clock, Delay, Timeout};
 
 /// Extensions for the [`Future`] trait.
-pub trait FutureExt: Future {
+pub trait FutureExt: Future + sealed::Sealed {
```
sandersaares: I do not understand how this is meant to work. Sealing a type means "only we can make implementations of this type", yet this appears to be an extension trait, which presumably means it is meant to extend any `Future`, including futures made by our customer code?

What am I missing here and what is the purpose of this "sealing"?

crates/tick/src/fmt/mod.rs
```diff
@@ -53,6 +66,9 @@ pub use iso_8601::Iso8601;
 pub use rfc_2822::Rfc2822;
 pub use unix_seconds::UnixSeconds;
 
+// Re-export SystemTimeExt from parent module
```
sandersaares: What is the purpose of this? It is not obvious to me why we do this.

[In other words: the comment on this line should be about "why" not about "what"; this is very `int i = 5; // Define integer i` style comment]

sgalkin (approved):

crates/tick/src/system_time_ext.rs
```diff
@@ -0,0 +1,89 @@
...
+#[cfg(any(feature = "fmt", test))]
+fn to_timestamp(system_time: SystemTime) -> jiff::Timestamp {
+    match jiff::Timestamp::try_from(system_time) {
+        Ok(timestamp) => timestamp,
+        Err(_) => to_timestamp_fallback(system_time),
+    }
+}
+
+#[cfg(any(feature = "fmt", test))]
+fn to_timestamp_fallback(system_time: SystemTime) -> jiff::Timestamp {
```
sandersaares: This could use a definition of what "fallback" means. What do we fall back "from" and "to"? What might be the situations where we might want to "fall back"?

sandersaares (changes requested):

crates/tick/src/future_ext.rs
martintmk: true, looking at this, the sealed pattern does not make sense here because we are restricting the trait to futures anyway.

crates/tick/src/system_time_ext.rs
martintmk: renamed to `to_timestamp_min_max`, this is for cases where system time cannot be converted to `Timestamp` and in that case we just display the outside jiff-supported ranges.

crates/tick/src/fmt/mod.rs
martintmk: useless comment indeed, dropped

sandersaares (approved):

PR #172, opened 2025-12-30
chore: Tweak compilation profile settings

### LINKED ISSUE #170: Stack traces from "coverage" check are missing line numbers
Unclear if it also affects other things but only the standard library lines here appear to have line numbers. Are we executing tests for coverage without generating debug information or something to that nature?

Example from https://github.com/microsoft/oxidizer/actions/runs/20592503815/job/59140206566?pr=169:

```text
thread 'system_time_ext::tests::display_ok' (2020) panicked at C:\Users\runneradmin\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\jiff-0.2.16\src\timestamp.rs:2833:41:
duration too big (negative)
stack backtrace:
   0: std::panicking::panic_handler
             at /rustc/806c2a35dcc1db7867864fea0a7a65554a5238b7/library\std\src\panicking.rs:689
   1: core::panicking::panic_fmt
             at /rustc/806c2a35dcc1db7867864fea0a7a65554a5238b7/library\core\src\panicking.rs:80
   2: core::panicking::panic_display
             at /rustc/806c2a35dcc1db7867864fea0a7a65554a5238b7/library\core\src\panicking.rs:259
   3: core::option::expect_failed
             at /rustc/806c2a35dcc1db7867864fea0a7a65554a5238b7/library\core\src\option.rs:2208
   4: <core::option::Option<std::time::SystemTime>>::expect
   5: <std::time::SystemTime as core::convert::From<jiff::timestamp::Timestamp>>::from
   6: core::sync::atomic::fence
   7: tick::system_time_ext::tests::display_ok::{closure#0}
   8: <tick::system_time_ext::tests::display_ok::{closure#0} as core::ops::function::FnOnce<()>>::call_once
   9: core::ops::function::FnOnce::call_once
             at C:\a\rust\rust\library\core\src\ops\function.rs:250
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

geeknoid:
- Make sure full debugging is explicitly enabled. Otherwise, cargo-llvm-cov seems to pick defaults that make it so full source file & line number info is not being collected. Fixes #170 

- While I was there, I had been meaning to tweak the release settings so we get better optimization during our benchmark runs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -162,9 +162,22 @@ multiple_crate_versions = "allow"
 missing_crate_level_docs = "warn"
 unescaped_backticks = "warn"
 
+# A bit of debugging support for release builds.
 [profile.release]
 debug = "line-tables-only"
 
+# Best perf possible for benchmarks
+[profile.bench]
+lto = "fat"
+codegen-units = 1
+
+# Best debugging experience possible for test builds. Even though this is the default, we specify it explicitly,
+# otherwise it appears that cargo-llvm-cov assumes different defaults which leads to only minimal debug info in
+# coverage builds.
+[profile.test]
+debug = "full"
+
+# No debug info for mutation testing to speed up compilation
 [profile.mutants]
 inherits = "test"
 debug = "none"


sandersaares (changes requested):
I would like us to have inline comments to justify each of these customized profile settings, so the next reader is aware of why the thing is the way it is.

martin-kolinek (approved):

Issue #174, opened 2025-12-31
Missing development environment setup info for cargo-spellcheck


sandersaares:
While we are in general missing a development environment setup guide and related scripting/tooling, for the most part it can be figured out on the fly (e.g. taking GitHub workflows as reference). However, `cargo spellcheck` appears not to be so easy to setup:

> cargo install cargo-spellcheck

results in

```text
warning: clang-sys@1.8.1: could not execute `llvm-config` one or more times, if the LLVM_CONFIG_PATH environment variable is set to a full path to valid `llvm-config` executable it will be used to try to find an instance of `libclang` on your system: "couldn't execute `llvm-config --prefix` (path=llvm-config) (error: program not found)"
error: failed to run custom build command for `clang-sys v1.8.1`

Caused by:
  process didn't exit successfully: `C:\Users\sasaares\AppData\Local\Temp\cargo-install1PHDae\release\build\clang-sys-2147579cc6ebb833\build-script-build` (exit code: 101)
  --- stdout
  cargo:warning=could not execute `llvm-config` one or more times, if the LLVM_CONFIG_PATH environment variable is set to a full path to valid `llvm-config` executable it will be used to try to find an instance of `libclang` on your system: "couldn't execute `llvm-config --prefix` (path=llvm-config) (error: program not found)"

  --- stderr

  thread 'main' (94696) panicked at C:\Users\sasaares\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\clang-sys-1.8.1\build\dynamic.rs:225:45:       
  called `Result::unwrap()` on an `Err` value: "couldn't find any valid shared libraries matching: ['clang.dll', 'libclang.dll'], set the `LIBCLANG_PATH` environment variable to a path where one of these files can be found (invalid: [])"
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
error: failed to compile `cargo-spellcheck v0.15.5`, intermediate artifacts can be found at `C:\Users\sasaares\AppData\Local\Temp\cargo-install1PHDae`.   
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
```

This might be as good as any place to start defining the development environment setup process.

martin-kolinek:
Tracked internally as work item 7326953.

PR #176, opened 2026-01-02
misc: Document and automate development environment setup

### LINKED ISSUE #173: cargo-spellcheck dictionary processing logic conflicts with Code Spell Checker (cSpell) VS code extension
`cargo-spellcheck` used in the GitHub workflows appears to require that the first line in a dictionary file be the line count (and perhaps has additional requirements). Code Spell Checker (cSpell), which we have so far used for IDE-integrated spell checking in VS Code, however, requires each line to be a word.

This means that updating dictionaries via cSpell will corrupt the dictionary (from the cargo-spellcheck point of view) because cSpell does not maintain any sort of line count.

> Error: The number of lines (293) does not match 00000000000000000000000000000000000000000000000.

[Aside, the hunspell format specification actually says the line count is *approximate* which cargo-spellcheck seems to violate here by requiring a match]

We need to converge the dictionary experience somehow and have a single dictionary file used both in the IDE and in the GitHub workflows.

@geeknoid did you already explore this topic when integrating cargo-spellcheck? Do you know of meaningful options to put the table?

sandersaares:
Adds `DEVELOPMENT.md` with local PC development environment setup instructions.

Adds a `justfile` with (incomplete set of) commands for various relevant activities.

Rewires the brains of the spell check logic so we do not need to babysit the dictionary - you can simply yolo it and `just spellcheck` will preprocess the dictionary before giving it to the nitpicky `cargo-spellcheck` utility. We are now required to use `just spellcheck` instead of `cargo spellcheck`.

Adds execution of all standalone examples into the `testing` check in the GitHub workflows.

Fixes #173, #174

Justfile commands:

```text
    build PROFILE='dev'
    careful FILTER=""         # Run unit and integration tests with extra standard libbrary debug assertions enabled.
    check PROFILE='dev'
    clippy PROFILE='dev'
    coverage-measure
    coverage-report
    deny
    docs
    examples PROFILE='dev'    # Run all stand-alone example binaries to ensure they complete with exit code 0.
    format
    format-check
    hack
    install-tools             # Installs the complete set of Rust toolchains and other tools for local PC development.
    miri FILTER=""            # Run unit and integration tests under Miri.
    miri-example EXAMPLE_NAME # Run an example binary under Miri.
    readme                    # Update README.md files for all packges.
    readme-check              # Check that README.md files for all packges are up to date.
    spellcheck                # Check the spelling of the workspace contents.
    test FILTER=""            # Runs unit and integration tests.
    test-docs FILTER=""       # Runs inline examples in documentation comments (doctests).
    test-more FILTER=""       # Same as 'test' but also run one iteration of benchmarks.
    udeps                     # Detect unused dependencies.
```


DEVELOPMENT.md
```diff
@@ -0,0 +1,111 @@
+# Development environment setup
+
+The standard development environment is a Windows PC, with Windows Subsystem for Linux (WSL) used for Linux development.
+
+> 💡 Development on a Linux or Mac PC is possible, though "at your own risk" - it is not expected that people verify compatibility with "Linux/Mac as primary workstation" scenarios when committing changes to the repo.
+
+The following software must be present on the PC:
+
+* [Latest version of LLVM](https://github.com/llvm/llvm-project/releases)
```
martintmk, line 10: is this indeed a requirement? I have never installed this and I can compile and work in the repo just fine.

just_basic.just
```diff
@@ -0,0 +1,110 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+build PROFILE='dev':
+    cargo build {{ target_package }} --profile {{ PROFILE }} --all-features --all-targets --locked
+
+# Run unit and integration tests with extra standard libbrary debug assertions enabled.
```
martintmk:
```suggestion
# Run unit and integration tests with extra standard library debug assertions enabled.
```

DEVELOPMENT.md
sandersaares: Yes, this is required by the spell checker, which is a particular sort of abomination. I am 50-50 on whether to accept this sort of dependency just for a spell checker, really. For now I stuck to documenting and avoided getting too annoyed by it :)

DEVELOPMENT.md
```diff
@@ -0,0 +1,111 @@
+# Development environment setup
+
+The standard development environment is a Windows PC, with Windows Subsystem for Linux (WSL) used for Linux development.
+
+> 💡 Development on a Linux or Mac PC is possible, though "at your own risk" - it is not expected that people verify compatibility with "Linux/Mac as primary workstation" scenarios when committing changes to the repo.
+
+The following software must be present on the PC:
```
Liamdoult: IMO the only *must* should be `cargo build` & `cargo test`. Everything else should be a quality of life improvement or for a specific documented reason and limited in scope. Especially if this is OSS.

This also answers the question of LLVM for spell check. It should be enforced by the pipeline, but a recommendation for the contributor. This needlessly complicates requirements for contribution. Especially with some of these windows only requirements (such as Visual Studio).

ps. `cargo build` does work out of the box on my mac machine. `cargo test` mostly works.

DEVELOPMENT.md
```diff
@@ -0,0 +1,111 @@
...
+  * Individual Components: C++ Clang Compiler for Windows
+* [Latest version of Visual Studio Code](https://code.visualstudio.com/Download) with extensions:
+  * C/C++ (ms-vscode.cpptools)
+  * rust-analyzer (rust-lang.rust-analyzer)
+  * WSL (ms-vscode-remote.remote-wsl)
+
+> 💡 Using an IDE other than Visual Studio Code is possible, though "at your own risk" - it is not expected that people verify compatibility with other IDEs when committing changes to the repo.
+
+* [Latest version of Git](https://git-scm.com/downloads/win) with components:
+  * Git LFS (Large File Storage)
```
Liamdoult, line 22: What is LFS being used for?

DEVELOPMENT.md
sandersaares: The development environment setup guide is intended to document the setup needed to exercise **every part** of the local PC development workflow. If someone wishes only to exercise a limited part like `cargo build` that is of course their right but I do not really feel it is worth our time to maintain a separate set of instructions for them.

I recognize that long setup instructions are annoying and a hindrance to easy contributions, of course. But this might be better seen as a motivation to simplify the tooling, rather than as a reason to provide different "levels" of functionality.

DEVELOPMENT.md
sandersaares: I believe we use it to store images.

DEVELOPMENT.md
martintmk: Ok, if this is not a hard requirement, we can live with this.

DEVELOPMENT.md
Liamdoult: > But this might be better seen as a motivation to simplify the tooling

This was more my point. That we should be striving to have the simplest tooling possible. Especially, to confirm with the industry standard, as much as reasonably possible.

justfiles/basic.just
```diff
@@ -0,0 +1,113 @@
+# Copyright (c) Microsoft Corporation.
```
geeknoid, line 1: Can these .just files be put into a subdirectory so as not to pollute the root of the repo so much?

DEVELOPMENT.md
geeknoid: I agree, this is this of requirements doesn't actual meet the bar of requirements. I ahev none of the tools listed here (latest LLVM, VS 2022, or VS Code) installed on my machine and I can build and use this repo just fine.

I think we should state the actual requirements as requirements: Minimum Rust version

Everything else should be listed under an "Additional Scenario" section which says "if you want to do X, then you need tool Y"

DEVELOPMENT.md
```diff
@@ -0,0 +1,111 @@
...
+
+The standard development environment is a Windows PC, with Windows Subsystem for Linux (WSL) used for Linux development.
+
+> 💡 Development on a Linux or Mac PC is possible, though "at your own risk" - it is not expected that people verify compatibility with "Linux/Mac as primary workstation" scenarios when committing changes to the repo.
+
+The following software must be present on the PC:
+
+* [Latest version of LLVM](https://github.com/llvm/llvm-project/releases)
+  * Select: ☑️ Add LLVM to the system PATH for the current user
+* [Visual Studio 2022](https://visualstudio.microsoft.com/downloads/) with components:
```
geeknoid: What value is Visual Studio for a Rust repo?

sandersaares: The C++ build tools that come with it are necessary to build some crates. I am not up to date on which ones need it in particular but since we have it in the private repo, odds are we also need it here. In theory, it is possible we no longer have (nor will introduce via private repo migration) such dependencies but I leave that investigation to a separate stage since this is something we have had in our internal setup instructions for a while now - this PR is more about documenting than about researching/innovating.

justfiles/basic.just
sandersaares: Good idea, will do.

DEVELOPMENT.md
sandersaares: Scenario-based setup guides are impractical to maintain - even a single full guide gets outdated but is at least somewhat manageable. This set of instructions is intended to describe a full setup. If you follow this guide, you can use all our development tooling. Partial scenarios or "pick your needs" are intentionally not part of this guide.

VS Code is the standard IDE in our team and the lowest common denominator in OSS, so by default I focus on it here, as it has always been in our private instructions. We may one day decide as a team to generalize away from VS Code but that day has not arrived today.

> I ahev none of the tools listed here (latest LLVM, VS 2022, or VS Code) installed on my machine and I can build and use this repo just fine.

The existence of this development setup guide does not mean that following alternative setup paths would not work. The only thing we guarantee is that following this guide will work. Deviate at your own risk.

If you can tell me how you installed `cargo-spellcheck` without LLVM, though, I would be very happy to throw that one out - it is only there because `cargo-spellcheck` refused to install without it (see #174).

DEVELOPMENT.md
geeknoid: You don't need these tools to build this repo, stating this as a requirement is a mistake which just makes it more likely someone will just ignore our repo because it's too much of a hassle.

sandersaares: I have verified that the instructions in `DEVELOPMENT.md` are accurate with regard to the Visual Studio requirement:

```
cargo install just --locked

...

   Compiling colorchoice v1.0.4
error: linker `link.exe` not found
  |
  = note: program not found

note: the msvc targets depend on the msvc linker but `link.exe` was not found

note: please ensure that Visual Studio 2017 or later, or Build Tools for Visual Studio were installed with the Visual C++ option.

note: VS Code is a different product, and is not sufficient.

error: could not compile `rayon-core` (build script) due to 1 previous error
```

That said, I observe that an editorial mistake kept the "Clang compiler" bullet point (part of an earlier draft that proved unworkable) - this is not correct and I have removed it.

sandersaares:
Verification found one bullet point that I was able to remove (no need for Clang tools in VS 2022). Otherwise the document appears correct.

DEVELOPMENT.md
martin-kolinek: > The existence of this development setup guide does not mean that following alternative setup paths would not work. The only thing we guarantee is that following this guide will work. Deviate at your own risk.

I personally think we should be a bit more careful with wording in this case. I definitely wouldn't want to discourage people from contributing by using the word must or requirement. I would expect a bit of softer language and positioning this as a suggestion rather than requirement would go a long way to alleviate the concerns raised above.

DEVELOPMENT.md
geeknoid: The actual requirement is to install the Microsoft C++ Build Tools. Those happen to be installed with VS, but they can also be installed separately for a much smaller less scary footprint.

DEVELOPMENT.md
sandersaares: I reviewed the development environment setup guide with this in mind and do not see any unreasonably strong language. It is a development environment setup guide. People deviating from setup guides should absolutely expect to be operating at their own risk - that is just how the world works, not something we impose on the world here. Happy to look at specific suggestions if you have some but at the moment I do not see anything actionable here.

DEVELOPMENT.md
sandersaares: You are correct - thank you for specific actionable suggestion! I have adjusted the guide to only include installation of the C++ build tools.

DEVELOPMENT.md
```diff
@@ -0,0 +1,113 @@
+# Development environment setup
+
+The standard development environment is a Windows PC, with Windows Subsystem for Linux (WSL) used for Linux development.
+After following this guide, you will be able to execute all the local PC development tooling associated with this repository.
+
+> 💡 Development on a Linux or Mac PC is possible, though "at your own risk" - it is not expected that people verify compatibility with "Linux/Mac as primary workstation" scenarios when committing changes to the repo.
```
martin-kolinek: This makes it sound like it's expected that people verify compatibility on Windows, which I don't believe should be the case.

DEVELOPMENT.md
```diff
@@ -0,0 +1,113 @@
...
+* [Latest version of LLVM](https://github.com/llvm/llvm-project/releases)
+  * Select: ☑️ Add LLVM to the system PATH for the current user
+* [Visual Studio Build Tools for C++](https://visualstudio.microsoft.com/visual-cpp-build-tools/) with components:
+  * Workload: Desktop development with C++
+* [Latest version of Visual Studio Code](https://code.visualstudio.com/Download) with extensions:
+  * C/C++ (ms-vscode.cpptools)
+  * rust-analyzer (rust-lang.rust-analyzer)
+  * WSL (ms-vscode-remote.remote-wsl)
+
+> 💡 Using an IDE other than Visual Studio Code is possible, though "at your own risk" - it is not expected that people verify compatibility with other IDEs when committing changes to the repo.
```
martin-kolinek: I don't really like this line - it is worded in an exclusionary way. One suggestion I'd have is to structure this file a bit differently. rather than saying "do development on Windows with VS Code or else...", we could say that you just need a common Rust setup, and have sections for different tested setups, one of which is Windows+WSL with VS Code.

DEVELOPMENT.md
```diff
@@ -0,0 +1,113 @@
+# Development environment setup
+
+The standard development environment is a Windows PC, with Windows Subsystem for Linux (WSL) used for Linux development.
+After following this guide, you will be able to execute all the local PC development tooling associated with this repository.
+
+> 💡 Development on a Linux or Mac PC is possible, though "at your own risk" - it is not expected that people verify compatibility with "Linux/Mac as primary workstation" scenarios when committing changes to the repo.
+
+The following software must be present on the PC:
```
martin-kolinek: I'd replace this sentence with something like "To be able to run all the tooling locally, you'll need the following tools installed".

martin-kolinek (changes requested):

DEVELOPMENT.md
martin-kolinek: By the way, that assumes msvc toolchain. I'd expect people to be able to use the windows gnu toolchain in which case they won't need MS C++ Build tools. Also, the build tools are part of [the Rust installation instructions](https://rust-lang.org/tools/install/), so I'm not sure how much we want to duplicate those.

DEVELOPMENT.md
martin-kolinek: Left some specific suggestions.

DEVELOPMENT.md
sandersaares: > I'd expect people to be able to use the windows gnu toolchain in which case they won't need MS C++ Build tools

This may be true but a guide needs to describe one path, not list possibilities, in order to be useful. I picked MSVC as the toolchain given that that's what Rust uses by default. This document intentionally does not describe alternative paths for doing things.

> Also, the build tools are part of [the Rust installation instructions](https://rust-lang.org/tools/install/), so I'm not sure how much we want to duplicate those.

Whatever the Rust installer does there is not enough to result in a working installation, unfortunately. Maybe a defect in the Rust installer - I did not investigate the reasons. See error above in thread.

DEVELOPMENT.md
sandersaares: I do not really expect us to be testing/validating different setups, only testing/validating one standard setup. Even keeping one standard setup maintained and working is a hassle and any promise we make regarding validating different setups would just be setting ourselves up for failure as the guides inevitably bitrot away.

That said, I am happy to soften the wording a bit here - edited.

DEVELOPMENT.md
sandersaares: Adjusted wording to refactor the statement in terms of scoping.

DEVELOPMENT.md
sandersaares: Adjusted

DEVELOPMENT.md
martin-kolinek: I guess the point I'd make is that most people aren't even going to keep this standard setup maintained and working. It's best effort, if that. So my preferred approach would be to keep this open and if someone wants to contribute an additional setup with the same maintenance model, it should be natural to do so.

In other words, based on the comments from folks above, not many people are following this guide, most just do git clone and cargo build, and I expect that to remain the same. I don't want to stop anyone from suggesting environment setups that worked for them, but I don't want to promote one of such setups to become the blessed one, discouraging all the others. There is nothing in the PR checks that would ensure these instructions stay up to date, so I don't see much value in having this blessed setup.

That said, I do think we should provide a general purpose way to run a reasonable subset of checks before publishing a PR, but I hope that can be achieved without saying you "need" this specific setup.

martin-kolinek (approved):

justfiles/setup.just
```diff
@@ -0,0 +1,39 @@
...
+        cargo-mutants@$env:CARGO_MUTANTS_VERSION `
+        cargo-nextest@$env:CARGO_NEXTEST_VERSION `
+        cargo-semver-checks@$env:CARGO_SEMVER_CHECKS_VERSION `
+        cargo-sort@$env:CARGO_SORT_VERSION `
+        cargo-udeps@$env:CARGO_UDEPS_VERSION `
+        cargo-workspaces@$env:CARGO_WORKSPACES_VERSION `
+        --locked
+
+    # This tool is not well maintained and fails to actually build if using locked dependencies. Okay then.
+    cargo install cargo-spellcheck@$env:CARGO_SPELLCHECK_VERSION
```
psandana, line 39: would this be a semver breaking issue?

DEVELOPMENT.md
martintmk: I'll merge this to unblock some tooling, but I have crated an issue so we can address this in follow-up.

justfiles/setup.just
martintmk: I don't understand how this could be a semver breaking issue?

(will merge, because this is not blocking anything)

psandana: It says that it fails to build if different set of deps are used, even in the compatible range. But agree this is not a stopper for this PR

PR #183, opened 2026-01-06
chore: Publish an update to bytesbuf so the docs on docs.rs are fixed

geeknoid:


crates/bytesbuf/src/lib.rs
```diff
@@ -126,7 +126,7 @@
 //!    or test code with no framework), you can create your own instance via [`GlobalPool::new()`].
 //!
 //! Once you have a memory provider, you can reserve memory from it by calling
-//! [`Memory::reserve()`] on it. This returns a [`BytesBuf`] with at least the requested
+//! [`Memory::reserve`](crate::mem::Memory::reserve) on it. This returns a [`BytesBuf`] with at least the requested
```
sandersaares, line 129: This change is not related to the PR - please avoid arbitrary modifications, as the API documentation already has the intended contents and these formatting changes in `lib.rs` are not necessary nor desirable.

crates/bytesbuf/Cargo.toml
```diff
@@ -4,7 +4,7 @@
 [package]
 name = "bytesbuf"
 description = "Types for creating and manipulating byte sequences."
-version = "0.2.0"
+version = "0.2.1"
```
sandersaares, line 7: Do we also need to bump versions of other crates or is this one the only one affected? [I believe Katerina might already be in process of doing this, by the way]

crates/bytesbuf/src/lib.rs
geeknoid: They are needed because the README file isn't being generated correctly.

crates/bytesbuf/Cargo.toml
geeknoid: No, we don't need to. bytesbuf if the only crate affected.

crates/bytesbuf/src/lib.rs
sandersaares: Please provide additional detail - I am not seeing the incorrect generation. What am I missing?

PR #184, opened 2026-01-06
chore: Update crates to fix docs

kate-shine:
Mainly to release changes from https://github.com/microsoft/oxidizer/commit/af97b4255ff21e2610a9a92f2b9b12832ee16c5d


kate-shine:
@microsoft-github-policy-service agree company="Microsoft"

sandersaares (commented):
I would have expected these to be patch version increments, not minor version increments, as the only purpose here should be to just jiggle docs.rs to build again by publishing an only-metadata-changes new version. Why are these minor version increments?

PR #189, opened 2026-01-07
feat: introduce layered crate

martintmk:
Summary

  Introduces layered, a modern service abstraction crate for building composable async services with middleware.

  Key features:
  - Service trait using native async fn with &self - no manual Future types or boxing required
  - Layer and Stack for composing middleware using ergonomic tuple syntax: (layer1, layer2, service).build()
  - Execute helper to convert closures into services
  - Intercept middleware for observing/modifying inputs and outputs (logging, metrics, validation)
  - DynamicService for type erasure when concrete types become unwieldy
  - Bidirectional Tower interoperability via the tower module

### DIFF
--- a/crates/layered/src/intercept.rs
+++ b/crates/layered/src/intercept.rs
@@ -39,12 +39,10 @@ use crate::Service;
 ///     Intercept::<String, String, _>::layer()
 ///         .on_input(|input| println!("request: {input}")) // input observers are called first
 ///         .on_input(|input| println!("another: {input}")) // multiple observers supported
-///         .debug_input() // convenience method to print inputs with `dbg!`
 ///         .modify_input(|input| input.to_uppercase()) // then inputs are modified
 ///         .modify_input(|input| input.to_lowercase()) // multiple modifications supported
 ///         .on_output(|output| println!("response: {output}")) // output observers called first
 ///         .on_output(|output| println!("another response: {output}")) // multiple observers supported
-///         .debug_output() // convenience method to print outputs with `dbg!`
 ///         .modify_output(|output| output.trim().to_string()) // then outputs are modified
 ///         .modify_output(|output| format!("result: {output}")), // multiple modifications supported
 ///     Execute::new(|input: String| async move { input }),
@@ -298,56 +296,6 @@ impl<In, Out> InterceptLayer<In, Out> {
     {
         self.modify_output.push(ModifyOutput(Arc::new(f)));
         self
-    }
-}
-
-impl<In: Debug, Out> InterceptLayer<In, Out> {
-    /// Adds debug logging for incoming inputs using `dbg!`.
-    ///
-    /// # Examples
-    ///
-    /// ```
-    /// # use layered::{Execute, Stack, Intercept, Service};
-    /// # async fn example() {
-    /// let stack = (
-    ///     Intercept::layer().debug_input(), // print input with dbg!
-    ///     Execute::new(|input: String| async move { input }),
-    /// );
-    ///
-    /// let service = stack.build();
-    /// let response = service.execute("input".to_string()).await;
-    /// # }
-    /// ```
-    #[must_use]
-    pub fn debug_input(self) -> Self {
-        self.on_input(|input| {
-            dbg!(input);
-        })
-    }
-}
-
-impl<In, Out: Debug> InterceptLayer<In, Out> {
-    /// Adds debug logging for outgoing outputs using `dbg!`.
-    ///
-    /// # Examples
-    ///
-    /// ```
-    /// # use layered::{Execute, Stack, Intercept, Service};
-    /// # async fn example() {
-    /// let stack = (
-    ///     Intercept::layer().debug_output(), // print outputs with dbg!
-    ///     Execute::new(|input: String| async move { input }),
-    /// );
-    ///
-    /// let service = stack.build();
-    /// let response = service.execute("input".to_string()).await;
-    /// # }
-    /// ```
-    #[must_use]
-    pub fn debug_output(self) -> Self {
-        self.on_output(|output| {
-            dbg!(output);
-        })
     }
 }
 
@@ -660,16 +608,6 @@ mod tests {
             Poll::Ready(Ok(())) => (),
             _ => panic!("Expected Poll::Ready(Ok(())), got {result:?}"),
         }
-    }
-
-    #[test]
-    fn debug_input_output() {
-        let stack = (
-            Intercept::layer().debug_input().debug_output(),
-            Execute::new(|s: String| async move { s }),
-        );
-        let svc = stack.build();
-        let _ = block_on(svc.execute("test".into()));
     }
 
     #[test]


crates/layered/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+//! ## Thread Safety
+//!
+//! All services must be [`Send`] + [`Sync`], and returned futures must be [`Send`].
+//! This ensures compatibility with multi-threaded async runtimes like Tokio.
+//!
+//! ## Features
+//!
+//! - **`intercept`** — Enables [`Intercept`] middleware
+//! - **`dynamic-service`** — Enables [`DynamicService`] for type erasure
+//! - **`tower-service`** — Enables Tower interoperability via the [`tower`] module
```
geeknoid:
```suggestion
//! - **`tower-service`** — Enables Tower interoperability via the [`tower`] crate
```

crates/layered/src/service.rs
```diff
@@ -0,0 +1,98 @@
...
+
+/// An async function `In → Out` that processes inputs.
+///
+/// This trait is the foundation for building composable services.
+/// Implement it directly for custom services, or use [`Execute`][crate::Execute]
+/// to wrap closures.
+///
+/// See the [crate documentation][crate] for usage examples and layer composition.
+#[cfg_attr(any(test, feature = "dynamic-service"), dynosaur::dynosaur(pub DynService = dyn(box) Service, bridge(none)))]
+pub trait Service<In>: Send + Sync {
```
geeknoid, line 12: So why is In a generic, but Out is an associated type? Couldn't you also have:

```rust
pub trait Service: Send + Sync {
    type In;
    type Out;
}
```

crates/layered/src/execute.rs
```diff
@@ -0,0 +1,93 @@
...
+
+use crate::Service;
+
+/// Wraps a function or closure as a [`Service`].
+///
+/// Use `Execute` to turn any async function into a service without defining a custom type.
+///
+/// # Examples
+///
+/// ```rust
```
geeknoid:
```suggestion
/// ```
```

geeknoid: Here and elsewhere in the crate.

crates/layered/src/intercept.rs
```diff
@@ -0,0 +1,662 @@
...
+/// # }
+/// ```
+///
+/// Advanced usage of `Intercept` allows you to modify and observe inputs and outputs:
+///
+/// ```rust
+/// # use layered::{Execute, Stack, Intercept, Service};
+/// # async fn example() {
+/// let execution_stack = (
+///     Intercept::<String, String, _>::layer()
```
geeknoid, line 39: Would this benefit from using an explicit builder pattern & builder terminology?

crates/layered/src/intercept.rs
```diff
@@ -0,0 +1,662 @@
...
+///     Execute::new(|input: String| async move { input }),
+/// );
+///
+/// let service = execution_stack.build();
+/// let response = service.execute("input".to_string()).await;
+/// # }
+/// ```
+#[derive(Clone, Debug)]
+pub struct InterceptLayer<In, Out> {
+    on_input: Vec<OnInput<In>>,
```
geeknoid, line 82: Should all these vectors be SmallVec?

martintmk: This is not intended to be used on hot path. Essentially, it's just a builder that one initializes at startup and then apply to build actual service.

crates/layered/src/intercept.rs
martintmk: I have chosen to use the `tower-layer` crate and terminology because it does not have any impact on core architecture but still gives us compatibility with tower-based ecosystem for free.

geeknoid: I'm not sure we're talking about the same thing.

I'm saying that Intercept::layer() with all the fluent method calls is in fact a builder. Shouldn't be this LayerBuilder::...build();? Or something along those lines?

crates/layered/src/lib.rs
martintmk: the `tower` is actually a module here, not a crate

crates/layered/src/service.rs
martintmk: Following the `tower::Service` design where it's also generic over input. We don't want to stray too much here because we still want to keep the designs close where it makes sense.

crates/layered/src/execute.rs
martintmk: done

crates/layered/src/intercept.rs
martintmk: the `LayerBuilder::...build();?` would produce the actual layer? One downside of this is that the code would be slightly more verbose. 

Rather than:

```
let stack = (
    Telemetry::layer()
    execute
);
```

We would have:
```
let stack = (
    Telemetry::builder().build()
    execute
);
```

Or just:

```
let stack = (
    Telemetry::builder()
    execute
);
```

But in the context of how this is used, I think layer is slightly better. (and also aligned with tower terminology)

rust-toolchain.toml
```diff
@@ -3,7 +3,7 @@
 
 [toolchain]
 # This needs to match RUST_LATEST in constants.toml.
-channel = "1.92"
+channel = "ms-prod-1.92"
```
Vaiz: is this an intended change?

crates/layered/README.md
```diff
@@ -0,0 +1,136 @@
...
+```rust
+use layered::Service;
+
+struct Greeter;
+
+impl Service<String> for Greeter {
+    type Out = String;
+
+    async fn execute(&self, name: String) -> Self::Out {
+        format!("Hello, {}!", name)
```
Vaiz:
```suggestion
        format!("Hello, {name}!")
```

crates/layered/README.md
```diff
@@ -0,0 +1,136 @@
...
+}
+```
+
+Use [`Execute`][__link3] to turn any async function into a service:
+
+```rust
+use layered::{Execute, Service};
+
+let greeter = Execute::new(|name: String| async move {
+    format!("Hello, {}!", name)
```
Vaiz:
```suggestion
        format!("Hello, {name}!")
```

crates/layered/README.md
```diff
@@ -0,0 +1,136 @@
...
+struct LogService<S>(S);
+
+impl<S, In: Send + std::fmt::Debug> Service<In> for LogService<S>
+where
+    S: Service<In>,
+{
+    type Out = S::Out;
+
+    async fn execute(&self, input: In) -> Self::Out {
+        println!("Input: {:?}", input);
```
Vaiz:
```suggestion
        println!("Input: {input:?}");
```

Vaiz: we should avoid using Debug formatting in examples

crates/layered/src/tower.rs
```diff
@@ -0,0 +1,267 @@
...
+pub struct AdapterLayer<L>(L);
+
+impl<L, S> Layer<S> for AdapterLayer<L>
+where
+    L: Layer<Adapter<S>> + Clone,
+{
+    type Service = Adapter<L::Service>;
+
+    fn layer(&self, inner: S) -> Self::Service {
+        // 1. Wrap the Oxidizer service to make it Tower-compatible
```
Vaiz: the mention of Oxidizer feels unnecessary

rust-toolchain.toml
martintmk: nope, thanks for catching

crates/layered/README.md
martintmk: true

crates/layered/src/intercept.rs
```diff
@@ -0,0 +1,708 @@
...
+/// Advanced usage of `Intercept` allows you to modify and observe inputs and outputs:
+///
+/// ```
+/// # use layered::{Execute, Stack, Intercept, Service};
+/// # async fn example() {
+/// let stack = (
+///     Intercept::<String, String, _>::layer()
+///         .on_input(|input| println!("request: {input}")) // input observers are called first
+///         .on_input(|input| println!("another: {input}")) // multiple observers supported
+///         .debug_input() // convenience method to print inputs with `dbg!`
```
Vaiz: `debug_input/output` sounds like something that should not be used in production. Would it make sense to gate it behind a feature flag?

Vaiz (approved):

crates/layered/src/intercept.rs
martintmk: it's something I have used internally, but I don't think we need to have it here nor expose publicly. Dropped.

Vaiz (approved):

crates/layered/src/intercept.rs
martintmk: I want to keep the initial version close to how we use it in oxidizer. We can refine the usage patterns and APIs in follow-up.

PR #192, opened 2026-01-13
feat: add AppError type for application level errors

Vaiz:
This pull request introduces `AppError`, a new application-level error type for the `ohno` crate, providing a simple catch-all error type for applications. 

To avoid the use of this type with libraries, `AppError` is gated behind `app_err` feature.

### DIFF
--- a/crates/ohno/src/app/macros.rs
+++ b/crates/ohno/src/app/macros.rs
@@ -3,63 +3,6 @@
 
 //! Macros for the `app` module.
 
-/// Return early with an [`AppError`](crate::AppError).
-///
-/// The macro accepts:
-/// - A string literal: `bail!("error message")`
-/// - An error expression: `bail!(MyAppError::new())`
-/// - A format string with arguments: `bail!("error: {value}")`
-///
-/// # Examples
-///
-/// ```rust
-/// use ohno::{AppError, bail};
-///
-/// fn check_value(x: i32) -> Result<(), AppError> {
-///     if x < 0 {
-///         bail!("value must be non-negative, got {x}");
-///     }
-///     Ok(())
-/// }
-/// ```
-///
-/// ```rust
-/// use ohno::{AppError, bail};
-///
-/// fn parse_config(data: &str) -> Result<(), AppError> {
-///    if data.is_empty() {
-///        bail!("config data cannot be empty");
-///    }
-///    Ok(())
-/// }
-/// ```
-///
-/// Bailing with an expression:
-///
-/// ```rust
-/// use ohno::{AppError, bail};
-///
-/// fn read_file(path: &str) -> Result<String, AppError> {
-///     bail!(std::io::Error::from(std::io::ErrorKind::PermissionDenied));
-/// }
-/// ```
-#[macro_export]
-#[cfg_attr(docsrs, doc(cfg(feature = "app-err")))]
-macro_rules! bail {
-    ($msg:literal $(,)?) => {
-        return Err($crate::AppError::new(format!($msg)))
-    };
-    ($fmt:expr, $($arg:tt)*) => {
-        return Err($crate::AppError::new(format!($fmt, $($arg)*)))
-    };
-    ($err:ident $(,)?) => {
-        return Err($crate::AppError::new($err))
-    };
-    ($err:expr $(,)?) => {
-        return Err($crate::AppError::new($err))
-    };
-}
-
 /// Construct an [`AppError`](crate::AppError) in place.
 ///
 /// The macro accepts:
@@ -112,4 +55,52 @@ macro_rules! app_err {
     ($err:expr $(,)?) => {
         $crate::AppError::new($err)
     };
+}
+
+/// Return early with an [`AppError`](crate::AppError).
+///
+/// The macro accepts:
+/// - A string literal: `bail!("error message")`
+/// - An error expression: `bail!(MyAppError::new())`
+/// - A format string with arguments: `bail!("error: {value}")`
+///
+/// # Examples
+///
+/// ```rust
+/// use ohno::{AppError, bail};
+///
+/// fn check_value(x: i32) -> Result<(), AppError> {
+///     if x < 0 {
+///         bail!("value must be non-negative, got {x}");
+///     }
+///     Ok(())
+/// }
+/// ```
+///
+/// ```rust
+/// use ohno::{AppError, bail};
+///
+/// fn parse_config(data: &str) -> Result<(), AppError> {
+///    if data.is_empty() {
+///        bail!("config data cannot be empty");
+///    }
+///    Ok(())
+/// }
+/// ```
+///
+/// Bailing with an expression:
+///
+/// ```rust
+/// use ohno::{AppError, bail};
+///
+/// fn read_file(path: &str) -> Result<String, AppError> {
+///     bail!(std::io::Error::from(std::io::ErrorKind::PermissionDenied));
+/// }
+/// ```
+#[macro_export]
+#[cfg_attr(docsrs, doc(cfg(feature = "app-err")))]
+macro_rules! bail{
+    ($($arg:tt)*) => (
+        return Err($crate::app_err!($($arg)*))
+    );
 }


crates/automation/src/lib.rs
```diff
@@ -48,14 +48,14 @@ pub fn list_packages(workspace_root: impl AsRef<Path>) -> Result<Vec<PackageMeta
         .arg("--no-deps")
         .current_dir(workspace_root.as_ref())
         .output()
-        .context("failed to execute cargo metadata")?;
+        .into_app_err("failed to execute cargo metadata")?;
```
martintmk, line 51: maybe just `app_err`?

crates/automation/src/lib.rs
```diff
@@ -48,14 +48,14 @@ pub fn list_packages(workspace_root: impl AsRef<Path>) -> Result<Vec<PackageMeta
...
 
     if !output.status.success() {
         let stderr = String::from_utf8_lossy(&output.stderr);
-        anyhow::bail!("cargo metadata failed: {stderr}");
+        ohno::bail!("cargo metadata failed: {stderr}");
```
martintmk, line 55:
```suggestion
        app_err!("cargo metadata failed: {stderr}");
```

martintmk: oki I see, we have both `bail!` and `app_err`, this is fine then

crates/ohno/src/app/error.rs
```diff
@@ -0,0 +1,119 @@
...
+
+    /// Returns the source error if this error wraps another error.
+    #[must_use]
+    pub fn source(&self) -> Option<&(dyn StdError + 'static)> {
+        StdError::source(&self.inner)
+    }
+
+    /// Finds the first source error of the specified type in the error chain.
+    #[must_use]
+    pub fn find_source<T: StdError + 'static>(&self) -> Option<&T> {
```
martintmk, line 143: could we also expose `root_source` that returns the root error?

crates/ohno/src/app/error.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// # Backtrace Capture
+    ///
+    /// Controlled by environment variables:
+    /// - `RUST_BACKTRACE=1` enables basic backtrace
+    /// - `RUST_BACKTRACE=full` enables full backtrace with all frames
+    pub fn backtrace(&self) -> &Backtrace {
+        self.inner.backtrace()
+    }
+
+    /// Returns top-level error message.
```
martintmk, line 167: So this does not include backtrace nor inner messages?

crates/ohno/src/lib.rs
```diff
@@ -220,13 +222,28 @@
...
+//! }
+//! ```
 
 #![doc(html_logo_url = "https://media.githubusercontent.com/media/microsoft/oxidizer/refs/heads/main/crates/ohno/logo.png")]
 #![doc(html_favicon_url = "https://media.githubusercontent.com/media/microsoft/oxidizer/refs/heads/main/crates/ohno/favicon.ico")]
 
 #[doc(hidden)]
 extern crate self as ohno;
 
+pub mod app;
```
martintmk: let's guard these with `app-err` feature. Libraries will almost never use the application errors.

martintmk (changes requested):

crates/automation/src/lib.rs
Vaiz: I prefer to have into as the main action here is transformation of the original error into AppError. The fact that it adds some additional context is not that important

crates/automation/src/lib.rs
Vaiz: yes, `app_err!` is for an easy in place construction, it can be used when you need to assign it to a variable. `bail!` macro also adds `return` that exits from function

crates/ohno/src/app/error.rs
Vaiz: sure, `root_cause` or `root_source`?

crates/ohno/src/app/error.rs
Vaiz: that's a bit tricky. It doesn't include backtrace and enrichment that was applied to that instance of `AppError`. But for source errors, we use their `Display` version which can be more verbose than expected.

I hope to improve that when https://github.com/rust-lang/rust/issues/99301 will get into stable

crates/ohno/src/app/error.rs
Vaiz: added `root_cause` method

crates/ohno/src/lib.rs
Vaiz: done

crates/automation/src/lib.rs
martintmk: oki, personally it just looks little ugly. Nothing to block PR about.

crates/ohno/src/app/error.rs
martintmk: thanks for context, it will be a way before we land that though

crates/ohno/Cargo.toml
```diff
@@ -23,6 +23,7 @@ allowed_external_types = ["ohno_macros::*"]
 all-features = true
 
 [features]
+app_err = []
```
martintmk:
```suggestion
app-err = []
```

to be aligned with other feature flags

martintmk (approved):

crates/automation/src/lib.rs
berkus: >main action here is transformation of the original error into AppError.

The main action _for this concrete call_ is to add context information, no?

crates/automation/src/lib.rs
```diff
@@ -41,21 +41,21 @@ pub struct Target {
 }
 
 /// List all workspace packages using `cargo metadata`
-pub fn list_packages(workspace_root: impl AsRef<Path>) -> Result<Vec<PackageMetadata>> {
+pub fn list_packages(workspace_root: impl AsRef<Path>) -> Result<Vec<PackageMetadata>, AppError> {
```
kate-shine, line 44: You could also feature gate your own Result type alias to simplify this in the same way as Anyhow does

crates/ohno/README.md
```diff
@@ -232,23 +233,45 @@ fn open_file(path: &str) -> Result<String, MyError> {
...
- [__cargo_doc2readme_dependencies_info]: ggGkYW0CYXSEGy4k8ldDFPOhG2VNeXtD5nnKG6EPY6OfW5wBG8g18NOFNdxpYXKEG6HS9SqShsaSGz42vAKwc2RWG7_xPji8SZ7NG3f-jaQ9sVJTYWSCgmRvaG5vZTAuMi4wgmtvaG5vX21hY3Jvc2UwLjIuMA
+ [__cargo_doc2readme_dependencies_info]: ggGkYW0CYXSEGy4k8ldDFPOhG2VNeXtD5nnKG6EPY6OfW5wBG8g18NOFNdxpYXKEG7xRMk7SDB66G1v2n-Rmf3nUGyeREro-k6dWGzjgBXgBcFi-YWSCgmRvaG5vZTAuMi4wgmtvaG5vX21hY3Jvc2UwLjIuMA
  [__link0]: https://doc.rust-lang.org/stable/std/?search=fmt::Display
  [__link1]: https://doc.rust-lang.org/stable/std/?search=fmt::Debug
- [__link10]: https://docs.rs/ohno_macros/0.2.0/ohno_macros/?search=enrich_err
+ [__link10]: https://doc.rust-lang.org/stable/std/macro.unreachable.html
  [__link11]: https://docs.rs/ohno_macros/0.2.0/ohno_macros/?search=enrich_err
- [__link12]: https://docs.rs/ohno/0.2.0/ohno/?search=Enrichable
+ [__link12]: https://docs.rs/ohno_macros/0.2.0/ohno_macros/?search=enrich_err
+ [__link13]: https://docs.rs/ohno/0.2.0/ohno/?search=Enrichable
```
berkus, line 267: This is an incredibly awkward link sorting order.

crates/ohno/examples/app/enrich_err.rs
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the MIT License.
+
+//! Demonstrates adding context to errors using `error_trace`.
+
+#![expect(clippy::unwrap_used, reason = "example code")]
+
+use ohno::AppError;
+use ohno::{bail, enrich_err};
+
+#[enrich_err("failed to load data for user {user_id}")]
```
berkus, line 11: any reason to use `enrich_err` instead of natural english `enrich_error`?

crates/ohno/src/app/macros.rs
```diff
@@ -0,0 +1,115 @@
...
+    ($fmt:expr, $($arg:tt)*) => {
+        return Err($crate::AppError::new(format!($fmt, $($arg)*)))
+    };
+    ($err:ident $(,)?) => {
+        return Err($crate::AppError::new($err))
+    };
+    ($err:expr $(,)?) => {
+        return Err($crate::AppError::new($err))
+    };
+}
```
kate-shine:
```suggestion
macro_rules! bail{
    ($($arg:tt)*) => (
        return Err(app_err!($($arg)*))
    );
}
```

Should work here

crates/ohno/src/test_util.rs
```diff
@@ -32,24 +34,22 @@ macro_rules! assert_error_message {
...
+        let test = move || {
+            if error_string == expected {
+                return ();
+            }
+            if let Some(remainder) = error_string.strip_prefix(expected) {
+                // backtrace, caused by, or error trace indicators
+                if remainder.starts_with("\n\nBacktrace:\n") || remainder.starts_with("\ncaused by: ") || remainder.starts_with("\n> ") {
+                    return ();
+                }
+            }
```
kate-shine, line 46: Maybe we could make this a bit leaner
```suggestion
if error_string.strip_prefix(expected).is_some_and(|remainder| remainder.starts_with("\n\nBacktrace:\n") || remainder.starts_with("\ncaused by: ") || remainder.starts_with("\n> ")) {
    return ();
}
```

crates/ohno/README.md
Vaiz: links are generated automatically by doc2readme

crates/ohno/examples/app/enrich_err.rs
Vaiz: to save two symbols I guess :)
that's what people wanted when we were discussing this API

crates/automation/src/lib.rs
Vaiz: considering that ohno is more general than AppError, I hesitate to add such an alias as it might be confusing for users. I would be down to add `AppResult` though

crates/ohno/src/test_util.rs
Vaiz: I prefer to have nested if instead of large closures.
either approach doesn't make much difference, so I'll the current one

kate-shine: I find nesting easier to get lost in, but that's just a personal opinion :)

crates/automation/src/lib.rs
kate-shine: AppResult would certainly work as well, although I wouldn't be worried about Result either - if you enable that feature, you declare intent to use AppError anyway, and if it's `type Result<T, E = AppError> = Result<T, E>`, you can always opt out of the AppError default anyway.

crates/ohno/src/app/macros.rs
Vaiz: applied

crates/automation/src/lib.rs
Vaiz: the problem with features is that they enabled across the whole workspace, so even if a library doesn't enable the feature it could be enabled by some other crate in the workspace
and next thing you know, you published a crate version that returns `ohno::Result` instead of `crate::Result` somewhere in public api

kate-shine: I don't think it's an issue. As long as they provide error type, it doesn't really matter for API, because alias isn't a different type. And if they don't, they'd need to explicitly use conversion methods like `into_app_err` to turn arbitrary error to `AppError`, which is way harder to do accidentally. But I don't insist on `Result` that much, `AppResult` is still fine.

kate-shine: (Obviously, it still shouldn't happen, but I feel like there is no possible functionality affecting problems with this)

Vaiz: > they'd need to explicitly use conversion methods like into_app_err

unfortunately, it's not the case. Any `std::error::Error` can be transformed into `AppError` using `?` operator

kate-shine: Good point

Vaiz: I'm not completely discard this idea, we might reconsider it later. Adding an alias is easier than removing :)

PR #193, opened 2026-01-13
feat: Add anyspawn crate with generic Spawner trait for spawn tasks for different async runtimes

schgoo:


crates/arty/src/spawner.rs
```diff
@@ -0,0 +1,191 @@
...
+/// });
+///
+/// match rx.await.unwrap() {
+///     Ok(value) => println!("Got {value}"),
+///     Err(e) => eprintln!("Task failed: {e}"),
+/// }
+/// # }
+/// ```
+#[derive(Debug)]
+pub enum Spawner {
```
martintmk:
```suggestion
pub struct Spawner(Inner)
```

I don't think we should be exposing enum. As a consumer of `Spawner`, I don't care how it was created, I just want to spawn tasks. Similar approach is used for `tick::Clock`,

crates/arty/src/spawner.rs
```diff
@@ -0,0 +1,191 @@
...
+            Self::Tokio => {
+                ::tokio::spawn(work);
+            }
+            Self::Custom(c) => (c.0)(Box::pin(work)),
+        }
+    }
+}
+
+/// Internal wrapper for custom spawn functions.
+pub struct CustomSpawner(Arc<SpawnFn>);
```
martintmk:
```suggestion
pub(crate) struct CustomSpawner(Arc<SpawnFn>);
```

internal?

crates/arty/src/spawner.rs
```diff
@@ -0,0 +1,191 @@
...
+///     let _ = tx.send(result);
+/// });
+///
+/// match rx.await.unwrap() {
+///     Ok(value) => println!("Got {value}"),
+///     Err(e) => eprintln!("Task failed: {e}"),
+/// }
+/// # }
+/// ```
+#[derive(Debug)]
```
martintmk:
```suggestion
#[derive(Debug, Clone)]
```

crates/arty/src/spawner.rs
```diff
@@ -0,0 +1,191 @@
...
+        F: Fn(BoxedFuture) + Send + Sync + 'static,
+    {
+        Self::Custom(CustomSpawner(Arc::new(f)))
+    }
+
+    /// Spawns an async task on the runtime.
+    ///
+    /// The task runs independently and its result is discarded. Use a channel
+    /// to retrieve results if needed.
+    pub fn spawn(&self, work: impl Future<Output = ()> + Send + 'static) {
```
martintmk:
```suggestion
    pub fn spawn_result<T: Send>(&self, work: impl Future<Output = T> + Send + 'static) {
```

Additional convenience method to also enable spawning result-based tasks? For tokio, this can be used directly, for custom spawner you can leverage something such as async one `futures::channel::oneshot` channel.

crates/arty/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+
+[package.metadata.docs.rs]
+all-features = true
+
+[features]
+default = ["tokio"]
+tokio = ["dep:tokio"]
+
+[dependencies]
+futures = { workspace = true, features = ["executor"] }
```
martintmk:
```suggestion
futures = { workspace = true, features = [] }
```

executor is kinda heavy, do we need to have it here by default?

crates/arty/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+tokio = { workspace = true, features = ["rt", "rt-multi-thread"], optional = true }
```
martintmk:
```suggestion
tokio = { workspace = true, features = ["rt"], optional = true }
```

could just rt be enough?

crates/arty/examples/tokio_basic.rs
```diff
@@ -0,0 +1,22 @@
...
+
+#[tokio::main]
+async fn main() {
+    let spawner = Spawner::Tokio;
+    let (tx, rx) = tokio::sync::oneshot::channel();
+
+    // Spawn a task that sends its result through a channel
+    println!("Spawning task...");
+    spawner.spawn(async move {
+        tokio::time::sleep(tokio::time::Duration::from_millis(50)).await;
```
martintmk: can we use tick here? :)

crates/arty/examples/tokio_basic.rs
```diff
@@ -0,0 +1,22 @@
...
+use arty::Spawner;
+
+#[tokio::main]
+async fn main() {
+    let spawner = Spawner::Tokio;
+    let (tx, rx) = tokio::sync::oneshot::channel();
+
+    // Spawn a task that sends its result through a channel
+    println!("Spawning task...");
+    spawner.spawn(async move {
```
martintmk:
```suggestion
    spawner.spawn_result(async move {
```

If this convenience method was added, the sample could be much simpler. (i don't really like the `spawn_result` name, I just wat to convey the idea)

crates/arty/src/spawner.rs
martintmk: https://microsoft.github.io/rust-guidelines/guidelines/libs/ux/index.html?highlight=clone#M-SERVICES-CLONE

crates/arty/src/spawner.rs
schgoo: So Inner is still an enum, right? We're just not exposing it.

crates/arty/Cargo.toml
schgoo: Nope, I don't actually need futures outside dev-dependencies anymore

crates/arty/Cargo.toml
schgoo: Yeah, I forgot to remove it

crates/arty/src/spawner.rs
```diff
@@ -180,10 +181,7 @@ impl Spawner {
     /// assert_eq!(result.await.unwrap(), 2);
     /// # }
     /// ```
-    pub fn run<T: Send + 'static>(
-        &self,
-        work: impl Future<Output = T> + Send + 'static,
-    ) -> oneshot::Receiver<T> {
+    pub fn run<T: Send + 'static>(&self, work: impl Future<Output = T> + Send + 'static) -> oneshot::Receiver<T> {
```
martintmk:
```
pub async fn run<T: Send + 'static>(&self, work: impl Future<Output = T> + Send + 'static) -> T
```

oneshot should be treated as implementation detail, no need to leak it outside

crates/arty/examples/tokio.rs
```diff
@@ -0,0 +1,31 @@
...
+    spawner.spawn({
+        let clock = clock.clone();
+        async move {
+            clock.delay(Duration::from_millis(10)).await;
+            println!("Background task completed!");
+        }
+    });
+
+    // Retrieve a result using run
+    let value = spawner.run(async { 1 + 1 }).await.unwrap();
```
martintmk: if passed future is does not return result, do we have to wrap it in `Result` ?

martintmk (approved):

crates/arty/src/spawner.rs
martintmk: yup, exactly, as what you did

crates/arty/examples/tokio.rs
schgoo: The oneshot returns Result<T, Canceled>. I'll have to think about what to do with the canceled case if we aren't exposing oneshot details

crates/arty/src/spawner.rs
schgoo: The problem is oneshot returns Result<T, Canceled>. What should we do if it's canceled? Basically, this happens if `work` panics. Is this the kind of place we would want to use `expect("spawned task panicked")`?

CHANGELOG.md
```diff
@@ -18,3 +18,4 @@ Please see each crate's change log below:
 - [`thread_aware_macros`](./crates/thread_aware_macros/CHANGELOG.md)
 - [`thread_aware_macros_impl`](./crates/thread_aware_macros_impl/CHANGELOG.md)
 - [`tick`](./crates/tick/CHANGELOG.md)
+- [`wing`](./crates/wing/CHANGELOG.md)
```
martintmk: this should be changed to `arty` (and also PR title)

README.md
```diff
@@ -40,6 +41,7 @@ These are the crates built out of this repo:
 - [`thread_aware_macros`](./crates/thread_aware_macros/README.md) - Macros for the `thread_aware` crate.
 - [`thread_aware_macros_impl`](./crates/thread_aware_macros_impl/README.md) - Macros for the `thread_aware` crate.
 - [`tick`](./crates/tick/README.md) - Provides primitives to interact with and manipulate machine time.
+- [`wing`](./crates/wing/README.md) - A generic future spawner compatible with any async runtime
```
martintmk:
```suggestion
- [`wing`](./crates/arty/README.md) - A generic future spawner compatible with any async runtime
```

geeknoid:
What's the link between your logo and this crate's use?

README.md
```diff
@@ -11,19 +11,21 @@
 
 This repository contains a set of crates that help you build robust highly scalable services in Rust.
 
-- [Crates](#crates)
-- [About this Repo](#about-this-repo)
+- [The Oxidizer Project](#the-oxidizer-project)
```
geeknoid: Why these changes here?

schgoo:
> What's the link between your logo and this crate's use?

Throwing the plane was like a fire-and-forget operation, also I had named this "wing" before. It probably isn't as relevant now. In fact, I might use it for uniflight, instead.

.spelling
```diff
@@ -38,6 +38,7 @@ btree_map
 buildable
 bytesbuf
 callee
+cancelled
```
geeknoid, line 40: Need to add 1 to the word count, which is on the first line of this file.

Don't ask me why, it's ridiculous. But it's needed.

README.md
schgoo: It was generated by my VS Code markdown extension. It automatically updates the table of contents to match the headers in the file. Should we keep it this way since it's more accurate?

.spelling
schgoo: Is it supposed to be equal to the number of lines in the file? Because then it would be 305? Also, the first line is a →. Is that going to break something?

crates/arty/README.md
```diff
@@ -0,0 +1,66 @@
+<div align="center">
+ <img src="./logo.png" alt="Arty Logo" width="96">
+
+# Arty
```
martin-kolinek: I kinda hoped that arty crate would be the actual runtime, and we could use something more generic for the abstraction layer (runtime-abstraction or something). Not a strong opinion, just feel like arty is a missed opportunity for just the abstraction layer.

crates/arty/src/spawner.rs
```diff
@@ -0,0 +1,258 @@
...
+    /// use arty::Spawner;
+    ///
+    /// # #[tokio::main]
+    /// # async fn main() {
+    /// let spawner = Spawner::tokio();
+    /// let result = spawner.run(async { 1 + 1 }).await;
+    /// assert_eq!(result, 2);
+    /// # }
+    /// ```
+    pub async fn run<T: Send + 'static>(&self, work: impl Future<Output = T> + Send + 'static) -> T {
```
martin-kolinek: How does this perform in comparison to tokio::spawn (which already returns an impl Future<T>)? Probably worth a benchmark as we don't want to provide an abstraction that slows things down unnecessarily.

crates/arty/README.md
schgoo: That makes sense. Ralf suggested I use arty. @ralfbiedert what do you think?

crates/arty/Cargo.toml
```diff
@@ -0,0 +1,36 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+[package]
+name = "arty"
+description = "A generic future spawner compatible with any async runtime"
```
ralfbiedert: I would call this a task spawner.

crates/arty/src/spawner.rs
schgoo: @martintmk was also suggesting we avoid using oneshot here for tokio, since it already has its own join handle and all that

crates/anyspawn/src/spawner.rs
```diff
@@ -0,0 +1,258 @@
...
+/// # }
+/// ```
+#[derive(Debug, Clone)]
+pub struct Spawner(SpawnerKind);
+
+#[derive(Debug, Clone)]
+enum SpawnerKind {
+    #[cfg(feature = "tokio")]
+    Tokio,
+    Custom(CustomSpawner),
```
ralfbiedert, line 101: Given that applications will decide whether they need this, I think this should be feature gated, so if you only ever use Tokio (or Custom) you don't pay for the other variants.

crates/arty/src/spawner.rs
```diff
@@ -0,0 +1,258 @@
...
+pub(crate) struct CustomSpawner(Arc<SpawnFn>);
+
+impl Debug for CustomSpawner {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("CustomSpawner").finish_non_exhaustive()
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
ralfbiedert: Can we move these tests to `tests/` please?

crates/anyspawn/src/spawner.rs
```diff
@@ -0,0 +1,258 @@
...
+///     .await;
+///
+/// match result {
+///     Ok(value) => println!("Got {value}"),
+///     Err(e) => eprintln!("Task failed: {e}"),
+/// }
+/// # }
+/// ```
+#[derive(Debug, Clone)]
+pub struct Spawner(SpawnerKind);
```
ralfbiedert, line 94: Given this has some slightly weird closures and might get more variants in the future, I think this should const-assert `Send`-iness.

crates/arty/src/spawner.rs
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    /// Spawns an async task on the runtime.
+    ///
+    /// The task runs independently and its result is discarded. Use
+    /// [`run`](Self::run) to retrieve results.
+    pub fn spawn(&self, work: impl Future<Output = ()> + Send + 'static) {
+        match &self.0 {
+            #[cfg(feature = "tokio")]
+            SpawnerKind::Tokio => {
```
geeknoid: I was expecting the implementation of tokio support to be just like the Custom type, except that the callbacks are provided by this library instead of by the caller. This would eliminate the need for match here. Basically, make all Spawners custom, but for tokio provide a "built-in custom"

crates/arty/src/spawner.rs
ralfbiedert: I'm not a big fan of `run`. I rather think your `spawn` should return its own `JoinHandle` (defined in `arty`) which you map for each of your variants. 

Also, I am wondering if we should give people on `arty` already an API  to express parallelism intent. Like have `spawn` and `spawn_anywhere` (names TBD). The former would make a best-effort to spawn on the same thread, while the latter would make a best-effort not to do that.

crates/arty/src/spawner.rs
schgoo: The benefit of doing it this way is that for tokio we can avoid the performance hit of allocating the future to heap. Although maybe tokio does that anyways, I'm not actually sure.

crates/anyspawn/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+authors.workspace = true
+license.workspace = true
+homepage.workspace = true
+repository.workspace = true
+
+[package.metadata.docs.rs]
+all-features = true
+
+[features]
+default = ["tokio"]
```
ralfbiedert: Default can be tokio, but would introduce `custom` feature here, compare above.

README.md
geeknoid: I'd rather nuke it, since the first entry in the table of contents ends up being the whole page, which is silly.

.spelling
geeknoid: Well, your PR is passing the spellcheck gate, so just leave it as-is...

crates/arty/README.md
geeknoid: any_arty?

ralfbiedert: @martin-kolinek you mean  the ox... umbrella becomes `arty`? I don't have super strong feelings either, would be onboard with that. I'm not a big fan of `runtime-abstraction` though, and also not really `wing`, the former is too long, the latter too unrelated. But let's discuss naming on Teams.

crates/arty/src/spawner.rs
schgoo: I was originally defining a JoinHandle type here, though that was back when I was also using a Spawner trait, too. Martin Tomka suggested simplifying it in this way. @martintmk any comments here?

schgoo: Ok I pushed a version with a `JoinHandle`. @martintmk is this any better than the way I was using JoinHandle before?

martintmk: Yup, this makes the Spawner more flexible as the returned future can now be named and stored. (`async T` vs `JoinHandle<T>`).

martin-kolinek: With this implementation, shouldn't spawn be implemented in terms of run? Or rather, I'd expect that with the current definition, we just need one function.

crates/anyspawn/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+//! # Features
+//!
+//! - `tokio` (default): Enables the [`Spawner::tokio`] constructor
+
+#![doc(html_logo_url = "https://media.githubusercontent.com/media/microsoft/oxidizer/refs/heads/main/crates/arty/logo.png")]
+#![doc(html_favicon_url = "https://media.githubusercontent.com/media/microsoft/oxidizer/refs/heads/main/crates/arty/favicon.ico")]
+
+mod spawner;
+
+pub use spawner::Spawner;
```
martin-kolinek, line 65: How does this compare with https://crates.io/crates/spawns? I assume we can avoid boxing for Tokio, so it's probably worth it, but we should look into that crate if we haven't already.

schgoo: Good find! From what I can see here are the main differences:
1. We have zero-cost for supported runtimes (i.e. Tokio right now, Arty in the future), spawns boxes everything.
2. Spawns has many more features i.e. task ids, cancellation API, pre-built compatibility with smol, async-global-executor
3. Spawns also has more complexity, you need to implement a trait, vs just using Spawner::custom(...)
4. Spawns stores the thread spawner in a thread-local context, whereas here we pass around a Spawner object. Obviously there are tradeoffs there. I'm not sure if having access to the Spawner provides any additional value. Would we ever need to mock it?

tl;dr Spawns has richer task management features, but also *probably* worse performance and greater complexity.

crates/arty/src/spawner.rs
schgoo: I added a benchmark to show the difference. There is no discernible overhead for Tokio.

crates/anyspawn/src/lib.rs
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#![cfg_attr(coverage_nightly, feature(coverage_attribute))]
+#![cfg_attr(docsrs, feature(doc_cfg))]
+#![warn(missing_docs)]
+
+//! Async runtime abstractions
```
geeknoid: Given the rebrand to the new crate name, perhaps this should be less grand?

crates/anyspawn/Cargo.toml
```diff
@@ -0,0 +1,53 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+[package]
+name = "anyspawn"
+description = "A generic task spawner compatible with any async runtime"
+version = "0.1.0"
+readme = "README.md"
+keywords = ["oxidizer", "async", "runtime", "futures"]
+categories = ["Asynchronous"]
```
geeknoid: Pretty sure the category slugs need to be lowervase.

crates/anyspawn/examples/custom.rs
```diff
@@ -0,0 +1,32 @@
...
+        })
+        .await;
+
+    // Retrieve a result by awaiting the JoinHandle
+    let handle = spawner.spawn(async { 1 + 1 });
+    let value = futures::executor::block_on(handle);
+    println!("Got result: {value}");
+
+    // Wait for background task
+    std::thread::sleep(Duration::from_millis(50));
```
martintmk: why do you need to keep this sleep here?

crates/anyspawn/examples/custom.rs
```diff
@@ -0,0 +1,32 @@
...
+    let () = spawner
+        .spawn(async {
+            std::thread::sleep(Duration::from_millis(10));
+            println!("Background task completed!");
+        })
+        .await;
+
+    // Retrieve a result by awaiting the JoinHandle
+    let handle = spawner.spawn(async { 1 + 1 });
+    let value = futures::executor::block_on(handle);
```
martintmk:
```suggestion
    let value = handle.await;
```

(we are already in async function)

crates/anyspawn/examples/tokio.rs
```diff
@@ -0,0 +1,33 @@
...
+            }
+        })
+        .await;
+
+    // Retrieve a result by awaiting the JoinHandle
+    let value = spawner.spawn(async { 1 + 1 }).await;
+    println!("Got result: {value}");
+
+    // Wait for background task
+    clock.delay(Duration::from_millis(50)).await;
```
martintmk: why is this needed here?

crates/anyspawn/src/spawner.rs
```diff
@@ -0,0 +1,187 @@
...
+    /// # async fn main() {
+    /// let spawner = Spawner::tokio();
+    /// let result = spawner.spawn(async { 42 }).await;
+    /// assert_eq!(result, 42);
+    /// # }
+    /// ```
+    #[must_use]
+    #[cfg(feature = "tokio")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "tokio")))]
+    pub fn tokio() -> Self {
```
martintmk: nit: This is the pattern we are using for constructor. Maybe align here too?

```suggestion
    pub fn new_tokio() -> Self {
```

and also `custom` -> `new_custom`.

crates/anyspawn/src/lib.rs
schgoo: Yes, good point.

crates/anyspawn/examples/custom.rs
schgoo: I don't 😅

crates/anyspawn/examples/tokio.rs
schgoo: Because I was making changes to too many different projects at once and forgot to delete this

martintmk (approved):

PR #202, opened 2026-01-16
feat!: allow to override backtrace policies per error

Vaiz:
This PR adds an ability to change backtrace collection policy for ohno errors.

`ohno::Error` derive macro now has additional `backtrace` attribute that can be set to `forced` or `disabled`

```rust
#[ohno::error]
#[backtrace(force)]
struct ForcedBacktraceError;

#[ohno::error]
#[backtrace(disabled)]
struct NoBacktraceError;
```

For manually created errors the policy can be set by `OhnoCoreBuilder::backtrace_policy`.

### DIFF
--- a/crates/ohno/src/core.rs
+++ b/crates/ohno/src/core.rs
@@ -67,6 +67,7 @@ impl OhnoCore {
 
     /// Creates a new [`OhnoCoreBuilder`](crate::OhnoCoreBuilder) for configuring an `OhnoCore` instance.
     #[must_use]
+    #[cfg_attr(test, mutants::skip)] // mutants replaces it with Default::default() which is equivalent to OhnoCoreBuilder::new()
     pub fn builder() -> crate::OhnoCoreBuilder {
         crate::OhnoCoreBuilder::new()
     }


crates/ohno/src/source.rs
```diff
@@ -5,7 +5,7 @@ use std::{error::Error as StdError, sync::Arc};
 
 /// The core error information - either a transparent error, a wrapped error, or none.
 #[derive(Debug, Clone)]
-pub enum Source {
+pub(crate) enum Source {
```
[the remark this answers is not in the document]
Vaiz: the module itself is private

ralfbiedert:
I sort of get why one might want to force backtraces on rare errors (although this might still cause performance issues if your guess is wrong), but preventing backtraces sounds like asking for trouble? Backtraces should not be enabled in production (unless rare circumstances) and lots of 3P backtraces possibly not being under our control, so when would that be applicable?

crates/ohno/src/lib.rs
```diff
@@ -95,6 +95,33 @@
 //! error (if any) is automatically shown as "Caused by:" in the error chain. If the inner error
 //! has no source, only the custom message is displayed.
 //!
+//! # Backtrace Policy
```
martin-kolinek, line 99: Do backtraces here refer to normal Rust backtraces, or the error traces created by the ohno::trace (or whatever it's called now) attribute?

Vaiz: it refers to `std::backtrace::Backtrace` which does some potentially heavy system calls

PR #203, opened 2026-01-16
feat: introduce seatbelt crate

martintmk:
Introduces seatbelt, a production-ready resilience middleware crate providing:

  - Retry: Automatic retries with configurable backoff (constant, linear, exponential) and jitter
  - Timeout: Cancellation of long-running operations with per-request overrides
  - Circuit Breaker: Prevents cascading failures by blocking requests when failure rates exceed thresholds

  Key features:
  - Runtime-agnostic design (works with any async runtime via tick::Clock)
  - Type-state pattern enforces correct configuration at compile time
  - OpenTelemetry metrics and structured logging support
  - Composable layers compatible with tower/layered middleware stacks

Example:

```rust
let clock = Clock::new_tokio();
  let context = ResilienceContext::new(&clock).name("my_service");

  let stack = (
      // Retry: recover from transient failures
      Retry::layer("retry", &context)
          .clone_input()
          .max_retry_attempts(3)
          .backoff(Backoff::Exponential)
          .recovery_with(|output: &Result<String, String>, _| match output {
              Ok(_) => RecoveryInfo::never(),
              Err(_) => RecoveryInfo::retry(),
          }),
      // Circuit breaker: stop calling failing services
      Breaker::layer("circuit", &context)
          .recovery_with(|output: &Result<String, String>, _| match output {
              Ok(_) => RecoveryInfo::never(),
              Err(_) => RecoveryInfo::retry(),
          })
          .rejected_input_error(|_, _| "circuit open".to_string()),
      // Timeout: bound operation duration
      Timeout::layer("timeout", &context)
          .timeout(Duration::from_secs(5))
          .timeout_output(|_| Err("timeout".to_string())),
      Execute::new(call_external_service),
  );

  let service = stack.into_service();
  let result = service.execute("request".to_string()).await;
```

### DIFF
--- a/crates/seatbelt/src/breaker/mod.rs
+++ b/crates/seatbelt/src/breaker/mod.rs
@@ -148,7 +148,7 @@
 //! | Minimum throughput | `100` executions | Minimum volume required before circuit can open | [`min_throughput`][BreakerLayer::min_throughput] |
 //! | Sampling duration | `30` seconds | Time window for calculating failure rates | [`sampling_duration`][BreakerLayer::sampling_duration] |
 //! | Break duration | `5` seconds | Duration circuit remains open before testing recovery | [`break_duration`][BreakerLayer::break_duration] |
-//! | Breaker isolation | Single global circuit | All inputs share the same circuit breaker state | [`breaker_id`][BreakerLayer::breaker_id] |
+//! | Breaker isolation | Shared circuit (default) | All inputs share the same circuit breaker state | [`breaker_id`][BreakerLayer::breaker_id] |
 //! | Half-open mode | `Reliable` | Gradual recovery with increasing probe percentages | [`half_open_mode`][BreakerLayer::half_open_mode] |
 //! | Enable condition | Always enabled | Circuit breaker protection is applied to all inputs | [`enable_if`][BreakerLayer::enable_if], [`enable_always`][BreakerLayer::enable_always], [`disable`][BreakerLayer::disable] |
 //!
@@ -251,8 +251,8 @@
 //!         .min_throughput(50)  // Require minimum 50 executions before considering circuit open
 //!         .sampling_duration(Duration::from_secs(60)) // Evaluate failures over 60-second window
 //!         .break_duration(Duration::from_secs(30))    // Stay open for 30 seconds before testing
-//!         // You can provide your own breaker ID logic if needed. The default is a single global
-//!         // circuit. By providing distinct IDs, you can have separate circuits for different inputs.
+//!         // You can provide your own breaker ID logic if needed. By default, all inputs share the
+//!         // same circuit. By providing distinct IDs, you can partition into separate circuits.
 //!         .breaker_id(|input| BreakerId::from(detect_breaker_id(input)))
 //!         // State change callbacks for monitoring and alerting
 //!         .on_opened(|output, _args| {
--- a/crates/seatbelt/src/breaker/layer.rs
+++ b/crates/seatbelt/src/breaker/layer.rs
@@ -250,7 +250,7 @@ impl<In, Out, S1, S2> BreakerLayer<In, Out, S1, S2> {
     /// function receives a reference to the input and returns a [`BreakerId`]
     /// identifying the circuit breaker instance to use.
     ///
-    /// **Default**: Single global circuit - all inputs share the same circuit breaker state
+    /// **Default**: Shared circuit - all inputs share the same circuit breaker state
     ///
     /// # Example
     ///


Cargo.toml
```diff
@@ -95,6 +96,11 @@ trybuild = { version = "1.0.114", default-features = false }
 typeid = { version = "1.0.3", default-features = false }
 windows-sys = { version = "0.61.2", default-features = false }
 xxhash-rust = { version = "0.8.15", default-features = false }
+fastrand = { version = "2.3.0", default-features = false, features = ["std"] }
```
geeknoid: Shoudl not have a feature specified here, do it in the crate's Cargo.toml. Same for http below.

geeknoid: Also, keep the dependencies sorted.

crates/seatbelt/Cargo.toml
```diff
@@ -0,0 +1,111 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+[package]
+name = "seatbelt"
+description = "Resilience and recovery mechanisms for services, applications, and libraries."
```
geeknoid: How about

"Resilience and recovery mechanisms for fallible operations.

crates/seatbelt/Cargo.toml
```diff
@@ -0,0 +1,111 @@
...
+version = "0.1.0"
+readme = "README.md"
+keywords = ["oxidizer", "resilience", "tower", "recovery", "retry", "circuit-breaker"]
```
geeknoid: Should "tower" be replaced with "layered"?

crates/seatbelt/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+#![cfg_attr(coverage_nightly, feature(coverage_attribute))]
+#![cfg_attr(docsrs, feature(doc_cfg))]
+#![doc(html_logo_url = "https://media.githubusercontent.com/media/microsoft/oxidizer/refs/heads/main/crates/seatbelt/logo.png")]
+#![doc(html_favicon_url = "https://media.githubusercontent.com/media/microsoft/oxidizer/refs/heads/main/crates/seatbelt/favicon.ico")]
+#![expect(
+    rustdoc::broken_intra_doc_links,
+    reason = "Too ugly to make 'live links' possible with the combination of features"
+)]
+
+//! Resilience and fault handling for applications and libraries.
```
geeknoid: Match the description in Cargo.toml and in the top-level README.md

crates/seatbelt/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+//!
+//! This crate helps applications handle transient faults gracefully through composable
```
geeknoid:
```suggestion
//! This crate helps you handle transient faults gracefully through composable
```

crates/seatbelt/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+//! - [`retry`]: Automatically retries failed operations with configurable backoff strategies.
+//! - [`circuit_breaker`]: Prevents cascading failures by stopping requests to unhealthy services.
+//!
+//! ## Features
+//!
+//! This crate supports several optional features that can be enabled to extend functionality:
+//!
+//! - `options`: Enables common APIs for building resilience middleware, including [`SeatbeltOptions`].
+//!   Requires [`tick`] for timing operations.
+//! - `service`: Re-exports common types for building middleware from [`layered`] crate.
```
geeknoid: How about calling this feature "layered"?

crates/seatbelt/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+
+#[cfg(any(feature = "retry", test))]
+pub mod retry;
+
+#[cfg(any(feature = "circuit-breaker", test))]
+pub mod circuit_breaker;
+
+/// Re-exported types from the [`layered`] crate.
+#[cfg(any(feature = "service", test))]
+pub mod service {
```
geeknoid: Also maybe call this module layered?

crates/seatbelt/src/retry/attempt.rs
```diff
@@ -0,0 +1,263 @@
...
+use std::fmt::Display;
+
+/// Represents a single attempt in a retry operation.
+///
+/// This struct tracks the current attempt index, and it provides methods to check if this is the
+/// first or last attempt.
+///
+/// The default attempt has:
+/// - `attempt`: 0 (first attempt, 0-based indexing)
+/// - `is_last`: true (indicating this is both the first and last attempt)
```
geeknoid: What? Why is is_last true for the first attempt?

crates/seatbelt/src/options/seatbelt_options.rs
```diff
@@ -0,0 +1,269 @@
...
+///
+/// > Note: The canonical, in‑depth description of the Oxidizer middleware / service
+/// > layering pattern lives in the [`layered`] crate documentation. Resilience
+/// > middleware in this crate is expected to follow that same pattern (layer types
+/// > implementing `Layer`, constructors taking the next service, no hidden global
+/// > state, etc.). If you need to evolve the pattern itself, update it in
+/// > `layered` and reference it here instead of diverging.
+#[derive(Debug)]
+#[non_exhaustive]
+pub struct SeatbeltOptions<In, Out> {
```
geeknoid: I worry that this is not serializable. Ultimately, I think we want Seatbelt to be configurable through JSON at the top. So I think we want to separate "options state" from things like Clock which is a runtime construct.

martintmk: This type was always intended to bundle shared runtime deps and configuration together. I am myself not super happy with the name either. What about one of these:

- `SeatbeltContext`
- `Context`
- `Pipeline`: super short, my thinking is that each resilience middleware is part of shared pipeline and we can just require to pass this shared pipeline when constructing resilience middlewares.

geeknoid: Looking at your usage examples, it feels a bit strange to be building this tuple and then calling build on it. I also don't like the fact the Execute thing means that a single pipeline cannot be used to do multiple different things.

Here's a chunk from your examples:

```rust
    let clock = Clock::new_tokio();
    let options = SeatbeltOptions::new(&clock)
        .pipeline_name("retry_advanced")
        .meter_provider(&meter_provider);

    // Define stack with retry layer
    let stack = (
        Retry::layer("my_retry", &options) 
        ... deleted for brevity ...
        Execute::new(send_request),
    );

    // Create the service from the stack
    let service = stack.build();

    let request = Request::builder().uri("https://example.com").body("value".to_string())?;

    match service.execute(request).await {
```

What if it looked like this instead:

```rust
    let clock = Clock::new_tokio();
    let pipeline = PipelineBuilder::new(&clock)
        .pipeline_name("retry_advanced")
        .meter_provider(&meter_provider);
        .add<Retry>("my_retry")...
        // as many .add as needed
    ).build();

    let request = Request::builder().uri("https://example.com").body("value".to_string())?;

    match pipeline.execute(|| send_request(request)).await {
```

Basically, you are building a pipeline using a PipelineBuilder. You specify the code to execute as a closure when executing the pipeline. So then I can use the pipeline with different closures as I want.

martintmk: Few notes:

When I initially started working in API design I did have something like `PipelineBuilder::new(&clock)`. The problem with this approach is that it requires tight coupling between individual middlewares and the `PipelineBuilder`. 

Additionally, due to our use of `Layer` our middleware will work with `ServiceBuilder` as defined by `https://docs.rs/tower/latest/tower/builder/struct.ServiceBuilder.html`.

The usage could be like:

```
ServiceBuilder::new()
    .layer(Retry::layer(...))
    .layer(Timeout::layer(...))
    .service(svc)
```

But the actual chaining is not that straightforward, because even this code `add<Retry>("my_retry")` requires additional customization of retries and it's difficult to do with just `PipelineBuilder`. We could do `add<Retry>("my_retry", |builder| { ... })` but I fear it would lead to callback spaghetti. 

Also this code:

```
pipeline.execute(|| send_request(request)).await
```

Does not scale too well because every call either incurs boxing or some magic around generics. It also distances our from  tower-service ecosystem and reduces the flexibility because we don't have access to input value anymore. 

IMHO, resilience middleware seeing and being able to modify the input value is a major improvement over Polly in c# that unlocks some major improvements (modifying inputs before retries, hedging for example)

crates/seatbelt/src/lib.rs
martintmk: I dropped the nesting, it's unnecessary and exported the types as is. Also this is now automatically exported whether any resilience middleware is enabled. (we do not have explicit "service" flag anymore).

crates/seatbelt/src/lib.rs
martintmk: refined the features, this one is dropped now.

crates/seatbelt/Cargo.toml
martintmk: updated

crates/seatbelt/src/options/seatbelt_options.rs
geeknoid: Well, then instead of SeatbeltOptions, I'd go with PipelineContext.

crates/seatbelt/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+        rustdoc::broken_intra_doc_links,
+        reason = "too ugly to make 'live links' possible with the combination of features"
+    )
+)]
+
+//! Resilience and recovery mechanisms for fallible operations.
+//!
+//! # Quick Start
+//!
+//! Add resilience to your services with just a few lines of code. **Retry** handles transient failures
```
geeknoid:
```suggestion
//! Add resilience to fallible operations, such as RPC calls over the network, with just a few lines of code. **Retry** handles transient failures
```

crates/seatbelt/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+//! let result = service.execute("input data".to_string()).await;
+//! # }
+//! # async fn my_string_operation(input: String) -> String {
+//! #     // Simulate processing that transforms the input string
+//! #     format!("processed: {}", input)
+//! # }
+//! # }
+//! ```
+//!
+//! > **Note**: Resilience middleware requires [`Clock`][tick::Clock] from the [`tick`] crate for timing
```
geeknoid: I'd move these two notes somewhere else. This is a bit too detailed for someone just trying to figure out what this crate does. Instead, I'd put a paragraph here that talks about the other kinds of resilience functionality the crate offers:

"In addition the classic retry and timeouts, this crate also delivers more sophisticated resilience patterns such as ..."

crates/seatbelt/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+//! > **Note**: This crate uses the [`layered`] crate for composing middleware. The middleware layers
+//! > can be stacked together using tuples and built into a service using the [`Stack`] trait.
+//!
+//! # Why?
+//!
+//! This crate provides production-ready resilience middleware with excellent telemetry for building
+//! robust distributed systems that can automatically handle timeouts, retries, and other failure
+//! scenarios.
+//!
+//! - **Runtime agnostic** - Works seamlessly across any async runtime. Use the same resilience
```
geeknoid: Make this the 3rd bullet. Being runtime-agnostic is not the most important thing this crate offers, right?

crates/seatbelt/src/breaker/mod.rs
```diff
@@ -0,0 +1,337 @@
+// Copyright (c) Microsoft Corporation.
```
geeknoid, line 1: This module is currently called "circuit" which I think is incorrect, this really ought to be circuit_breaker.

crates/seatbelt/src/options/seatbelt_options.rs
martintmk: done, the type is now called `PipelineContext`.

crates/seatbelt/src/breaker/mod.rs
martintmk: done

crates/seatbelt/Cargo.toml
martintmk: done

crates/seatbelt/src/retry/attempt.rs
martintmk: if single attempt is made,  it's both first and last one. Checking the last attempt is used to avoid some potentially expensive calls that are made in case the input needs to be prepared for retry. If we know there are no more attempts made, we can avoid this call. 

Ultimately, the retry strategy determines and creates the attempt, because it does know whether particular attempt is last or not.

crates/seatbelt/src/lib.rs
martintmk: true, updated

crates/seatbelt/src/lib.rs
martintmk: done, moved to overview section

crates/seatbelt/src/breaker/service.rs
```diff
@@ -0,0 +1,520 @@
+// Copyright (c) Microsoft Corporation.
```
geeknoid, line 1: I would definitely prefer for the source file names to match the type names (circuit_breaker.rs) rather than the concept name (service.rs). Makes it hard to find stuff when you're looking.

crates/seatbelt/src/retry/attempt.rs
geeknoid: I think the comment is confusing. How about instead:

```
/// - `attempt`: 0 (first attempt, 0-based indexing)
/// - `is_first`: true
/// - `is_last`: true
```

crates/seatbelt/src/retry/attempt.rs
```diff
@@ -0,0 +1,264 @@
...
+/// assert_eq!(last_attempt.index(), 2);
+///
+/// // Use the default attempt (single-shot operation)
+/// let default_attempt = Attempt::default();
+/// assert_eq!(default_attempt.index(), 0);
+/// assert!(default_attempt.is_first());
+/// assert!(default_attempt.is_last());
+/// ```
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Attempt {
```
geeknoid, line 44: Shouldn't this be in the retry module?

crates/seatbelt/src/shared/backoff.rs
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+/// Defines the backoff strategy used by resilience middleware for retry operations.
+///
+/// Backoff strategies control how delays between retry attempts are calculated, providing
+/// different approaches to spacing out retries to avoid overwhelming failing systems while
+/// balancing responsiveness and resource utilization.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum Backoff {
```
geeknoid: Shouldn't this be in the retry module?

crates/seatbelt/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+//! # async fn my_string_operation(input: String) -> String {
+//! #     // Simulate processing that transforms the input string
+//! #     format!("processed: {}", input)
+//! # }
+//! # }
+//! ```
+//!
+//! # Why?
+//!
+//! This crate provides production-ready resilience middleware with excellent telemetry for building
```
geeknoid, line 70: Preface with something like:

```
Communicating over a network is inherently fraught with problems. The network can go down at any time,
sometimes for a millisecond or two. The endpoint you're connecting to may crash or be rebooted,
network configuration may change from under you, etc. To deliver a robust experience to users, and to
achieve 5 or more 9s of availability, it is imperative to implement robust resilience patterns to
mask these transient failures.
```

crates/seatbelt/src/shared/backoff.rs
martintmk: My intention is to re-use this type and also `Attempt` for hedging so I just put it into the root.

crates/seatbelt/src/timeout/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+//! let service = stack.build();
+//! let result = service.execute("input".to_string()).await;
+//! # Ok(())
+//! # }
+//! # async fn my_operation(input: String) -> Result<String, io::Error> { Ok(input) }
+//! ```
+//!
+//! # Configuration
+//!
+//! The [`TimeoutLayer`] uses a type state pattern to enforce that all required properties are
```
geeknoid, line 41: So what's the benefit of using this pattern, rather then just requiring the parameters on the layer function in the first place?

```rust
Timeout::layer("timeout", &context, Duration::from_secs(30), |_| io::Error::new(io::ErrorKind::TimedOut, "operation timed out"));
```

That's a lot less typing...

crates/seatbelt/src/timeout/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+//! // Build and execute the service
+//! let service = stack.build();
+//! let result = service.execute("quick".to_string()).await?;
+//! # let _result = result;
+//! # Ok(())
+//! # }
+//! # async fn execute_unreliable_operation(input: String) -> Result<String, io::Error> { Ok(input) }
+//! ```
+//!
+//! ## Incomplete Configuration
```
geeknoid: I don't think you need to give examples of things that don't compile, that's a bit odd...

crates/seatbelt/src/timeout/mod.rs
martintmk: If you look at `TimeoutLayer` you can notice there are multiple ways on how the required parameters can be configured. This all depends on combination of generics. For example, if output is result-based one can call `TimeoutLayer::timeout_error`, or just `TimeoutLayer::timeout_output` if full flexibility is necessary. 

Requiring all required parameters to be specified when constructing the layer removes that flexibility or can lead to many layer constructors to be exposed. Same thing applies to circuit-breaker and retries. So rather than deal with many constructors, I embraced the builder approach for everything.

crates/seatbelt/src/timeout/mod.rs
martintmk: this was to demonstrate the type-safe pattern that does not allow to build incomplete middleware. Shall I drop it?

crates/seatbelt/src/lib.rs
martintmk: nice. added

crates/seatbelt/src/breaker/service.rs
martintmk: i tried this, but my preference was the unified structure because I know that under every resilience middleware there is a know strucure:

- middleware_module
  - service
  - layer
  - args
  - callbacks
  - telemetry
  - mod

This makes it really easy to orient around in this structure with one downside being that it makes the files less searchable.  But personally, I would like keep this skeleton. This work nicely for this kind of repeatable pattern. (unless there are strong objections)

crates/seatbelt/src/timeout/mod.rs
martintmk: Same thing with `RetryLayer`, depending on the scenario once can choose from these required methods:

- `RetryLayer::clone_input_with`
- `RetryLayer::clone_input`
- `RetryLayer::recovery_with`
- `RetryLayer::recovery`

Supporting all these combinations in constructor would be tricky and kinda verbose.

crates/seatbelt/src/timeout/mod.rs
geeknoid: Yeah, I don't think it's needed. I could show you many things that won't compile :-)

crates/seatbelt/src/timeout/mod.rs
geeknoid: I definitely see the value of the type state approach in general. My comment is specifically about required parameters. If they're truly required values, then using the type state pattern for those just seems like an extra burden, more keystrokes. And I think it's less intuitive when learning to use an API.

It looks as though I can create a timeout thingy easily. Oh, but I get a (in this case fairly cryptic) compile time error telling me something is wrong. I need to figure out that I must call these other two functions in order to supply parameters.

When it's just there in the constructor as an explicit parameter, it's unambiguous.

crates/seatbelt/src/shared/backoff.rs
geeknoid: It is bad form to reexport a type in different modules in the same crate? So you could have retry::Attempt and hedging::Attempt? It might be odd, but from a doc standpoint it increases discoverability IMO.

crates/seatbelt/src/timeout/mod.rs
martintmk: dropped

crates/seatbelt/src/shared/backoff.rs
martintmk: since this is not stable and we don't have hedging anyway, I just re-exported these types in the `retry` module. It's indeed cleaner now. Once the `hedging` module is added, we can also re-export it there.

crates/seatbelt/src/retry/attempt.rs
martintmk: done

ralfbiedert (commented):
Notes from discussion:

- [x] `PipelineContext` -> `ResilienceContext` (or so)
- [x] Hide `Set`, `NotSet`
- [x] Rename + document typestate variables
- [x] Remove re-exports of `layered`
- [ ] `.build()` -> `(in)to_service` (in layered)

crates/seatbelt/src/timeout/mod.rs
martintmk: We talked with Ralf regarding state pattern and he is generally fine with it. We agreed on doing some simplification regarding the name of the type states. 

Regarding the required parameters. Having a constructor to set these and abandon type-state pattern would work if there was only a single way to set these, but this is not the case here. These parameters can be set using multiple methods and these is no simple way to set these using a constructor.

crates/seatbelt/src/retry/attempt.rs
martintmk: yeah, indeed better. changed

crates/seatbelt/src/retry/args.rs
```diff
@@ -0,0 +1,161 @@
...
+    }
+}
+
+/// Arguments for the [`recovery_with`][super::RetryLayer::recovery_with] callback function.
+///
+/// Provides context for recovery classification.
+#[derive(Debug)]
+pub struct RecoveryArgs<'a> {
+    pub(super) attempt: Attempt,
+    pub(super) clock: &'a Clock,
```
geeknoid, line 39: Since clocks are cheap to clone and share common internals, why are you dealing with clock references here? That forces you to deal with lifetimes.

crates/seatbelt/src/retry/backoff copy.rs
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) Microsoft Corporation.
```
geeknoid: delete this file?

crates/seatbelt/src/retry/backoff.rs
```diff
@@ -0,0 +1,696 @@
...
+
+    /// Exponential backoff strategy that doubles delays with each attempt.
+    ///
+    /// **Example with `2s` base delay:** `2s, 4s, 8s, 16s, 32s, ...`
+    Exponential,
+}
+
+// The delay generation follows the Polly V8 implementation:
+//
+// https://github.com/App-vNext/Polly/blob/452b34ee1e3a45ccce156a6980f60901a623ee67/src/Polly.Core/Retry/RetryHelper.cs#L3
```
geeknoid: I'd rather we stand on our own and not reference Polly V8 in general. We should be the authoritative source at this point if anything :-)

crates/seatbelt/src/retry/constants.rs
```diff
@@ -0,0 +1,33 @@
...
+/// choice for transient faults in distributed systems and pairs well with jitter
+/// to avoid thundering herds.
+pub(super) const DEFAULT_BACKOFF: Backoff = Backoff::Exponential;
+
+/// Base delay for the backoff schedule; conservative 2 seconds by default.
+///
+/// A `2s` starting delay prevents aggressive retry storms during partial outages
+/// while still enabling fast recovery for short-lived failures. Workloads with
+/// different needs can override this via configuration.
+pub(super) const DEFAULT_BASE_DELAY: Duration = Duration::from_secs(2);
```
geeknoid: This is way too long, this should probably be 10ms.

THe vast majority of communication failures are due to instantaneous blips in the network rsther than outages. If you're service A calling service B and you end up waiting for 2 seconds to retry, then service A's SLA would be completely violated and it's likely that the caller of service A will have abandoned the operstion.

crates/seatbelt/src/retry/backoff.rs
```diff
@@ -0,0 +1,696 @@
...
+/// The default factor used for exponential backoff calculations for cases where jitter is not applied.
+const EXPONENTIAL_FACTOR: f64 = 2.0;
+
+/// Defines the backoff strategy used by resilience middleware for retry operations.
+///
+/// Backoff strategies control how delays between retry attempts are calculated, providing
+/// different approaches to spacing out retries to avoid overwhelming failing systems while
+/// balancing responsiveness and resource utilization.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum Backoff {
```
geeknoid, line 22: So what is the rationale for supporting Constant and Linear? When would someone want to use those? I'm wondering whether we should support these.

crates/seatbelt/src/retry/backoff.rs
```diff
@@ -0,0 +1,696 @@
...
+///
+/// - Successive delays are not a direct function of the immediately previous
+///   delay. Instead, each step samples a random phase (`t = attempt + U[0,1)`)
+///   and advances a smooth curve; we only take the delta from the previous
+///   position on that curve. This weakens correlation between consecutive
+///   samples and reduces synchronization across many callers.
+///
+/// What does `v2` mean?
+///
+/// - It refers to the second-generation formulation from
```
geeknoid: Let's get rid of this Polly stuff.

crates/seatbelt/src/retry/backoff.rs
```diff
@@ -0,0 +1,696 @@
...
+
+    Duration::try_from_secs_f64(secs).unwrap_or(Duration::MAX)
+}
+
+#[derive(Debug, Clone)]
+pub(super) struct BackoffOptions {
+    pub backoff_type: Backoff,
+    pub base_delay: Duration,
+    pub max_delay: Option<Duration>,
+    pub use_jitter: bool,
```
geeknoid, line 187: What is the use case for supporting non-jitter retries? Should we support this at all?

crates/seatbelt/src/retry/layer.rs
```diff
@@ -0,0 +1,786 @@
...
+
+    /// Configures infinite retry attempts.
+    ///
+    /// This setting will cause the operation to be retried indefinitely until it succeeds
+    /// or the retry is aborted by other means (e.g., cancellation, timeout).
+    ///
+    /// **Warning**: Use with caution as this can cause infinite loops if the operation
+    /// consistently fails.
+    #[must_use]
+    pub fn infinite_retry_attempts(mut self) -> Self {
```
geeknoid: What's the scenario for this? This just seems like a bad idea.

crates/seatbelt/src/retry/layer.rs
```diff
@@ -0,0 +1,786 @@
...
+    /// This call replaces any previous condition.
+    ///
+    /// **Default**: Always enabled
+    ///
+    /// # Arguments
+    ///
+    /// * `is_enabled` - Function that takes a reference to the input and returns
+    ///   `true` if retry protection should be enabled for this request
+    #[must_use]
+    pub fn enable_if(mut self, is_enabled: impl Fn(&In) -> bool + Send + Sync + 'static) -> Self {
```
geeknoid, line 228: I wonder about the real-world utility of all these callbacks, given that they are all synchronous. If there's a need to touch any complex state in order to implement these callbacks, the application would need to deal with all sorts of complexities to get this to compose properly.

There's no way to make these async-friendly?

crates/seatbelt/src/retry/service.rs
```diff
@@ -0,0 +1,652 @@
...
+        // Only ever delay if we have a next attempt
+        if matches!(flow_control, ControlFlow::Continue(_)) {
+            self.clock.delay(retry_delay).await;
+        }
+
+        flow_control
+    }
+}
+
+enum RecoverableKind {
```
geeknoid: Do we need this enum?

crates/seatbelt/src/context.rs
```diff
@@ -0,0 +1,165 @@
...
+    logs_enabled: bool,
+    _in: std::marker::PhantomData<fn() -> In>,
+    _out: std::marker::PhantomData<fn() -> Out>,
+}
+
+impl<In, Out> ResilienceContext<In, Out> {
+    /// Create a context with a clock. Initializes with `name = "default"`.
+    pub fn new(clock: impl AsRef<Clock>) -> Self {
+        Self {
+            clock: clock.as_ref().clone(),
```
geeknoid, line 30: Should DEFAULT_PIPELINE_NAME to DEFAULT_CONTEXT_NAME?

Also, would it make sense to always require the user to supply a name instead of making it optional and supporting a default?

crates/seatbelt/src/context.rs
```diff
@@ -0,0 +1,165 @@
...
+        Self {
+            meter: Some(crate::metrics::create_meter(provider)),
+            ..self
+        }
+    }
+
+    /// Enable structured logging for resilience events.
+    #[must_use]
+    #[cfg(any(feature = "logs", test))]
+    pub fn enable_logs(self) -> Self {
```
geeknoid: In other APIs in this crate, you use a different pattern for these kinds of booleans:

```rust
pub fn use_jitter(&self, use_jitter: bool)
```

I think we probably want to use the same pattern as use_jitter, which makes it considerably easier to use the API when the behavior is defined in config since you can just pass the boolean from configuration to the function. When all you have is "enable_logging", then you need to have different code completely to deal with the "with logging" vs. "Without logging" cases, which breaks the fluent programming model.

crates/seatbelt/src/retry/layer.rs
```diff
@@ -0,0 +1,786 @@
...
+    ///
+    /// This is required because Rust's ownership model doesn't allow reusing
+    /// the same input value across multiple attempts.
+    ///
+    /// # Arguments
+    ///
+    /// * `clone_fn` - Function that takes a reference to the input and [`CloneArgs`]
+    ///   containing context about the retry attempt, and returns an optional cloned input
+    #[must_use]
+    pub fn clone_input_with(
```
geeknoid, line 140: Feels strange to me to support this pattern of "we'll let you clone even though you don't support the Clone trait".

Also, I don't understand the comment about the Rust ownership system. When you say "This is required", the "This" means "cloning of inputs" or the fact we're invoking a callback function?

crates/seatbelt/src/breaker/args.rs
```diff
@@ -0,0 +1,161 @@
...
+use crate::circuit_breaker::PartitionKey;
+
+/// Arguments for the [`recovery_with`][super::CircuitLayer::recovery_with] callback function.
+///
+/// Provides context for recovery classification in the circuit breaker.
+#[derive(Debug)]
+#[non_exhaustive]
+pub struct RecoveryArgs<'a> {
+    pub(crate) partition_key: &'a PartitionKey,
+    pub(crate) clock: &'a Clock,
```
geeknoid, line 17: As in the retry case, why is this a clock reference?

crates/seatbelt/src/circuit_breaker/partition_key.rs
```diff
@@ -0,0 +1,206 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use std::borrow::Cow;
+use std::collections::hash_map::DefaultHasher;
+use std::fmt::Display;
+use std::hash::Hasher;
+
+/// Key that identifies a partition for which a separate circuit breaker instance is maintained.
```
geeknoid: Can you give a practical example of the expected scope of a partition? Should you have a partition for every service you interact with, or for every distinct kind of APIs you call, or every call to an API, or...?

And what is the 'lifetime' of a partition? If I use integers, can I reuse integers over time? Or should they be monotonically increasing?

crates/seatbelt/src/circuit_breaker/mod.rs
```diff
@@ -0,0 +1,315 @@
...
+//! - [`RecoveryKind::Retry`][crate::RecoveryKind::Retry]
+//! - [`RecoveryKind::Unavailable`][crate::RecoveryKind::Unavailable]
+//!
+//! # Partitioning
+//!
+//! Circuit breakers can maintain separate circuit states for different logical groups of requests
+//! by providing a [`partition_key`][CircuitLayer::partition_key] function. This allows
+//! the creation of multiple independent circuits based on the input properties.
+//!
+//! For example, a typical scenario where partitioning is useful is HTTP request where the partition key
```
geeknoid: I wonder if the whole model would be more cohesive if instead of talking about partitions, you talked about circuits?

PartitionKey becomes CircuitId. As a developer, you create a circuit, against which you apply a circuit breaker.

crates/seatbelt/src/breaker/mod.rs
```diff
@@ -0,0 +1,315 @@
...
+//! - [`RecoveryKind::Unavailable`][crate::RecoveryKind::Unavailable]
+//!
+//! # Partitioning
+//!
+//! Circuit breakers can maintain separate circuit states for different logical groups of requests
+//! by providing a [`partition_key`][CircuitLayer::partition_key] function. This allows
+//! the creation of multiple independent circuits based on the input properties.
+//!
+//! For example, a typical scenario where partitioning is useful is HTTP request where the partition key
+//! is extracted from the request scheme, host, and port. This allows isolation of circuit states
```
geeknoid: I think you probably want to be a bit more prescriptive here. What are meaningful partitions? Are partitions long-lived? What are examples of bad partitions?

crates/seatbelt/src/circuit_breaker/mod.rs
```diff
@@ -0,0 +1,315 @@
...
+//! The circuit breaker middleware uses the following default values when optional configuration
+//! is not provided:
+//!
+//! | Parameter | Default Value | Description | Configured By |
+//! |-----------|---------------|-------------|---------------|
+//! | Failure threshold | `0.1` (10%) | Circuit opens when failure rate exceeds this percentage | [`failure_threshold`][CircuitLayer::failure_threshold] |
+//! | Minimum throughput | `100` requests | Minimum request volume required before circuit can open | [`min_throughput`][CircuitLayer::min_throughput] |
+//! | Sampling duration | `30` seconds | Time window for calculating failure rates | [`sampling_duration`][CircuitLayer::sampling_duration] |
+//! | Break duration | `5` seconds | Duration circuit remains open before testing recovery | [`break_duration`][CircuitLayer::break_duration] |
+//! | Partitioning | Single global circuit | All requests share the same circuit breaker state | [`partition_key`][CircuitLayer::partition_key] |
```
geeknoid: Is that a good idea? If I can't talk to A for a while, I also won't be able to talk to B and C?

crates/seatbelt/src/retry/layer.rs
martintmk: Once there is a need to have async callbacks, we will just expose `enable_if_async` (or so) overloads. 

But we should be careful about promoting asynchrony here, as this can lead to lower performance and additional allocations because this is on a hot path. Imho, from the user perspective, these callbacks should be really fast and if the conditions are dependent on an async state, it should be cached by caller.

crates/seatbelt/src/retry/layer.rs
martintmk: This is for case of some background jobs that is executed infrequently but it always needs to be alive. Infinite retries are appropriate for that case.  

For example: background sync of list of tenants to local cache

crates/seatbelt/src/retry/backoff.rs
martintmk: :( (I wanted to keep the memory alive :D)

crates/seatbelt/src/retry/constants.rs
martintmk: I think this might be case of `RetryLayer::fast_first_retry(Duration::from_ms(10))` configuration switch that overrides the delay for the very first retry.

crates/seatbelt/src/retry/args.rs
martintmk: I would not consider cloning the clock as super cheap, this still incurs atomic operation and this is done on a hot path.
While dealing with lifetimes is annoying for some scenarios, this is not it. Lifetime is not visible for calling user, nor does it infects the call chain so I think it's fine here.

crates/seatbelt/src/circuit_breaker/mod.rs
martintmk: these break duration (and all other config) apply to circuit partitions. IE, if partition A is broken, B is not affected.

crates/seatbelt/src/circuit_breaker/mod.rs
martintmk: Maybe `BreakerId` then?

crates/seatbelt/src/retry/service.rs
martintmk: indeed not required, dropped

crates/seatbelt/src/context.rs
martintmk: true, renamed to `use_logs` and `use_metrics`.

crates/seatbelt/src/circuit_breaker/mod.rs
martintmk: renamed

crates/seatbelt/src/breaker/mod.rs
martintmk: refined the docs

crates/seatbelt/src/retry/constants.rs
geeknoid: I don't consider 10ms to be particularly fast.

As a general rule, the service wants to set the total execution time for any of its requests to never exceed its own SLA, which will typically be less than 1 second.

I think different environments affect this somewhat. In a service->service world where both services involved are in the same environment (e.g. both in Azure) then the retry latency should be smaller. If the traffic has to travel over the Internet however, then you probably want a longer initial retry and longer SLA period to accommodate longer latencies.

crates/seatbelt/src/retry/layer.rs
geeknoid: I think in that case, you want the ultimate loop to be in the application and not in the infra, to enable better diagnostics and to avoid the infra being the source of infinite loop in the case of failure.

I suggest to take this out and not encourage the pattern. But if the user wants to do this, they can just use some ridiculously large retry count and timeout values.

crates/seatbelt/src/retry/backoff.rs
martintmk: I have seen uses of these across .NET resilience, particular for some low-latency clients. I would keep these, to at least show we do support some configurability in this area.

crates/seatbelt/src/retry/backoff.rs
martintmk: dropped mention of polly from the code

crates/seatbelt/src/retry/constants.rs
martintmk: Since this is a new lib, not burdened by legacy, I think changing the default to 10ms is fine. It indeed is better for S2S scenarios. I have also included a note to use different default for client to service scenarios.

crates/seatbelt/src/retry/backoff copy.rs
martintmk: done

crates/seatbelt/src/retry/layer.rs
martintmk: One concrete scenario is HttpRequest that contains streamed request body. The body cannot implement the `Clone` trait because in this scenario cloning is not possible. (we cannot buffer potentially GBs of data into memory). 

On the other hand, there are cases where the HTTP request contains buffered bodies that can be cloned. So the cloning is indeed conditional based on the state of request body. For this reason we also need to be aware of conditional clone. We also have a shortcut for trivially cloneable inputs.

I have refined the docs for the method.

crates/seatbelt/src/retry/layer.rs
martintmk: Dropped. If  anything, we can add this back when someone requests it.

crates/seatbelt/src/circuit_breaker/partition_key.rs
martintmk: Refined the docs (see `BreakerId`). 

> And what is the 'lifetime' of a partition?

Currently, partition is long-lived inner state of circuit breaker. We are not dealing with cleanup right now. No re-using of keys if you want to have distinct partitions.

crates/seatbelt/src/context.rs
martintmk: renamed

crates/seatbelt/Cargo.toml
```diff
@@ -0,0 +1,118 @@
...
+    "tick::clock::Clock",
+    "tower_layer::Layer",
+]
+
+[package.metadata.docs.rs]
+all-features = true
+
+[features]
+default = []
+timeout = []
```
ralfbiedert, line 36: Should we really have these features? 

I don't think the 3 modules have a meaningful impact on compile time, and are an inherent feature of the crate. That one extra dependency (`fastrand`) is lightweight and has zero deps, while your other mandatory dependencies each come with a handful. 

Essentially, you force people to think (and make them go back and forth fixing their Cargo.toml) without a compile-time benefit.

crates/seatbelt/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+//!             _ => RecoveryInfo::never(),
+//!         }),
+//!     // Timeout middleware: Cancels operations that take too long
+//!     Timeout::layer("timeout", &context)
+//!         .timeout_output(|_| "operation timed out".to_string())
+//!         .timeout(Duration::from_secs(30)),
+//!     // Your core business logic
+//!     Execute::new(my_string_operation),
+//! )
+//!     .into_service();
```
ralfbiedert, line 51: I think this example should show the import `use layered::Stack`. The other imports are mostly self-evident (e.g., `clock` from the used types, but given `into_service` is an otherwise invisible extension trait the examples should point at its origin.

crates/seatbelt/src/breaker/half_open_mode.rs
```diff
@@ -0,0 +1,171 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use std::time::Duration;
+
+use crate::breaker::constants::MIN_SAMPLING_DURATION;
+use crate::breaker::engine::probing::ProbesOptions;
+
+/// Defines the behavior of the circuit breaker when transitioning from half-open to closed state.
```
ralfbiedert: Can we shorten this to 

Behavior of the circuit breaker when transitioning from half-open to closed state.

crates/seatbelt/benches/breaker.rs
```diff
@@ -0,0 +1,60 @@
...
+use alloc_tracker::{Allocator, Session};
+use criterion::{Criterion, criterion_group, criterion_main};
+use futures::executor::block_on;
+use layered::{Execute, Service, Stack};
+use seatbelt::breaker::Breaker;
+use seatbelt::{RecoveryInfo, ResilienceContext};
+use tick::Clock;
+
+#[global_allocator]
+static ALLOCATOR: Allocator<std::alloc::System> = Allocator::system();
```
ralfbiedert, line 13: Why are we doing this?

martintmk: to track allocations in benchmarks

crates/seatbelt/src/breaker/half_open_mode.rs
martintmk: done

crates/seatbelt/src/lib.rs
martintmk: done

crates/seatbelt/Cargo.toml
martintmk: each of the modules is quite chunky, for cases when someone does not need circuit breaker, we should not force them to compile it. I have made a change though to enable all middlewares by default for straightforward scenarios.

crates/seatbelt/src/circuit_breaker/mod.rs
geeknoid: But you mention here "single global circuit" and that's what I'm referring to when I ask "is this a good idea".

martintmk: Ah yes, these are the defaults. We cannot assume default partitioning strategy for arbitrary inputs. For some common workloads, from example HTTP stack, we do actually provide different defaults. In case of HTTP requests, the partition key is the combination of (scheme + host + port).

crates/seatbelt/src/breaker/mod.rs
```diff
@@ -0,0 +1,312 @@
...
+//! The circuit breaker middleware uses the following default values when optional configuration
+//! is not provided:
+//!
+//! | Parameter | Default Value | Description | Configured By |
+//! |-----------|---------------|-------------|---------------|
+//! | Failure threshold | `0.1` (10%) | Circuit opens when failure rate exceeds this percentage | [`failure_threshold`][BreakerLayer::failure_threshold] |
+//! | Minimum throughput | `100` executions | Minimum volume required before circuit can open | [`min_throughput`][BreakerLayer::min_throughput] |
+//! | Sampling duration | `30` seconds | Time window for calculating failure rates | [`sampling_duration`][BreakerLayer::sampling_duration] |
+//! | Break duration | `5` seconds | Duration circuit remains open before testing recovery | [`break_duration`][BreakerLayer::break_duration] |
+//! | Breaker isolation | Single global circuit | All inputs share the same circuit breaker state | [`breaker_id`][BreakerLayer::breaker_id] |
```
geeknoid: As I mentioned elsewhere, I'm worried that having the defaults be "Single global circuit" will lead to incorrect behavior in the app. If the app doesn't override the defaults, they will get interference between interactions with one service and a different one.

Could we eliminate the notion of "single global circuit" and instead go the opposite route? Every interaction to a given service is tracked as an individual circuit?

martintmk: > Could we eliminate the notion of "single global circuit" and instead go the opposite route? Every interaction to a given service is tracked as an individual circuit?

IMHO I would not do this, in most scenarios single circuit is actually the right scenario. Imagine a some HTTP client that communicates with a single server. Circuit partitioning only becomes relevant when you actually have multiple endpoints to communicate with and even that is a advanced hedging scenario. 

What we see in most cases is indeed a single global circuit. Also there is no notion of "service instance". Service, as circuit breaker sees it is just some transparent inner handler and it cannot make any assumptions about the partitioning strategy. (nor is CB responsible for creation of inner services)

geeknoid: It feels as though we have a terminology problem.

When the default is "single global circuit", what do you mean by "global"? When I read "global" in my mind I think "global to the process". I think you mean something different, right?

martintmk: I see now, maybe the term "global" is wrong. There are no process-level statics in seatbelt. When talking about global circuit, I mean the internal circuit instance that is used across all service executions.

This is the default behavior, unless partitioning is enabled  through customizing the `Breaker::breaker_id` callback.

martintmk: (refined the docs)

PR #207, opened 2026-01-19
feat(layered): add typed InterceptFuture for tower Service impl

martintmk:
Replace `Pin<Box<dyn Future<...>>>` with a named `InterceptFuture<Out>` type in the tower Service implementation for Intercept middleware.

### DIFF
--- a/.ado/pipelines/publish-testing.yml
+++ b/.ado/pipelines/publish-testing.yml
@@ -1,0 +1,31 @@
+trigger:
+- main
+
+parameters:
+- name: validationOsList
+  type: object
+  default:
+  - linux
+- name: validationProfiles
+  type: object
+  default:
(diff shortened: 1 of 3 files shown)


crates/layered/src/intercept.rs
```diff
@@ -136,6 +140,28 @@ where
     }
 }
 
+/// Future returned by [`Intercept`] when used as a tower [`Service`](tower_service::Service).
+#[cfg(any(feature = "tower-service", test))]
+pub struct InterceptFuture<Out> {
```
psandana, line 145: Is it an `InterceptedFuture`?

psandana: Oh, now I checked it is part of the Intercept feature. 👍

PR #228, opened 2026-01-23
chore: simplify the tick internals

martintmk:
Dropping `GlobalState` since it's unnecessary layer.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,13 +1,23 @@
 version: 2
 updates:
-  - package-ecosystem: "cargo"
-    directory: "/"
-    schedule:
-      interval: "monthly"
-    # only security updates will be opened automatically
-    open-pull-requests-limit: 0
+- package-ecosystem: "cargo"
+  directory: "/"
+  schedule:
+    interval: "weekly"
+    day: "monday"
+    time: "07:00"
+    timezone: "UTC"
+  commit-message:
+    prefix: "chore: "
+  ignore:
+    # ignore minor updates for tokio as they are semver compatible
+    - dependency-name: "tokio"
+      versions: ["1.x"]
+    # ignore patch updates for all dependencies
+    - dependency-name: "*"
+      update-types: ["version-update:semver-patch"]
 
-  - package-ecosystem: "github-actions"
-    directory: "/"
-    schedule:
-      interval: "monthly"
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+    interval: "monthly"


crates/tick/src/runtime/clock_driver.rs
```diff
@@ -39,7 +39,7 @@ impl ClockDriver {
         match next_timer {
             Some(next) => Ok(Some(next)),
             // Check if this is the last reference to the clock state
-            None if Arc::strong_count(&self.0) == 1 => Err(ClockGone::new()),
+            None if self.0.ownership_count() == 1 => Err(ClockGone::new()),
```
Vaiz: if you add `is_unique` method, it would be more specific and less error-prone

crates/tick/src/state.rs
```diff
@@ -38,7 +52,7 @@ pub(crate) struct SynchronizedTimers {
     // with `RefCell`. The `RefCell` variant is around 7% faster. In practice, in real applications,
     // the difference is negligible. The real performance improvement comes from isolating the `Clock` to each thread.
     // This reduces lock contention and provides linear scalability.
-    timers: Mutex<Timers>,
+    timers: Arc<Mutex<Timers>>,
```
Vaiz, line 56: https://crates.io/crates/triomphe - in case we want to shave off a couple of bytes

crates/tick/src/runtime/clock_driver.rs
martintmk: the `is_unique` is little ambiguous, I would stick with current name.

crates/tick/src/state.rs
martintmk: will keep this in mind, but want to keep the amount of external deps to minimum for now

crates/tick/src/runtime/clock_driver.rs
ralfbiedert: I generally prefer if we align with `std` naming. This one is a bit tricky, `Arc` has a stable `strong_count` (which I could see becoming `ownership_count` here). It also has an unstable `is_unique`. 

If we think it ever make sense to query for actual `ownership_count` I'm inclined to keep it. 

If not, I think `is_unique` is a better name. Then we should go for that and only revisit if the `Arc` item weren't stabilized.

martintmk: FYI, I need similar APIs for `thread_aware` crate to see if there are any owners of an instance for the current thread. Without it, I cannot adopt it here. 

(will change to is_unique since I am now in the minority :D)

PR #231, opened 2026-01-26
chore: publish ohno 0.3.0

Vaiz:
## [0.3.0] - 2026-01-26

- ✨ Features

  - add `AppError::downcast_ref` and `Into<std::error::Error>` ([#225](https://github.com/microsoft/oxidizer/pull/225))

### DIFF
--- a/scripts/release-crate.ps1
+++ b/scripts/release-crate.ps1
@@ -436,6 +436,38 @@ function Write-Changelog {
     Write-Host "✅ Changelog created at '$changelogFile'."
 }
 
+function Update-Readme {
+    param(
+        [string]$crateName,
+        [string]$crateFolder
+    )
+
+    $readmeTemplate = Join-Path $crateFolder "../README.j2"
+    if (-not (Test-Path $readmeTemplate)) {
+        Write-Warning "README template not found at '$readmeTemplate'. Skipping README generation."
+        return
+    }
+
+    if (-not (Test-CommandExists -command "cargo-doc2readme")) {
+        Write-Warning "cargo-doc2readme is not installed. Skipping README generation. Install with: cargo install cargo-doc2readme"
+        return
+    }
+
+    Write-Host "📝 Updating README.md..."
+    Push-Location $crateFolder
+    try {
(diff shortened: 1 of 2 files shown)


scripts/generate-docs.ps1
```diff
@@ -1,7 +0,0 @@
-# Copyright (c) Microsoft Corporation.
```
geeknoid, line 1: Did you mean to delete this script? I think Ralf just added this recently.

scripts/release-crate.ps1
```diff
@@ -69,6 +69,7 @@ $script:TypeGroupMapping = @{
 
 # Maps the final group key to a user-friendly header in the changelog.
 $script:HeaderNameMapping = @{
+    'breaking'      = '⚠️ Breaking';
```
geeknoid, line 72: This keyword needs to be added to the list of terms the list of allowed keywords for the pr-title gate in main.yml

scripts/generate-docs.ps1
Vaiz: yes, it's replaced with `just docs`

scripts/release-crate.ps1
Vaiz: that's not a keyword used in PR titles, breaking changes are indicated by adding `!` after change type, e.g. `feat!:`

geeknoid: Does this work in our setup? I didn't implement this functionality in the support scripts I think.

Vaiz: it follows this convention: https://www.conventionalcommits.org/en/v1.0.0/#summary
and the CI doesn't complain about such titles, for example #225

geeknoid: It's not about the convention, it's about our implementation of the convention, which is all handled via scripts and GitHub workflows.

Vaiz: what other tooling do we currently have that depends on commit messages?

- PR title checker is already working.
- version bump script is mostly addressed in this PR.

is there anything else that might need to be updated?

PR #237, opened 2026-01-27
chore: Update checkout action version in codeql.yml

berkus:
To fix dependabot GH action versions to include the full version number, all versions should be formatted identically - in this case, first encountered version was shortened to major version, so dependabot shortened all other versions to major in #232 

With this change the updated dependabot PR should include full versions for all GH actions.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -57,7 +57,7 @@ jobs:
         # your codebase is analyzed, see https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/codeql-code-scanning-for-compiled-languages
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yml
```diff
@@ -57,7 +57,7 @@ jobs:
         # your codebase is analyzed, see https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/codeql-code-scanning-for-compiled-languages
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@v4.0.0
```
geeknoid, line 60: This should be 6.0.1 I think.

berkus: That will be updated to 6.0.2 by dependabot in #232 (the one I'm fixing here)

Vaiz (approved):

PR #246, opened 2026-02-09
chore: Improve our crate's repository property.

geeknoid:
Turns out the repository property in a crate's metadata doesn't
just have to point to a repo's root. It can point directly to a
specific subdirectory. So we can leverage that such that clicking
on the repo link on crates.io will bring the user straight to
the right place in our repo.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,6 @@ rust-version = "1.88"
 authors = ["The Oxidizer Project Authors"]
 license = "MIT"
 homepage = "https://github.com/microsoft/oxidizer"
-repository = "https://github.com/microsoft/oxidizer"
 
 # Adhere to the following rules when adding dependencies:
 # * All dependencies specified here need to have default-features turned off - individual crates can enable the
--- a/crates/ohno_macros/Cargo.toml
+++ b/crates/ohno_macros/Cargo.toml
@@ -14,7 +14,7 @@ rust-version.workspace = true
 authors.workspace = true
 license.workspace = true
 homepage.workspace = true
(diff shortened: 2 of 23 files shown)


Cargo.toml
```diff
@@ -11,7 +11,6 @@ rust-version = "1.88"
 authors = ["The Oxidizer Project Authors"]
 license = "MIT"
 homepage = "https://github.com/microsoft/oxidizer"
-repository = "https://github.com/microsoft/oxidizer"
```
Vaiz, line 14: it might be better to keep the default repo in case someone forgets to manually set it in a new crate

crates/ohno_macros/Cargo.toml
```diff
@@ -14,7 +14,7 @@ rust-version.workspace = true
 authors.workspace = true
 license.workspace = true
 homepage.workspace = true
-repository.workspace = true
+repository = "https://github.com/microsoft/oxidizer/tree/main/crates/ohno_macros"
```
Vaiz, line 17: I don't see much value in pointing to macros crate, maybe this should point to ohno instead?

Vaiz (approved):

crates/ohno_macros/Cargo.toml
geeknoid: This link shows up when the user is on crates.io looking at the ohno_macros crate, so I think the link on that page should point to the same crate.

Cargo.toml
geeknoid: The add_crates.ps1 script ensures this is always populated.

PR #248, opened 2026-02-11
feat(bytesbuf): Improve how memory block metadata is exposed in API

sandersaares:
# Mission

* Ensure that BytesView and BytesBuf expose slice memory block metadata as a first-class citizen whenever iterating through their slices.
* Refactor BytesView and BytesBuf "get slices" methods to use iterators instead of "copy to array" style for more flexible API (user code can do the copy from iterator to array if it wants to).
* Add `Send + Sync` bounds to memory block metadata to ensure that they do not cause futures to become `!Send` if references are stored (in practice, all foreseeable metadata types we expect to implement are going to be thread-safe immutable data).

This fills a gap in functionality (hard to access metadata for vectored writes into a BytesBuf), makes the API more flexible by preferring iterators and improves compatibility with metadata processing in futures.

# BytesView API changes

The current situation is:

* BytesView exposes .first_slice_meta() and .iter_slice_metas(), which are a good start but it is lacking an "iterate over slices and metas together" operation.
* BytesView exposes .slices() and .io_slices() which copy the individual slices into arrays of slices. This is cumbersome because sometimes you just want an iterator, so the copy is unnecessary.

Change BytesView as follows:

* The existing BytesView::iter_slice_metas() is removed.
* The existing BytesView::io_slices() is removed.
* The existing BytesView::slices() is removed.
* We replace the last one with BytesView::slices() -> BytesViewSlices<'a>, which is an iterator that iterates over tuples of (slice, meta).

Expected usage (pseudocode):

```
for (data, meta) in bytes.slices() {
    ...
}
```

If the user wants to still copy the slices into an array, they can just fill the array in this loop.

We will not provide any replacement for .io_slices(). This was anyway a legacy function of dubious value. It remains available through `bytes::Buf` when the `bytes-compat` feature is enabled.

# BytesBuf API changes

The current situation is:

* BytesBuf exposes no memory block metadata at all about the slices it allows writes into! This is an important functionality gap.

Change BytesBuf as follows:

* Add BytesBuf::first_unfilled_slice_meta().

Because BytesBuf only directly provides access to the first unfilled slice (intentionally), we do not need a mechanism to iterate through the metadatas (or slices) on BytesBuf itself.
However, it also provides concurrent access to multiple slices through the vectored write mechanism in BytesBufVectoredWrite.

Change BytesBufVectoredWrite as follows:

* Rename `iter_slices_mut()` to just `slices_mut()` to simplify the name (as long as we are already making breaking changes).
* Change the BytesBufRemaining returned from slices_mut() to iterate over tuples of (slice, meta) similar to what we did with BytesView.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -271,6 +271,7 @@ Tokio's
 toml
 toolchain
 toolchains
+topologies
 trait
 TTL
 TTLs
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -410,9 +410,9 @@ checksum = "773648b94d0e5d620f64f280777445740e61fe701025087ec8b57f45c791888b"
 
 [[package]]
 name = "cpulist"
-version = "1.0.3"
+version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "10921f832bb6e2075c6910ab0be7b33e413029ae6d74b01f049fc6cb237362e6"
+checksum = "008abe0a6734be222147bad0a6233cbe22e558bba7f3a35c90d85fd90dd25cd0"
 dependencies = [
  "itertools 0.14.0",
  "new_zealand",
@@ -648,9 +648,9 @@ checksum = "77ce24cb58228fbb8aa041425bb1050850ac19177686ea6e0f41a70416f56fdb"
 
 [[package]]
 name = "folo_ffi"
-version = "0.1.9"
+version = "0.1.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b2c4a9efbd06c6c8500f5d97a20159996679088bdb5f598a863960bedf3f2574"
+checksum = "bb929208a4b0187184f37cc84c90c42de5abd4d95a881dc46cf79a418ac0dc36"
 
 [[package]]
 name = "fragile"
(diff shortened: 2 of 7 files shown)


crates/bytesbuf/src/buf.rs
```diff
@@ -1101,6 +1129,18 @@ impl<'a> Iterator for BytesBufRemaining<'a> {
...
+        let meta_with_a = {
+            let meta = span_builder.block().meta();
+
+            // SAFETY: The metadata reference points into the block's heap allocation, not into
+            // the span builder's stack memory. We transmute it to 'a immediately so the immutable
+            // borrow of `span_builder` is released before the mutable borrow below.
+            // The metadata is valid for 'a because the BlockRef implementation guarantees metadata
+            // lives as long as any clone of the memory block, and we hold an exclusive reference
+            // to the BytesBuf for the lifetime 'a.
+            unsafe { mem::transmute::<Option<&dyn BlockMeta>, Option<&'a dyn BlockMeta>>(meta) }
```
Vaiz, line 1141: we should check if there is a workaround without unsafe

crates/bytesbuf/CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
...
+  - automatically publish release notes ([#247](https://github.com/microsoft/oxidizer/pull/247))
+
+- 🧩 Miscellaneous
+
+  - Merge remote-tracking branch 'origin/main' into u/sasaares/bytesbuf-expose-meta
+  - Update readme
+  - trait BlockMeta: Any + Send + Sync + Debug
+  - Skip mutations that cause infinite loops
+  - Improve test coverage
+  - Improve bytesbuf memory block metadata API UX
```
Vaiz: could you clean that up
ideally, release should be done in a separate pr

crates/bytesbuf/src/buf.rs
sandersaares: Do you have a specific suggestion?

crates/bytesbuf/CHANGELOG.md
sandersaares: Reverted this commit, will do as separate PR

crates/bytesbuf/src/buf.rs
Vaiz: nope :(

sandersaares: OK. Well, there's always tomorrow.

PR #249, opened 2026-02-11
feat: remove implicit `Sized` requirement from `BytesBuf::as_write` method

Vaiz:
this allows to use `BytesBuf::as_write` with `dyn Memory`


sandersaares (approved):
Good catch! Do we have the same issue elsewhere?

PR #252, opened 2026-02-12
feat: add tower-service compatibility to seatbelt

martintmk:
As title says. Seatbelt can now interop natively with tower-based ecosystem. The downside is decreased ergonomics and performance.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -277,7 +277,7 @@ dependencies = [
 
 [[package]]
 name = "bytesbuf_io"
-version = "0.2.0"
+version = "0.3.0"
 dependencies = [
  "bytesbuf",
  "futures",
@@ -1873,7 +1873,7 @@ dependencies = [
 
 [[package]]
 name = "thread_aware"
-version = "0.6.1"
+version = "0.6.2"
 dependencies = [
  "many_cpus",
  "mutants",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ homepage = "https://github.com/microsoft/oxidizer"
 # local dependencies
 anyspawn = { path = "crates/anyspawn", default-features = false, version = "0.1.0" }
 bytesbuf = { path = "crates/bytesbuf", default-features = false, version = "0.3.0" }
-bytesbuf_io = { path = "crates/bytesbuf_io", default-features = false, version = "0.2.0" }
+bytesbuf_io = { path = "crates/bytesbuf_io", default-features = false, version = "0.3.0" }
 data_privacy = { path = "crates/data_privacy", default-features = false, version = "0.10.1" }
 data_privacy_macros = { path = "crates/data_privacy_macros", default-features = false, version = "0.9.0" }
 data_privacy_macros_impl = { path = "crates/data_privacy_macros_impl", default-features = false, version = "0.9.0" }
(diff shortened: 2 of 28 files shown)


crates/seatbelt/src/breaker/service.rs
```diff
@@ -95,6 +101,76 @@ where
     }
 }
 
+/// Future returned by [`Breaker`] when used as a tower [`Service`](tower_service::Service).
+#[cfg(any(feature = "tower-service", test))]
+pub struct BreakerFuture<Out> {
+    inner: Pin<Box<dyn Future<Output = Out> + Send>>,
```
schgoo, line 110: Why do we need to implement this future? Is it about the future size issue?

martintmk: Not really, it's about future API evolution. If we expose it as `Pin<Box<dyn Future<Output = Out> + Send>>`, it implies that this future is always boxed. Having a dedicated future type allows us to optimize it later. 

Extreme case is manually writing the future, but at some point Rust will enable to name the anonymous futures and then we can just change the implementation to be efficient and ergonomic without boxing the future.

crates/seatbelt/src/lib.rs
```diff
@@ -102,6 +102,21 @@
 //! - [`retry`] - Middleware that automatically retries failed operations.
 //! - [`breaker`] - Middleware that prevents cascading failures.
 //!
+//! # Tower Compatibility
+//!
+//! All resilience middleware ([`timeout::Timeout`], [`retry::Retry`], [`breaker::Breaker`]) are
+//! compatible with the Tower ecosystem when the `tower-service` feature is enabled. This allows
+//! you to use `tower::ServiceBuilder` to compose middleware stacks:
+//!
+//! ```rust,ignore
```
sandersaares: Is this `ignore` because doctest cannot guarantee existence of tower dependency? I vaguely remember some limitations like that, just asking to make sure there is a reason (maybe would be good to document it as `# // ignore because ...`).

crates/seatbelt/src/breaker/service.rs
```diff
@@ -95,6 +101,76 @@ where
...
+{
+    type Response = Res;
+    type Error = Err;
+    type Future = BreakerFuture<Result<Res, Err>>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: Req) -> Self::Future {
```
sandersaares, line 149: I am troubled by the maintainability aspects of having this logic duplicated for `tower-service` and for `layered`.

If true reuse is not possible, we should at least leave very big bold comments saying "HEY COPILOT PAY ATTENTION: this function is one of two logic-equivalent functions in this file, make sure changes are always synchronized between the two" or something.

I suggest even adding a statement about this to `seatbelt/src/AGENTS.md` and telling it to pay special attention in this crate.

crates/seatbelt/src/lib.rs
martintmk: it was just AI being lazy, fixed

crates/seatbelt/src/breaker/service.rs
martintmk: indeed, added more comments and AGENTS.md with this info

PR #254, opened 2026-02-12
feat(bytesbuf): GlobalPool uses smaller memory blocks for smaller reservation requests

sandersaares:
feat(bytesbuf): GlobalPool uses smaller memory blocks for smaller reservation requests

Previously, we used 64KB blocks, always. This was just a placeholder implementation, really, so now we take a first step forward by adding 1KB, 4KB and 16KB block sizes. We use whatever is the smallest block that fits for the requested memory reservation (anything above 64KB gets multiple 64KB blocks).

This can help avoid wasted memory if we have a lot of small allocations. Not specifically the case with our I/O benchmarks so far but since bytesbuf is more generic than just I/O, this can become of some relevance when we start talking about data processing (e.g. serialization of URL templates).

Also added some basic metrics to GlobalPool to allow us to understand memory reservation efficiency and drive future design decisions in this domain with a larger degree of data-basis.

On the flip side, this additional logic in reserve() does slow down memory reservation somewhat (20% for large reservations). I have not seen our memory pool show up on profiler traces so far, so hopefully this change is insignificant enough to have a real world impact but let's keep it in mind in case we notice something.

### DIFF
--- a/crates/thread_aware/src/cell/mod.rs
+++ b/crates/thread_aware/src/cell/mod.rs
@@ -450,6 +450,34 @@ where
 }
 
 impl<T, S: Strategy> Arc<T, S> {
+    /// Gets the number of strong references to the value in the current thread/affinity.
+    ///
+    /// This method returns the strong reference count for the underlying [`sync::Arc`]
+    /// that holds the value for the current affinity, excluding any internal references
+    /// held by the storage for deduplication purposes. Each affinity maintains its own
+    /// separate value with its own reference count.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use thread_aware::{Arc, PerCore};
+    ///
+    /// let arc = Arc::<_, PerCore>::new(|| 42);
+    /// assert_eq!(Arc::strong_count(&arc), 1);
+    ///
+    /// let arc2 = arc.clone();
+    /// assert_eq!(Arc::strong_count(&arc), 2);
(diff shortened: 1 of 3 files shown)


crates/bytesbuf/src/mem/global.rs
```diff
@@ -101,115 +109,166 @@ struct GlobalPoolInner {
...
-                let meta = BlockMeta {
-                    block_pool: Arc::clone(&self.block_pool),
-                    handle,
-                    ref_count: AtomicUsize::new(1),
-                };
+        // Pick the smallest sub-pool that fits, then use only that block size for the entire
+        // reservation. For requests exceeding the largest block size, use multiple blocks of
+        // the largest size. Using uniform block sizes avoids imbalances when repeated
+        // reservations are not perfectly aligned with block size boundaries.
+        if min_bytes <= 1024 {
```
martintmk, line 139:
```suggestion
        if min_bytes <= BUFF_THRESHOLD_1024 {
```

Shall we introduce constants for this?

sandersaares: Feels a little too `const RED: &str = "RED"` to me - if the constant is just a re-statement of the literal, I do not really feel the value.

Ideally, I would like it to be some `const SIZES: [...]` but that did not seem to work out linguistically, Rust is not that expressive (at least at compile time).

Vaiz (approved):

PR #255, opened 2026-02-12
feat: implement thread_aware in tick 

martintmk:
this makes the Clock natively compatible with single-threaded runtimes

### DIFF
--- a/crates/tick/src/clock.rs
+++ b/crates/tick/src/clock.rs
@@ -658,6 +658,7 @@ mod tests {
         assert_eq!(clock_2.state.timers_len(), 0);
         assert_eq!(driver_1.state.timers_len(), 1);
         assert_eq!(driver_2.state.timers_len(), 0);
+        assert_eq!(clock_1.clone().relocated(source, pinned_2).state.timers_len(), 0);
 
         // register the timer on clock 2
         let mut fut_2 = Box::pin(clock_2.delay(Duration::from_secs(100)));


geeknoid (approved):

crates/tick/src/runtime/inactive_clock.rs
```diff
@@ -57,13 +66,7 @@ impl InactiveClock {
...
     #[must_use]
     pub fn activate(self) -> (Clock, ClockDriver) {
-        let mut state = self.0;
-
-        if matches!(state, ClockState::System(_)) {
-            // if this is system clock, create a new instance to avoid sharing timers
-            state = ClockState::new_system();
-        }
-
+        let state = self.0;
```
sandersaares: This PR is slightly annoying to review because there is no context for what `.0` means in all the different places it appears. I humbly suggest that we should not use tuple structs outside throwaway test code.

crates/tick/src/clock.rs
```diff
@@ -143,6 +145,12 @@ use crate::timers::TimerKey;
 #[derive(Debug, Clone)]
 pub struct Clock(pub(crate) ClockState);
 
+impl ThreadAware for Clock {
+    fn relocated(self, source: thread_aware::affinity::MemoryAffinity, destination: thread_aware::affinity::PinnedAffinity) -> Self {
```
sandersaares: The `extra::verbose::type::paths` seem out of place. I suggest being more generous with `use` statements to make such lines easier to read.

crates/tick/src/clock.rs
```diff
@@ -143,6 +145,12 @@ use crate::timers::TimerKey;
 #[derive(Debug, Clone)]
 pub struct Clock(pub(crate) ClockState);
 
+impl ThreadAware for Clock {
```
sandersaares, line 173: I think thread-aware behavior is important enough to document in a type's API documentation if it has or could theoretically have an impact on user-visible behavior. How does the type handle being relocated across threads? For example, do clones stay connected across threads (e.g. in `ClockControl` scenarios)?

It is not super clear to me what might be the exact considerations worth documenting here but it does seem like there is potential for different interpretations/implementations so some clarity feels desirable.

martintmk: true, documented across the crate

crates/tick/src/clock.rs
martintmk: updated

crates/tick/src/runtime/inactive_clock.rs
martintmk: I refined these and moved to regular fields

sandersaares (approved):

PR #259, opened 2026-02-13
chore: release a new major version of tick and seatbelt crate

martintmk:


sandersaares:
In the same PR (which I guess gets squashed to a single commit, yeah). I have not tried it myself but I do not immediately see why it would not work. Probably it works.

Vaiz:
> > > Once released, I'll bump all these.
> > 
> > 
> > Why not all together?
> 
> You mean releasing multiple versions that are interconnected withing a single commit? Will that work?

yes, it works
moreover, I think we should prefer doing this instead of publishing crates one by one

crates/seatbelt/Cargo.toml
```diff
@@ -4,7 +4,7 @@
 [package]
 name = "seatbelt"
 description = "Resilience and recovery mechanisms for fallible operations."
-version = "0.2.0"
+version = "0.3.0"
```
Vaiz, line 7: why breaking?

martintmk: new major version of `tick` crate

Vaiz: could you add to changelog what was broken and how to migrate to a new version?

Vaiz: > new major version of tick crate

as long as it is not reexported through seatbelt, it should be fine

martintmk: > could you add to changelog what was broken and how to migrate to a new version?

This information is already contained with the changelog for `tick` crate. Do you see any reason to duplicate it here?

you don't even have to re-export, exposing it in your public APIs forces you to use that particular major version

Vaiz: got it, yeah, that's unfortunate
are there any plans to wrap such types to reduce the number of potential breaking changes?

martintmk: At some point, we will have to treat these core types as foundational and avoid unnecessary wrapping (and also avoid breaking changes)

PR #262, opened 2026-02-15
feat(tick): rename MIN to UNIX_EPOCH

martintmk:
To better align the naming with `SystemTime::UNIX_EPOCH`. This is breaking change, but we are releasing a new major version in #259 anyway.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -277,7 +277,7 @@ dependencies = [
 
 [[package]]
 name = "bytesbuf_io"
-version = "0.2.0"
+version = "0.3.0"
 dependencies = [
  "bytesbuf",
  "futures",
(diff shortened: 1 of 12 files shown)


sandersaares (approved):

Vaiz (approved):
it might be the time to start treating our public API as more stable
in that case a better solution is to deprecate `MIN` and introduce `UNIX_EPOCH`

martintmk:
> it might be the time to start treating our public API as more stable in that case a better solution is to deprecate `MIN` and introduce `UNIX_EPOCH`

Deprecation incurs some technical dept. I feel we are not yet at the stage where adoption is high enough to varant this. (albeit I am treating sem-ver incompatible changes very seriously)

sandersaares:
> it might be the time to start treating our public API as more stable

We should explicitly signal this with a `1.0.0` for each crate when we choose to treat our APIs as stable. Today feels too early, though.

PR #265, opened 2026-02-16
feat: Introduce the templated_uri crate family

kate-shine:
- Improved UriSafeString handling. Added infallible method for creation of url encoded string, and moved the current constructor to try_new(), which validates content and creates UriSafeString if it's valid (for handling of pre-encoded values)

Fixes:
- BaseUri stripped the path prefix in some cases
- BaseUri wasn't stripping the query when constructed from existing Uri
- Uri Unsafe string was handling the redaction wrong, and redacted content even when it shouldn't.

Improvements:
- 100 % code coverage and mutation testing
- Streamlining of some macro code

### DIFF
--- a/crates/templated_uri/src/base_uri.rs
+++ b/crates/templated_uri/src/base_uri.rs
@@ -600,7 +600,7 @@ impl BaseUri {
         parts.authority = Some(self.authority().clone());
         parts.path_and_query = Some(full_path);
 
-        http::Uri::from_parts(parts).map_err(ValidationError::caused_by)
+        http::Uri::from_parts(parts).map_err(Into::into)
     }
 }
 
--- a/crates/templated_uri/src/base_uri/origin.rs
+++ b/crates/templated_uri/src/base_uri/origin.rs
@@ -131,7 +131,7 @@ impl std::str::FromStr for Origin {
     /// Returns a [`ValidationError`] if the string is not a valid origin
     /// (missing scheme, unsupported scheme, or invalid authority).
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let uri: http::Uri = s.parse().map_err(ValidationError::caused_by)?;
+        let uri: http::Uri = s.parse().map_err(ValidationError::from)?;
         let scheme = uri.scheme().ok_or_else(|| ValidationError::caused_by("missing scheme"))?.clone();
         let authority = uri
             .authority()
--- a/crates/templated_uri/src/macros.rs
+++ b/crates/templated_uri/src/macros.rs
@@ -36,17 +36,10 @@ pub use templated_uri_macros::UriParam;
 /// # Example
 ///
 /// ```
-/// # use std::fmt::Display;
 /// use templated_uri::UriUnsafeParam;
 ///
 /// #[derive(UriUnsafeParam)]
 /// struct PathSegment(String);
-///
-/// impl Display for PathSegment {
-///     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-///         write!(f, "{}", self.0)
-///     }
-/// }
 /// ```
 ///
 /// This allows `PathSegment` to be used in URI templates with `{+param}` syntax where


crates/templated_uri/Cargo.toml
```diff
@@ -0,0 +1,66 @@
...
+
+edition.workspace = true
+rust-version.workspace = true
+authors.workspace = true
+license.workspace = true
+homepage.workspace = true
+repository = "https://github.com/microsoft/oxidizer/tree/main/crates/templated_uri"
+
+[package.metadata.cargo_check_external_types]
+allowed_external_types = [
```
geeknoid, line 20: I worry that we're having to expose all of these external types here. Are we OK coupling this crate to the http crate?

kate-shine: I have the same worry, but http crate is an industry standard and we need to consider it everywhere anyway. I don't think we can avoid interfacing with it.

geeknoid: I'd rather we introduce our own types that mirror these, and include crate features to make it easier to interop with the http crate's URI type and the stand-alone url crate. So we still play nice, but it's a separate feature.

crates/templated_uri/src/uri_safe.rs
```diff
@@ -99,32 +127,48 @@ impl UriSafeString {
...
-            // Reserved characters in RFC 6570:
-            // - Template delimiters: '{', '}'
-            // - Reserved gen-delims (RFC 3986): ':', '/', '?', '#', '[', ']', '@'
-            // - Reserved sub-delims (RFC 3986): '!', '$', '&', '\'', '(', ')', '*', '+', ',', ';', '='
-            if "{}/:?#[]@!$&'()*+,;=".contains(c) {
-                return Err(UriSafeError {
-                    invalid_char: c,
-                    position: i,
-                });
+    pub fn new_raw(raw: impl Into<String>) -> Result<Self, UriSafeError> {
```
geeknoid: When should someone use this instead of `new` above? Should this an unsafe function? Maybe `new_unchecked` would be more idiomatic?

crates/templated_uri/Cargo.toml
kate-shine: I'd certainly like that, although the amount of work behind that to make it work well is quite non-trivial. It would be possible to start by newtype wrapping http and forwarding API to it, but I'd prefer to not do that and switch to our own types as a future breaking change that fully replaces it instead.

The reason is that there is some weirdness in http behavior which they can't get rid of anymore for backwards compatibility reasons. We will be able to drop that compatibility when we switch types, but if we wrap those types now, I'm afraid that we will conserve the behavior.

Would it be ok to introduce this as a breaking change in future major version instead of now as a part of open sourcing?

crates/templated_uri/src/uri_safe.rs
kate-shine: I can imagine using this the moment you already have an urlencoded string, which can happen. New `new()` would reencode % symbols and mess up the url.

I don't like `unchecked` explicitly checks though, the old `new` is now infallible due to inclusion of url encoding.

Maybe 

`new` -> `new_encode` or just `encode`
`new_raw` -> `new`

?

Although the reason I went for `new` is that I'd like it to be a prioritized way to construct this.

Honestly, urlencoding is a mess. I hate that you can't safely do a second encoding pass on an already encoded string.

crates/templated_uri/Cargo.toml
geeknoid: Yeah, we can do it in a follow-on PR, but I'd rather if "future" means "next week" rather than "sometime in the next few years".

At this point, most coding should be handled by AI at lightspeed, so coding time should generally no longer be much of a limiting factor for us.

martintmk: Let's be clear here. Us doing wrappers over `http` types break almost all compatibility we have with existing ecosystem. This also has direct impact over our existing adopters which require this interop. 

IMHO, I see these wrappers as unnecessary and just bound to introduce more friction. The http crate  works really well for us (outside of few quirks).

kate-shine: Yep, that was my reason for not the wrap in a first place. It would also keep the oddities intact, especially the ones that pain me the most, like PathAndQuery's inconsistent handling of path's first slash.

And it's true that while I wished few times for some alternative to http types, there is a reason why basically nobody does it apart from outliers like Actix. When I was working on a larger Actix based codebase, the fact that their http client's types were incompatible with the rest of Rust ecosystem made it a pain to work with, forcing us to do conversions all the time, and it was one of biggest reasons why we ended up switching from awc (Actix Web Client) to Reqwest.

crates/templated_uri/src/uri_safe.rs
kate-shine: Did some renaming

crates/templated_uri/src/uri_safe.rs
```diff
@@ -0,0 +1,438 @@
...
+}
+
+impl Error for UriSafeError {}
+
+/// A wrapper around String that guarantees the inner value is safe to use in URI templates.
+///
+/// According to RFC 6570, certain characters are reserved and must be percent-encoded.
+/// This type ensures its content doesn't contain those reserved characters.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct UriSafeString(Cow<'static, str>);
```
martintmk:
```suggestion
pub struct UriString(Cow<'static, str>);
```

or `UriParameter`?

crates/templated_uri/src/uri_safe.rs
```diff
@@ -0,0 +1,438 @@
...
+    impl Sealed for NonZeroU128 {}
+    impl Sealed for NonZeroUsize {}
+    impl Sealed for IpAddr {}
+    impl Sealed for Uuid {}
+    impl<T> Sealed for Sensitive<T> where T: Sealed {}
+    impl<T> Sealed for &T where T: Sealed + ?Sized {}
+}
+
+/// Marks types that, when [`Display`ed](std::fmt::Display), are valid for URI use.
+pub trait UriSafe: private::Sealed + Display + Debug {}
```
martintmk: Can we unseal this trait?

I actually encountered a scenario where I have my own type that I want to mark as UriSafe.

crates/templated_uri/src/uri_safe.rs
kate-shine: `UriParameter` is too general I think, I mean, u32 can also be a UriParameter. If anything, it sounds like a trait name instead. 

`UriString` can be misunderstood for the whole Uri, but I also like that it implies that it's a string wrapper. 

`UriStringParam` is longer, but feels most precise of them all. I have another idea though that I'd like to check.

crates/templated_uri/src/uri_safe.rs
kate-shine: I'm not sure about that. It exists to provide type guarantees, if we unseal it, someone can easily implement it for String and bypass all protection :/ What I could do is leverage the percent encoded string type as intermediary in some way, I'll check that.

martintmk: >if we unseal it, someone can easily implement it for String and bypass all protection :/ 

Wouldn't foreign trait restriction prevent this?


My point here is to allow this pattern:

``` rust
struct MyUrl {
   tenant: Tenant
}
```

Where `Tenant` is just something like `struct Tenant(Uuid)`. I don't think we should force everyone to use `UriSafeString` (or whetever the final name is) if owner of business domain can design their own uri-safe types.

In the end it's their responsibility. If the declare the parameter safe and it contains invalid values, the Uri creation will fail.

kate-shine: I'm refactoring the code and getting rid of that trait completely right now, turning it into `UriSafe<T>(T)` instead, and making sure it's safe on construction.

As for your own type, I didn't realize it at first (It comes from refactoring that was introduced earlier by Ralf), you can just implement `UriFragment` for it now. If you are wrapping a type that is already `UriSafe`, you should be able to just derive it, otherwise you need to implement it manually and use `as_uri_safe` method to convert it to known UriSafe type.

kate-shine: So, for your `struct Tenant(Uuid)`

It should already be possible to 

```rust
#[derive(UriFragment)
struct Tenant(Uuid)
```

crates/templated_uri_macros_impl/src/lib.rs
```diff
@@ -0,0 +1,1295 @@
...
+use crate::struct_template::struct_template;
+use crate::uri_param::{uri_param_impl, uri_unsafe_param_impl};
+
+macro_rules! bail {
+    ($span:ident, $msg:expr) => {
+        crate::bail!($span, $msg, )
+    };
+    ($span:ident, $msg:expr, $($args:tt),*) => {
+        return syn::Error::new_spanned($span, format!($msg, $($args,)*)).to_compile_error()
+    };
```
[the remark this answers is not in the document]
kate-shine: This is false, traling comma in format compiles just fine.

PR #274, opened 2026-02-18
feat(bytesbuf): make BytesBufWriter owning via into_writer()

### LINKED ISSUE #269: feat(bytesbuf): Improve ergonomics of BytesBuf::as_writer()
* Return a named type so it can be more easily stored in fields.
* Implement `BufWrite`. We cannot use `std::io::BufWriter` because it needlessly allocates an intermediate buffer. A `BytesBuf` writer is already buffered!

sandersaares:
## Summary

`BytesBuf` now exposes an owning writer cycle:

`BytesBuf::into_writer(memory) -> BytesBufWriter<M> -> BytesBufWriter::into_inner() -> BytesBuf`

This replaces the previous referencing writer adapter and keeps write behavior unchanged.

## Changes

- Renamed the write adapter type from `BytesBufWrite` to `BytesBufWriter` and kept it publicly re-exported from the crate root.
- Changed `BytesBuf::as_write()`/`writer()` style borrowing API to an owning `BytesBuf::into_writer(memory)` API.
- Refactored `BytesBufWriter` to own both `BytesBuf` and memory provider `M` by value, and added `into_inner()` to recover the buffer.
- Added `impl<M: Memory + ?Sized> Memory for &M` so reference memory handles (including `&dyn Memory`) work with `into_writer`.
- Renamed internal module file `write_adapter.rs` to `buf_writer.rs` to match the type role.
- Included the existing repository-level `AGENTS.md` guidance update in this PR diff (no manual changelog edits).

## Validation

- `cargo test -p bytesbuf` (AzureLinux3.0 WSL)
  - 153 unit tests passed
  - 57 doc-tests passed

Closes #269

### DIFF
--- a/crates/bytesbuf/src/mem/memory.rs
+++ b/crates/bytesbuf/src/mem/memory.rs
@@ -43,3 +43,50 @@ impl<M: Memory + ?Sized> Memory for &M {
         (*self).reserve(min_bytes)
     }
 }
+
+#[cfg_attr(coverage_nightly, coverage(off))]
+#[cfg(test)]
+mod tests {
+    use std::sync::atomic::{AtomicUsize, Ordering};
+
+    use super::Memory;
+    use crate::BytesBuf;
+    use crate::mem::testing::TransparentMemory;
+
+    #[derive(Debug)]
+    struct CountingMemory {
+        reserve_calls: AtomicUsize,
+        inner: TransparentMemory,
+    }
+
+    impl CountingMemory {
+        fn new() -> Self {
+            Self {
+                reserve_calls: AtomicUsize::new(0),
+                inner: TransparentMemory::new(),
+            }
+        }
+    }
+
+    impl Memory for CountingMemory {
+        fn reserve(&self, min_bytes: usize) -> BytesBuf {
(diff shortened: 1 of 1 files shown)


crates/bytesbuf/src/buf.rs
```diff
@@ -933,8 +933,8 @@ impl BytesBuf {
     /// # Ok::<(), std::io::Error>(())
     /// ```
     #[inline]
-    pub fn as_write<M: Memory + ?Sized>(&mut self, memory: &M) -> impl std::io::Write {
-        BytesBufWrite::new(self, memory)
+    pub fn writer<'m, M: Memory + ?Sized>(&mut self, memory: &'m M) -> BytesBufWriter<'_, 'm, M> {
```
martintmk:
```suggestion
    pub fn into_writer<M: Memory>(self, memory: M) -> BytesBufWriter<M> {
```

And then also introduce BytesBufWriter::into_inner that gives back the `BytesBuf`.


Can we work with this API? I fear current lifetime-based approach won't work with pooled objects.

sandersaares: Updated.

martintmk (approved):

PR #275, opened 2026-02-19
feat(tick): custom debug implementations

martintmk:
the default debug implementation was too noisy

### DIFF
--- a/crates/tick/src/clock_control.rs
+++ b/crates/tick/src/clock_control.rs
@@ -797,7 +797,7 @@ mod tests {
     }
 
     #[test]
-    #[cfg(not(miri))]
+    #[cfg_attr(miri, ignore)]
     fn debug_ok() {
         let system = SystemTime::UNIX_EPOCH + Duration::from_secs(123);
         let control = ClockControl::new_at(system);
@@ -809,7 +809,7 @@ mod tests {
     }
 
     #[test]
-    #[cfg(not(miri))]
+    #[cfg_attr(miri, ignore)]
     fn debug_negative_offset_ok() {
         let system = SystemTime::UNIX_EPOCH - Duration::from_secs(123);
         let control = ClockControl::new_at(system);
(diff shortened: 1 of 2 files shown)


crates/tick/src/clock_control.rs
```diff
@@ -67,14 +67,25 @@ use crate::{Clock, thread_aware_move};
...
 pub struct ClockControl {
     /// Clock control requires controlling the passage of time across threads.
     /// For this reason, we need to use a mutex to ensure that state is consistent
     /// across all threads.
     state: Arc<Mutex<State>>,
 }
 
+impl std::fmt::Debug for ClockControl {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let state = self.state.lock().expect("acquiring lock must always succeed");
```
sandersaares: This is just `unwrap()` in more verbose form. Is there a reason we expect acquiring the lock to always succeed? If so, we should document that reason in the expect-message (i.e. what makes it impossible for this lock to get poisoned).

martintmk: improved, the lock is not even needed here

crates/tick/src/clock_control.rs
```diff
@@ -791,4 +795,25 @@ mod tests {
         // The timer should still be registered since we couldn't advance further to reach it
         assert_eq!(control.timers_len(), 1);
     }
+
+    #[test]
+    #[cfg(not(miri))]
```
sandersaares: I suggest `#[cfg_attr(miri, ignore)]` because this one does not generate dead code lint spam (it is still compiled, just not executed).

martintmk: This is cool, zed had trouble with `#[cfg(not(miri))]` code and disabled it. This even fixes that issue! Super cool!

sandersaares (approved):

PR #278, opened 2026-02-20
fix(ohno): address some correctness and usability findings

geeknoid:
Correctness:
- Fix compounding name bug in generate_unique_field_name (ohno_core_1_2 instead of ohno_core_2 on repeated collision)
- Emit clear compile error when #[ohno::error] is applied to enums
- Add #[track_caller] to Option<T> IntoAppError implementations

Usability:
- Make enrichments()/enrichment_messages() iterate in chronological order (oldest-first), consistent with format_error display order

### DIFF
--- a/crates/ohno/src/core.rs
+++ b/crates/ohno/src/core.rs
@@ -103,14 +103,14 @@ impl OhnoCore {
         self.data.backtrace.as_backtrace()
     }
 
-    /// Returns an iterator over the enrichment information in reverse order (most recent first).
+    /// Returns an iterator over the enrichment information in chronological order (oldest first).
     pub fn enrichments(&self) -> impl Iterator<Item = &EnrichmentEntry> {
-        self.data.enrichment.iter().rev()
+        self.data.enrichment.iter()
     }
 
-    /// Returns an iterator over just the enrichment messages in reverse order (most recent first).
+    /// Returns an iterator over just the enrichment messages in chronological order (oldest first).
     pub fn enrichment_messages(&self) -> impl Iterator<Item = &str> {
-        self.data.enrichment.iter().rev().map(|ctx| ctx.message.as_ref())
+        self.data.enrichment.iter().map(|ctx| ctx.message.as_ref())
     }
 
     /// Formats the main error message without backtrace and error enrichment.
@@ -305,7 +305,7 @@ mod tests {
         error.add_enrichment(EnrichmentEntry::new("ctx1", "test.rs", 1));
         error.add_enrichment(EnrichmentEntry::new("ctx2", "test.rs", 2));
         let messages: Vec<_> = error.enrichment_messages().collect();
-        assert_eq!(messages, vec!["ctx2", "ctx1"]);
+        assert_eq!(messages, vec!["ctx1", "ctx2"]);
     }
 
     #[test]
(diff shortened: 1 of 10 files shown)


crates/ohno/src/core.rs
```diff
@@ -103,14 +103,14 @@ impl OhnoCore {
         self.data.backtrace.as_backtrace()
     }
 
-    /// Returns an iterator over the enrichment information in reverse order (most recent first).
+    /// Returns an iterator over the enrichment information in chronological order (oldest first).
     pub fn enrichments(&self) -> impl Iterator<Item = &EnrichmentEntry> {
-        self.data.enrichment.iter().rev()
+        self.data.enrichment.iter()
```
martin-kolinek, line 108: @vaiz, this looked intentional in the original version - what was the reason to go against the format_error! convention?

crates/automation/src/lib.rs
```diff
@@ -52,7 +52,7 @@ pub fn list_packages(workspace_root: impl AsRef<Path>) -> Result<Vec<PackageMeta
 
     if !output.status.success() {
         let stderr = String::from_utf8_lossy(&output.stderr);
-        ohno::bail!("cargo metadata failed: {stderr}");
+        ohno::bail!("cargo metadata failed: {}", stderr);
```
martin-kolinek: I'm a bit surprised this doesn't cause a clippy warning.

martin-kolinek (approved):

crates/automation/src/lib.rs
geeknoid: I'm guessing clippy doesn't see this as a formatting thing.

sandersaares (approved):

crates/automation/src/lib.rs
Vaiz: clippy doesn't know about non-standard macros like `bail!`

crates/ohno_macros/src/error_type_attr.rs
```diff
@@ -45,11 +45,16 @@ use crate::utils::generate_unique_field_name;
...
+    TokenStream::from(error_impl(&mut input))
+}
 
-    add_fiasko_error_derive(&mut input);
-    add_ohno_core_field(&mut input);
+fn error_impl(input: &mut DeriveInput) -> proc_macro2::TokenStream {
+    if let Err(err) = add_ohno_core_field(input) {
+        return err.to_compile_error();
+    }
+    add_fiasko_error_derive(input);
```
Vaiz, line 55: that's a fiasco :D

crates/ohno_macros/src/error_type_attr.rs
```diff
@@ -149,19 +158,52 @@ mod tests {
...
             enum TestError {
                 Variant1,
                 Variant2,
             }
         };
 
-        let original = input.to_token_stream().to_string();
-        crate::error_type_attr::add_ohno_core_field(&mut input);
+        let result = crate::error_type_attr::add_ohno_core_field(&mut input);
+        assert!(result.is_err());
```
Vaiz, line 170: I find such checks pointless, `unwrap_err` on the next line does it for you

crates/automation/src/lib.rs
Vaiz: btw, why do we need this change? did it break?

crates/ohno/src/core.rs
Vaiz: do you mean `OhnoCore::format_error` method?
originally, `format_error` supposed to output traces in reverse order, but I couldn't find a way to order traces from inner errors

I don't think there will be an issue if we change order here

crates/automation/src/lib.rs
Vaiz: I don't see any changes in bail macro, @geeknoid could you rollback this?

geeknoid: Undone.

PR #285, opened 2026-02-24
build: Delta configuration file

tekian:


geeknoid (approved):

.delta.toml
```diff
@@ -0,0 +1,84 @@
+# cargo-delta configuration file
+# This file shows the default configuration values used by cargo-delta
```
sandersaares: I assume the purpose of adding this file to the repo is not to show the default configuration values, so this comment feels wrong.

tekian: Quite right, thanks Sander! Removed.

sandersaares (approved):

PR #290, opened 2026-02-27
feat(seatbelt): ResilienceContext is now ThreadAware

martintmk:
I found a need for it. Minimal changes so far, until more scenarios are uncovered. We can then start putting these on top of middleware services.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -949,9 +949,9 @@ checksum = "92ecc6618181def0457392ccd0ee51198e065e016d1d527a7ac1b6dc7c1f09d2"
 
 [[package]]
 name = "jiff"
-version = "0.2.18"
+version = "0.2.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e67e8da4c49d6d9909fe03361f9b620f58898859f5c7aded68351e85e71ecf50"
+checksum = "b3e3d65f018c6ae946ab16e80944b97096ed73c35b221d1c478a6c81d8f57940"
 dependencies = [
  "jiff-static",
  "jiff-tzdb-platform",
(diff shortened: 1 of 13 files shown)


crates/seatbelt/src/context.rs
```diff
@@ -23,6 +27,15 @@ pub struct ResilienceContext<In, Out> {
     _out: std::marker::PhantomData<fn() -> Out>,
 }
 
+impl<In, Out> ThreadAware for ResilienceContext<In, Out> {
+    fn relocated(mut self, source: MemoryAffinity, destination: PinnedAffinity) -> Self {
+        // Only clock is thread-aware for now. At some point, we also want
+        // telemetry to be tread-aware too.
+        self.clock = self.clock.relocated(source, destination);
```
Vaiz, line 34: derive macro is more reliable
for example when a new field is added, there are less chances developer forgets to update this piece of code

crates/seatbelt/Cargo.toml
```diff
@@ -27,6 +27,9 @@ allowed_external_types = [
     "tick::clock::Clock",
     "tower_layer::Layer",
     "tower_service::Service",
+    "thread_aware::core::ThreadAware",
+    "thread_aware::affinity::MemoryAffinity",
+    "thread_aware::affinity::PinnedAffinity",
```
Vaiz, line 32: that's unfortunate that implementing a trait requires to expose 3 external types :(
@Darksecond , is there a way to improve that?

crates/seatbelt/src/context.rs
martintmk: I don't actually don't want to depend on derive feature, it pulls syn which is quite heavy.

Vaiz: afiak, it's not a big deal, it's hard to find a project that doesn't have `syn` in their dependency tree

martintmk: true, but I rather be conservative here. If I find I am repeating myself too often, I'll switch to it here.

crates/seatbelt/Cargo.toml
martintmk: I don't think we can do anything here, all these types are required.

PR #293, opened 2026-02-27
feat(tick): indicate whether the clock is alive or not using Debug

martintmk:
I found cases where timers are not advancing. This PR improves the `Debug` implementation with `alive` fields that tells this information.


crates/tick/src/timers.rs
```diff
@@ -45,13 +45,19 @@ pub(crate) struct Timers {
     /// The [`Waker`] represents the task awaiting the timer.
     wakers: BTreeMap<TimerKey, Waker>,
     last_discriminator: u32,
+    // Determines whether the timers are currently being advanced.
```
sandersaares, line 49: Might be useful to help understanding for people not super familiar with this crate to also explain here what might be the reasons for this being `false`.

martintmk: added more details

PR #294, opened 2026-02-27
feat(seatbelt): introduce Fallback middleware

martintmk:
Vibe-coded with few back and forth. Opus can follow existing patterns pretty easily, so adding new middleware as this is quite straightforward.

### DIFF
--- a/crates/seatbelt/src/breaker/engine/engine_telemetry.rs
+++ b/crates/seatbelt/src/breaker/engine/engine_telemetry.rs
@@ -4,8 +4,6 @@
 use tick::Clock;
 
 #[cfg(any(feature = "metrics", feature = "logs", test))]
-use crate::TelemetryString;
-#[cfg(any(feature = "metrics", feature = "logs", test))]
 use crate::breaker::CircuitState;
 #[cfg(any(feature = "metrics", test))]
 use crate::breaker::telemetry::*;
@@ -21,14 +19,14 @@ pub(crate) struct EngineTelemetry<T> {
     #[cfg(any(feature = "metrics", feature = "logs", test))]
     pub(super) telemetry: TelemetryHelper,
     #[cfg(any(feature = "metrics", feature = "logs", test))]
-    pub(super) breaker_id: TelemetryString,
+    pub(super) breaker_id: crate::TelemetryString,
     #[cfg(any(feature = "metrics", feature = "logs", test))]
     pub(super) clock: Clock,
 }
 
 impl<T> EngineTelemetry<T> {
     #[cfg(any(feature = "metrics", feature = "logs", test))]
-    pub fn new(inner: T, telemetry: TelemetryHelper, breaker_id: TelemetryString, clock: Clock) -> Self {
+    pub fn new(inner: T, telemetry: TelemetryHelper, breaker_id: crate::TelemetryString, clock: Clock) -> Self {
         Self {
             inner,
             telemetry,
@@ -38,7 +36,7 @@ impl<T> EngineTelemetry<T> {
     }
 
     #[cfg(not(any(feature = "metrics", feature = "logs", test)))]
-    pub fn new(inner: T, _telemetry: TelemetryHelper, _breaker_id: TelemetryString, _clock: Clock) -> Self {
+    pub fn new(inner: T, _telemetry: TelemetryHelper, _breaker_id: crate::TelemetryString, _clock: Clock) -> Self {
         Self { inner }
     }
 }
--- a/crates/seatbelt/src/utils/telemetry_helper.rs
+++ b/crates/seatbelt/src/utils/telemetry_helper.rs
@@ -1,14 +1,12 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-use crate::TelemetryString;
-
 #[derive(Debug, Clone)]
 pub(crate) struct TelemetryHelper {
     #[cfg(any(feature = "metrics", feature = "logs", test))]
-    pub(crate) pipeline_name: TelemetryString,
+    pub(crate) pipeline_name: crate::TelemetryString,
     #[cfg(any(feature = "metrics", feature = "logs", test))]
-    pub(crate) strategy_name: TelemetryString,
+    pub(crate) strategy_name: crate::TelemetryString,
     #[cfg(any(feature = "metrics", test))]
     pub(crate) event_reporter: Option<opentelemetry::metrics::Counter<u64>>,
     #[cfg(any(feature = "logs", test))]


crates/seatbelt/src/fallback/layer.rs
```diff
@@ -0,0 +1,293 @@
...
+    should_fallback: Option<ShouldFallback<Out>>,
+    fallback_action: Option<FallbackAction<Out>>,
+    enable_if: EnableIf<In>,
+    telemetry: TelemetryHelper,
+    _state: PhantomData<fn(In, S1, S2) -> Out>,
+}
+
+impl<In, Out> FallbackLayer<In, Out, NotSet, NotSet> {
+    #[must_use]
+    pub(crate) fn new(name: std::borrow::Cow<'static, str>, context: &crate::ResilienceContext<In, Out>) -> Self {
```
sandersaares: Some needlessly verbose type paths: std::borrow::Cow, crate::ResilienceContext. We can use `use` statements -they are free.

crates/seatbelt/src/fallback/callbacks.rs
```diff
@@ -0,0 +1,54 @@
...
+    /// Invoke the fallback action.
+    pub(crate) async fn call(&self, out: Out, args: FallbackActionArgs) -> Out {
+        match self {
+            Self::Sync(f) => f.call(out, args),
+            Self::Async(f) => f.call(out, args).await,
+        }
+    }
+}
+
+impl<Out> Clone for FallbackAction<Out> {
```
Vaiz, line 47: can u derive that?

crates/seatbelt/src/fallback/layer.rs
```diff
@@ -0,0 +1,293 @@
...
+    /// This call replaces any previous fallback action.
+    #[must_use]
+    pub fn fallback_output(self, value: Out) -> FallbackLayer<In, Out, S1, Set>
+    where
+        Out: Clone + Sync,
+    {
+        self.fallback(move |_, _| value.clone())
+    }
+
+    /// Sets an asynchronous fallback action.
```
sandersaares, line 128: Should we leave a note about safe composition here? e.g. "keep this inside the timeout layer so a timeout here gets aborted + retried correctly" or similar?

crates/seatbelt/src/fallback/layer.rs
```diff
@@ -0,0 +1,293 @@
...
+    }
+}
+
+impl<In, Out, S> Layer<S> for FallbackLayer<In, Out, Set, Set> {
+    type Service = Fallback<In, Out, S>;
+
+    fn layer(&self, inner: S) -> Self::Service {
+        let shared = FallbackShared {
+            enable_if: self.enable_if.clone(),
+            should_fallback: self.should_fallback.clone().expect("should_fallback must be set in Ready state"),
```
sandersaares: I like that the expect-messages here are positively-worded (instead of "is not" negative wording) but it sounds a bit like a requirement ("must be") whereas in reality we know it to be a fact (due to typestate) so even better would be "xyz is set - enforced by typestate pattern" (aside, it is not clear to me what a `Ready` state is - I do not see this term in this file, sounds like hallucination)

crates/seatbelt/src/lib.rs
```diff
@@ -6,7 +6,14 @@
...
 #![cfg_attr(
-    not(all(feature = "retry", feature = "timeout", feature = "breaker", feature = "metrics", feature = "logs")),
+    not(all(
+        feature = "retry",
+        feature = "timeout",
+        feature = "breaker",
+        feature = "fallback",
+        feature = "metrics",
+        feature = "logs"
+    )),
```
Vaiz, line 16: just replace it with allow

crates/seatbelt/Cargo.toml
```diff
@@ -36,10 +36,11 @@ allowed_external_types = [
 all-features = true
 
 [features]
-default = ["timeout", "retry", "breaker"]
+default = ["timeout", "retry", "breaker", "fallback"]
 timeout = []
 retry = ["dep:fastrand"]
 breaker = ["dep:fastrand"]
+fallback = []
```
Vaiz, line 43: is there any benefit to keeping it optional?

crates/seatbelt/src/fallback/service.rs
```diff
@@ -0,0 +1,311 @@
...
+    async fn handle_fallback(&self, output: Out) -> Out {
+        let new_output = self.fallback_action.call(output, FallbackActionArgs {}).await;
+
+        #[cfg(any(feature = "metrics", test))]
+        if self.telemetry.metrics_enabled() {
+            use crate::utils::{EVENT_NAME, PIPELINE_NAME, STRATEGY_NAME};
+
+            self.telemetry.report_metrics(&[
+                opentelemetry::KeyValue::new(PIPELINE_NAME, self.telemetry.pipeline_name.clone()),
+                opentelemetry::KeyValue::new(STRATEGY_NAME, self.telemetry.strategy_name.clone()),
```
Vaiz, line 198: afaik, if those names are regular `String`, it will do an additional allocation
it's better to keep `otel::Value` in `TelemetryHelper` type, or maybe even KeyValue

```rust
pub(crate) struct TelemetryHelper {
    #[cfg(any(feature = "metrics", feature = "logs", test))]
    pub(crate) pipeline_name: std::borrow::Cow<'static, str>,
    #[cfg(any(feature = "metrics", feature = "logs", test))]
    pub(crate) strategy_name: std::borrow::Cow<'static, str>,
```

crates/seatbelt/src/fallback/service.rs
```diff
@@ -0,0 +1,311 @@
...
+                opentelemetry::KeyValue::new(EVENT_NAME, super::telemetry::FALLBACK_EVENT_NAME),
+            ]);
+        }
+
+        #[cfg(any(feature = "logs", test))]
+        if self.telemetry.logs_enabled {
+            tracing::event!(
+                name: "seatbelt.fallback",
+                tracing::Level::WARN,
+                pipeline.name = %self.telemetry.pipeline_name,
```
Vaiz, line 208: I think you don't have to apply `Display` for string types

crates/seatbelt/src/fallback/service.rs
Vaiz: it might be useful to have our own lightweight primitive for such cases

```rust
pub enum TelemetryString {
   Static(&'static str),
   Arc(Arc<str>),
}

impl From<TelemetryString> for otel::Value {..}
```

crates/seatbelt/src/fallback/callbacks.rs
martintmk: It won't work if `Out` is not `Clone`. But explicit implementation works even for that case.

crates/seatbelt/src/fallback/layer.rs
martintmk: I feel this belongs more to general guidelines on the crate-level docs.

crates/seatbelt/src/fallback/layer.rs
martintmk: simplified

crates/seatbelt/src/fallback/layer.rs
martintmk: added

crates/seatbelt/src/fallback/service.rs
martintmk: I have introduced `TelemetryString` type alias, later we can extend it to custom type.

crates/seatbelt/src/fallback/service.rs
martintmk: in `tracing` this won't work, I need to put % ahead of `Cow<'static, str>`

crates/seatbelt/src/fallback/layer.rs
martintmk: improved (model just follow patterns from other middleware which was obsolete)

crates/seatbelt/Cargo.toml
martintmk: I want to be consistent, and if someone doesn't want fallback why compile it

crates/seatbelt/src/lib.rs
martintmk: unfortunately, there is no other way to express this intent:

"when all features are enabled, I want to see broken links"

Vaiz: imo, it doesn't worth the effort, just always allow broken links and it won't bother you

crates/seatbelt/src/fallback/service.rs
Vaiz: you should be able to call `as_str()` on Cow object

crates/seatbelt/src/lib.rs
martintmk: I want my docs to be always valid with proper-cross links when viewed on docs.rs. To me, it's worth extra boilerplate.

crates/seatbelt/src/fallback/service.rs
martintmk: for me, I rather keep it as it is (feels less verbose)

Vaiz (approved):

PR #295, opened 2026-03-01
feat: Add the file and sync_thunk crates

geeknoid:


crates/file/examples/basic_read_write.rs
```diff
@@ -0,0 +1,33 @@
...
+//!
+//! Demonstrates binding to a directory with [`Root::bind`] and using
+//! the [`Directory`] capability to read and write files.
+
+use file::Root;
+
+#[tokio::main]
+async fn main() -> std::io::Result<()> {
+    let tmp = tempfile::tempdir()?;
+    let dir = Root::bind(tmp.path()).await?;
```
sandersaares: If we envision `bind_tokio()` or `bind_foo()` here in the future to support different runtimes and I/O subsystems, should we also mangle the base name to avoid it being accidentally used as the default? `bind_fallback()` or `bind_std()` or `bind_threadpool()` or something?

crates/file/examples/file_types.rs
```diff
@@ -0,0 +1,41 @@
...
+
+#[tokio::main]
+async fn main() -> std::io::Result<()> {
+    let tmp = tempfile::tempdir()?;
+    let dir = Root::bind(tmp.path()).await?;
+
+    // WriteOnlyFile — can write, cannot read.
+    let mut wf = WriteOnlyFile::create(&dir, "data.txt").await?;
+    let mut buf = wf.reserve(64);
+    buf.put_slice(*b"written via WriteOnlyFile");
```
sandersaares, line 22: Can be shortened slightly as `BytesView::copied_from_slice(b"blah", &wf)`

crates/file/examples/file_types.rs
```diff
@@ -0,0 +1,41 @@
...
+    wf.write(buf.consume_all()).await?;
+    wf.flush().await?;
```
sandersaares, line 24: What are flush() semantics here? Is it just a pass-through to OS "flush" APIs? Or does it imply that the file types have some inherent internal buffers in the Rust universe? I assume the former but we should be careful not to give the wrong impression here.
We should also try support generic buffering layers built on `Read` and `Write` traits using something like `BufRead`, `BufWrite `and that we already have in the private repo, as well as via `ConcurrentRead` and `ConcurrentWrite` (ref `ConcurrentBufRead` in private repo) perhaps along the lines of #292 (comments welcome).

crates/file/examples/open_options.rs
```diff
@@ -0,0 +1,60 @@
...
+
+use file::{OpenOptions, Root};
+
+#[tokio::main]
+async fn main() -> std::io::Result<()> {
+    let tmp = tempfile::tempdir()?;
+    let dir = Root::bind(tmp.path()).await?;
+
+    // Create a new file for reading and writing.
+    let mut rw = OpenOptions::new()
```
sandersaares, line 20: File sharing options would also be useful to have and demonstrate (are concurrent readers/writers allowed).

crates/file/examples/positional_types.rs
```diff
@@ -0,0 +1,54 @@
...
+
+use file::{PositionalFile, ReadOnlyPositionalFile, Root, WriteOnlyPositionalFile};
+
+#[tokio::main]
+async fn main() -> std::io::Result<()> {
+    let tmp = tempfile::tempdir()?;
+    let dir = Root::bind(tmp.path()).await?;
+
+    // WriteOnlyPositionalFile — write-only access at explicit offsets.
+    let wf = WriteOnlyPositionalFile::create(&dir, "pos.txt").await?;
```
sandersaares, line 20: We should try ensure that this still remains compatible with `Read`, `Write`, `ConcurrentRead`, `ConcurrentWrite` traits. I suppose that could be layered on as an "operation" layer?

Something like `fn write_from(&self, offset) -> WriteHead<'a>` which implements `Write` and `ConcurrentWrite` for example, incrementing the offset internally as writes are accumulated?

This might also be a simple way to structure the forward-only file types internally, as limiting layers over the positional ones? If the wiring hooks up right.

crates/file/examples/streaming_io.rs
```diff
@@ -0,0 +1,43 @@
...
+        wf.write(buf.consume_all()).await?;
+    }
+    wf.flush().await?;
+    println!("wrote 10 chunks");
+
+    // Stream the file back in small pieces.
+    let mut rf = ReadOnlyFile::open(&dir, "stream.bin").await?;
+    let mut total = 0usize;
+    loop {
+        let chunk = rf.read_max(16).await?;
```
sandersaares, line 34: Not entirely sure what semantics of `read_max` are based on the example but is this not just reinventing stuff from [`ReadExt`](https://docs.rs/bytesbuf_io/latest/bytesbuf_io/trait.ReadExt.html) or `Read` traits? It is not clear to me why this is not just using the traits from `bytesbuf_io` for I/O functions.

crates/file/examples/basic_read_write.rs
geeknoid: Renamed.

PR #297, opened 2026-03-02
feat(seatbelt): improve telemetry

martintmk:
just a bit of cleanup to prepare for next set of changes

### DIFF
--- a/crates/bytesbuf/src/mem/global.rs
+++ b/crates/bytesbuf/src/mem/global.rs
@@ -516,7 +516,7 @@ mod tests {
     }
 
     #[test]
-    #[cfg(not(miri))] // Miri runtime scales with memory access size, so this takes forever.
+    #[cfg_attr(miri, ignore)] // Miri runtime scales with memory access size, so this takes forever.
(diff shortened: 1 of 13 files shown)


crates/seatbelt/src/utils/attributes.rs
```diff
@@ -41,4 +53,19 @@ mod tests {
     fn test_event_name_is_expected() {
         assert_eq!(EVENT_NAME, "resilience.event.name");
     }
+
+    #[test]
+    fn test_attempt_index_is_expected() {
+        assert_eq!(ATTEMPT_INDEX, "resilience.attempt.index");
```
Vaiz, line 59: kinda the same as `assert!(2==2)`

sandersaares: I feel offended by these tests.

martintmk: sorry folks :D

martintmk: will think about addressing, but it's not super urgent

PR #300, opened 2026-03-04
chore(seatbelt): get rid of warnings caused by hidden docs

martintmk:
These warnings become annoying enough to just show these items in docs:

image

Later if analysis is fixed, we can revert

### DIFF
--- a/crates/seatbelt/src/typestates.rs
+++ b/crates/seatbelt/src/typestates.rs
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+//! Type state markers for builder patterns.
+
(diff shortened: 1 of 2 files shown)


crates/seatbelt/src/typestates.rs
```diff
@@ -4,11 +4,9 @@
 /// A flag indicating that the required property is set.
 #[non_exhaustive]
 #[derive(Debug)]
-#[doc(hidden)]
 pub struct Set;
```
sandersaares, line 9: Where will these show up in the docs? Should we have some `seatbelt::typestates` module to hide them in, so they are at least "out of sight"?

martintmk: good idea, added

PR #323, opened 2026-03-13
fix: bump cargo-llvm-cov version to provide accurate results for codecov

### LINKED ISSUE #318: codecov ignores `coverage(off)`
`ohno/src/test_util.rs` already includes `#![cfg_attr(coverage_nightly, coverage(off))]` at the top of the file, but codecov ignores it and issues a warning about low code coverage. currently, we use a temporary fix where the file is explicitly added to [`codecov.yml`](https://github.com/microsoft/oxidizer/pull/315/changes#diff-3e878d9bdc47f29a0ab909a7db939acf08d2d2f1aaba81e6f897b32361fa40db)

Report: https://app.codecov.io/gh/microsoft/oxidizer/pull/315/blob/crates/ohno/src/test_util.rs

psandana:
Fixes #318 

cargo-llvm-cov was outdated, not producing accurate results for coverage in some portions of the workspace. 

Bumping version to 0.8.4

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -10,3 +10,5 @@ coverage:
         base: auto
 ignore:
 - "crates/testing_aids/**/*"
+# Macro expansion causes coverage to be inaccurate in these files, so we ignore them for now. We can re-enable them once we have a better solution for macro coverage.
+- "crates/ohno/src/test_util.rs"


codecov.yml
```diff
@@ -10,5 +10,5 @@ coverage:
         base: auto
 ignore:
 - "crates/testing_aids/**/*"
-# this is a temporary fix, for some reason codecov ignores cfg option in the file itself
+# Macro expansion causes coverage to be inaccurate in these files, so we ignore them for now. We can re-enable them once we have a better solution for macro coverage.
```
geeknoid, line 13: I'm confused. Does upgrading LLVM_COV's version actually fix the issue?

psandana: There are 2 issues. That's why is confusing. Macros are not well supported as they produce code, thus their sources are never really executed. This is a known limitation, and what this file is aiming for.

The second, is that some portions of the report were not marked as visited, like ohno backtrace.rs. This is fixed with the bump of the llvm-cov version.

PR #331, opened 2026-03-20
chore: don't run doc tests on miri

martintmk:
Many documentation examples use APIs that are not supported under Miri, mainly proc-macro crates in Tokio. Rather than polluting doc examples with `#[cfg(miri)]` guards or stripping them down to the point where they lose instructional value, we simply exclude doctests from Miri execution.

This is already reflected in CI (`--lib --tests`) and the `just miri` command (which uses `nextest`, a doctest-free runner). This PR makes the intent explicit and documented.


.github/workflows/main.yml
```diff
@@ -197,7 +197,7 @@ jobs:
         run: cargo +${{ env.RUST_NIGHTLY }} careful nextest run --all-features --workspace --color always --target ${{ matrix.target }}
       - name: Miri
         if: success() || failure()
-        run: cargo +${{ env.RUST_NIGHTLY }} miri test --all-features --workspace
+        run: cargo +${{ env.RUST_NIGHTLY }} miri test --all-features --workspace --lib --tests
```
psandana, line 200: do we need to test examples too?

martintmk: examples are not tested for miri I believe

schgoo (approved):

PR #336, opened 2026-03-24
feat: move Clock::timed method to tick

schgoo:
For cachet this helper function was added to a `Clock` extension to time async futures and return the result and duration (primarily for telemetry). I'm proposing moving this to the tick crate.


psandana (approved):

crates/tick/src/timed.rs
```diff
@@ -19,12 +19,11 @@ use crate::Stopwatch;
 /// # Examples
 ///
 /// ```
-/// use std::time::Duration;
-///
 /// use tick::{Clock, TimedResult};
 ///
 /// # async fn example(clock: &Clock) {
 /// let TimedResult { result, duration } = clock.timed(async { 42 }).await;
+/// println!("Result: {}, Duration: {:?}", result, duration);
```
psandana: leftover?

psandana (approved):

crates/tick/src/timed.rs
```diff
@@ -0,0 +1,68 @@
...
+/// use tick::{Clock, TimedResult};
+///
+/// # async fn example(clock: &Clock) {
+/// let TimedResult { result, duration } = clock.timed(async { 42 }).await;
+/// assert_eq!(result, 42);
+/// assert!(duration >= Duration::from_millis(0));
+/// # }
+/// ```
+#[derive(Debug, Clone, Copy)]
+pub struct TimedResult<R> {
```
sandersaares, line 33: A type named `XyzResult` that is not a type alias for `std::result::Result` is confusing - "results" in Rust tend to be fairly specifically type alises of `std::result::Result` with scenario-specific type parameters. We should avoid the word "result" when we are simply referring to outputs or return values.

Suggested name `TimedOutput` to align with `Future<Output = ...>` naming.

crates/tick/examples/timed_result.rs
```diff
@@ -0,0 +1,25 @@
...
+use tick::{Clock, TimedResult};
+
+#[tokio::main]
+async fn main() {
+    // Create a clock for the Tokio runtime.
+    let clock = Clock::new_tokio();
+
+    // Start some background work that returns a result after a delay.
+    let background_job = async {
+        clock.delay(std::time::Duration::from_millis(10)).await;
```
sandersaares, line 16: This is an oddly verbose type path. Do we really need `std::time::Duration` spelled out? Would be more readable if we used some more `use` statements.

sandersaares (changes requested):

schgoo:
Closing this until we have additional use cases

PR #350, opened 2026-03-31
refactor: Make CacheTier::len()/is_empty() async

schgoo:
Some types of distributed caches will need to make async calls. To prepare for that and to make it non-blocking this change converts `CacheTier::len()` to async. This is high priority because it is a breaking API change, and it needs to get made before we really start to onboard consumers. Making this change requires many updates to existing tests, and to implementations of `CacheTier`.

Since len() is used on every insert and invalidate to record the size of the cache to telemetry, this change also removes `record_size` until we can refactor it into configuration, instead of calling it on every `insert`.


Everod52 (approved):

crates/cachet/src/cache.rs
```diff
@@ -342,19 +342,25 @@ where
...
-        self.storage.len()
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the underlying storage tier fails.
+    pub async fn len(&self) -> Result<Option<u64>, Error> {
+        self.storage.len().await
     }
 
     /// Returns `true` if the cache **appears** to contain no entries.
```
psandana:
```suggestion
    /// Returns `Ok(true)` if the cache **appears** to contain no entries.
```

crates/cachet_service/src/adapter.rs
```diff
@@ -95,9 +95,9 @@ where
         }
     }
 
-    fn len(&self) -> Option<u64> {
+    async fn len(&self) -> Result<Option<u64>, Error> {
         // Service-based tiers typically don't expose length information
-        None
+        Ok(None)
```
psandana: I'm not fully convinced than returning an `Ok(None)` is descriptive enough. Instead of returning a `Result<Option<u64>, Error`, I would just return an `Result<u64, Error>` and in the Error encode two kind of errors: a catastrophic one, or a "Non-exposed length information" or "Length information not available".

That would simplify the API a lot. Not sure how frequent is your `Ok(None)`, but carries not information about what that means.

crates/cachet_tier/src/dynamic.rs
```diff
@@ -64,8 +64,8 @@ impl<K: Send + Sync, V: Send> CacheTier<K, V> for DynamicCache<K, V> {
         self.0.clear().await
     }
 
-    fn len(&self) -> Option<u64> {
-        self.0.len()
+    async fn len(&self) -> Result<Option<u64>, Error> {
+        self.0.len().await
```
psandana, line 69: Do we need to await here or just let the future to pass by? (I'm a bit rusty on this topic, so it is actually an open question)

crates/cachet_service/src/adapter.rs
schgoo: Added `LenError` with an `Unsupported` value here instead of `Option`

crates/cachet_tier/src/dynamic.rs
schgoo: Do you mean instead of returning Result we return a Future directly and make this not async? What's the benefit?

psandana: IIRC, a call to `await` the outside `len()`, will call on the inner `len()` and `await` on it. If we could return the Future in the inner `len()`, then you only require one `await` call on the original caller of the outside `len()`.

What I'm not sure if the compiler optimizes this or not. Anyhow, I thought it was easier to accomplish but requires transforming the inner `len()` into a normal function that returns a Future. So, I'm resolving this conversation.

crates/cachet/src/cache.rs
```diff
@@ -326,7 +326,7 @@ where
 
     /// Returns an **approximate** count of entries, if supported by the underlying storage.
     ///
-    /// Returns `None` if the underlying storage does not support size tracking.
+    /// Returns `Ok(None)` if the underlying storage does not support size tracking.
     ///
     /// # Approximation
```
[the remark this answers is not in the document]
psandana: Now it does not return option, right?

crates/cachet_tier/src/error.rs
```diff
@@ -267,4 +302,25 @@ mod tests {
         let error = Error::from_message("permanent failure");
         assert_eq!(error.recovery(), RecoveryInfo::never());
     }
+
+    #[test]
+    fn failed_lenerror_from_error() {
```
psandana: nit:

```suggestion
    fn failed_len_error_from_error() {
```

crates/cachet_tier/src/error.rs
```diff
@@ -267,4 +302,25 @@ mod tests {
...
+        let error = Error::from_message("permanent failure");
+        let len_err: LenError = error.clone().into();
+        assert_eq!(LenErrorKind::Failed, len_err.kind);
+        assert_eq!(error.to_string(), len_err.source().expect("should have source error").to_string());
+    }
+
+    #[test]
+    fn unsupported_lenerror_has_unsupported_kind() {
```
psandana:
```suggestion
    fn unsupported_len_error_has_unsupported_kind() {
```

crates/cachet_tier/src/error.rs
```diff
@@ -267,4 +302,25 @@ mod tests {
...
+        let len_err = LenError::unsupported();
+        assert_eq!(LenErrorKind::Unsupported, len_err.kind);
+    }
+
+    #[test]
+    fn lenerror_converts_to_error() {
```
psandana:
```suggestion
    fn len_error_converts_to_error() {
```

crates/cachet_tier/src/tier.rs
```diff
@@ -52,7 +51,7 @@ pub trait CacheTier<K, V>: Send + Sync {
 
     /// Returns an **approximate** count of entries, if the implementation supports it.
     ///
-    /// Returns `None` for implementations that do not track size.
+    /// Returns `Ok(None)` for implementations that do not track size.
```
psandana: To fix

crates/cachet_tier/src/tier.rs
```diff
@@ -31,9 +31,8 @@ use crate::{CacheEntry, Error};
 /// monotonicity guarantee only applies to reads and writes through the same
 /// Rust object.
 ///
-/// `len` and `is_empty` have default implementations:
+/// `len` has a default implementation:
 /// - `len`: Returns `None` (not all tiers track size)
```
psandana: To fix

crates/cachet/src/cache.rs
```diff
@@ -342,19 +342,23 @@ where
...
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the underlying storage tier fails.
+    pub async fn len(&self) -> Result<u64, LenError> {
+        self.storage.len().await
     }
 
-    /// Returns `true` if the cache **appears** to contain no entries.
+    /// Returns Ok(`true`) if the cache appears to contain no entries.
```
psandana:
```suggestion
    /// Returns Ok(`true`) if the cache contains no entries.
```

crates/cachet/src/cache.rs
```diff
@@ -342,19 +340,25 @@ where
...
+    /// # Errors
+    ///
+    /// Returns `Err(SizeError::unsupported())` if the underlying storage does not support size tracking.
+    /// Returns an error if the underlying storage tier fails.
+    pub async fn len(&self) -> Result<u64, SizeError> {
+        self.storage.len().await
     }
 
-    /// Returns `true` if the cache **appears** to contain no entries.
+    /// Returns Ok(`true`) if the cache appears to contain no entries.
```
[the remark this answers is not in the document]
psandana: This is a good comment @schgoo

crates/cachet_memory/tests/cache.rs
```diff
@@ -143,20 +143,20 @@ async fn clear_returns_ok() {
...
 
 #[cfg_attr(miri, ignore)]
-#[test]
-fn len_returns_some_not_none() {
+#[tokio::test]
+async fn len_returns_ok() {
     // Moka's entry_count() is eventually consistent, so we can't assert exact
     // counts immediately after insert. But we can verify that len() returns
-    // Some (not None), which catches the mutation `len -> None`.
+    // Ok (not Unsupported), which catches the mutation `len -> unsupported`.
```
psandana: this confuses me. Is this correct?

psandana: Is it Err()?

schgoo: No, it returns Ok because len is implemented for the memory cache.

psandana: Ok(not UnSupported)?

schgoo: Removing this test because it is redundant now.

psandana (approved):

PR #362, opened 2026-04-09
feat(http_extensions): add request/response timeouts and refactor body builder

martintmk:
Adds response-level and body-level idle timeouts to `http_extensions`, introduces `HttpBodyOptions` for per-body configuration, and restructures the body module.

### Timeouts

- `TimeoutBody` wraps streaming bodies with an idle timer that resets on each frame. On timeout, returns an error and drops the inner body; subsequent polls return `None` consistent with `is_end_stream()`.
- `ResponseTimeout` / `BodyTimeout` extension types attach to requests for middleware to enforce.
- `HttpRequestBuilder` gains `response_timeout()`, `body_timeout()`, and `timeout()` convenience methods.

### Body options

- New `HttpBodyOptions` with `timeout` and `buffer_limit` fields; per-call options merge on top of builder-level defaults.
- `body()` and `stream()` now accept `&HttpBodyOptions`; replaces the old `external()` method.

### Refactor

- Splits `body.rs` into `body/{mod, builder, options, timeout_body}.rs`.
- Renames `creator`/`create_body` to `builder` throughout.
- Fixes doc examples to be self-contained and runnable.

### Testing

All new types and methods are covered: `TimeoutBody` happy/timeout/post-timeout paths, `BodyTimeout`/`ResponseTimeout`, `HttpBodyOptions` merge semantics, and the new builder methods.


Vaiz (approved):

crates/http_extensions/src/body/builder.rs
```diff
@@ -0,0 +1,619 @@
...
+/// - Runtime-specific optimizations
+/// - Easier testing with the `test-util` feature
+///
+/// # Examples
+///
+/// ```
+/// # use http_extensions::{HttpBody, HttpBodyBuilder};
+/// # fn example(create_body: &HttpBodyBuilder) {
+/// // Create different body types
+/// let text_body: HttpBody = create_body.text("Hello world");
```
sandersaares: `create_body` sounds like legacy naming. Presumably it should just be `builder` now?

crates/http_extensions/src/body/builder.rs
```diff
@@ -0,0 +1,619 @@
...
+///
+/// - Memory pooling for better performance
+/// - Runtime-specific optimizations
+/// - Easier testing with the `test-util` feature
+///
+/// # Examples
+///
+/// ```
+/// # use http_extensions::{HttpBody, HttpBodyBuilder};
+/// # fn example(create_body: &HttpBodyBuilder) {
```
sandersaares: Is it possible to make the example actually run and not just compile? That would be preferable - example code that merely compiles is easy to go wrong. We should make sure our examples actually execute.

At minimum, if we really cannot find a way to run API docs example code, we should tag such examples with `no_run` so there is a corresponding alert in the docs GUI that the code has not been run as part of testing.

This comment applies to several example blocks I see here.

crates/http_extensions/src/body/builder.rs
```diff
@@ -0,0 +1,619 @@
...
+    /// to the global per-thread memory used by [`HttpBodyBuilder::new`].
+    #[must_use]
+    pub fn with_custom_memory(memory: impl MemoryShared) -> Self {
+        Self {
+            memory: MemoryWrapper::Opaque(Unaware(OpaqueMemory::new(memory))),
+            response_buffer_limit: None,
+        }
+    }
+
+    /// Sets the response buffer limit for all bodies created by this builder.
```
sandersaares: This description is not informative and just repeats the name of the function. What is a "response buffer limit"? The type does not appear to explain this term anywhere. It is not clear to me what this is for, how it is meant to be used or what impact it has on HTTP client behavior (whether it causes some things to become errors, or just works as an optimization, or something else entirely).

crates/http_extensions/src/body/builder.rs
```diff
@@ -0,0 +1,619 @@
...
+    /// a timeout error is returned.
+    ///
+    /// The deadline is computed once at construction time; successive polls see a shrinking
+    /// remaining time rather than a fixed per-poll timeout.
+    ///
+    /// If `timeout` is so large that adding it to the current instant overflows, the timeout
+    /// is silently dropped and the body behaves as if created via [`custom_body`][Self::custom_body].
+    /// In practice such a duration (e.g. [`Duration::MAX`][std::time::Duration::MAX]) is far longer than any realistic
+    /// connection lifetime, so the timeout would never fire anyway.
+    pub fn custom_body_with_timeout<B>(&self, body: B, timeout: std::time::Duration, clock: &Clock) -> HttpBody
```
sandersaares: Feels strange to see such a combined method (and also the repetition with `stream_with_timeout()`). Why is it not something like `builder.timeout(..).custom_body(..)`? That would feel more "builder like". Is there a Rustological limitation here?

crates/http_extensions/src/body/builder.rs
```diff
@@ -0,0 +1,619 @@
...
+    /// entire data reception. If the body is not fully received within the specified duration
+    /// a timeout error is returned.
+    ///
+    /// The deadline is computed once at construction time; successive polls see a shrinking
+    /// remaining time rather than a fixed per-poll timeout.
+    ///
+    /// If `timeout` is so large that adding it to the current instant overflows, the timeout
+    /// is silently dropped and the body behaves as if created via [`custom_body`][Self::custom_body].
+    /// In practice such a duration (e.g. [`Duration::MAX`][std::time::Duration::MAX]) is far longer than any realistic
+    /// connection lifetime, so the timeout would never fire anyway.
```
sandersaares: This last paragraph feels a bit too "impelmentation details". If the user cannot distinguish the difference between "the timeout is enforced but can never be reached" and "the timeout is not enforced" we might as well say nothing and pretend we enforce even Duration::MAX timeouts because nobody can prove us wrong. I suggest moving this paragraph from API documentation to an inline comment on the `match` below. This makes for a simpler to understand API surface.

crates/http_extensions/src/body/builder.rs
```diff
@@ -0,0 +1,619 @@
...
+    ///     }
+    /// }
+    ///
+    /// # fn example(create_body: &HttpBodyBuilder) {
+    /// // Create HttpBody from your custom body
+    /// let custom_body = CustomBody(vec![1, 2, 3, 4]);
+    /// let body = create_body.custom_body(custom_body);
+    /// # }
+    /// ```
+    pub fn custom_body<B>(&self, body: B) -> HttpBody
```
sandersaares: Naming suggestion: `body()`

crates/http_extensions/src/body/timeout_body.rs
```diff
@@ -0,0 +1,327 @@
...
+            *this.current_delay = None;
+            return Poll::Ready(result);
+        }
+
+        // Inner body is pending — enforce the deadline via a delay.
+        // Reuse the existing delay if we are re-polled without the inner body
+        // making progress, or create a new one for the remaining time.
+        let delay = this.current_delay.get_or_insert_with(|| Delay::new(this.clock, remaining));
+
+        if Pin::new(delay).poll(cx).is_ready() {
```
sandersaares, line 72: I observe here that we take advance of the fact that `Delay: Unpin`. Is this a promise we are intentionally making in `tick`, across all possible runtimes it may support, past and future? Just wanted to check that we are not depending on a coincidence.

crates/http_extensions/src/request_timeout.rs
```diff
@@ -0,0 +1,90 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use std::time::Duration;
+
+/// A request-level timeout that can be attached to HTTP requests as an extension.
+///
+/// This timeout represents the maximum time allowed for the entire request/response cycle,
+/// including receiving the response headers and reading all data from the HTTP body. This
```
sandersaares: I assume this means "HTTP response body"? Would be good to clarify.

crates/http_extensions/src/request_timeout.rs
```diff
@@ -0,0 +1,90 @@
...
+/// is different from the body timeout that only limits the time spent streaming the
```
sandersaares: What does "the body timeout" refer to exactly? Might be good to be more explicit (can we link something?).

crates/http_extensions/src/http_response_builder.rs
```diff
@@ -414,7 +414,7 @@ impl HttpResponseBuilder<'_> {
     where
         B: http_body::Body<Data = BytesView, Error: Into<HttpError>> + Send + 'static,
     {
-        let body = self.creator.external(body);
```
sandersaares, line 417: The word "creator" here also feels outdated if we really call it a "builder" in public API.

sandersaares (commented):
Left some comments. A few are not really related to this PR but I took the opportunity to provide feedback for whatever showed up in the diff anyway.

crates/http_extensions/src/body/timeout_body.rs
martintmk: Essentially yes, if type is `Unpin` it's now bound with API contract and removing it is a breaking change.

crates/http_extensions/src/body/builder.rs
martintmk: will do that in a separate follow-up pr

crates/http_extensions/src/body/builder.rs
martintmk: fixed

crates/http_extensions/src/body/builder.rs
martintmk: addressed, now we are using BodyOptions

crates/http_extensions/src/http_response_builder.rs
martintmk: adressed

crates/http_extensions/src/request_timeout.rs
martintmk: reworked

crates/http_extensions/src/request_timeout.rs
martintmk: also reworked, we now have a dedicated ResponseTimeout and BodyTimeout

crates/http_extensions/Cargo.toml
```diff
@@ -86,7 +86,7 @@ serde_core = { workspace = true }
 serde_json = { workspace = true, features = ["std"] }
 static_assertions = { workspace = true }
 templated_uri = { workspace = true, features = ["uuid"] }
-tick = { workspace = true, features = ["test-util"] }
+tick = { workspace = true, features = ["test-util", "tokio"] }
 tokio = { workspace = true, features = ["macros", "rt-multi-thread", "net"] }
```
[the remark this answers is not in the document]
martintmk: no true, this is fixed already

crates/http_extensions/src/body/builder.rs
```diff
@@ -0,0 +1,597 @@
...
+    /// When using this method, the memory is shared across all threads as opposed
+    /// to the global per-thread memory used by [`HttpBodyBuilder::new`].
+    #[must_use]
+    pub fn with_custom_memory(memory: impl MemoryShared, clock: &Clock) -> Self {
+        Self {
+            memory: MemoryWrapper::Opaque(Unaware(OpaqueMemory::new(memory))),
+            clock: clock.clone(),
+            options: BodyOptions::default(),
+        }
+    }
```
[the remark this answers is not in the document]
martintmk: yes, this is breaking change

crates/http_extensions/src/http_request_builder.rs
```diff
@@ -235,6 +237,36 @@ impl<R> HttpRequestBuilder<'_, R> {
...
+    }
+
+    /// Sets both the response timeout and the body timeout to the same duration.
+    ///
+    /// This is a convenience method equivalent to calling both
+    /// [`response_timeout`](Self::response_timeout) and
+    /// [`body_timeout`](Self::body_timeout) with the same value.
+    pub fn timeout(self, duration: Duration) -> Self {
+        self.response_timeout(duration).body_timeout(duration)
+    }
```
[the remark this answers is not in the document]
martintmk: indeed, this needs to be adopted in fetch, outside of scope here

sandersaares (approved):

PR #366, opened 2026-04-10
feat: introduce ohno::ErrorLabel

martintmk:
This PR introduces a new `ErrorLabel` type that encapsulates the telemetry-friendly error information.

Initially, I introduced an http-only `HttpErrorLabel` type, but then realized this might be universal enough to be ohno.

```rust
use ohno::ErrorLabel;

// From a static string
let label: ErrorLabel = "timeout".into();
assert_eq!(label, "timeout");

// Dotted chain from parts
let label = ErrorLabel::from_parts(["http", "client", "timeout"]);
assert_eq!(label, "http.client.timeout");
```


crates/http_extensions/src/error_label.rs
```diff
@@ -0,0 +1,204 @@
...
+
+impl Default for HttpErrorLabel {
+    fn default() -> Self {
+        Self(Cow::Borrowed(""))
+    }
+}
+
+impl From<ErrorKind> for HttpErrorLabel {
+    fn from(s: ErrorKind) -> Self {
+        format!("io.{}", s.to_string().replace(' ', "_")).into()
```
Vaiz: NIT: it's better to format into string and then do replacement as it avoids intermediate allocation

```rust
format!("io.{s}").replace(' ', "_")).into()
```

Vaiz (approved):

crates/http_extensions/src/error_label.rs
```diff
@@ -0,0 +1,383 @@
...
+    /// let http_err = HttpError::other(io_err, RecoveryInfo::retry(), "connect");
+    /// let label = HttpErrorLabel::from_error_chain(&http_err);
+    /// assert_eq!(label, "connect.io.connection_refused");
+    /// ```
+    #[must_use]
+    pub fn from_error_chain(error: &(dyn Error + 'static)) -> Self {
+        let chain = successors(Some(error), |e| (*e).source()).filter_map(get_label_from_error);
+
+        Self::from_parts(chain)
+    }
```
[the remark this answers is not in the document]
martintmk: will be handled by caller

crates/http_extensions/src/error_label.rs
```diff
@@ -0,0 +1,381 @@
...
+
+fn get_label_from_error(error: &(dyn Error + 'static)) -> Option<HttpErrorLabel> {
+    if let Some(err) = error.downcast_ref::<std::io::Error>() {
+        return Some(err.kind().into());
+    }
+
+    if let Some(err) = error.downcast_ref::<HttpError>() {
+        return Some(err.label().clone());
+    }
+
```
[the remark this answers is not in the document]
martintmk: i rather keep the new variants visible on new rust releases that add error kind and then just make perf-related minor version upgrade

crates/error_label/src/lib.rs
```diff
@@ -0,0 +1,381 @@
...
+        let mut result = match parts.next() {
+            Some(first) => String::from(first.as_ref()),
+            None => return Self::default(),
+        };
+        for part in parts {
+            result.push('.');
+            result.push_str(part.as_ref());
+        }
+        Self(Cow::Owned(result))
+    }
```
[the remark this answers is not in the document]
martintmk: true, I'll handle single case variant

crates/http_extensions/src/error_label.rs
```diff
@@ -0,0 +1,342 @@
...
+    #[test]
+    fn from_error_chain_http_wrapping_io() {
+        let io_err = std::io::Error::new(std::io::ErrorKind::BrokenPipe, "pipe");
+        let http_err = HttpError::from(io_err);
+        let label = HttpErrorLabel::from_error_chain(&http_err);
+        // HttpError converted from io::Error gets label "broken_pipe",
+        // and the inner io::Error also contributes "broken_pipe".
+        assert_eq!(label, "broken_pipe.broken_pipe");
+    }
```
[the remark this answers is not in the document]
martintmk: addressed

crates/error_label/src/lib.rs
```diff
@@ -0,0 +1,400 @@
...
+    /// ```rust
+    /// # use error_label::ErrorLabel;
+    /// let label = ErrorLabel::from_parts(["http", "timeout"]);
+    /// assert_eq!(label, "http.timeout");
+    ///
+    /// let label = ErrorLabel::from_parts(["a", "b", "c"]);
+    /// assert_eq!(label, "a.b.c");
+    /// ```
+    #[must_use]
+    pub fn from_parts(parts: impl IntoIterator<Item = impl AsRef<str>>) -> Self {
```
Vaiz: how it's compatible with "Low-cardinality" part? it probably leads to the opposite effect

sgalkin: Should depth be limited too? Each step might increase cardinality a lot.

crates/ohno/src/error_label.rs
```diff
@@ -0,0 +1,400 @@
...
+use std::borrow::Cow;
+use std::collections::hash_set::HashSet;
+use std::error::Error;
+use std::fmt;
+use std::io::ErrorKind;
+use std::iter::successors;
+
+/// A low-cardinality label for an error, useful for metrics and logging.
+///
+/// # Cardinality requirements
```
Vaiz, line 18: what about allowed symbols, enforcing lower case, etc?

sgalkin (approved):

crates/ohno/src/error_label.rs
```diff
@@ -0,0 +1,400 @@
...
+        let mut parts = parts.into_iter().filter(|v| !v.as_ref().is_empty());
+        let mut result = match parts.next() {
+            Some(first) => String::from(first.as_ref()),
+            None => return Self::default(),
+        };
+        for part in parts {
+            result.push('.');
+            result.push_str(part.as_ref());
+        }
+        Self(Cow::Owned(result))
```
Vaiz, line 116: I really think it should cache already allocated labels to avoid additional allocations
maybe we could use SymbolTable or something similar

crates/ohno/src/error_label.rs
martintmk: at this point, I don't have strong need to enforce particular characters, it would just create additional overhead.  We can fine-tune by usage later.

Vaiz: what about automatically transforming everything to lowercase to avoid having two distinct labels that differs only in case, e.g. `Fmt_Error` and `fmt_error`

crates/ohno/src/error_label.rs
martintmk: I would leave this as perf improvement later, maybe also introduce a `compact_str` support that would reduce allocations for smaller labels (most common use-case).

crates/error_label/src/lib.rs
martintmk: changed the signature to indicate this also requires something that is also `ErrorLabel`. This preserves the semantics for each part.

martintmk: for large error chains, this this might be a problem. Practically, I have not seen cases where this would balloon out of proportions. Practically, our values would be something like:

- `connect.hyper.timeout`
- `connect.hyper.addr_unavailable`

crates/ohno/src/error_label.rs
martintmk: done

crates/ohno/src/error_label.rs
```diff
@@ -0,0 +1,572 @@
...
+        &self.0
+    }
+}
+
+impl From<ErrorKind> for ErrorLabel {
+    /// Creates a label from an IO error kind.
+    ///
+    /// Maps each [`ErrorKind`] variant to a `snake_case` string label suitable for use as a
+    /// metric tag. Unrecognized variants (e.g. future additions to [`ErrorKind`]) are converted
+    /// using their [`Display`] representation with spaces replaced by underscores.
```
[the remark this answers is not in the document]
martintmk: already fixed

crates/http_extensions/src/error.rs
```diff
@@ -333,7 +333,11 @@ mod tests {
         let error = HttpError::from(std::io::Error::other("test"));
         assert_eq!(error.message(), "test");
         assert_eq!(error.recovery(), RecoveryInfo::never());
-        assert_eq!(error.label(), "io");
+        assert_eq!(error.label(), "other");
+
+        let error = HttpError::from(std::io::Error::new(std::io::ErrorKind::BrokenPipe, "some message"));
+        assert_eq!(error.recovery(), RecoveryInfo::retry());
+        assert_eq!(error.label(), "broken_pipe");
```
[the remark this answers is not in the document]
martintmk: this is intended

crates/ohno/src/error_label.rs
```diff
@@ -0,0 +1,500 @@
...
+    /// ```rust
+    /// # use ohno::ErrorLabel;
+    /// let label = ErrorLabel::from_parts(["http", "timeout"]);
+    /// assert_eq!(label, "http.timeout");
+    ///
+    /// let label = ErrorLabel::from_parts(["a", "b", "c"]);
+    /// assert_eq!(label, "a.b.c");
+    /// ```
+    #[must_use]
+    pub fn from_parts(parts: impl IntoIterator<Item = impl Into<Self>>) -> Self {
```
Vaiz, line 106: does the order matter here? we can reduce cardinality by sorting parts

crates/ohno/src/error_label.rs
```diff
@@ -0,0 +1,500 @@
...
+    ///
+    /// let label = ErrorLabel::from(ErrorKind::ConnectionRefused);
+    /// assert_eq!(label, "connection_refused");
+    /// ```
+    fn from(kind: ErrorKind) -> Self {
+        Self::from_io(kind)
+    }
+}
+
+const fn is_valid_label_char(b: u8) -> bool {
```
Vaiz, line 289:
```rust
const fn is_valid_label_char(b: u8) -> bool {
    matches!(b, b'0'..=b'9' | b'a'..=b'z' | b'_' | b'.')
}
```

crates/ohno/src/error_label.rs
martintmk: yup, the order does matter, we want to preserve it. Usually it goes from lower cardinality to higher ones:

- `request.connect.connection_refused`

This tells us that sending the request failed because the connection failed to created due to refusal. This is created from the chain of 3 errors.

martintmk: We can also have:

`request.connection_reset`

This tells us that the request failed on already established connection because it was reset.

crates/ohno/src/error_label.rs
martintmk: refined the implementation, used match arm for better clarity

Vaiz (approved):

Vaiz (approved):

PR #369, opened 2026-04-10
feat(cachet_memory): Add LRU eviction policy for in memory cache

schgoo:
Adding LRU eviction policy support for the `InMemoryCache`

This is simply an abstraction around the `moka::policy::EvictionPolicy`. We do not want consumers to have to take a direct dependency on `moka`.


Everod52 (approved):

crates/cachet_memory/src/policy.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    pub(crate) fn into_moka_policy(self) -> moka::policy::EvictionPolicy {
+        match self.inner {
+            EvictionPolicyInner::TinyLfu => moka::policy::EvictionPolicy::tiny_lfu(),
+            EvictionPolicyInner::Lru => moka::policy::EvictionPolicy::lru(),
+        }
+    }
+}
+
+impl fmt::Debug for EvictionPolicy {
```
Vaiz, line 54: could you add Display impl as well?

Vaiz (approved):

crates/cachet_memory/src/policy.rs
schgoo: @Vaiz I can do that. So I understand, when is it recommended that we implement Display as well as Debug?

schgoo: Wait, actually, you can't implement both, can you? Doesn't the `fmt` method conflict? Nevermind I just tested it and it does work somehow.

Vaiz: > Wait, actually, you can't implement both, can you?

you sure can


rules are simple:
- add `impl Display` if you expect something to be logged (debug should not be used in telemetry)
- add `impl Debug` almost always, especially for public types; you never know when you might need to dump a structure to figure out an obscure bug

Vaiz (approved):

PR #370, opened 2026-04-12
feat(recoverable): support conversion from std::io::error::ErrorKind

martintmk:
I have found already multiple places where this code is duplicated. It's better to have some implicit opinionated defaults.


crates/recoverable/src/io.rs
```diff
@@ -0,0 +1,88 @@
...
+    fn from(kind: ErrorKind) -> Self {
+        match kind {
+            ErrorKind::WouldBlock
+            | ErrorKind::TimedOut
+            | ErrorKind::ConnectionReset
+            | ErrorKind::ConnectionAborted
+            | ErrorKind::NotConnected
+            | ErrorKind::ConnectionRefused
+            | ErrorKind::AddrInUse
+            | ErrorKind::AddrNotAvailable
```
sgalkin, line 67: how this unavailable is different from unavailable states defined on L68?

sgalkin: for example difference between "interface temporarily unavailable" and "network interface is down" is not 100% clear

sgalkin (approved):

crates/recoverable/src/io.rs
martintmk: The "AddrNotAvailable" is commonly related with port exhaustion. While I see a case for it being unavailable classification, our partners see this as a common retry scenario. I'll keep it as is, see how it behaves and we can refine later.

sgalkin (approved):

martin-kolinek (approved):

PR #373, opened 2026-04-14
feat(layered): relax overly restrictive static bounds

martintmk:
There are few places where the 'static requirement is unnecessarily enforced.


Vaiz (approved):

sandersaares (approved):
I suggest adding some tests with non-`'static` data. I see that some bounds were removed but I see no evidence it actually works.

sandersaares:
I see a bunch of tests added now but aren't those all still `'static`? Am I missing the non-`'static` references somewhere?

martintmk:
> I see a bunch of tests added now but aren't those all still `'static`? Am I missing the non-`'static` references somewhere?

The explicit type annotation makes it clear that we are dealing with non-static strings:

```rust
    let context: ResilienceContext<&str, &str> = ResilienceContext::new(&clock);
```

And just to make sure, I did update the "references.cs" composite test:

```rust
    let input = "hello".to_string();
    let service = stack.into_service();
    let output = service.execute(&input).await;
```

sandersaares:
> The explicit type annotation makes it clear that we are dealing with non-static strings:

I do not see any non-`'static` lifetimes listed in the code you pasted, though. Am I missing something?

> And just to make sure, I did update the "references.cs" composite test:

OK, that pattern does look believable, though.

sandersaares (approved):

crates/layered/src/intercept.rs
```diff
@@ -374,15 +374,15 @@ impl<In, Out, S> crate::Layer<S> for InterceptLayer<In, Out> {
     }
 }
 
-struct OnInput<In>(Arc<dyn Fn(&In) + Send + Sync>);
+struct OnInput<In>(Arc<dyn for<'a> Fn(&'a In) + Send + Sync>);
```
afoxman, line 377:
```suggestion
struct OnInput<'a, In>(Arc<dyn Fn(&'a In) + Send + Sync>);
```
I think something like this will get you closer, but the hang-up is that the closure lifetime is much longer than &In... and the closure could borrow and hold using the ref. 

A stateless function, on the other hand, can't do that. So it solves the problem, but takes away closures. And users who need stateful calls would have to put their "context" data in `In` which is gross. Sorry, all I have is bad news :|

A cheaper, backward compatible fix would be to document that this isn't supported :)

afoxman (approved):
Thanks for refining the lifetimes all over the crate!

martintmk:
currently, we haven't found a way to make this work with references without poisoning the service lifetime. The go-to solution is to use owned types for now.

crates/layered/src/intercept.rs
martintmk: I am starting to realize that this indeed might not be possible with current Rust :(

sgalkin (approved):

martintmk:
this is not fixable at this point

PR #377, opened 2026-04-15
feat(cachet): add serialization support with PostcardEncoder/PostcardCodec

schgoo:
# Summary

Adds a serialization boundary for the cache pipeline, enabling typed <K, V> caches to automatically serialize keys and values into
BytesView for downstream tiers (e.g., Redis, Memcached, or any byte-oriented store).

### Serialization codecs (`cachet::transform::serialize`)

- BincodeEncoder - one-directional `T -> BytesView` via bincode, writing directly into pool-backed `BytesBuf` memory
- BincodeCodec - bidirectional `T <-> BytesView` for value round-tripping

### Builder API

- `.serialize()` on `CacheBuilder` and `FallbackBuilder` - applies a bincode serialization boundary, returning a `TransformBuilder` that accepts `BytesView`-typed fallback tiers
- `promotion_policy()` and `time_to_refresh()` on `TransformBuilder` - previously these were hardcoded to `always()` and None; now configurable at the transform boundary

### Supporting changes

- `DistributedCacheTier` - new marker trait for tiers that operate on serialized `BytesView` data
- `CacheEntry` `PartialEq` - custom impl comparing only value + ttl (excludes cached_at), relaxed bound from `V: PartialEq + Eq` to `V`: `PartialEq`
- `BytesView` Hash - required for use as cache keys in hash-map-based tiers
- Add `impl Into<CacheEntry<V>>` for insert to simplify semantics

### Documentation

- Added serialize feature to the Cargo features table in lib.rs
- Added "Serialization Boundary" example section in lib.rs module docs
- Added doc example on `.serialize()` builder method
- New `examples/serialization.rs` standalone example

### Tests

- Integration test verifying full round-trip: insert typed values -> serialize to bincode bytes -> invalidate primary -> fetch from serialized fallback -> deserialize back to typed values
- Asserts exact bincode byte encoding to detect any serialization format changes

### Usage
```rust
let redis = RedisCacheTier::new("redis://localhost:6379");

let cache = Cache::builder::<String, String>(clock)
    .memory()
    .serialize()
    .fallback(
        Cache::builder::<BytesView, BytesView>(clock)
            .storage(redis) // any DistributedCacheTier
    )
    .promotion_policy(FallbackPromotionPolicy::always())
    .build();

// Typed on the outside, serialized bytes on the inside.
cache.insert("user:1", "Alice").await?;
let val = cache.get("key").await?;
```


crates/cachet/src/builder/transform.rs
```diff
@@ -192,34 +197,54 @@ where
     }
 }
 
-// ── Sealed + CacheTierBuilder ──
+// TODO - Dedupe Transform/Fallback builders
```
afoxman: nit: is this TODO tracked? don't want to lose it...

crates/cachet/src/transform/serialize.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+impl<T: Serialize + Send + Sync> Encoder<T, BytesView> for BincodeCodec {
+    fn encode(&self, value: &T) -> Result<BytesView, Error> {
+        encode(value)
+    }
+}
+
+fn encode<T: Serialize + Send + Sync>(value: &T) -> Result<BytesView, Error> {
+    let mut writer = BytesBuf::new().into_writer(GlobalPool::new());
```
afoxman: Should the pool selection be controllable through config? Is pool selection and reuse even a thing? I'm new here.

afoxman (approved):
This all looks good to me. Maybe look at using feature gates for things like serialization in this PR, which bring in more dependencies. This could be a future task - no need to hold up this PR.

crates/cachet/src/transform/serialize.rs
schgoo: I actually need to look more deeply at this serialization impl. The primary issue is the bincode crate is no longer maintained. No other crate provides streaming reads and zero intermediate allocations.

crates/bytesbuf/src/view.rs
```diff
@@ -864,6 +865,18 @@ impl<const LEN: usize> PartialEq<BytesView> for &[u8; LEN] {
 
 impl Eq for BytesView {}
 
+impl Hash for BytesView {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        // Hash all bytes in logical order, consistent with PartialEq.
+        // We also hash the cached length as an additional input, but equality and hashing are
+        // both defined by the logical byte sequence rather than how it is segmented into spans.
+        self.len.hash(state);
+        for (slice, _) in self.slices() {
```
Vaiz: what is the second argument? why it doesn't matter for hashing?

crates/cachet_tier/src/distributed.rs
```diff
@@ -0,0 +1,16 @@
...
+
+use crate::CacheTier;
+
+/// A [`CacheTier`] that stores serialized bytes.
+///
+/// This is a marker trait for cache tiers that communicate over the network
+/// (e.g., Redis, Memcached) where keys and values are already serialized to
+/// [`BytesView`]. Use the `.serialize()` builder method to insert a serialization
+/// boundary before a distributed tier.
+pub trait DistributedCacheTier: CacheTier<BytesView, BytesView> {}
```
Vaiz: I'm not sure if having an additional trait helps in anyway
can we just use `CacheTier<BytesView, BytesView>` as a requirement?

crates/cachet_tier/src/entry.rs
```diff
@@ -139,6 +139,15 @@ impl<V> CacheEntry<V> {
     }
 }
 
+impl<V> PartialEq for CacheEntry<V>
+where
+    V: PartialEq,
+{
+    fn eq(&self, other: &Self) -> bool {
+        self.value() == other.value() && self.ttl == other.ttl()
```
Vaiz: interesting question, when you compare two entries, do you care about TTL or it's just additional metadata that doesn't matter?

I assume in most cases you want to be explicit if you compare only value, or value + ttl, so maybe it's better to remove `PartialEq` implementation completely?

crates/cachet/src/transform/serialize.rs
```diff
@@ -0,0 +1,59 @@
...
+
+use serde::{Serialize, de::DeserializeOwned};
+
+/// An encoder that serializes values to [`BytesView`] using bincode (one-directional).
+///
+/// Implements `Encoder<T, BytesView>` for any `T: Serialize + Send + Sync`.
+///
+/// For bidirectional serialization/deserialization, use [`BincodeCodec`].
+#[derive(Debug, Clone, Copy)]
+pub struct BincodeEncoder;
```
Vaiz: who is responsible for keeping serialization compatible between different versions? or at least detecting that format has changed instead of crashing the service

potentially, this also can open a way for security vulnerabilities

crates/bytesbuf/src/view.rs
schgoo: It is the slice index. I didn't think we would want the hash to change if the bytes were segmented differently. There is maybe some inefficiency in self.slices() cloning the BytesView and bumping the ref-count for all the slices (which the PartialEq impl accounts for)

crates/cachet/src/builder/transform.rs
schgoo: Yes, it is.

crates/cachet/src/transform/serialize.rs
schgoo: Also, good catch on the pool selection. This is actually the worst possible thing to do; we should be reusing a pool rather than creating a new one for every encode. I think for now we'll use a thread-local pool and not force this decision on the user. It's a serialization implementation detail.

crates/cachet_tier/src/distributed.rs
schgoo: Yeah, we probably don't need it yet

crates/cachet/src/transform/serialize.rs
schgoo: There are three separate issues here:
1. Format compatibility: cache data is ephemeral. A format change should not affect local caches. For distributed/persisted caches...
2. Deserialization errors: deserialization errors will return a Result. Here we can either handle it ourselves or let the user handle it. I think error handling in cachet needs some sizable improvements in general, this is probably in scope for that
3. Security: this is a very good point, especially with postcard, which uses varint encoding. We could embed a format version, or add a payload checksum, or both. I'll look into it

schgoo: I think the security question will actually answered by the next PRs for compression/encryption. If the user is encoding security-sensitive data, they should also be encrypting it. A format change would fail to decrypt, triggering a cache miss

Vaiz: what about malicious data that exploit bugs in deserialization to execute payload's code?
the moment we start providing serialization/deserialization algorithms, we open a door for all kinds of problems.

Here is an example with JSON library in dotnet - https://github.com/advisories/GHSA-8g4q-xg66-9fp4

disclaimer: it's not my area of expertise, so maybe I'm overreacting here

schgoo: Hmm, I'm not sure this example is relevant. That's not to say there isn't some way deserialization could get exploited here. For example, since postcard uses varint decoding, a fixed length prefix could cause issues. Of course, that would assume that the bad actor already has access to the cache and is adding invalid entries. This would be solved by encryption, network isolation, etc.

crates/cachet_tier/src/entry.rs
schgoo: Good call. We should just be comparing the dereferenced value, rather than having a weird Eq implementation like this.

crates/cachet/examples/serialization.rs
```diff
@@ -0,0 +1,42 @@
...
+    let key = "greeting".to_string();
+
+    // Insert a typed value — it's serialized to BytesView before reaching L2.
+    cache
+        .insert(key.clone(), CacheEntry::new("Hello, world!".to_string()))
+        .await
+        .expect("insert failed");
+
+    // Get returns a typed value — deserialized from BytesView if fetched from L2.
+    let value = cache.get(&key).await.expect("get failed");
```
Vaiz: cache should always return the value we just inserted
```suggestion
let value = cache.get(&key).await.expect("get failed").unwrap();
```

crates/cachet/src/serialize/codec.rs
```diff
@@ -0,0 +1,187 @@
...
+#[derive(Debug, Clone)]
+pub struct PostcardCodec;
+
+impl<T: Serialize + Send + Sync> Encoder<T, BytesView> for PostcardEncoder {
+    fn encode(&self, value: &T) -> Result<BytesView, Error> {
+        encode(value)
+    }
+}
+
+impl<T: Serialize + Send + Sync> Encoder<T, BytesView> for PostcardCodec {
```
Vaiz, line 66: NIT: for better readability, it's  better to group struct definition and impl blocks together. Here `PostcardCodec` definition is placed in between `PostcardEncoder` definition and `impl PostcardEncoder`

crates/cachet/src/serialize/codec.rs
```diff
@@ -0,0 +1,187 @@
...
+    fn encode(&self, value: &T) -> Result<BytesView, Error> {
+        encode(value)
+    }
+}
+
+fn encode<T: Serialize + Send + Sync>(value: &T) -> Result<BytesView, Error> {
+    // TODO make Cache thread aware so we can simply store the pool in the encoder
```
Vaiz: you can use [thread_local](https://docs.rs/thread_local/latest/thread_local/) crate for this

crates/cachet/src/serialize/codec.rs
```diff
@@ -0,0 +1,187 @@
...
+    // Until then we need this to avoid creating a new pool for every encode call, which would be
+    // very expensive
+    thread_local! {
+        static POOL: GlobalPool = GlobalPool::new();
```
Vaiz: creating ad-hoc memory pools in random places may lead to unexpected increase in memory consumption

another problem - `bytesbuf::mem::GlobalPool` never releases the memory

crates/cachet/src/serialize/codec.rs
```diff
@@ -0,0 +1,187 @@
...
+        let bytes = to_contiguous(&value);
+        let Some((version, payload)) = bytes.split_first() else {
+            return Ok(DecodeOutcome::SoftFailure("empty payload"));
+        };
+
+        if *version != FORMAT_VERSION {
+            return Ok(DecodeOutcome::SoftFailure("format version mismatch"));
+        }
+
+        match postcard::from_bytes(payload) {
```
Vaiz, line 88: I wonder if 1 byte shift may affect performance of accessing this data because now access to that memory is not aligned

crates/cachet/src/transform/codec.rs
```diff
@@ -39,15 +56,121 @@ pub trait Encoder<From, To>: Send + Sync {
...
+}
+
+impl<A, B> TransformEncoder<A, B> {
+    /// Creates a new `TransformEncoder` from a fallible closure.
+    pub fn new<EncodeError>(encode_fn: impl Fn(&A) -> Result<B, EncodeError> + Send + Sync + 'static) -> Self
+    where
+        EncodeError: std::error::Error + Send + Sync + 'static,
+    {
+        Self {
+            encode_fn: Box::new(move |a| encode_fn(a).map_err(|e| Error::from_source(e))),
```
Vaiz: NIT: `map_err(Error::from_source)`

crates/cachet/src/transform/testing.rs
```diff
@@ -0,0 +1,107 @@
...
+/// ```
+/// use cachet::{MockCodec, DecodeOutcome};
+///
+/// // A codec that always succeeds
+/// let codec = MockCodec::<i32>::value();
+///
+/// // A codec that always returns a soft failure
+/// let codec = MockCodec::<i32>::soft_failure("version mismatch");
+/// ```
+pub struct MockCodec<T> {
```
Vaiz, line 28: is there much value in making this type publicly available?
I can see how it can lead to multiple breaking release just because we slightly tweaked its behavior

geeknoid (approved):

crates/cachet/src/serialize/codec.rs
schgoo: Yeah I think what I need to do is just init the GlobalPool per cache and pass it in here

crates/cachet/src/serialize/codec.rs
schgoo: It shouldn't be necessary, actually, GlobalPoll already uses thread_aware under the covers, I shouldn't need thread_local if I just have one per cache instead of this.

crates/cachet/src/serialize/codec.rs
schgoo: This shouldn't be an issue. `from_bytes` processes bytes sequentially. Since &[u8] has an alignment of 1, the 1-byte offset should not have a performance impact.

crates/cachet/src/transform/testing.rs
schgoo: I could be convinced not to expose it; it's unlikely any I/O would be happening in one of these implementations

crates/cachet/examples/serialization.rs
```diff
@@ -34,9 +34,6 @@ async fn main() {
         .expect("insert failed");
 
     // Get returns a typed value — deserialized from BytesView if fetched from L2.
-    let value = cache.get(&key).await.expect("get failed");
-    match value {
-        Some(e) => println!("get({key}): {}", e.value()),
-        None => println!("get({key}): not found"),
-    }
+    let value = cache.get(&key).await.expect("get failed").unwrap();
+    println!("get({key}): {:?}", value);
```
Vaiz: NIT: we should avoid using debug formatting, especially in examples
```suggestion
    println!("get({key}): {value}");
```

Vaiz (approved):

Vaiz (approved):

Vaiz (approved):

PR #382, opened 2026-04-17
feat(http_extensions): fine grained error labels

martintmk:
Give more insights into the telemtry.

### DIFF
--- a/crates/http_extensions/src/error.rs
+++ b/crates/http_extensions/src/error.rs
@@ -14,12 +14,12 @@ use recoverable::{Recovery, RecoveryInfo};
 use thread_aware::ThreadAware;
 use thread_aware::affinity::{MemoryAffinity, PinnedAffinity};
 
+use crate::HttpRequest;
 use crate::error_labels::{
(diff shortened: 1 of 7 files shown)


crates/http_extensions/src/error.rs
```diff
@@ -258,6 +257,29 @@ impl HttpError {
...
         self.request.take().map(|holder| *holder.into_inner())
     }
+
+    /// Resolves the error label for pre-defined set of errors.
+    ///
+    /// This method recognizes the following error types:
+    /// - [`HttpError`]
+    /// - [`JsonError`][crate::json::JsonError]
+    /// - [`std::io::Error`]
+    pub fn resolve_error_label(error: &(dyn std::error::Error + 'static)) -> Option<ErrorLabel> {
```
Vaiz: I don't think it belongs to `HttpError` type as it tries to cast to other error types
maybe it could be a standalone function

it's logic also looks pretty random and covers very niche case. it's not clear why would someone use this instead of writing a similar one by hand. Even worse, it could be used in the code not related to HTTP in anyway

martintmk: fair point, hidden from API

Vaiz (approved):

sgalkin (approved):

PR #385, opened 2026-04-20
feat(ohno)!: Labeled::label() returns ErrorLabel by value

martintmk:
Change the return type of Labeled::label() from &ErrorLabel to ErrorLabel, allowing callers to take ownership without a separate clone.

Update all Labeled implementations in templated_uri and http_extensions. Bump versions of all transitively affected crates:
- ohno 0.3.2 -> 0.4.0
- templated_uri 0.1.2 -> 0.2.0
- http_extensions 0.3.1 -> 0.4.0
- cachet_tier 0.1.0 -> 0.2.0
- cachet_memory 0.1.0 -> 0.2.0
- cachet_service 0.1.0 -> 0.2.0
- cachet 0.1.0 -> 0.2.0
- bytesbuf_io 0.4.0 -> 0.5.0


crates/ohno/src/error_label.rs
```diff
@@ -10,7 +10,7 @@ use std::iter::successors;
 /// Trait for errors that carry an [`ErrorLabel`].
 pub trait Labeled {
     /// Returns the label attached to this error.
-    fn label(&self) -> &ErrorLabel;
+    fn label(&self) -> ErrorLabel;
```
martintmk, line 13: Owned error labels now allocate on every lookup

Changing `Labeled::label(&self)` from a borrowed label to an owned `ErrorLabel` makes implementations clone their stored label on every lookup. Static labels remain allocation-free, but composite labels created by `from_parts` or `from_error_chain` use an owned string, so metrics/logging lookups allocate repeatedly. Consider preserving a borrowed accessor for stored labels and exposing a separate owned/computed-label path.

martintmk (commented):
One non-blocking performance concern: owned error labels may allocate on repeated lookups. No vote because this is my own PR.

PR #389, opened 2026-04-23
feat(http_extensions): introduce a new "routing" module 

martintmk:
## Add `routing` module for resolving request `BaseUri`

Adds a `routing` module to `http_extensions` that centralizes resolution of the `BaseUri` (scheme + authority) for outgoing requests, while callers continue to specify path/query independently.

### API

- `Router` with constructors: `default`, `fixed`, `fallback` (primary/secondary based on previous attempt's `RecoveryInfo`), and `custom` (closure over `RouterContext`).
- `BaseUriConflict` policy (via `Router::conflict_policy`, default `UseOriginal`) for when both the target `Uri` and the router supply a `BaseUri`.
- `RouterContext` carrying per-attempt state (attempt index, last-attempt flag, prior `RecoveryInfo`) supplied by the caller's resilience layer.

### Example

Multi-region failover — swap to the secondary region when the primary returns `RecoveryKind::Unavailable`:

```rust
let router = Router::fallback(
    BaseUri::from_static("https://api-westus.example.com"),
    BaseUri::from_static("https://api-eastus.example.com"),
);

// Per attempt, the resilience layer builds a context and resolves the URI.
let ctx = RouterContext::new()
    .with_attempt(1)
    .with_previous_recovery(prev_recovery_info);
let uri = router.resolve_request_uri(&ctx, request)?;
```


crates/http_extensions/src/body/builder.rs
```diff
@@ -336,6 +336,12 @@ impl Memory for MemoryWrapper {
     }
 }
 
+impl AsRef<Clock> for HttpBodyBuilder {
```
Vaiz, line 339: when is it helpful?
it's a bit weird that builder type becomes a struct that you can pass around as a clock

crates/http_extensions/src/routing/routing.rs
```diff
@@ -0,0 +1,509 @@
...
+    /// Returns [`HttpError::validation`] when the policy is [`BaseUriConflict::Fail`] and
+    /// either:
+    /// - the target [`Uri`] already has a [`BaseUri`] and the routing also produces one; or
+    /// - the target [`Uri`] has neither a [`BaseUri`] nor a path.
+    #[expect(
+        clippy::needless_pass_by_value,
+        reason = "while not consuming the context, we might do it at some point"
+    )]
+    pub fn create_uri(&self, ctx: RoutingContext, uri: Uri) -> Result<Uri, HttpError> {
+        let routed = self.resolve(&ctx);
```
Vaiz: this can be done after uri validation

crates/http_extensions/src/routing/router.rs
```diff
@@ -0,0 +1,509 @@
...
+        if existing.is_none() && path.is_none() && self.conflict_policy == BaseUriConflict::Fail {
+            return Err(HttpError::validation_with_label(
+                "the target uri cannot be empty; provide a base uri or a path such as `/...`",
+                LABEL_URI_MISSING,
+            ));
+        }
+
+        // if new base uri is not available, return existing uri
+        let Some(routed) = routed else {
+            return Ok(Uri::from_parts(existing, path));
```
Vaiz, line 185: I wonder if we can avoid deconstruction of initial uri, and return it here instead of creating a new instance

crates/http_extensions/src/routing/routing.rs
```diff
@@ -0,0 +1,509 @@
...
+        *request.uri_mut() = resolved.try_into()?;
+        Ok(())
+    }
+
+    /// Resolves the [`BaseUri`] for the current request, if any.
+    fn resolve(&self, ctx: &RoutingContext) -> Option<BaseUri> {
+        match self.resolver.as_ref() {
+            Resolver::Empty => None,
+            Resolver::Fixed(base_uri) => Some(base_uri.clone()),
+            Resolver::Custom(f) => f(ctx),
```
Vaiz: I would try to avoid calling an unknown function if we already now that there is a conflict with KeepExisting/Fail policy

Vaiz (approved):

crates/http_extensions/src/body/builder.rs
martintmk: I have a type that owns the `HttpBodyBuilder` but also require `Clock`. Rather than duplicating and owning the `Clock` I can just retrieve it from the builder.

And FYI, the `HttpBodyBuilder` should be seen as a service that crates a `HttpBody`, so it's not a typical builder that's only used once.

crates/http_extensions/src/routing/router.rs
martintmk: The deconstruction is very cheap anyway, I wouldn't keep it as it is now.

crates/http_extensions/src/routing/routing.rs
martintmk: true

crates/http_extensions/src/routing/routing.rs
martintmk: I tried to apply, but the conflict can be only evaluated after new base URI is produced. We need to call this.

crates/http_extensions/src/body/builder.rs
Vaiz: can we keep Clock private by exposing it via pub(crate) method?

crates/http_extensions/src/routing/routing.rs
```diff
@@ -0,0 +1,510 @@
...
+    ///
+    /// Otherwise, the primary [`BaseUri`] is used. This is intended for
+    /// scenarios where the primary endpoint becomes unavailable but requests
+    /// can still be served by a fallback endpoint.
+    ///
+    /// [`RecoveryInfo`]: recoverable::RecoveryInfo
+    /// [`RecoveryKind::Unavailable`]: recoverable::RecoveryKind::Unavailable
+    #[must_use]
+    pub fn fallback(primary: BaseUri, fallback: BaseUri) -> Self {
+        Self::custom(move |ctx| Some(if use_fallback(ctx) { fallback.clone() } else { primary.clone() }))
```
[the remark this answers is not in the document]
martintmk: indeed, default conflict policy for fallback is changed

crates/http_extensions/src/body/builder.rs
martintmk: not really, as this is accessed in `fetch` so internal APIs wont be visible.

crates/http_extensions/src/routing/routing.rs
```diff
@@ -0,0 +1,540 @@
...
+    ///
+    /// - the request's existing URI cannot be converted to a [`Uri`],
+    /// - [`Routing::create_uri`] fails (e.g., a [`BaseUriConflict::Fail`] conflict), or
+    /// - the resolved [`Uri`] cannot be converted back to an [`http::Uri`].
+    pub fn update_request_uri(&self, ctx: RoutingContext, request: &mut HttpRequest) -> Result<(), HttpError> {
+        // Clone the URI rather than taking it: if any step below fails we leave the
+        // request's URI untouched.
+        let uri: Uri = request.uri().clone().try_into()?;
+        let resolved = self.create_uri(ctx.with_request(request), uri)?;
+        *request.uri_mut() = resolved.try_into()?;
```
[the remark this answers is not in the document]
martintmk: true, we try to avoid this by attaching original structured Uri to the request

crates/http_extensions/examples/custom_client.rs
```diff
@@ -4,22 +4,27 @@
 //! An example of a custom HTTP client using `http_extensions`.
```
sandersaares, line 4: The inclusion of this routing stuff into a "custom client" example seems excessive overloading of a single example, making it hard to follow - examples should be focused on a single thing. It might be valuable to create a separate example for the routing feature.

crates/http_extensions/src/routing/mod.rs
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Routing primitives for resolving the [`BaseUri`] of an outgoing request.
+//!
+//! A [`Routing`] decides which [`BaseUri`] should be attached to a target [`Uri`]
```
sandersaares: "A [`Routing`] decides" does not English for me. I suggest `Router` as the name if this is the way we are meant to think about the type. A thing that decides routing is a router.

crates/http_extensions/src/routing/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+//! before it is sent. This is useful when a library or middleware needs to centralize
```
sandersaares: This explanation is not understandable without additional context. Can you introduce some real-world example where this might be desirable? What would be the different BaseUris and why would one want to use multiple of them, and based on what might one want to vary the choices?

crates/http_extensions/src/routing/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+//! the resolution of the destination of HTTP requests while still allowing callers to
+//! express the rest of the request (path, query, ...) independently.
+//!
+//! # Construction
+//!
+//! - [`Routing::default`] - returns no [`BaseUri`] (the target [`Uri`] is used as-is).
```
sandersaares: Terminology confusion: this section talks about "returning" things, which is a property of functions. But this is a struct, not a function. Structs do not return things.

The summary at the top talks about "resolving" which might be a more appropriate term.

crates/http_extensions/src/routing/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+//!   receives a [`RoutingContext`].
+//!
+//! # Conflict resolution
+//!
+//! When the target [`Uri`] passed to [`Routing::create_uri`] already carries a
+//! [`BaseUri`] and the routing also produces one, [`Routing`] uses the configured
+//! [`BaseUriConflict`] policy to decide what to do. The policy can be set with
+//! [`Routing::conflict_policy`] and defaults to [`BaseUriConflict::KeepExisting`].
+
+#[expect(unused_imports, reason = "simplifies the docs")]
```
sandersaares: Feels a strange way to do docs hyperlinking. I suppose it works but I would just do it entirely in the docs domain as:

```rust
//! [`BaseUri`]: template_uri::BaseUri
```

Granted, the `use` way might be better generalizable to repeated doc blocks, so I can see the appeal.

crates/http_extensions/src/routing/routing.rs
```diff
@@ -0,0 +1,596 @@
...
+    /// Keep the [`BaseUri`] already present on the target [`Uri`] (the default).
+    #[default]
+    KeepExisting,
+
+    /// Return an error when the target [`Uri`] already has a [`BaseUri`] and the
+    /// routing also produces a conflicting [`BaseUri`].
+    Fail,
+
+    /// Replace the target's [`BaseUri`] with the one produced by the routing.
+    Override,
```
sandersaares: Ambiguous terminology - which one overrides the other? Unclear from name. I suggest `KeepTarget` or something to that effect.

crates/http_extensions/src/routing/routing.rs
```diff
@@ -0,0 +1,596 @@
...
+use crate::error_labels::{LABEL_URI_CONFLICT, LABEL_URI_MISSING};
+use crate::{HttpError, HttpRequest};
+
+/// Strategy used by [`Routing::create_uri`] when both the target [`Uri`] and the
+/// routing produce a [`BaseUri`].
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+pub enum BaseUriConflict {
+    /// Keep the [`BaseUri`] already present on the target [`Uri`] (the default).
+    #[default]
+    KeepExisting,
```
sandersaares: Ambiguous terminology - the whole point of a conflict is that there are two "existing" base URIs. Which one of the existing ones does it keep? Unclear from name. I suggest `KeepRouting` or something to that effect.

crates/http_extensions/src/routing/routing.rs
```diff
@@ -0,0 +1,596 @@
...
+/// );
+/// ```
+///
+/// Pick a [`BaseUri`] dynamically:
+///
+/// ```
+/// use http_extensions::routing::{Routing, RoutingContext};
+/// use templated_uri::{BaseUri, Uri};
+///
+/// let routing = Routing::custom(|_ctx| Some(BaseUri::from_static("https://api.example.com")));
```
sandersaares: `ctx` is an uncomfortable shorterning of "context" as it lacks keyboard and linguistic flow. I suggest `cx` which is the common style in Rust, or just `context`.

crates/http_extensions/src/routing/routing.rs
```diff
@@ -0,0 +1,596 @@
...
+#[derive(Debug, Clone, Default)]
+pub struct Routing {
+    resolver: Arc<Resolver>,
+    conflict_policy: BaseUriConflict,
+}
+
+impl Routing {
+    /// Creates a [`Routing`] that always returns the given [`BaseUri`].
+    #[must_use]
+    pub fn base_uri(base_uri: BaseUri) -> Self {
```
sandersaares: Naming feels a bit odd as it seems to describe the data type of the input instead of the logical handling. I suggest `Routing::constant()` or something like that.

crates/http_extensions/src/routing/router.rs
```diff
@@ -0,0 +1,596 @@
...
+    }
+
+    /// Creates a [`Routing`] that selects between a primary and a fallback [`BaseUri`]
+    /// based on the previous attempt's [`RecoveryInfo`] and the current attempt's
+    /// position in the retry sequence.
+    ///
+    /// The first attempt always uses the primary [`BaseUri`]. On subsequent
+    /// attempts, the fallback [`BaseUri`] is used when either:
+    ///
+    /// - the previous attempt's [`RecoveryInfo`] reports
```
sandersaares: The API documentation here is only understandable if the reader knows what a `RecoveryInfo` is and how it relates the the HTTP request process. Maybe some more context could be provided to help paint a fuller picture even for readers that do not know what that is? Specifically, how does a `RecoveryInfo` even enter the routing workflow, where does it come from? Does the caller need to provide it somewhere? What are the caller's responsibilities in this domain?

crates/http_extensions/src/routing/routing.rs
```diff
@@ -0,0 +1,596 @@
...
+
+    /// Returns `true` when this [`Routing`] may resolve to more than one
+    /// [`BaseUri`] across attempts.
+    ///
+    /// Resilience layers can use this to decide whether retrying a request that
+    /// previously failed with an unavailable endpoint is worthwhile: if
+    /// alternatives exist, a subsequent attempt may be routed to a different
+    /// endpoint and succeed.
+    ///
+    /// Returns `true` for [`Routing::fallback`] and [`Routing::custom`] (which
```
sandersaares: Might be worth documenting here (even with internal comments) why we expect the assumption of `Custom` having alternatives to be a safe and sensible thing to do, as it is not strictly guaranteed. Is there anything that can go wrong with assuming this or is it generally safe, and if so why?

crates/http_extensions/src/routing/routing.rs
```diff
@@ -0,0 +1,596 @@
...
+    ///
+    /// Returns [`HttpError::validation`] when the policy is [`BaseUriConflict::Fail`] and
+    /// either:
+    /// - the target [`Uri`] already has a [`BaseUri`] and the routing also produces one; or
+    /// - the target [`Uri`] has neither a [`BaseUri`] nor a path.
+    #[expect(
+        clippy::needless_pass_by_value,
+        reason = "while not consuming the context, we might do it at some point"
+    )]
+    pub fn create_uri(&self, ctx: RoutingContext, uri: Uri) -> Result<Uri, HttpError> {
```
sandersaares: I have seen three different terms used for this operation now: resolve, return, create. Let's be consistent, please.

crates/http_extensions/src/routing/router.rs
```diff
@@ -0,0 +1,596 @@
...
+    fn resolve(&self, ctx: &RoutingContext) -> Option<BaseUri> {
+        match self.resolver.as_ref() {
+            Resolver::Empty => None,
+            Resolver::Fixed(base_uri) => Some(base_uri.clone()),
+            Resolver::Custom(f) => f(ctx),
+        }
+    }
+}
+
+// --- Private items below ---
```
sandersaares: Seems like a fairly nonstandard decorator - I would prefer not to have such things, especially as they become out of date very fast (nobody bothers looking at these, really).

crates/http_extensions/src/routing/router_context.rs
```diff
@@ -0,0 +1,139 @@
...
+/// Context passed to the closure of [`Routing::custom`] when resolving a [`BaseUri`].
+///
+/// This type is intentionally opaque so that fields can be added in the future without
+/// breaking the closure signature. Use the getters to inspect the available
+/// information about the current request attempt.
+///
+/// [`Routing::custom`]: super::Routing::custom
+/// [`BaseUri`]: templated_uri::BaseUri
+#[derive(Debug, Clone)]
+#[non_exhaustive]
```
sandersaares: This attribute does not do anything as the fields are not public.

sandersaares (commented):
Design looks sensible. I left some minor UX/maintainability/readability comments.

sandersaares:
The PR description is not understandable to me - perhaps it needs an update. The two first "alternatives" also seem to have equal code, so I do not think the code is even right.

crates/http_extensions/src/routing/routing.rs
martintmk: Renamed to `UseOriginal`, `UseRouted`, `Fail`

crates/http_extensions/examples/custom_client.rs
martintmk: reverted, we can keep the example simple

crates/http_extensions/src/routing/mod.rs
martintmk: +1

crates/http_extensions/src/routing/routing.rs
martintmk: used the term "resolve*"

crates/http_extensions/src/routing/routing.rs
martintmk: used `fixed`

crates/http_extensions/src/routing/mod.rs
martintmk: done

crates/http_extensions/src/routing/routing.rs
martintmk: made this explicit, caller must say when creating custom router whether it can provide alternative URLs

crates/http_extensions/src/routing/mod.rs
martintmk: updated

crates/http_extensions/src/routing/router.rs
martintmk: refined and added more information in "Recovery Context" in the routing module

tekian (approved):

crates/http_extensions/src/http_request_builder.rs
```diff
@@ -352,8 +352,15 @@ impl<R> HttpRequestBuilder<'_, R> {
...
+        // Attach both the templated `Uri` and its `PathAndQuery`:
+        //   - `Uri` preserves the caller's unrouted target so
+        //     `Router::resolve_request_uri` can re-route from it on every retry.
+        //   - `PathAndQuery` backs `RequestExt::path_and_query` and
+        //     `ExtensionsExt::uri_template_label`.
         let path = uri.to_path_and_query();
-        let mut request = self.builder.uri(http::Uri::try_from(uri)?).body(body)?;
+        let http_uri = http::Uri::try_from(uri.clone())?;
+        let mut request = self.builder.uri(http_uri).body(body)?;
+        request.extensions_mut().insert(uri);
```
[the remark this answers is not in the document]
martintmk: caller inserting Uri into extensions is not something we expect to be done

sandersaares (approved):

PR #391, opened 2026-04-23
feat(templated_uri)!: API review and overall cleanup

martintmk:
Broad API refinement of the `templated_uri` crate. Many breaking changes — review call sites against the updated surface.
 
 ### ⚠️ Breaking changes
 
 - **Type renames** — dropped the `Uri` prefix throughout:
   - `UriPath` → `PathAndQuery`
   - `UriTemplate` → `PathAndQueryTemplate`
   - `UriSafe*` → `Escaped*`
   - `UriParam` → `Escape`, `UriUnsafeParam` → `Raw`
   - `ValidationError` → `UriError`
   - `Uri::to_http_path()` → `Uri::to_path_and_query()`
 - **`Origin` / `BaseUri`** now accept any URI scheme (not just HTTP/HTTPS).
   - `Origin::port()` returns `Option<u16>`.
   - `Origin::try_from_parts` replaced by infallible `Origin::from_parts`.
 - **`PathAndQuery` redaction** — now mirrors `Uri`: implements `RedactedDisplay` / `RedactedDebug`, and `to_string()` returns `Sensitive<String>`.
 - **`PathAndQueryTemplate` trait refined** — added `render()`; removed `Display`, `to_uri_string()`, `into_uri()`; swapped meanings of `template()` and 
`format_template()`.
 - **Re-exports** — removed the `http` re-export module. `Authority` and `Scheme` are now re-exported at the crate root; `Parts` is no longer re-exported. Use 
`http::uri::PathAndQuery` directly when needed.
 
 ### 📚 Documentation
 
 - New `_documentation::recipes` module with usage recipes.
 - Polished examples (`uri_templating`, `classified_templating`) and README.
 
 ### ✅ Tests
 
 - Added coverage for renamed and new APIs; updated UI/compile-fail tests.


crates/templated_uri/src/escaped.rs
Vaiz, line 1: since we do such broad renames, maybe we rename `Safe` into `Valid`? My association with Safe that this uri is safe to visit, not that it uses right symbols.

My suggestions: `ValidUri` or `ValidatedUri`

though, a better option is to rename `UriSafe` into `Uri`, and `Uri` into `UriString` or something

martintmk: True, I'll apply this.

kate-shine: Not gonna lie, naming this was an ongoing nightmare. I don't really like ValidUri either, because for example `foo/bar` is a totally valid Uri part, but uri unsafe for our templating purpose.

The distinction is "Allows reserved characters" and "Doesn't allow reserved characters". "Safe" was chosen because allowing reserved characters makes it safe from injection and accidental use of character that breaks the Uri.

Although, after latest changes that introduced automatic escaping, "UriEscaped" might be a good choice as well.

martintmk: The term "Escaped" wins for me. It's pretty clear what it does. Will make the change.

Safe < Valid < Escaped

crates/templated_uri/src/uri_escaped.rs
```diff
@@ -0,0 +1,492 @@
...
+/// (`escape`, `try_new`, `from_static`) to create instances.
+pub type UriEscapedString = UriEscaped<Cow<'static, str>>;
+
+/// Error returned when a string is not a valid URI-escaped string.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct UriEscapeError(&'static str);
+
+impl fmt::Display for UriEscapeError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(self.0)
```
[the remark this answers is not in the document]
martintmk: it's intentional, we can add it later if necessary

crates/templated_uri/src/uri.rs
```diff
@@ -62,89 +57,101 @@ impl Default for Uri {
...
-    ///
-    /// # Errors
+    /// Returns the path and query as a [`PathAndQuery`] if present.
     ///
-    /// Returns a [`ValidationError`] if the path and query cannot be validated.
-    pub fn to_path_and_query(&self) -> Result<Option<PathAndQuery>, ValidationError> {
-        self.path_and_query.as_ref().map(TargetPathAndQuery::to_path_and_query).transpose()
+    /// Conversion errors are suppressed and returned as `None`. In practice
+    /// this should never happen: a templated path that fails to materialize
+    /// into a valid `PathAndQuery` indicates a programming error in the
```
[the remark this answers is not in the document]
martintmk: it's intended to make APIs simpler, practically the conversion will almost never fail

crates/templated_uri/src/escaped.rs
martintmk: pushed polished changes, please take a look

crates/templated_uri/src/escape.rs
```diff
@@ -0,0 +1,162 @@
...
+use crate::{Escaped, EscapedString};
+
+/// Marks types usable from templates, e.g., `/get/{foo}`.
+pub trait Escape {
+    /// Returns this value wrapped in [`Escaped`], proving it is properly escaped for URI use.
+    fn escape(&self) -> Escaped<impl Display>;
+}
+
+/// Marks types with possibly dodgy content usable from templates, e.g., `/get/{+foo}`.
+pub trait UnescapedDisplay {
```
kate-shine: The 'Display' part feels wrong. Can't it be Param as before? Or just `Unescaped` or even better
`UntreatedParam`/ `Untreated`, as it ultimately should still contain escaped data, but user should handle that
manually (Which allows use of aforementioned Reserved characters).

"Display" here is the internal implementation, not the trait purpose. (Of course, people know that rust
template macros use Display internally, but here it also just doesn't pair well with 'Escaped' - shouldn't it be
`EscapedDisplay` then?)

crates/templated_uri/src/path.rs
```diff
@@ -0,0 +1,223 @@
...
+#[derive(Clone)]
+pub struct Path(PathInner);
+
+#[derive(Clone)]
+enum PathInner {
+    Static(Sensitive<PathAndQuery>),
+    Templated(Arc<dyn PathTemplate>),
+}
+
+impl Path {
```
kate-shine: This naming might be confusing - in Uri, Path doesn't include the Query part of Uri, but we do. This might create wrong expectations about its content, but also cause issues in future, if we ever want to implement methods to extract respective parts from it.

I suggest `PathAndQuery` similarly to http

crates/http_extensions/src/extensions/request_ext.rs
```diff
@@ -1,34 +1,34 @@
...
 use http::Request;
-use templated_uri::uri::TargetPathAndQuery;
+use templated_uri::Path;
 
 /// Extensions for HTTP requests.
 pub trait RequestExt: sealed::Sealed {
-    /// Returns the path and query associated with this request, if any.
-    fn path_and_query(&self) -> Option<&TargetPathAndQuery>;
+    /// Returns the URI path associated with this request, if any.
+    fn path(&self) -> Option<&Path>;
```
kate-shine: Same issue as already mentioned, `URI path` is a specific term, and explicitly doesn't contain optional query.

crates/templated_uri/src/escape.rs
Vaiz: `Unescaped` sounds weird, how about human readable? that's what unescaped essentially means, right?

crates/templated_uri/src/origin.rs
```diff
@@ -0,0 +1,419 @@
...
+    /// the well-known HTTP and HTTPS schemes the default port is inferred from
+    /// the scheme (`80` for `http`, `443` for `https`). For all other schemes
+    /// without an explicit port this method returns `None`.
+    #[must_use]
+    pub fn port(&self) -> Option<u16> {
+        if let Some(port) = self.authority.port_u16() {
+            return Some(port);
+        }
+
+        if self.scheme == Scheme::HTTP {
```
Vaiz: how we decide which schemes are "well-known"? that's definitely not the full list

maybe it's better to return None if port is not explicitly assigned for better consistency

crates/templated_uri/src/origin.rs
```diff
@@ -0,0 +1,419 @@
...
+    /// assert_eq!(format!("{}", origin), "https://example.com");
+    ///
+    /// let custom_port = Origin::from_static("https://example.com:8443");
+    /// assert_eq!(format!("{}", custom_port), "https://example.com:8443");
+    /// ```
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}://", self.scheme)?;
+
+        match (self.scheme.as_str(), self.authority.port_u16()) {
+            (s, Some(HTTP_DEFAULT_PORT)) if s == Scheme::HTTP.as_str() => write!(f, "{}", self.authority.host()),
```
Vaiz, line 194: I would expect the port to be present if it was explicitly assigned even if it's a default one

crates/templated_uri/src/path_and_query.rs
```diff
@@ -0,0 +1,223 @@
...
+
+impl RedactedDisplay for PathAndQuery {
+    #[cfg_attr(test, mutants::skip)] // Do not mutate display output.
+    fn fmt(&self, engine: &RedactionEngine, f: &mut Formatter<'_>) -> fmt::Result {
+        match &self.0 {
+            PathAndQueryInner::Static(classified_pq) => {
+                // We can't use to_string in redaction because it automatically prepends a slash if the path doesn't start with one.
+                // as_str doesn't do that, so we declassify to get the inner PathAndQuery and then use as_str.
+                let reclassified = Sensitive::new(classified_pq.declassify_ref().as_str(), classified_pq.data_class().clone());
+                f.write_str(&engine.redacted_to_string(&reclassified))
```
Vaiz:
```suggestion
          RedactedDisplay::fmt(&**reclassified, engine, f)
```

crates/templated_uri/src/escape.rs
martintmk: I have also discussed with Ralf, and changed this to `RawDisplay`, would that work with you?

martintmk: Renamed to RawDisplay and update the docs so the intent is more clear.

crates/templated_uri/src/origin.rs
martintmk: Just based on our common scenarios, where HTTP and HTTPS is the major use-case.

crates/templated_uri/src/origin.rs
martintmk: this is for display purposes, where we don't really want to show default ports for HTTP/HTTPS schemes (breaking change)

crates/templated_uri/src/origin.rs
Vaiz: why not return `None`?

martintmk: In SDK we used this inference at multiple places, O I thought I should preserve it. But we can drop it as well, and just add that inference on other places. (to be more explicit)

crates/templated_uri/src/escape.rs
martintmk: discussed offline, renamed to just `Raw`

Vaiz (approved):

kate-shine (approved):

PR #397, opened 2026-04-29
refactor(cachet): Rename FallbackPromotionPolicy to InsertPolicy and move it to CacheWrapper instead of on the FallbackCache

schgoo:
The `FallbackPromotionPolicy` could not adequately serve the purpose of allowing patterns like negative caching. Usage of `get_or_insert` would insert into all levels of the cache regardless of the policy set on a fallback, since the policy was only checked on a get hit. This change moves policy checks to the `CacheWrapper` implementation of `insert`, so that it is always applied to each tier individually. It also renames it to `InsertPolicy` for clarity.


crates/cachet/src/lib.rs
```diff
@@ -255,7 +254,11 @@ pub use cachet_tier::{CacheEntry, CacheTier, Error, Result, SizeError};
 #[doc(inline)]
 pub use cachet_tier::{CacheOp, MockCache};
 #[doc(inline)]
-pub use fallback::FallbackPromotionPolicy;
+pub use policy::InsertPolicy;
+
+/// Deprecated: Use [`InsertPolicy`] instead.
+#[deprecated(since = "0.2.0", note = "renamed to InsertPolicy")]
+pub type FallbackPromotionPolicy<V> = InsertPolicy<V>;
```
Vaiz, line 261: as I understand, your PR already contains breaking changes, maybe we don't need to keep this alias?

Vaiz (approved):

geeknoid (approved):

crates/cachet/src/lib.rs
schgoo: Yeah that's probably fair. I'll remove it when I actually bump the version.

PR #398, opened 2026-04-30
ci: switch wild linker to wild-linker/action@0.8.0

Vaiz:
The action moved from davidlattimore/wild-action to the wild-linker org (see https://github.com/wild-linker/action). Pin to 0.8.0.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -113,4 +113,4 @@ runs:
 
     - name: Install Wild Linker
       if: inputs.enable-wild-linker == 'true'
-      uses: davidlattimore/wild-action@0.7.0
+      uses: wild-linker/action@0.8.0


.github/actions/setup/action.yml
```diff
@@ -113,4 +113,4 @@ runs:
 
     - name: Install Wild Linker
       if: inputs.enable-wild-linker == 'true'
-      uses: davidlattimore/wild-action@0.7.0
+      uses: wild-linker/action@0.8.0
```
psandana, line 116: curious about this change. Is it acknowledged this new action repo?

psandana (approved):

.github/actions/setup/action.yml
Vaiz: yes, if you go https://github.com/davidlattimore/wild-action it will redirect you to the new repo with the same maintainer

geeknoid (approved):

PR #407, opened 2026-05-07
chore: enforce nightly formatting

Vaiz:


martin-kolinek (approved):

ralfbiedert (approved):

geeknoid:
I don't think this is a good idea. It means that it would require devs to have nightly bits installed in order to be able to reliably make changes for this repo.

Vaiz:
> I don't think this is a good idea. It means that it would require devs to have nightly bits installed in order to be able to reliably make changes for this repo.

we need to enforce this if we want to avoid PRs with hundreds changed files just because developer applied nightly formatting on the whole repo

ralfbiedert:
@copilot resolve the merge conflicts in this pull request

Vaiz:
@copilot resolve merge conflicts in this pull request

martintmk (approved):

PR #409, opened 2026-05-08
fix(cachet): Fix broken seatbelt dev dependency path

schgoo:
The release pipeline doesn't sort release priority based on dev-deps, but it still checks for those packages before releasing. When Seatbelt was updated to 0.4.5 in #387, it tried to check for the 0.4.5 release before actually releasing it, and got stuck. This PR changes it to a path ref to avoid this issue.

### DIFF
--- a/crates/cachet/Cargo.toml
+++ b/crates/cachet/Cargo.toml
@@ -68,7 +68,7 @@ opentelemetry = { workspace = true, features = [
 ] }
 opentelemetry_sdk = { workspace = true, features = ["logs", "testing"] }
 recoverable = { workspace = true }
-seatbelt = { workspace = true, features = ["retry", "tower-service"] }
+seatbelt = { path = "../seatbelt", features = ["retry", "tower-service"] }
 testing_aids = { path = "../testing_aids" }
 tick = { workspace = true, features = ["test-util", "tokio"] }
 tokio = { workspace = true, features = ["rt"] }


crates/cachet/Cargo.toml
```diff
@@ -4,7 +4,7 @@
 [package]
 name = "cachet"
 description = "A composable, customizable multi-tier caching library with rich feature support."
-version = "0.2.0"
+version = "0.2.1"
```
psandana: No need to bump as 0.2.0 was not released

crates/cachet/Cargo.toml
```diff
@@ -68,7 +68,7 @@ opentelemetry = { workspace = true, features = [
 ] }
 opentelemetry_sdk = { workspace = true, features = ["logs", "testing"] }
 recoverable = { workspace = true }
-seatbelt = { workspace = true, features = ["retry", "tower-service"] }
+seatbelt = { path = "../seatbelt", features = ["retry", "tower-service"] }
```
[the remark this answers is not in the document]
psandana: you can set also default-features = false if reeaally needed. I don't really think this would affect.

crates/cachet/CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Changelog
 
+## [0.2.1] - 2026-05-08
```
psandana: No new release as 0.2.0 has not been released

Cargo.toml
```diff
@@ -26,7 +26,7 @@ homepage = "https://github.com/microsoft/oxidizer"
 anyspawn = { path = "crates/anyspawn", default-features = false, version = "0.4.0" }
 bytesbuf = { path = "crates/bytesbuf", default-features = false, version = "0.4.2" }
 bytesbuf_io = { path = "crates/bytesbuf_io", default-features = false, version = "0.4.0" }
-cachet = { path = "crates/cachet", default-features = false, version = "0.2.0" }
+cachet = { path = "crates/cachet", default-features = false, version = "0.2.1" }
```
psandana: no new version

psandana (changes requested):

psandana (approved):

crates/cachet/Cargo.toml
ralfbiedert, line 71: This feels like a hack working around a bug, or am I missing something?

AdomasBekeras (approved):

crates/cachet/Cargo.toml
schgoo: The bug is in the release pipeline. It checks for the updated release of seatbelt, before it can actually release it, and then fails.

PR #412, opened 2026-05-11
chore(templated_uri)!: release a new major version

martintmk:


crates/templated_uri/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## [0.2.0] - 2026-05-11
+
+- ⚠️ Breaking
+
+  - API review and overall cleanup ([#391](https://github.com/microsoft/oxidizer/pull/391))
```
sgalkin: One thing is missing on API - I can't find a way to self assert Uri safety for my type without going through re-encoding (and allocation)

sgalkin (approved):

crates/templated_uri/CHANGELOG.md
kate-shine: If your underlying type is String that doesn't require escaping, you can implement `Encode` and return `EncodedString::try_new()` - it shouldn't reallocate.

But, it's true that we should also implement a wrapper over `Cow<T>` that won't have .encode() method, just a dynamic and static validating constructors.

kate-shine (approved):

crates/templated_uri/CHANGELOG.md
martintmk: I believe this already works such as:

```rust
    let pre_constructed = "some-value".to_string();
    let escaped = EscapedString::try_new(pre_constructed).unwrap();
```

So if you pass owned string, the `EscapedString` won't re-allocated again.

kate-shine:
I hope I will be able to finish https://github.com/microsoft/oxidizer/pull/408 today, if it can wait - but it's fine if not, it doesn't change public api, so we can release it as a minor version

kate-shine (approved):

PR #424, opened 2026-05-14
fix(cachet): lower tick dependency from ^0.2.2 to >=0.2.1

codingsnoop:
tick 0.2.2 was never published to the OxidizerDependencies registry, causing cachet 0.2.0 consumers to fail resolution. This patch release widens the tick requirement to >=0.2.1,<0.4.0 so that tick 0.2.1 (which exists in the registry) satisfies the constraint.

### DIFF
--- a/crates/cachet/Cargo.toml
+++ b/crates/cachet/Cargo.toml
@@ -4,7 +4,7 @@
 [package]
 name = "cachet"
 description = "A composable, customizable multi-tier caching library with rich feature support."
-version = "0.2.0"
+version = "0.2.1"
 readme = "README.md"
 keywords = ["oxidizer", "caching", "concurrency"]
 categories = ["caching", "concurrency"]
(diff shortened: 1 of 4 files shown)


codingsnoop:
@microsoft-github-policy-service agree company="Microsoft"

Vaiz:
@martin-kolinek is working on the fix for publishing pipeline. Once it's fixed, all crates will published

codingsnoop:
> @martin-kolinek is working on the fix for publishing pipeline. Once it's fixed, all crates will published

Yes I think I don't need to merge this...using as workaround locally for the time being until publishing fixed.

geeknoid (approved):

crates/cachet/Cargo.toml
```diff
@@ -60,7 +60,7 @@ pin-project-lite = { workspace = true }
 postcard = { workspace = true, optional = true }
 serde = { workspace = true, optional = true, features = ["derive"] }
 thread_aware = { workspace = true }
-tick = { workspace = true, features = [] }
+tick = { path = "../tick", default-features = false, version = ">=0.2.1, <0.4.0", features = [] }
```
ralfbiedert, line 63: This feels doubly wrong. For one, it breaks the `workspace` convention, It also asserts semver compatibility over 2 major versions, which is hard to review without knowledge of both crates _and_ their version history.

ralfbiedert (changes requested):

PR #427, opened 2026-05-18
feat(data_privacy)!: introduce data_privacy_core 

Vaiz:
Breaking changes in data_privacy crate can lead to cascading breaking changes across all consumer crates. This PR decouples redaction engines from basic data_privacy types so we can have more stable crate for classification, and have more freedom when we do breaking changes in redaction itself.


crates/data_privacy/src/lib.rs
```diff
@@ -178,29 +179,23 @@
...
+// Needed for the `taxonomy` macro to be able to use `data_privacy` instead of `crate` in examples.
 extern crate self as data_privacy;
 
-mod classified;
-mod data_class;
-mod macros;
-mod redacted;
+// Re-export everything from data_privacy_core.
+pub use data_privacy_core::{
+    Classified, DataClass, IntoDataClass, RedactedDebug, RedactedDisplay, RedactedToString, Redactor, Sensitive, classified, taxonomy,
```
ralfbiedert: This didn't preserve the macro documentation.

Image

crates/data_privacy_macros_impl/src/crate_path.rs
```diff
@@ -0,0 +1,44 @@
...
+use quote::quote;
+
+/// Returns the token stream for the crate path to use in generated code.
+///
+/// Checks the consumer's `Cargo.toml` for `data_privacy` first (since it re-exports core),
+/// then falls back to `data_privacy_core`.
+#[must_use]
+#[cfg_attr(coverage_nightly, coverage(off))]
+pub fn resolve() -> TokenStream {
+    use proc_macro_crate::{FoundCrate, crate_name};
```
ralfbiedert: Not sure how I feel about this over just saying `data_privacy` is the way to go. It's a large crate, but it doesn't help lowering our dependency count, adding 8 more or so to our list (7 of them uniquely for this) for a use case we could say we just don't support.

ralfbiedert (approved):

crates/data_privacy_macros_impl/src/crate_path.rs
Vaiz: I'm actually expect people to use `data_privacy_core` crate in libraries if they define reusable types. for example, a library defines `TenantId`, it doesn't need to have a dependency on full data_privacy crate

ralfbiedert: Like in the `oxidizer` case this would be an antipattern, at least when it comes to trying to make proc macros work. Random example, serde:

> In crates that derive an implementation of Serialize or Deserialize, you must depend on the [serde](https://crates.io/crates/serde) crate, not serde_core.

Vaiz: maybe crate names are not the best,
what if we change `data_privacy_core` -> `data_privacy` and `data_privacy` -> `data_privacy_redactors` ?

in that case, we can avoid re-exporting macros all together

ralfbiedert: That's a possible conversation, but that seems to conflict with the premise of this PR. I'm all for creating a `_core` crate of stable items to avoid breakage. In this example though you seem to imply that everything in data_privacy then is stable except redactors? 

Now that I think about it, I think what would improve this PR is actually moving out the macros of `_core`. These are mainly crate-specific, and we can happily rename macro attributes or change behavior over crate versions, without breaking overall ecosystem composability.

Vaiz: I like that idea, what would you say if the only things that stays in core crate are those:

**Structs**
DataClass

**Traits** (all of them)
Classified
IntoDataClass
RedactedDebug
RedactedDisplay
RedactedToString
Redactor


I'm not totally sure about [Sensitive](https://docs.rs/data_privacy/latest/data_privacy/struct.Sensitive.html), but I think it's fine to keep it out of core

Vaiz: @ralfbiedert, wdyt?

crates/data_privacy/src/lib.rs
Vaiz: fixed

crates/data_privacy_macros_impl/src/crate_path.rs
Vaiz: applied

crates/data_privacy_core/src/redactor.rs
```diff
@@ -0,0 +1,35 @@
...
+/// strategies (e.g. hash-based or replacement-based redactors) implement this
+/// trait. Custom implementations are possible for testing or specialized
+/// scenarios.
+pub trait Redactor {
+    /// Returns whether this redactor would modify output for the given data class.
+    ///
+    /// Implementations may return `false` when no transformation is applied — for
+    /// example, when a strategy operates in pass-through mode, or when redaction has
+    /// been explicitly suppressed for the given class.
+    fn would_redact(&self, data_class: &DataClass) -> bool;
```
ralfbiedert: Small nit as you're also releasing a new version, can we rename this to just `redacts` (in the general theme of shortening names)?

ralfbiedert (approved):

crates/data_privacy_macros_impl/src/crate_path.rs
ralfbiedert: > I'm not totally sure about [Sensitive](https://docs.rs/data_privacy/latest/data_privacy/struct.Sensitive.html), but I think it's fine to keep it out of core

Yes I agree with your approach to keep it out for now. `Sensitive` is meant to be an ad-hoc thing, mostly used crate-internally.

crates/data_privacy_core/src/redactor.rs
Vaiz: done

geeknoid:
What is the concrete motivation for this extra layer of complexity? Is there something concrete or this is purely theoretical?

Vaiz:
> What is the concrete motivation for this extra layer of complexity? Is there something concrete or this is purely theoretical?

in telemetry, we rely on types implementing `Classified` and `RedactedDisplay` but it only works when telemetry crate and user's crates use the same version of `data_privacy`, otherwise traits become incompatible. with growing number of crates that depend on `data_privacy`, coordinated upgrade becomes much harder, so we gonna see mixed versions of `data_privacy` which will cause compilation issues

this change is a bit preliminary, but I do expect that it would save us some headache when we need to do a breaking change in some of the redactors for example

AdomasBekeras (approved):

PR #429, opened 2026-05-18
chore: Bump cachet_memory to 0.2.0

codingsnoop:
Publishes the eviction_policy API that was added after the initial 0.1.0 release.


schgoo:
Why do we need to bump this?

*edit* we talked offline, the eviction policy changes never got released

schgoo:
Added to #430

PR #438, opened 2026-05-21
feat(templated_uri): add effective_port and make port return explicit port

martintmk:
BaseUri::port() and Origin::port() now return only the explicit port from the authority, without inferring scheme defaults. The new effective_port() method returns the explicit port or falls back to the IANA-registered default (80 for http, 443 for https).

### DIFF
--- a/crates/templated_uri/src/base_uri.rs
+++ b/crates/templated_uri/src/base_uri.rs
@@ -300,12 +300,11 @@ impl BaseUri {
         Self { origin, path: self.path }
     }
 
-    /// Returns the port of this [`BaseUri`].
+    /// Returns the explicit port of this [`BaseUri`].
     ///
-    /// Returns the explicit port from the authority when present. For HTTP and
-    /// HTTPS, the well-known default port is inferred from the scheme when no
-    /// port is specified. For other schemes without an explicit port this
-    /// method returns `None`.
+    /// Returns the explicit port from the authority when present, or `None`
+    /// when no port is specified. Default ports are not inferred from the
+    /// scheme; use [`BaseUri::effective_port`] for that.
     ///
     /// # Examples
     ///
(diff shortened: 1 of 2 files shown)


crates/templated_uri/src/origin.rs
```diff
@@ -99,12 +99,54 @@ impl Origin {
...
+    /// // HTTPS default is used when no port is specified.
+    /// let origin = Origin::from_static("https://example.com");
+    /// assert_eq!(origin.effective_port(), Some(443));
+    ///
+    /// // HTTP default is used when no port is specified.
+    /// let origin = Origin::from_static("http://example.com");
+    /// assert_eq!(origin.effective_port(), Some(80));
+    /// ```
+    #[must_use]
+    pub fn effective_port(&self) -> Option<u16> {
```
ralfbiedert, line 138: Nit, shouldn't it be the other way around with slightly different names?

- `port()` -> return effective port 
- `explicit_port()` -> return explicit port if given

ralfbiedert (approved):

crates/templated_uri/src/origin.rs
martintmk: I would suggest keeping the behavior of http crate that does not assume any ports.

sgalkin:
Why is this needed? It feels confusing to have 2 functions solving the same problem of getting port out of uri.

crates/templated_uri/src/origin.rs
sgalkin: maybe `explicit_port()` should be called `port_of_default()` (as derived from schema) or maybe even move it out as a free function?

martintmk: This solves two problems, sometimes you only want port as is it specified, so no magic replacements, being very explicit.

However, there are cases where port is inferred (as is the case for HTTP/HTTPS scheme), this is the use-case for `effective_port`. Mostly, this information is useful for telemetry, or when creating TCP connections where port is mandatory.

martintmk: > port_of_default()
I think it's better for caller to provide the replacement value rather us pushing something unsuitable. 

In my scenarios, I am using -1 for telemetry:

```rust
uri.effective_port().map_or(-1, i64::from)
```

crates/templated_uri/src/base_uri.rs
```diff
@@ -300,12 +300,11 @@ impl BaseUri {
...
     }
 
-    /// Returns the port of this [`BaseUri`].
+    /// Returns the explicit port of this [`BaseUri`].
     ///
-    /// Returns the explicit port from the authority when present. For HTTP and
-    /// HTTPS, the well-known default port is inferred from the scheme when no
-    /// port is specified. For other schemes without an explicit port this
-    /// method returns `None`.
+    /// Returns the explicit port from the authority when present, or `None`
```
sandersaares, line 305: What makes sense in my little head is `fn port() -> Option<u64>` being the effective port and `fn raw_port() -> Option<u64>` being the port-as-written.

The URI types for me are logical types - the primary use case does NOT care what was written where, what matters is what ends up being used. So the effective port is the main use case, not the raw string contents. The raw string contents are some advanced niche use case. Therefore, I agree that `effective_port()` seems like it should be the default and any raw-field-values should be shoved away behind more cumbersome APIs.

What `http` does is completely immaterial in my view - we should not feel ourselves bound by poor choices made by others in the Rust ecosystem, as there are so very many of those.

martintmk: > The URI types for me are logical types - the primary use case does NOT care what was written where, what matters is what ends up being used. 

IMHO the following instances:

```rust
    let origin1 = Origin::from_static("https://example.com");
    let origin2 = Origin::from_static("https://example.com:8080");
```

Should not have the same representation. 

In `origin2` I explicitly set the port and this should be reflected appropriately in the model. In the current API iteration, the `Origin::port` tells us whether port was explicitly provided, while `Origin::effective_port` does additional port inference that is not part of the logical model.

PR #450, opened 2026-05-26
feat: introduce fetch_tls crate

martintmk:
Introduces `fetch_tls`, a new crate that provides backend-agnostic TLS
configuration for HTTP clients. It separates *what* TLS behavior an
application wants from *which* TLS implementation actually provides it:
applications describe their TLS requirements once, and the consuming
HTTP client (or other library) decides which backend to materialize at
runtime.

### Two perspectives

- **Applications** use `TlsOptions` / `TlsOptionsBuilder` to describe
  what they want: pin a specific backend, wrap an already-built
  backend, or leave the choice to the consuming library.
- **Libraries** use `TlsBackendBuilder` to turn a `TlsOptions` into a
  ready-to-use `TlsBackend`. The library contributes the
  environment-specific pieces (e.g. the rustls crypto provider and
  default server certificate verifier) and chooses a backend when the
  application did not pin one.

Applications describe *what* they want without picking a backend:

``` rust
use fetch_tls::{ClientIdentity, TlsOptions};
use http::Version;

// Simplest case: defaults, backend chosen by the HTTP client library.
let tls = TlsOptions::default();

// Customize shared options without committing to rustls or native-tls.
let tls = TlsOptions::builder()
    .supported_http_versions(&[Version::HTTP_2])
    .client_identity(ClientIdentity::from_pem(cert_pem, key_pem)?)
    .build();
```

Libraries that consume `TlsOptions` own the backend choice via
`TlsBackendBuilder`:

```rust
use std::sync::Arc;
use fetch_tls::TlsBackendBuilder;
use rustls::crypto::aws_lc_rs;

// Configure once, per HTTP client.
let backend_builder = TlsBackendBuilder::new()
    .configure_rustls(
        Arc::new(aws_lc_rs::default_provider()),
        Arc::new(MyDefaultServerCertVerifier),
    )
    // Optional: override the default backend promoted by `configure_rustls`.
    .defaults_to_native_tls();

// Materialize whatever the application gave us into a ready-to-use backend.
let backend = backend_builder.build_backend(tls)?;
```


crates/fetch_tls/src/backend.rs
```diff
@@ -0,0 +1,55 @@
...
+///
+/// Typically produced by [`TlsBackendBuilder`](super::TlsBackendBuilder);
+/// construct directly only when wrapping a backend you have already built.
+#[derive(Debug, Clone)]
+#[allow(
+    clippy::allow_attributes,
+    clippy::large_enum_variant,
+    reason = "configuration object; boxing would clutter the public API without performance benefit"
+)]
+pub enum TlsBackend {
```
ralfbiedert, line 25: Small nit, should this really be a public enum with its variants, or rather something opaque? The only thing that'd speak for an enum were if another crate needed to match on this, but that sounds brittle if the different variants are `cfg`'ed, right?

ralfbiedert (approved):

crates/fetch_tls/src/backend.rs
martintmk: I didn't want to over-complicate over what we have internally. We explicitly expose something very similar, so I wanted to preserve the same capabilities. (we do not shy away from the fact that we support  either rustls or native tls). Once we go closer to stabilization we can think about refining these APIs.

> The only thing that'd speak for an enum were if another crate needed to match on this, but that sounds brittle if the different variants are cfg'ed, right?

Indeed, but anyone consuming `TlsBackend` need to be aware of both `rustls` or `native-tls` features or just use discards if some backend variant is not accounted for.

tekian (approved):

PR #451, opened 2026-05-26
feat(templated_uri): add try_effective_port method

martintmk:
Adds a fallible variant of effective_port on Origin and BaseUri that returns a UriError when the authority does not specify a port and the scheme has no well-known default known to this crate (http, https).

(noticed that this could be useful when using the effective_port in our codebase)


sgalkin (approved):
Please address both Copilot comment - remove crate:: from UriError and add negative test case

ralfbiedert (approved):

PR #454, opened 2026-05-26
feat: add configurable ttl on stampede protected cache, eviction telemetry

codingsnoop:
This pull request introduces version 0.5.1 of the `cachet` crate, adding major new features for per-entry TTL control and improved eviction telemetry. The most significant changes include new `get_or_insert_with` and `try_get_or_insert_with` APIs for flexible cache entry creation, support for emitting `cache.eviction` telemetry events, and updates to documentation and internal structures to reflect these enhancements.

**New APIs for Per-Entry TTL Control:**

- Added `get_or_insert_with` and `try_get_or_insert_with` methods to the `Cache` type, allowing users to provide closures that return `CacheEntry<V>`, which enables setting custom TTLs per entry on cache miss. These methods support both infallible and fallible computation functions, and integrate with stampede protection and concurrency mechanisms. [[1]](diffhunk://#diff-c5cfe68284c3a1649b02a21ea117ff42421685631bb5d53f533af48b72acd09dR440-R601) [[2]](diffhunk://#diff-e2d218227ee125e9cea08516795e31f38572d96d587023c4078fc7b9727f6fb9R3-R8)

**Eviction Telemetry Enhancements:**

- Introduced a new `max_capacity` builder method on `CacheBuilder` to specify a capacity threshold for emitting `cache.eviction` telemetry events when an insert causes an eviction. This does not enforce storage limits but enables monitoring cache evictions.
- Added a new `cache.eviction` event to the telemetry system, including updates to documentation, event constants, and the `CacheTelemetry` interface to record these events. [[1]](diffhunk://#diff-f4ff7d297b285dd84021eedce9d83c43fbc6714bd70dc7bdc5b140358337e29bR88-R90) [[2]](diffhunk://#diff-6db4b6e2955ab62ea9f1167fb1888399fbc028dc9015fdeadd6389e866cf3f80R168-R173) [[3]](diffhunk://#diff-0464feb9c43fcde112a91037dc00123a9e4cf78ed75e241075ccf95d092d5275L257-R257) [[4]](diffhunk://#diff-75e748e4cca077594f89a01fbb5a5e87fbbf02b5aedcfc84616f3385b26b730aL259-R259)

**Documentation and Versioning Updates:**

- Updated the crate version to 0.5.1 in `Cargo.toml` and all documentation links, and added release notes to the `CHANGELOG.md` describing the new features. [[1]](diffhunk://#diff-ecf183d1fc0103513082a7d0f795dec0ca1b459aea6e2b5e95ce951d52dc1355L7-R7) [[2]](diffhunk://#diff-75e748e4cca077594f89a01fbb5a5e87fbbf02b5aedcfc84616f3385b26b730aL268-R289) [[3]](diffhunk://#diff-e2d218227ee125e9cea08516795e31f38572d96d587023c4078fc7b9727f6fb9R3-R8)
- Expanded documentation and examples for the new APIs and telemetry event, including in the main crate docs and README. [[1]](diffhunk://#diff-75e748e4cca077594f89a01fbb5a5e87fbbf02b5aedcfc84616f3385b26b730aL259-R259) [[2]](diffhunk://#diff-0464feb9c43fcde112a91037dc00123a9e4cf78ed75e241075ccf95d092d5275L257-R257) [[3]](diffhunk://#diff-c5cfe68284c3a1649b02a21ea117ff42421685631bb5d53f533af48b72acd09dR440-R601) [[4]](diffhunk://#diff-2a096004576836b51d07627e82054e5a2a3e0c58ef87a94172dff5420cde2f00R230-R260)

**Internal and Test Infrastructure:**

- Extended internal merger structures and test utilities to support the new APIs and the `max_capacity` parameter, ensuring comprehensive coverage and correct behavior. [[1]](diffhunk://#diff-c5cfe68284c3a1649b02a21ea117ff42421685631bb5d53f533af48b72acd09dR30-R32) [[2]](diffhunk://#diff-c5cfe68284c3a1649b02a21ea117ff42421685631bb5d53f533af48b72acd09dR46-R48) [[3]](diffhunk://#diff-2a096004576836b51d07627e82054e5a2a3e0c58ef87a94172dff5420cde2f00R47) [[4]](diffhunk://#diff-2a096004576836b51d07627e82054e5a2a3e0c58ef87a94172dff5420cde2f00R61) [[5]](diffhunk://#diff-2a096004576836b51d07627e82054e5a2a3e0c58ef87a94172dff5420cde2f00R95) [[6]](diffhunk://#diff-b0ee5d09e942d88862f7413b5d31993cf077d1e9c1159940b0a9d092e0d86f58L45-R53) [[7]](diffhunk://#diff-d14eac843c4a852b29cf364a3a8f49fea5d76bfd1c5c1e2fc40463104d110dc6L196-R196) [[8]](diffhunk://#diff-d14eac843c4a852b29cf364a3a8f49fea5d76bfd1c5c1e2fc40463104d110dc6R449) [[9]](diffhunk://#diff-3867ac465f1a5910b6eca38e47b0c06bdafbdff353ed5172e34526f05a48eca1R388) [[10]](diffhunk://#diff-3867ac465f1a5910b6eca38e47b0c06bdafbdff353ed5172e34526f05a48eca1L478-R479) [[11]](diffhunk://#diff-3867ac465f1a5910b6eca38e47b0c06bdafbdff353ed5172e34526f05a48eca1L510-R519) [[12]](diffhunk://#diff-f4ff7d297b285dd84021eedce9d83c43fbc6714bd70dc7bdc5b140358337e29bR120)

These changes provide more flexibility and observability for users of the `cachet` crate, especially for advanced caching scenarios requiring per-entry expiration and eviction monitoring.


crates/cachet/src/builder/cache.rs
```diff
@@ -224,6 +227,37 @@ impl<K, V, CT> CacheBuilder<K, V, CT> {
         self
     }
 
+    /// Sets the maximum capacity for eviction telemetry.
```
schgoo: The name does imply that this enforces capacity limits on the underlying storage, even if the comments do not. Also, I think this has limited benefit without being enforced in the underlying storage. I think it would make more sense to make this an internal implementation detail. You can implement the telemetry in the `InMemoryCache`, but if we're going to put it here we need some way to enforce it. Essentially, you'd have to add some logic in the `CacheWrapper` to check that `CacheTier::len() > max_capacity` and then invalidate.

crates/cachet/src/cache.rs
```diff
@@ -433,6 +437,168 @@ where
...
+    ///         let value = 42; // expensive computation
+    ///         let ttl = Duration::from_secs(300); // determined by response
+    ///         CacheEntry::expires_after(value, ttl)
+    ///     })
+    ///     .await?;
+    /// assert_eq!(*entry.value(), 42);
+    /// # Ok::<(), cachet::Error>(())
+    /// # };
+    /// ```
+    pub async fn get_or_insert_with<Q, Fut>(&self, key: &Q, f: impl FnOnce() -> Fut + Send) -> Result<CacheEntry<V>, Error>
```
schgoo, line 493: `with` isn't particularly descriptive. What about `get_or_insert_entry`?

schgoo: It also might get confused with things like Moka's [`get_with`](https://docs.rs/moka/latest/moka/future/struct.Cache.html#method.try_get_with), which does the same thing as our `get_or_insert`

crates/cachet/src/builder/cache.rs
schgoo: Just saw that you are actually checking that in the wrapper. Again, if we're going to put this in the public API, we should enforce it. It will get complicated, though, because we will probably need to tie the max_capacity policy to the underlying storage's implementation of it. For moka, to get the benefits of the eviction policy, we have to set the max_capacity using the moka cache builder. This is why I didn't originally add max_capacity to the API. In order to implement it, we need to implement eviction policies. To implement those, we need to maintain some sort of index on what is in the cache. Essentially, we're reimplementing many core features. We might want to do that eventually, but I don't think this works in a silo.

crates/cachet/src/wrapper.rs
```diff
@@ -144,9 +147,19 @@ where
             return Ok(());
         }
 
+        // Check if inserting will cause an eviction (cache at capacity).
+        let at_capacity = if let Some(max_cap) = self.max_capacity {
+            self.inner.len().await.map_or(false, |len| len >= max_cap)
```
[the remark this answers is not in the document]
schgoo: This is a good suggestion, and I believe is what the auth cache actually checks (i.e. did the length change). It's worth considering, though, that `len` is not always cheap. In some cases, it may have to make an external service call (for example, with a redis cache)

crates/cachet/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## [0.6.0] - 2026-05-21
```
schgoo: Technically there are no breaking changes here, you are simply adding a new feature. You could bump this to 0.5.1

schgoo (changes requested):

schgoo:
I think the low-hanging fruit implementation for this is to implement a Moka [eviction listener](https://docs.rs/moka/latest/moka/notification/type.ListenerFuture.html), and then use that to emit the event you want

crates/cachet/src/builder/cache.rs
ralfbiedert: Isn't this more like a `capacity_alert` or so?

crates/cachet/src/cache.rs
ralfbiedert: Comparing to `std`, I think the `_with` suffix is quite fitting, since it implies a closure producing the value. I'd be more tripped by this returning a `Result` vs. the `try_` variant below signalling the _closure_ is fallible, but given that other things can be fallible upstream this might just be a domain-specific idiom.

schgoo: @ralfbiedert `get_or_insert` already takes a closure. The `_with` method here returns a `Result<CacheEntry<V>, Error>` instead of `get_or_insert`'s `Result<V, Error>`. I'm just saying that it might get confused with how `_with` is used in Moka. Since the `try_` methods in the API allow the input closure to be fallible, instead of the method itself, should those be named differently?

codingsnoop: The Result return type is coming from the Cache Operation not the closure - I'll leave the naming for now.

crates/cachet/src/builder/cache.rs
schgoo: Yeah, that is a better descriptor for it. If we differentiate it from `max_capacity` that could be ok. I still worry a bit about the cost of many `len()` calls in a hot path, though.

crates/cachet/src/builder/cache.rs
```diff
@@ -172,6 +231,37 @@ impl<K, V, CT> CacheBuilder<K, V, CT> {
...
+    ///
+    /// let clock = Clock::new_tokio();
+    /// let cache = Cache::builder::<String, i32>(clock)
+    ///     .with_eviction_telemetry()
+    ///     .memory_with(|b| b.max_capacity(1_000))
+    ///     .build();
+    /// ```
+    #[cfg(feature = "memory")]
+    #[must_use]
+    pub fn with_eviction_telemetry(mut self) -> Self {
```
schgoo: The issue with this is that, if they have the "memory" feature enabled, they can also set this on other kinds of tiers (which don't have this hooked up). Could we just put this on `InMemoryBuilder`? The new `memory_with` method is an excellent idea. Ideally it should look something like this:

```rust
let cache = Cache::builder::<String, i32>(clock)
    .memory_with(|b| b.max_capacity(1000).with_eviction_telemetry())
    .build();
```

I imagine there will be some issues with `CacheTelemetry` not being available in the `cache_memory` crate. There are several ways to solve this.

1. Create a separate telemetry stack in the memory crate.
2. Go with Ralf's suggestion to use your first draft, but rename the method to `CacheBuilder::capacity_alert(usize)` or something.
3. Refactor telemetry into a separate crate.

Options 1 and 2 are pretty good, I think. I'm less convinced about option 3

crates/cachet/src/eviction.rs
```diff
@@ -0,0 +1,100 @@
...
+    /// because the hook is keyed to the first build of a builder.
+    pub(crate) fn init(&self, telemetry: CacheTelemetry, name: CacheName) {
+        let _ = self.state.set(HookState { telemetry, name });
+    }
+
+    /// Invoked by the in-memory tier on each removal.
+    ///
+    /// Only `Size` and `Expired` causes are reported as evictions; `Explicit`
+    /// and `Replaced` are user-initiated and already accounted for by the
+    /// `cache.invalidated` / `cache.inserted` events.
```
schgoo: This should be documented in the public API

crates/cachet/src/wrapper.rs
```diff
@@ -445,4 +445,26 @@ mod tests {
         let result = wrapper.clear().await;
         result.unwrap_err();
     }
+
+    #[cfg_attr(miri, ignore)]
+    #[tokio::test]
+    async fn wrapper_does_not_emit_eviction_event_directly() {
```
schgoo: This test seems like the AI covering its tracks, but doesn't test anything useful.

crates/cachet/Cargo.toml
```diff
@@ -4,7 +4,7 @@
 [package]
 name = "cachet"
 description = "A composable, customizable multi-tier caching library with rich feature support."
-version = "0.5.0"
+version = "0.6.0"
```
schgoo: Again, no breaking changes, bump the patch version instead

crates/cachet_memory/src/builder.rs
```diff
@@ -218,6 +240,50 @@ impl<K, V, H> InMemoryCacheBuilder<K, V, H> {
...
+    /// `Size` for capacity-driven evictions, `Expired` for TTL/TTI expirations,
+    /// `Explicit` for [`invalidate`](cachet_tier::CacheTier::invalidate) or
+    /// [`clear`](cachet_tier::CacheTier::clear) calls, and `Replaced` for inserts
+    /// that overwrote an existing key.
+    ///
+    /// The listener runs on the cache's background maintenance task. Keep the
+    /// closure cheap; expensive work should be offloaded to a separate task.
+    ///
+    /// Replacing an already-registered listener is supported; only the most
+    /// recently configured listener is invoked.
```
schgoo: But not by us, right? We aren't exposing that

crates/cachet_memory/src/builder.rs
```diff
@@ -218,6 +240,50 @@ impl<K, V, H> InMemoryCacheBuilder<K, V, H> {
         self
     }
 
+    /// Registers a listener that is called when an entry is removed from the cache.
+    ///
+    /// The listener receives a [`RemovalCause`] indicating why the entry was removed:
+    /// `Size` for capacity-driven evictions, `Expired` for TTL/TTI expirations,
+    /// `Explicit` for [`invalidate`](cachet_tier::CacheTier::invalidate) or
+    /// [`clear`](cachet_tier::CacheTier::clear) calls, and `Replaced` for inserts
```
schgoo, line 259: For `clear`, how many eviction hooks are triggered?

crates/cachet_memory/src/notification.rs
```diff
@@ -0,0 +1,56 @@
...
+}
+
+pub(crate) fn from_moka(cause: moka::notification::RemovalCause) -> RemovalCause {
+    match cause {
+        moka::notification::RemovalCause::Expired => RemovalCause::Expired,
+        moka::notification::RemovalCause::Size => RemovalCause::Size,
+        moka::notification::RemovalCause::Explicit => RemovalCause::Explicit,
+        moka::notification::RemovalCause::Replaced => RemovalCause::Replaced,
+    }
+}
```
schgoo, line 34: nit: slightly more idiomatic

```suggestion
impl From<moka::notification::RemovalCause> for RemovalCause {
    fn from(cause: moka::notification::RemovalCause) -> RemovalCause {
        match cause {
            moka::notification::RemovalCause::Expired => RemovalCause::Expired,
            moka::notification::RemovalCause::Size => RemovalCause::Size,
            moka::notification::RemovalCause::Explicit => RemovalCause::Explicit,
            moka::notification::RemovalCause::Replaced => RemovalCause::Replaced,
        }
    }
}
```

crates/cachet/tests/eviction.rs
```diff
@@ -0,0 +1,58 @@
...
+use testing_aids::LogCapture;
+use tick::Clock;
+use tracing_subscriber::Registry;
+use tracing_subscriber::layer::SubscriberExt;
+
+/// Inserting past the configured `max_capacity` of the underlying moka cache
+/// must eventually emit a `cache.eviction` event for the size-based removals.
+#[cfg_attr(miri, ignore)]
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn memory_size_eviction_emits_telemetry() {
```
schgoo, line 21: How flaky is this in practice? Have copilot run it 100 times and see if it fails for any of them

schgoo (changes requested):

crates/cachet_memory/src/builder.rs
codingsnoop: yes, rm from comment

crates/cachet_memory/src/builder.rs
codingsnoop: Moka fires the listener once per entry (all with Explicit), deferred via background sweep. Our hook ignores Explicit, so no cache.eviction/cache.expired events fire from clear — the wrapper emits a single cache.cleared event instead, which avoids duplicating signals.

crates/cachet_memory/src/notification.rs
codingsnoop: The issue is that we don't want to expose moka as part of cachet's public api and make users take a dependency. we can't privatize the from impl - this is why I have the private from method and enum duplication.

crates/cachet/tests/eviction.rs
codingsnoop: couldn't produce a failure locally..

crates/cachet/CHANGELOG.md
codingsnoop: fixed

crates/cachet/src/builder/cache.rs
codingsnoop: I moved it into the memory crate

crates/cachet/src/eviction.rs
codingsnoop: added

crates/cachet_memory/src/lib.rs
schgoo, line 49: We should probably add some documentation to the lib.rs/README

schgoo (commented):
This is better. I think I would prefer that we allow options to specify which events get emitted, but we can do that later. Let's make sure the documentation is all up to date and then it's good to go

schgoo (approved):

ralfbiedert (approved):

crates/cachet/src/cache.rs
```diff
@@ -428,7 +432,172 @@ where
...
+    async fn do_try_get_or_insert_with<E, Fut>(&self, key: &K, f: impl FnOnce() -> Fut) -> Result<CacheEntry<V>, Error>
+    where
+        E: std::error::Error + Send + Sync + 'static,
+        Fut: Future<Output = Result<CacheEntry<V>, E>>,
+    {
+        if let Some(entry) = self.storage.get(key).await? {
+            return Ok(entry);
+        }
+        let mut entry = f().await.map_err(Error::from_source)?;
+        entry.ensure_cached_at(self.clock.system_time());
```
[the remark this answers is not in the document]
codingsnoop: to address in follow up

schgoo (approved):

PR #456, opened 2026-05-27
fix: gate gungraun to linux

psandana:
With this we allow to exclude the benchmarks that requires gungraun if the feature is not enabled


ralfbiedert (approved):

sandersaares (commented):
What specific problem is this addressing? There are different problems related to Gungraun and I am not confident that this is really the right approach to tidy things up here.

Feature flags for purely repo-specific development workflows feel like a design smell - nobody is going to babysit their feature flags, they will either specify none (`cargo test`) or all (`cargo test --all-features`) but not hand-pick some specific ones. And they will pick one of these two options regardless of whether they have Gungraun installed or not, so this can still end up with errors.

If this is addressing a problem of "gungraun benches fail on Windows" then I suggest to `cfg(linux)` those tests and make accessing them on Windows a no-op. Reference implementation: https://github.com/folo-rs/folo/blob/main/packages/awaiter_set/benches/awaiter_set_cg.rs

If this is addressing a problem of "gungraun benches fail on Linux if I do not have Gungraun installed" then my preferred fix is to simply require Gungraun on Linux by adding Valgrind to our standard development environment setup in `DEVELOPMENT.md` and `gungraun-runner` into `just install-tools` as a required tool. Partial reference implementation: https://github.com/folo-rs/folo/blob/083d512b7ca6d5ac6c2dd8e8896e2cd119efdcf4/justfiles/just_setup.just#L37

This way we can get by without any feature flags.

Or is there a third problem I am unaware of that needs a feature flag, where it is truly necessary?

It is also possible I misunderstand how this works or how it is meant to be used (which would suggest we need to add some docs in this PR). Feel free to educate me if so.

sandersaares (changes requested):

sandersaares (approved):
We should also add `apt install valgrind` to `DEVELOPMENT.md` (similar to what we have in our internal wiki dev env setup page).

psandana:
> We should also add `apt install valgrind` to `DEVELOPMENT.md` (similar to what we have in our internal wiki dev env setup page).

Will do in another PR. This one is getting bigger and bigger.

PR #465, opened 2026-06-02
feat: cancelable crate

afoxman:
This PR introduces a new crate named `cancelable`.

It models C# cancelation tokens and sources in Rust, aimed at improving the accuracy of C# -> Rust code translation.


ralfbiedert:
I'm not particularly fond of the naming and structure of this, is there a reason we don't align this more with how Tokio does it?

crates/cancelable/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+//! let source = CancellationTokenSource::new();
+//! source.subscribe(|| println!("Operation canceled"));
+//! source.cancel();
+//! # }
+//! ```
+//!
+//! # Futures
+//!
+//! The [`WithCancellationExt`] trait adds a
+//! [`with_cancellation`](WithCancellationExt::with_cancellation) method
```
sandersaares: I often see AI insert this `with_` prefix in many places, so I guess that is where it comes from, but that does not feel like some standard pattern to actually follow. Just `future.cancelable(token)` feels more Rusty to me.

crates/cancelable/src/token.rs
```diff
@@ -0,0 +1,538 @@
...
+        Self {
+            canceled: AtomicBool::new(canceled),
+            subscribers: Mutex::new(if canceled { None } else { Some(Vec::new()) }),
+        }
+    }
+
+    /// Returns `true` if cancellation has been requested.
+    #[must_use]
+    fn is_cancelled(&self) -> bool {
+        self.canceled.load(Ordering::Acquire)
```
sandersaares, line 64: I have found it a very useful practice to document on top of every atomic operation why the selected memory ordering is the correct one to use on that line (similar to safety comments). This is a fairly trivial flag-raising case but still, good practice is good practice.

crates/cancelable/src/token.rs
```diff
@@ -0,0 +1,538 @@
...
+    /// Returns `true` on success, signaling that the caller is responsible for notifying subscribers.
+    fn try_set_cancelled(&self) -> bool {
+        self.canceled
+            .compare_exchange(false, true, Ordering::AcqRel, Ordering::Acquire)
+            .is_ok()
+    }
+
+    /// Signal cancellation and notify subscribers
+    fn cancel_and_notify(&self) {
+        if self.try_set_cancelled() {
```
sandersaares: Code that contains logic pyramids (increasing indentation) is harder to read. I recommend flipping the condition so we return early if `false` instead.

crates/cancelable/src/token.rs
```diff
@@ -0,0 +1,538 @@
...
+            let subscribers = match self.subscribers.lock() {
+                // Lock, take contents, and unlock
+                Ok(mut guard) => guard.take().unwrap_or_default(),
+
+                // Lock has been poisoned, which means we can't read the subscriber list.
```
sandersaares: It is not clear to me why we treat a poisoned lock here as equivalent to a canceled cancellation token. Maybe there is a reason for this but it is not immediately obvious - feels like something worth documenting with at least inline comments on the mutex if there is some reason here.

If we do not expect the lock to ever be poisoned, this "poisoned lock" logic path just creates untestable code and we would be better off IMO by just expecting it to never be poisoned and guarding the mutex with `.expect("we do not execute any code that could panic while holding this lock")`.

crates/cancelable/src/token.rs
```diff
@@ -0,0 +1,537 @@
...
+            let weak = source.token.weak_ref();
+
+            for parent in parents {
+                let weak = Weak::clone(&weak);
+                parent.inner.subscribe(Box::new(move || {
+                    if let Some(inner) = weak.upgrade() {
+                        inner.cancel_and_notify();
+                    }
+                }));
+            }
```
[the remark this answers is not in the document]
sandersaares: This sounds like an important weakness in the current implementation. Long-lived parent tokens are quite common. We probably do need to allow for an unregistration when the child source is dropped, which implies some sort of deeper lifecycle relationship between the two (e.g. `Source` becomes `Source { core: Arc<SourceCore> }`, the latter of which is `Weak<T>` shared to some kind of `SubscriberToken` so when child does `drop(SubscriberToken)` it deregisters the subscription? Just first idea off the top of my head)

crates/cancelable/src/token.rs
```diff
@@ -0,0 +1,538 @@
...
+    #[must_use]
+    pub fn linked(parents: &[CancellationToken]) -> Self {
+        let source = Self::new();
+
+        if !parents.is_empty() {
+            let weak = source.token.weak_ref();
+
+            for parent in parents {
+                let weak = Weak::clone(&weak);
+                parent.inner.subscribe(Box::new(move || {
```
sandersaares: Is it possible to avoid the boxing? I suppose if this were private API, we could simply make it concrete but given that `subscribe()` is public API it of course needs to support arbitrary closures.

In some other hot paths that take arbitrary closures, we have used a thread-local `infinity_pool::BlindPool` to store the closures in a per-thread object pool, which (at least under the ailing Windows default memory allocator) offers better performance than boxing.

sandersaares (changes requested):
Looks nice! A lot simpler than I feared it would be and I like how well it enables .NET parity to be modeled. The subscriber leak that Copilot raised is a concern, though.

crates/cancelable/src/token.rs
```diff
@@ -0,0 +1,538 @@
...
+
+/// Controls cancellation for one or more [`CancellationToken`]s
+///
+/// Create a source, distribute tokens via [`token()`](Self::token), and call
+/// [`cancel()`](Self::cancel) when the operation should stop.
+///
+/// Dropping a `CancellationTokenSource` does **not** cancel its tokens.
+/// Outstanding tokens simply remain in their current state.
+#[derive(Debug, Default)]
+pub struct CancellationTokenSource {
```
martintmk, line 245: I know that this goes directly from C#, but maybe we could try to simplify these names?

`CancelToken` + `Canceller`

crates/cancelable/src/token.rs
martintmk: imho, what we found out is that small allocations usually are not that big of an issue as we thought they are.

Why not keep the dependencies slim and only optimize once we determine this causes real issues?

sandersaares: > imho, what we found out is that small allocations usually are not that big of an issue as we thought they are.

I am not convinced of this (yet?). Just saying to clarify this is not universally held opinion (in context of the "we" in there). Entirely possible it is still true, of course.

crates/cancelable/src/future.rs
```diff
@@ -0,0 +1,183 @@
...
+
+use pin_project::pin_project;
+
+use crate::CancellationToken;
+
+/// Error returned when a [`WithCancellation`] future detects that its
+/// associated [`CancellationToken`] has been canceled.
+#[ohno::error]
+#[display("operation was canceled")]
+pub struct Canceled {}
```
schgoo, line 34: Super nit: We're using Cancellation (two l's) and Cancelable/Canceled (one l) at the same time

crates/cancelable/src/lib.rs
afoxman: I like this naming better and have updated the code

crates/cancelable/src/token.rs
afoxman: There are 2 places I touch the subscriber list -- here and below in subscribe(). I think your comment refers to the code below in subscribe() even though it is up here in cancel_and_notify.

I had panics in both places via expect(..) and then took them out before finalizing the PR. I though it might be overly heavy-handed, even though quietly ignoring the poisoned lock doesn't sit well with me either. 

I suppose what you are imparting here is that when the time comes for a panic, don't try to paper over it. Best to be up front and loud when the alarm needs to sound. I've seen this be very successful as an OOM -> crash strategy in complex code bases like Office. I like it better but have been reticent to use it in Rust so far because people see unwrap/expect and get upset.

crates/cancelable/src/token.rs
afoxman: Yes. I came up with expanding the definition of a Subscriber to be an enum of a boxed closure or a weak ref to another token. Effectively, I pushed the closure from 211 out to real code, and call it only when that type of subscription is in place.

crates/cancelable/src/token.rs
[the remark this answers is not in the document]
afoxman: I agree for linked tokens, but not so much for external callback closures.

What I've done is split out the registration of subscribers, tracking external closures separately from linked tokens (both in a single enum). When a child source is dropped, it visits each of its parent sources, and unlinks itself.

The primary use case is met -- long-lived parents don't grow infinitely over time as children come and go. The only thing that can make them grow is someone deliberately pumping a ton of callback closures into them. These cannot "unsubscribe". I noted that explicitly in the docs. If that somehow becomes a problem, we can always revisit.

crates/cancelable/src/token.rs
afoxman: The big question for me is "Will AI be able to figure out that the different names match up with the C# classes?" I have no idea how discoverable this stuff is, or how much weight it will carry with the agents bringing in things like crate docs and other code samples.

Without a good answer, having the names match feels safe? Maybe it's a delusion. And the choice is pretty permanent so let's talk it through a bit.

You've used agents to build things - likely more than me. What's your feeling on discoverability?

crates/cancelable/src/future.rs
```diff
@@ -0,0 +1,190 @@
...
+        let this = self.project();
+
+        // Check cancellation before running the inner future so we can
+        // short-circuit without performing unnecessary work.
+        if this.token.is_cancelled() {
+            return Poll::Ready(Err(Canceled::new()));
+        }
+
+        match this.inner.poll(cx) {
+            Poll::Ready(output) => Poll::Ready(Ok(output)),
```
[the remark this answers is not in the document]
afoxman: this analysis is wrong

crates/cancelable/src/token.rs
```diff
@@ -0,0 +1,830 @@
...
+        if !self.try_set_cancelled() {
+            // Already canceled by someone else. They will notify.
+            return;
+        }
+
+        // Lock, take subscribers, and unlock
+        let subscribers = self
+            .subscribers
+            .lock()
+            .expect("subscriber lock is poisoned")
```
sandersaares: I see you took my advice on not papering over panics, as you put it. An important part of this, however, is to make it clear to the reader why we can be comfortable with adding a potential panic into the code. In this case, the reason is (AFAIUI) that (unless we create a bug somewhere), there is no way for the lock to become poisoned because we do not call out to any uncontrolled code (e.g. user callbacks) under the lock.

This needs to be conveyed to the next reader, though. I suggest using the expect-message for that. In its current form "lock is poisoned" it does not add much more value than the panic stack trace itself. Such an `expect()` is sort of an `unwrap()` in disguise. We can extract a lot more value if we explain here the reason we feel it is fine to panic. I recommend this as a general principle for `.expect()` that we never expect to actually get hit - to have the message explain why it can never happen. Lock poisoning is a common case but this really applies generally.

crates/cancelable/src/token.rs
```diff
@@ -0,0 +1,830 @@
...
+        // Release the lock, then notify immediately.
+        drop(guard);
+        callback.notify();
+    }
+
+    /// Remove the linked child token from the list of subscribers.
+    ///
+    /// This is a no-op if cancellation has already occurred (the list is `None`).
+    ///
+    /// # Panics
```
sandersaares, line 128: We should only document panics that the caller can actually cause. If we never expect the lock to be poisoned (which is the case as far as I can tell), this panic can never occur, so we should not document it.

Instead, we should just suppress the lint that says "you need to document your panics", with an explanation that there are only unrechable panics in here.

crates/cancelable/src/token.rs
sandersaares: Agent discoverability is black magic! All I hear is stories of how people fail to get agents to do the right thing, rarely about how they succeed. I have also had pretty hit or miss experiences but one thing I observe is that there is a clear trend toward doing the right thing more consistently over time. Considering that agents and techniques to use agents constantly evolve, we probably do not need to worry too much about pampering them - the models (or rather, the harnesses that drive them) are likely to get better and cover any gaps.

More important IMO is to keep the resulting code human-understandable to people coming from .NET, in which case familiarity matters more than for agents. The agent has the patience to read documentation to understand differences, after all, but far fewer humans are going to do that.

The naming is a bit long in the Rust universe but not objectionable to me. Any alternatives I can come up with have other negatives. Feels fine overall.

sandersaares (approved):

codingsnoop (approved):

crates/cancelable/src/future.rs
```diff
@@ -0,0 +1,190 @@
...
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+
+        // Check cancellation before running the inner future so we can
+        // short-circuit without performing unnecessary work.
+        if this.token.is_cancelled() {
+            return Poll::Ready(Err(Canceled::new()));
+        }
+
+        match this.inner.poll(cx) {
```
schgoo, line 113: If the inner future returns `Pending`, then we're stuck here until something wakes the task. For some I/O tasks, this means it won't wake until the I/O returns (or a timeout or something). Essentially, it might not get the cancellation message (i.e. call `is_cancelled`) until the I/O task has already completed.

schgoo: I think we need to create our own Waker that will wake the task when cancellation occurs, then subscribe to it when inner.poll returns Pending

crates/cancelable/src/future.rs
afoxman: yeah you can thank the spellchecker and clippy for that 😒

crates/cancelable/src/future.rs
afoxman: The design goal is "cooperative" cancellation, not preemptive.

We could do that, but it would be a whole different animal. And it should use a different name since this one aligns with the naming/behavior from C#. 

I did intend for this crate to expanded with these kinds of ideas - so maybe a different struct for an interruptible / pre-emptive cancel is a good addition.

crates/cancelable/src/future.rs
schgoo: You can add words to .spelling!

schgoo (approved):

ralfbiedert (changes requested):
Still same comment I left earlier. I think this needs a motivation why it looks like the C# version, and not like Tokio's flavor of that. In the end we try to solve Rust problems, not C# problems.

afoxman:
> Still same comment I left earlier. I think this needs a motivation why it looks like the C# version, and not like Tokio's flavor of that. In the end we try to solve Rust problems, not C# problems.

I don't know Tokio well enough to know what this means. Can you give some examples of things you'd like changed? Is it runtime differences? Naming? Different data modeling? Contractual differences?

ralfbiedert:
> I don't know Tokio well enough to know what this means. Can you give some examples of things you'd like changed? Is it runtime differences? Naming? Different data modeling? Contractual differences?

Tokio has a [CancellationToken](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html) in the `tokio_util` crate; additional [documentation is here in the 'Shutdown' section](https://tokio.rs/tokio/topics/shutdown).

My criticism is that this API looks like a port from C#, for example having CT and CTS, how linked tokens are created, ... If at the end of a design analysis this is still the best API then of course it's fine (although CTS is too much of a name for Rust IMO), but I haven't seen any conversation where this was compared against Tokio's or any other Rust-based approach. 

Very generally, the first reflex for crates like this should be looking into "how does Rust solve this C# problem", not "how does C# solve this C# problem". What would be good is looking at what crates already exist out there. There is at least Tokio's, maybe `smol` tried solving this too?

afoxman:
tokio_util::CancellationToken looks a lot like what I've got here for my CancellationToken. APIs even have mostly the same names. I don't have a drop-guard, but that can be added to my design if it is needed.

So I think what you're hinting at is that you don't like the idea of a CancellationTokenSource. This exists because I offer more than Tokio's solution -- I let callers register for callbacks so they can define their own cancellation logic rather than embedding it into the task being canceled.

It's a different approach. That's why it is a bit different.

I'd argue this does in fact solve a big Rust problem, and Tokio's cancellation token is evidence that something like this is needed. My impl is different, which is one reason it exists. Further, though, and more importantly, the problem of code translation from C# to Rust is front and center for Microsoft. It is a worthy problem to solve, and this helps with that, giving developers and AI agents something very familiar to work with - something missing from the Rust ecosystem.

schgoo (approved):

schgoo:
@ralfbiedert I'm approving to unblock @afoxman for now. If you're game, I think we should have a deeper discussion about your concerns here, and what recommendations you have for the API.

afoxman:
It sounds like this needs more higher-level discussion, so I'm going to land it in an internal repo for now. After we have some consensus on how to proceed, I can get this PR landed here (with any needed changes).

ralfbiedert:
> It sounds like this needs more higher-level discussion,

Yes, let's have a call.

> I'd argue this does in fact solve a big Rust problem, and Tokio's cancellation token is evidence that something like this is needed. 

If we looked at existing implementations and decided it's better, sure. 

> Further, though, and more importantly, the problem of code translation from C# to Rust is front and center for Microsoft. It is a worthy problem to solve, and this helps with that, giving developers and AI agents something very familiar to work with

I don't buy that argument. Relative to all other code, cancellation is a tiny corner of the problem space. Everything else in that picture already looks and acts differently, for good reason. 

- C# has reflection, Rust doesn't,
- C# has a GC, Rust doesn't,
- C#'s `static` and link-model is different from Rust's,
- C#'s async runtime interaction with the language works differently in Rust,
- C# doesn't have an equivalent of Rust's `Future` drop cancellation (it couldn't even express that if it wanted to due to lack of move semantics)

When porting infrastructure code, the goal can't be to have APIs that look identical so "people have something familiar to worth with"; there is so much impedance mismatch in the rest of the language that I don't even know what that would mean at scale.

crates/cancelable/src/token.rs
```diff
@@ -0,0 +1,870 @@
...
+    Linked(Weak<Inner>),
+}
+
+impl Subscriber {
+    fn notify(self) {
+        match self {
+            Self::External(f) => f(),
+            Self::Linked(weak) => {
+                if let Some(inner) = weak.upgrade() {
+                    inner.cancel_and_notify();
```
geeknoid: **Linked cancellation can exhaust the thread stack** — *Correctness · Medium · High*

`CancellationTokenSource::cancel()` enters `Inner::cancel_and_notify()`, which invokes each linked subscriber synchronously; this branch then calls `inner.cancel_and_notify()` again. A valid chain therefore consumes one Rust call frame per linked source. The current test stops at three levels, so a sufficiently deep chain can stack-overflow (typically aborting the process) instead of delivering cooperative cancellation.

The other sites in the path are [`cancel`](https://github.com/microsoft/oxidizer/blob/ff067d44db3ee71b4a85b45bb56986c10cbf7772/crates/cancelable/src/token.rs#L299-L310) and the subscriber loop in [`cancel_and_notify`](https://github.com/microsoft/oxidizer/blob/ff067d44db3ee71b4a85b45bb56986c10cbf7772/crates/cancelable/src/token.rs#L79-L97).

**Direction:** propagate linked cancellation with an explicit worklist/queue (or otherwise bound recursion) while preserving exactly-once notification.

**Done when:** cancellation depth no longer consumes one call frame per link, and a regression test exercises a chain beyond a deliberately small thread stack; measuring fan-out/depth also resolves the unguarded performance path.

crates/cancelable/src/token.rs
```diff
@@ -0,0 +1,870 @@
...
+        // wait for the thread to start running
+        while counter.load(Ordering::Relaxed) == 0 {
+            assert!(std::time::Instant::now() < deadline, "thread did not start in time");
+            std::hint::spin_loop();
+        }
+
+        thread_handle
+    }
+
+    #[test]
```
geeknoid: **The concurrent state machine is only tested as a visibility check** — *Testing · Medium · High*

This test starts a polling reader and then cancels, but it never interleaves the operations that mutate `subscribers`. No current test would catch a callback being lost or invoked twice when `subscribe()` races `cancel_and_notify()`, or a linked callback firing after `CancellationTokenSource::drop()` races parent cancellation. Those are the exact invariants implemented jointly by the atomic flag and mutex.

**Direction:** add deterministic concurrency coverage (for example, a `loom`-compatible synchronization shim or barrier-controlled hooks) for cancel-vs-subscribe and parent-cancel-vs-linked-drop, asserting exact-once delivery or successful unregistration for every ordering.

**Done when:** tests force both orderings of each race and fail if a subscriber is skipped, duplicated, or invoked after the linked source has won the unregister race.

crates/cancelable/src/lib.rs
```diff
@@ -0,0 +1,72 @@
...
+//! let source = CancellationTokenSource::new();
+//! let token = source.token();
+//!
+//! let result = async { 42 }.cancelable(token).await?;
+//! assert_eq!(result, 42);
+//! # Ok(())
+//! # }
+//! ```
+
+pub mod future;
```
geeknoid: **`CancelableExt` has two public import paths** — *Conformance · Low · High*

Making `future` public exposes `cancelable::future::CancelableExt`, while the next line re-exports the same item as `cancelable::CancelableExt`. That gives one public item two canonical paths, contrary to [M-SINGLE-ITEM-PATH](https://microsoft.github.io/rust-guidelines/guidelines/ai/#M-SINGLE-ITEM-PATH), and makes generated docs and agent-assisted imports ambiguous before this API is published.

**Direction:** choose one canonical path—typically keep the module private and re-export both `CancelableExt` and `Canceled` at the crate root, or remove the root re-export.

**Done when:** every public item in this module is reachable through exactly one documented import path.

geeknoid (commented):
Complete static-only five-pass review of the new `cancelable` crate: all 150 security, 162 correctness, 182 testing, 183 performance, and 143 conformance classes were assessed. Nothing was built or executed; generated `Cargo.lock` was excluded per scope.

The main new issue is stack growth through synchronous recursive linked-token propagation. I also called out the absence of deterministic tests for the atomic/mutex interleavings and one adopted API-path guideline deviation.

The cancellation state transition itself is otherwise coherent: subscription is serialized with draining, callbacks run after the mutex is released, linked registrations use weak references, and source drop unregisters linked children. The security pass found no attacker-controlled trust boundary or unsafe/FFI surface.

Prior feedback about post-poll `Ready` semantics, wake behavior, callback unregistration, panic behavior, doctests/imports, spelling, and lock scope was not repeated.

PR #466, opened 2026-06-02
chore: Tidy cargo dependencies to unbreak publishing

sandersaares:
## Motivation

Untidy dev-dependency declarations caused breakage in release process due to resulting in a wrong order of dependency publishing.

## Summary

Applied [`/ss-cargo-deps-tidy`](https://github.com/sandersaares/ss-ai/blob/main/skills/ss-cargo-deps-tidy/SKILL.md) to tidy Cargo dependencies in the workspace.

The headliner is **Stage 3**: same-workspace dev-dependencies are now path-only references, which is required for the publishing-order resolver. The remaining stages are additional Cargo.toml tidying performed for consistency.

| Stage | What changed |
|---|---|
| 2 — Style | Converted 100+ `foo.workspace = true` and similar dotted-key declarations to inline-table style across 24 crate manifests; converted the one `[dev-dependencies.syn]` sub-table to inline-table; padded short version requirements (`"0.8"` → `"0.8.0"`) for ahash, async-once-cell, bytemuck, dashmap, rstest, zerocopy. |
| **3 — Same-workspace dev-deps** | **Converted 36 dev-dependencies referencing workspace members via `workspace = true` to `path = "../<crate>"` form (features preserved). This is the fix needed to unbreak publishing order resolution.** |
| 4 — Version centralization | Moved `smol` from `anyspawn`'s dev-dependencies into `[workspace.dependencies]`; `anyspawn` now uses `workspace = true`. |
| 5 — Default features off | No change — all workspace deps already had `default-features = false`. |
| 6 — Workspace-level features | No change. The only flagged item was `postcard = { features = ["use-std"] }` which is the hyphenated alt phrasing of the allow-listed `use_std`. Did not hoist `std` / `alloc` to workspace level because `multitude` is a `#![no_std]` crate that shares deps (serde, bytes) with std-using crates. |
| 7 — Minimum-version verification | Bumped the minimum versions discovered to be broken when pinned via `=x.y.z`: |

| Dep | Was | Now | Reason |
|---|---|---|---|
| `ahash` | `0.8` | `0.8.4` | 0.8.0 yanked |
| `hyper` | `1.6.0` | `1.7.0` | `wiremock` 0.6.5 requires `hyper >= 1.7` |
| `new_zealand` | `1.0.1` | `1.0.4` | `many_cpus` 2.1.0 requires `new_zealand >= 1.0.4` |
| `zerocopy` | `0.8` | `0.8.26` | `FromZeros` derive on `#[repr(C, align(N))]` structs needs >= 0.8.26 (bisected) |

## Verification

All passed on Windows with the temporarily-pinned `=x.y.z` minimum versions, then reverted to the bumped baselines above:

- `cargo check --workspace --all-targets`
- `cargo check --workspace --all-targets --release`
- `cargo check --workspace --all-targets --all-features`
- `cargo clippy --workspace --all-targets --all-features`

### DIFF
--- a/crates/fundle/Cargo.toml
+++ b/crates/fundle/Cargo.toml
@@ -9,11 +9,11 @@ readme = "README.md"
 keywords = ["oxidizer", "di"]
 categories = ["rust-patterns"]
 
-edition.workspace = true
-rust-version.workspace = true
-authors.workspace = true
-license.workspace = true
-homepage.workspace = true
+edition = { workspace = true }
+rust-version = { workspace = true }
+authors = { workspace = true }
+license = { workspace = true }
+homepage = { workspace = true }
 repository = "https://github.com/microsoft/oxidizer/tree/main/crates/fundle"
 
 [package.metadata.cargo_check_external_types]
@@ -23,10 +23,10 @@ allowed_external_types = ["fundle_macros::*"]
 all-features = true
 
 [dependencies]
-fundle_macros.workspace = true
+fundle_macros = { workspace = true }
 
 [dev-dependencies]
-trybuild.workspace = true
+trybuild = { workspace = true }
 
 [lints]
 workspace = true
--- a/crates/http_extensions/Cargo.toml
+++ b/crates/http_extensions/Cargo.toml
@@ -8,11 +8,11 @@ version = "0.4.2"
 readme = "README.md"
 keywords = ["oxidizer", "http", "extensions", "client", "server"]
 categories = ["network-programming"]
-edition.workspace = true
-rust-version.workspace = true
-authors.workspace = true
-license.workspace = true
-homepage.workspace = true
+edition = { workspace = true }
+rust-version = { workspace = true }
+authors = { workspace = true }
+license = { workspace = true }
+homepage = { workspace = true }
 repository = "https://github.com/microsoft/oxidizer/tree/main/crates/http_extensions"
 
 [package.metadata.cargo_check_external_types]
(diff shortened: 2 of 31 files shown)


crates/uniflight/Cargo.toml
```diff
@@ -9,11 +9,11 @@ readme = "README.md"
 keywords = ["oxidizer", "coalescing", "stempede", "singleflight", "deduplication"]
```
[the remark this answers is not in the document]
sandersaares: While true, going to skip this because I do not want to go for another 3 hour build loop for a typo.

crates/fundle/Cargo.toml
```diff
@@ -9,11 +9,11 @@ readme = "README.md"
...
-edition.workspace = true
-rust-version.workspace = true
-authors.workspace = true
-license.workspace = true
-homepage.workspace = true
+edition = { workspace = true }
+rust-version = { workspace = true }
+authors = { workspace = true }
+license = { workspace = true }
+homepage = { workspace = true }
```
martin-kolinek, line 16: Why the more verbose syntax? Does that have an effect on dependency resolution as well?

crates/http_extensions/Cargo.toml
```diff
@@ -57,42 +57,42 @@ json = ["dep:serde_core", "dep:serde_json"]
...
+bytesbuf = { path = "../bytesbuf", features = ["bytes-compat", "test-util"] }
+criterion = { workspace = true }
+data_privacy = { path = "../data_privacy" }
 futures = { workspace = true, features = ["std", "executor"] }
 hyper = { workspace = true, features = ["client", "http1"] }
 hyper-util = { workspace = true, features = ["server-auto", "tokio"] }
-layered = { workspace = true, features = ["intercept", "dynamic-service"] }
-mutants.workspace = true
-ohno = { workspace = true, features = ["test-util", "app-err"] }
+layered = { path = "../layered", features = ["intercept", "dynamic-service"] }
```
martin-kolinek, line 84: I was really hoping that without cyclic deps we wouldn't run into issues with just using workspace deps for dev-dependencies.  Intuitively, I'd expect that without cycles, dev-dependencies coming from workspace with standard versions should work - where does this intuition break down?

martin-kolinek (approved):

crates/fundle/Cargo.toml
sandersaares: Just for uniformity. Having different syntax mixed (because sometimes you need to add `features` etc) is just ugly and inconsistent.

crates/http_extensions/Cargo.toml
sandersaares: I do not know why it works this way. I simply have observed that with path-only dev-dependencies it works and the other way around it does not work.

sgalkin (approved):

PR #471, opened 2026-06-04
feat: Rewrite the multitude crate

geeknoid:
- Makes Arena Send (was !Send).
- Removes the Rc family of smart pointers.
- Slight performance increase in certain cases.

Public types removed: Rc, RcStr, RcUtf16Str, ArcStr, BoxStr.

Production-code unsafe surface (`src/`, excluding `#[cfg(test)]`
modules and doc comments):

| Category       | Before | After | Delta |
|----------------|-------:|------:|------:|
| unsafe fn      |    160 |    82 |   -78 |
| unsafe { ... } |    681 |   195 |  -486 |
| unsafe impl    |     23 |    16 |    -7 |


crates/multitude/src/box.rs
```diff
@@ -89,146 +67,46 @@ use crate::rc::Rc;
 /// assert_eq!(*b, vec![1, 2, 3, 4]);
 /// ```
 pub struct Box<T: ?Sized, A: Allocator + Clone = Global> {
-    ptr: InLocalChunk<T, A>,
+    /// Pointer to the contained value, which lives in a 64K-aligned
+    /// [`SharedChunk`]'s payload. The chunk header is recovered by
+    /// walking backwards within the value's `CHUNK_ALIGN` tile.
+    ptr: NonNull<T>,
+    /// Marker for variance, dropck, and the !Send/!Sync property
```
ralfbiedert: Why is this unconditionally `!Send`? Given that `Arena` is `Send` now, I should be able to allocate a box on thread 1 and then move the arena to thread 2, right? Symmetry-wise, this should be the same as moving the Box instead, as in either case the deallocation has to cross thread boundaries?

crates/multitude/src/box.rs
```diff
@@ -13,70 +13,48 @@ use core::fmt::{self, Debug, Display, Formatter, Pointer};
...
-/// soundness.
-///
-/// If you need `Send` + `Sync` ownership of an arena value, use
-/// [`Arc`](crate::Arc) (via
-/// [`Arena::alloc_arc`](crate::Arena::alloc_arc)) instead. Convert from
-/// a [`Box<T, A>`] to an [`Rc<T, A>`](crate::Rc) via
-/// [`Box::into_rc`](Self::into_rc) when you need shared, immutable
-/// access (also `!Send`/`!Sync`).
+/// [`Box`] is always `!Send` and `!Sync`, even when `T` is both. If
+/// you need `Send`/`Sync` ownership of an arena value, use
```
ralfbiedert: This feels off. People should use `Box` when they need single ownership, and `Arc` if they need shared ownership.

ralfbiedert:
Maybe this is still WIP, but for now `Arena` is still `!Send`:

image

crates/multitude/src/box.rs
geeknoid: Fixed

ralfbiedert (approved):

psandana (approved):

PR #472, opened 2026-06-04
fix: re-release dependents of data_privacy 0.12 breaking change so they are compatible with latest data_privacy

Vaiz:
interesting detail, `templated_uri` uses `data_privacy::Sensitive` which was not moved to `data_privacy_core`. As a result, breaking changes in `data_privacy` crate cascade through all depended crates. Maybe we should move `Sensitive` to core as well


sandersaares (approved):

sandersaares (changes requested):
The title says "bump data_privacy" but the content does not actually change version of data_privacy, instead changing version of other crates. What is happening here?

sandersaares:
Okay I am starting to get the picture - there was `data_privacy 0.12` released but for some reason the dependents were not released. Why not? Ah, because #427 did not include them - it seems to be manually incremented version number, not via the script that cares for the dependency tree. I associate this with your question "what is the right way to release GitHub packages" in chat yesterday - this was just human error in not using the script.

sandersaares (approved):

Vaiz:
@copilot resolve the merge conflicts in this pull request

ralfbiedert:
Hm, just had a look and I even think I originally wrote the `to_string() -> Sensitive<>` pattern. The semantics of `Sensitive` changed a bit over time, originally this was meant to be an internal type only, but it also made sense to reach for it when revealing stringy information. I suppose we could move it to core, but then `data_privacy` ends up being a redactors crate + re-exports.

ralfbiedert (approved):

Vaiz:
superseded by #477

PR #478, opened 2026-06-05
feat(thread_aware): add feature-gated ThreadAware impls for 3rd-party crate types

martin-kolinek:
Adds opt-in `ThreadAware` implementations for commonly-used 3rd-party types behind per-crate Cargo features. By default this crate brings no additional dependencies; users enable only the features they need.

## Motivation

Discussed in the Oxidizer Contributors channel (thread by Evgenii Shutov), and inspired by ox-sdk!5243054 which adds `ThreadAware` for `UuidId<T>` / `UuidString` wrappers around `uuid::Uuid`. Many users wrap or re-export 3rd-party primitive/value types and need them to satisfy `ThreadAware` to flow through APIs that require it. Doing this in user crates requires the newtype dance; offering the impls here behind opt-in features removes that friction without forcing the dependencies on anyone.

## Design

Each supported crate gets its own Cargo feature; enabling it pulls in that crate as an optional dependency and adds blanket `ThreadAware` impls for its plain value types via a small private macro.

Feature names follow this convention (documented in `Cargo.toml`) so a future major version of the wrapped crate can land additively without forcing a breaking release of `thread_aware`:

* Stable `1.x` (or any other stable major) → bare crate name (e.g. `bytes`, `http`, `uuid`).
* `N.x` for `N >= 2` → `<crate><N>` (e.g. `bytes2` if `bytes 2.x` ever lands).
* `0.x` → `<crate>0<minor>` (e.g. `jiff02` for `jiff 0.2.x`).

| Feature | Types |
| --- | --- |
| `bytes` | `Bytes`, `BytesMut` |
| `http` | `StatusCode`, `Version`, `Method`, `HeaderName`, `HeaderValue`, `HeaderMap<HeaderValue>`, `Uri`, `uri::Authority`, `uri::Scheme`, `uri::PathAndQuery`, `uri::Port<T>`, `Error`, `uri::InvalidUri`, `Request<T: ThreadAware>`, `Response<T: ThreadAware>` |
| `jiff02` | `Timestamp`, `Span`, `SignedDuration`, `civil::Date`, `civil::Time`, `civil::DateTime`, `civil::ISOWeekDate` |
| `uuid` | `Uuid` |

All inert value types (`StatusCode`, `Method`, `HeaderName`, `HeaderValue`, `Bytes`, `Uri`, jiff/uuid primitives, etc.) get a no-op `relocate`. `HeaderMap<HeaderValue>` is also treated as inert because its values are inert. `Request<T>::relocate` and `Response<T>::relocate` forward to the body only — iterating header values would be wasted work since `HeaderValue::relocate` is no-op, and `http::Extensions` carries `Any` payloads that cannot be relocated generically.

### Intentional omissions

- `jiff::Zoned` (carries a TimeZone reference; semantics warrant explicit user choice)
- `chrono` and `time` (deferred per reviewer feedback; can be added later as `chrono04` / `time03` if needed)

### Files

- `crates/thread_aware/src/third_party/{mod,bytes,http,jiff02,uuid}.rs` — one private submodule per supported crate version, each gated by `#[cfg(any(test, feature = "..."))]`. `mod.rs` defines a private `impl_noop_thread_aware!` macro used by submodules with inert value types.
- The wrapped crates are declared inline in `crates/thread_aware/Cargo.toml` so this crate owns its version pins and no workspace-level deps are added.
- Each submodule has unit tests that statically assert `ThreadAware`/`Send`/`Sync` and a smoke test on the threaded runtime.

### `allowed_external_types`

Because impls on foreign types make those types reachable through this crate's public surface, `cargo_check_external_types` (run with `--all-features` in CI) flags them. Explicit type paths are listed in `Cargo.toml` (rather than wildcards) for visibility.

## Verification

- `cargo clippy --all-features --all-targets -- -D warnings` ✅
- `cargo test --all-features` ✅ (127 unit + integration + 14 doctests)
- `cargo test --tests --no-default-features --workspace` ✅
- `just format`, `just readme`, `just spellcheck`, `cargo sort --check --grouped --workspace` ✅


martin-kolinek:
&#x1F916; Fixed in 404dcdb: made `third_party` and its feature-gated submodules `pub` so the doc link resolves to this crate's module and users can navigate to it.

---

crates/thread_aware/Cargo.toml
```diff
@@ -27,11 +65,27 @@ default = ["derive"]
 derive = ["dep:thread_aware_macros"]
 threads = ["dep:many_cpus"]
 
+# Optional `ThreadAware` impls for 3rd-party crate types. Each feature pulls in
+# only that one crate as an optional dependency. See `third_party::*` modules.
+bytes = ["dep:bytes"]
```
martintmk, line 79: While I like this change due improved ergonomics, one thing that worries me is massive exposure to these crates making breaking changes and thus thread_aware making breaking releases too. 

Since this is such foundational crate, are we OK with this?

martin-kolinek: &#x1F916; Good concern. Some context and a proposed mitigation:

**Current exposure.** All 6 features are opt-in and off by default, so a consumer is only affected if they explicitly enable e.g. `thread_aware/chrono` — and at that point they already depend on `chrono` themselves. A `chrono` major bump forces *them* to migrate regardless, so the additional friction `thread_aware` adds is "wait for `thread_aware` to release a matching version" rather than "I now have to migrate because of `thread_aware`".

**The real risk.** Yes, this still couples `thread_aware`'s SemVer to whoever moves fastest among `bytes`/`chrono`/`http`/`jiff`/`time`/`uuid`. For a foundational crate that is undesirable.

**Proposed mitigation: version-suffixed features.** Adopt the same pattern many ecosystem bridge crates use (`rand_core`-style, `serde_with`, etc.):

- `chrono-0_4`, `uuid-1`, `http-1`, `time-0_3`, `jiff-0_2`, `bytes-1`

Each suffixed feature pins to a single major. When `chrono 1.0` ships we can add `chrono-1` *additively* without removing `chrono-0_4`. `thread_aware` then only needs a breaking release when it wants to drop an old-major feature, not every time a wrapped crate bumps.

Trade-off: feature names are uglier and the README needs to spell out the mapping.

Want me to refactor to that scheme before merging, or keep the plain names and accept the coupling? Happy either way — your call.

---

martintmk: I suggest dropping chrono and time, these should be treated as legacy APIs as ecosystem alligns to jiff. Also no need to complicate the features since thread_aware is not stable anyway, but we should make some note about using the version suffixes before stabilization.

crates/thread_aware/src/third_party/http.rs
```diff
@@ -0,0 +1,36 @@
...
+//!
+//! Enable with the `http` Cargo feature.
+//!
+//! Only the fully inert value types are covered. `HeaderName`, `HeaderValue`,
+//! `Uri`, etc. internally hold `Bytes` (and therefore an `Arc`) and deserve
+//! separate, more nuanced handling rather than a blanket no-op.
+
+use ::http::{Method, StatusCode, Version};
+
+impl_noop_thread_aware!(StatusCode, Version, Method);
```
martintmk: I would do the same for Request, Response, HeaderMap, HeaderValue, HeaderName.

crates/thread_aware/Cargo.toml
Vaiz: if we call features `http_v1`, `bytes_v1`, it would be a matter of adding new features when there are breaking changes in those crates

Vaiz (approved):

crates/thread_aware/src/third_party/time.rs
```diff
@@ -0,0 +1,41 @@
...
+    assert_impl_all!(Time: ThreadAware, Send, Sync, Copy);
+    assert_impl_all!(PrimitiveDateTime: ThreadAware, Send, Sync, Copy);
+    assert_impl_all!(OffsetDateTime: ThreadAware, Send, Sync, Copy);
+    assert_impl_all!(UtcOffset: ThreadAware, Send, Sync, Copy);
+    assert_impl_all!(Duration: ThreadAware, Send, Sync, Copy);
+    assert_impl_all!(Weekday: ThreadAware, Send, Sync, Copy);
+    assert_impl_all!(Month: ThreadAware, Send, Sync, Copy);
+
+    #[test]
+    #[cfg(feature = "threads")]
```
sandersaares: I believe our general pattern is the other way around - we always gate items behind `any(test, feature = "foo")` so tests do not need features manually turned on to test everything and test functions themselves do not need gating (we should probably document this pattern in agents.md).

sandersaares (commented):
> time::Instant (requires std feature on time)

I am not sure why this is an omission. Do we really intend to support no-std? Feels unnecessary. If someone is using our SDK, we can assume they are using the standard library.

crates/thread_aware/Cargo.toml
sandersaares: > Also no need to complicate the features since thread_aware is not stable anyway, but we should make some note about using the version suffixes before stabilization.

If we add the suffixes we should add them now. "We can fix it later" is rarely true and if we do not do it right from the start, it is unlikely we ever will.

crates/thread_aware/README.md
```diff
@@ -100,12 +100,27 @@ from types that implement [`ThreadAware`][__link23].
   environments: `default-features = false`.
 * **`threads`**: Enables features mainly used by async runtimes for OS interactions.
 
+### 3rd-party crate impls
+
+The following opt-in features provide no-op [`ThreadAware`][__link24] implementations
+for inert value types from popular 3rd-party crates. Each feature adds only
+its named crate as an optional dependency; by default none are enabled and
```
sandersaares: Features can add dependencies, but the wording "feature adds ... an optional dependency" seems wrong. If the feature is enabled, the dependency is not optional anymore.

crates/thread_aware/README.md
```diff
@@ -100,12 +100,27 @@ from types that implement [`ThreadAware`][__link23].
...
+the crate brings in no extra dependencies. See [`third_party`][__link25] for the list
+of types covered by each feature.
+
+* **`bytes`**: Impls for `bytes::Bytes`, `bytes::BytesMut`.
```
sandersaares, line 120: What alternatives are there to us doing this? Is the choice between this or everyone rolling their own newtypes?

martin-kolinek: Yes, pretty much.

crates/thread_aware/Cargo.toml
martin-kolinek: &#x1F916; Done in 4b4664b3. Adopted the version-suffixed feature scheme (`bytes_v1`, `http_v1`, `jiff_v0_2`, `uuid_v1`) so future major bumps can be supported additively without forcing a breaking release of `thread_aware`. Also dropped `chrono` and `time` per your suggestion.

---

crates/thread_aware/src/third_party/http.rs
martin-kolinek: &#x1F916; Done in 4b4664b3. Added impls for `HeaderName`, `HeaderValue`, `HeaderMap<T>`, `Request<T>`, `Response<T>`. `HeaderMap`/`Request`/`Response` propagate `relocate` to their inner `T` (mirroring `Vec<T>`/`Box<T>`); the headers themselves are no-op like the other inert types.

---

crates/thread_aware/src/third_party/time.rs
martin-kolinek: &#x1F916; Done in 4b4664b3. Switched the per-feature `mod` declarations to `#[cfg(any(test, feature = "..."))]` and moved `bytes`, `http`, `jiff`, `uuid` to unconditional `dev-dependencies` so `cargo test` exercises every third-party impl without needing any feature flag.

---

crates/thread_aware/README.md
martin-kolinek: &#x1F916; Done in 4b4664b3. Reworded the `# Features` block: dropped 'optional dependency' phrasing; now says enabling a feature pulls that crate in as a dependency.

---

crates/thread_aware/README.md
martin-kolinek: &#x1F916; Confirming Martin's answer: the practical alternative is every downstream crate writing its own newtype wrapper around each foreign type (`MyUuid(uuid::Uuid)` etc.) just to satisfy the `ThreadAware` bound, which is the friction this PR removes. Providing the impls here behind opt-in, version-suffixed features keeps the default dependency footprint at zero.

---

crates/thread_aware/src/third_party/http.rs
```diff
@@ -0,0 +1,77 @@
...
+            value.relocate(source, destination);
+        }
+    }
+}
+
+impl<T: ThreadAware> ThreadAware for Request<T> {
+    fn relocate(&mut self, source: Option<Affinity>, destination: Affinity) {
+        self.body_mut().relocate(source, destination);
+    }
+}
```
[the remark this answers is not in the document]
martin-kolinek: :robot: GHCP: Good catch. `Request<T>::relocate` now iterates `headers_mut().values_mut()` and calls `relocate` on every header value before relocating the body, matching the `HeaderMap<T>` propagation behavior. Pushed in `fa83ffd4`.

_via [Road to Gas Town](https://aka.ms/roadtogastown) :robot:_

crates/thread_aware/src/third_party/http.rs
```diff
@@ -0,0 +1,77 @@
...
+
+impl<T: ThreadAware> ThreadAware for Response<T> {
+    fn relocate(&mut self, source: Option<Affinity>, destination: Affinity) {
+        self.body_mut().relocate(source, destination);
+    }
+}
```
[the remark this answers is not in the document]
martin-kolinek: :robot: GHCP: Same fix applied to `Response<T>`: `relocate` now propagates to every header value and then the body. Pushed in `fa83ffd4`.

_via [Road to Gas Town](https://aka.ms/roadtogastown) :robot:_

crates/thread_aware/src/lib.rs
```diff
@@ -92,6 +92,24 @@
 //!   environments: `default-features = false`.
 //! * **`threads`**: Enables features mainly used by async runtimes for OS interactions.
 //!
+//! ## 3rd-party crate impls
```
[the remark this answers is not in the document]
martin-kolinek: :robot: GHCP: Updated the PR description to reflect the actual implementation: version-suffixed feature names (`bytes_v1`/`http_v1`/`jiff_v0_2`/`uuid_v1`), the additional `http` types (`HeaderName`, `HeaderValue`, `HeaderMap<T>`, `Request<T>`, `Response<T>`), and the new omissions list (`chrono`/`time` deferred per Martin's feedback).

_via [Road to Gas Town](https://aka.ms/roadtogastown) :robot:_

crates/thread_aware/src/third_party/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+// Licensed under the MIT License.
+
+//! `ThreadAware` implementations for selected 3rd-party crate types.
+//!
+//! Each submodule is gated behind a Cargo feature whose name embeds the major
+//! (or 0.x minor) of the wrapped crate — for example `bytes_v1`, `http_v1`,
+//! `jiff_v0_2`, `uuid_v1`. Enabling a feature pulls in that crate as a
+//! dependency and exposes `ThreadAware` impls for inert, self-contained types
+//! from it. By default no such features are enabled, so the crate stays
+//! dependency-free.
```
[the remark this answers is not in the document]
martin-kolinek: GHCP: Good catch — reworded the module doc to clarify that the no-feature default avoids pulling in any of the *wrapped* third-party crates (rather than implying the crate has zero dependencies, which is not true since `derive` is on by default).

crates/thread_aware/src/third_party/http_v1.rs
```diff
@@ -0,0 +1,109 @@
...
+
+//! `ThreadAware` impls for [`http`] (1.x) types.
+//!
+//! Enable with the `http_v1` Cargo feature.
+//!
+//! Inert value types (`StatusCode`, `Method`, `Version`, `HeaderName`,
+//! `HeaderValue`) get a no-op `relocate`. Container types (`HeaderMap<T>`,
+//! `Request<T>`, `Response<T>`) propagate `relocate` to every element they own;
+//! `Request<T>` and `Response<T>` relocate both their header values and body,
+//! mirroring how this crate handles `Vec<T>` and `Box<T>`.
```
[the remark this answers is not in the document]
martin-kolinek: GHCP: Agreed — `http::Extensions` holds `Any` values with type-erased payloads, so we can't relocate them generically. Added a doc note calling that out explicitly so callers stashing thread-affine state in extensions know they need to relocate it themselves.

crates/thread_aware/src/third_party/http_v1.rs
```diff
@@ -0,0 +1,109 @@
...
+        map.insert("x-one", HeaderValue::from_static("one"));
+        map.insert("x-two", HeaderValue::from_static("two"));
+        map.relocate(Some(affinities[0]), affinities[1]);
+        assert_eq!(map.get("x-one").unwrap(), "one");
+        assert_eq!(map.get("x-two").unwrap(), "two");
+    }
+
+    #[test]
+    #[cfg(feature = "threads")]
+    fn request_relocate_propagates_to_body() {
```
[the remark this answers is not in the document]
martin-kolinek: GHCP: Renamed to `request_relocate_propagates_to_body_and_headers` to match what the test actually exercises.

crates/thread_aware/Cargo.toml
```diff
@@ -27,17 +51,36 @@ default = ["derive"]
...
+http_v1 = ["dep:http"]
+jiff_v0_2 = ["dep:jiff"]
+uuid_v1 = ["dep:uuid"]
+
 [dependencies]
 many_cpus = { workspace = true, optional = true }
 
 thread_aware_macros = { workspace = true, optional = true }
 
+bytes = { workspace = true, optional = true }
```
sandersaares: We need to version-fix these dependency references to match the versions in the feature declarations, otherwise this is bound to create confusion when we upgrade past a breaking change.

Workspace should to have these dependencies explicitly versioned in their name to signal "fixed major version":

```toml
[workspace.dependencies]
bytes_v1 = { version = "0.1.0", package = "bytes" }
```

And we need to reference them by their explicitly versioned names in the package manifest:

```toml
[dependencies]
bytes_v1 = { workspace = true, optional = true }
```

crates/thread_aware/src/third_party/http_v1.rs
```diff
@@ -0,0 +1,162 @@
...
+    /// Used by mutation-testing-killing tests to detect that container impls
+    /// actually delegate to their inner elements rather than no-op'ing —
+    /// otherwise mutants that replace the body of `relocate` with `()` would
+    /// survive when the inner type's `relocate` itself has no observable
+    /// effect (as it does for `HeaderValue` or `Vec<u8>`).
+    #[cfg(feature = "threads")]
+    #[derive(Default)]
+    struct Counter(u32);
+
+    #[cfg(feature = "threads")]
```
sandersaares: I believe I already left a comment earlier that the ox-sdk standard pattern is the other way around - in tests we always consider all feature flags enabled, via `any(test, feature = "foo")` on the cfg-gated API surface itself, and no cfg-guards in the test logic itself.

Maybe this pattern is violated in existing code of the thread_aware package - I suggest fixing the existing code to follow the pattern instead of spamming cfg-guards all over test code.

sandersaares (changes requested):

crates/thread_aware/src/third_party/http_v1.rs
```diff
@@ -0,0 +1,162 @@
...
+    fn relocate(&mut self, source: Option<Affinity>, destination: Affinity) {
+        for value in self.values_mut() {
+            value.relocate(source, destination);
+        }
+    }
+}
+
+impl<T: ThreadAware> ThreadAware for Request<T> {
+    fn relocate(&mut self, source: Option<Affinity>, destination: Affinity) {
+        for value in self.headers_mut().values_mut() {
```
martintmk: Going through every header map value just to do no-op  is quite wasteful.

crates/thread_aware/src/third_party/http_v1.rs
```diff
@@ -0,0 +1,162 @@
...
+            value.relocate(source, destination);
+        }
+        self.body_mut().relocate(source, destination);
+    }
+}
+
+impl<T: ThreadAware> ThreadAware for Response<T> {
+    fn relocate(&mut self, source: Option<Affinity>, destination: Affinity) {
+        for value in self.headers_mut().values_mut() {
```
martintmk: same, only body relocation is necessary

crates/thread_aware/src/third_party/http_v1.rs
```diff
@@ -0,0 +1,162 @@
...
+
+use ::http::header::{HeaderMap, HeaderName, HeaderValue};
+use ::http::{Method, Request, Response, StatusCode, Version};
+
+use crate::ThreadAware;
+use crate::affinity::Affinity;
+
+impl_noop_thread_aware!(StatusCode, Version, Method, HeaderName, HeaderValue);
+
+impl<T: ThreadAware> ThreadAware for HeaderMap<T> {
```
martintmk: default header map `HeaderMap<T: HeaderValue>` should be no-op.

crates/thread_aware/src/lib.rs
```diff
@@ -142,6 +160,7 @@
 mod cell;
 mod core;
 mod impls;
+pub mod third_party;
```
ralfbiedert: Why are these `pub`? Just for the impls to exist / work it doesn't need to be public, and the documentation there doesn't provide any benefit either, esp. given that we don't really special-document `ThreadAware` behavior for anything else in our own crates.

ralfbiedert (changes requested):

crates/thread_aware/Cargo.toml
martin-kolinek: GHCP: Good call. Added new version-suffixed workspace deps (`bytes_v1`, `http_v1`, `jiff_v0_2`, `uuid_v1` with `package = ...`) and switched `thread_aware` to use them. Kept the existing `bytes`/`http`/`jiff`/`uuid` workspace entries since 11 other crates depend on them — adding aliases instead of renaming avoids unrelated churn.

crates/thread_aware/src/third_party/http_v1.rs
martin-kolinek: GHCP: Removed all `#[cfg(feature = "threads")]` guards from individual tests. `pinned_affinities` is unconditionally `pub`, so the guards were stale dead weight.

crates/thread_aware/src/third_party/http_v1.rs
martin-kolinek: GHCP: Dropped the header iteration. `Request<T>::headers_mut()` is `HeaderMap<HeaderValue>` and `HeaderValue::relocate` is no-op, so iterating was pure waste. `Request<T>::relocate` now only relocates the body.

crates/thread_aware/src/third_party/http_v1.rs
martin-kolinek: GHCP: Same fix for `Response<T>` — body-only relocation now.

crates/thread_aware/src/third_party/http_v1.rs
martin-kolinek: GHCP: Narrowed the impl from `HeaderMap<T: ThreadAware>` to `HeaderMap<HeaderValue>` and routed it through the noop macro — no iteration. The narrower bound matches how `http` itself uses `HeaderMap`.

crates/thread_aware/src/lib.rs
martin-kolinek: GHCP: Made `third_party` and its submodules private. The `ThreadAware` impls still apply via the trait, and there's no API surface that needs to be exposed.

crates/thread_aware/src/lib.rs
```diff
@@ -142,6 +160,7 @@
 mod cell;
 mod core;
 mod impls;
+mod third_party;
 mod wrappers;
```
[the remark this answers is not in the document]
martin-kolinek: GHCP: Fixed in 5c8e40ec — replaced the broken `[`third_party`]` link in the crate-level docs with an inline summary of each feature's types and the body-only relocation semantics. README regenerated via `just readme`.

crates/thread_aware/src/third_party/http_v1.rs
```diff
@@ -0,0 +1,115 @@
...
+    fn relocate(&mut self, source: Option<Affinity>, destination: Affinity) {
+        self.body_mut().relocate(source, destination);
+    }
+}
+
+impl<T: ThreadAware> ThreadAware for Response<T> {
+    fn relocate(&mut self, source: Option<Affinity>, destination: Affinity) {
+        self.body_mut().relocate(source, destination);
+    }
+}
```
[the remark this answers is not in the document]
martin-kolinek: GHCP: PR description was out of date; updated it to match the current design. The body-only behaviour is intentional (Martin requested it earlier in this review): `HeaderMap<HeaderValue>` is treated as inert because `HeaderValue::relocate` is no-op, so iterating it is pure waste, and `Request<T>`/`Response<T>` only need to relocate their body for the same reason.

crates/thread_aware/src/third_party/http_v1.rs
```diff
@@ -0,0 +1,115 @@
...
+//! Enable with the `http_v1` Cargo feature.
+//!
+//! All inert value types in `http` (`StatusCode`, `Method`, `Version`,
+//! `HeaderName`, `HeaderValue`) get a no-op `relocate`. `HeaderMap<HeaderValue>`
+//! — the only header map shape the `http` crate produces — is also treated as
+//! inert, so iterating its entries would be wasted work.
+//!
+//! `Request<T>` and `Response<T>` propagate `relocate` to their body only.
+//! Their headers are `HeaderMap<HeaderValue>` and therefore already no-op.
+//!
```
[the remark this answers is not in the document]
martin-kolinek: GHCP: Reworded the module docs in 5c8e40ec to describe what this crate's impls do (HeaderMap<HeaderValue> is no-op because HeaderValue is no-op; Request/Response forward only the body) rather than what shape `http` produces.

crates/thread_aware/src/third_party/http.rs
```diff
@@ -0,0 +1,115 @@
...
+        map.insert("x-one", HeaderValue::from_static("one"));
+        map.insert("x-two", HeaderValue::from_static("two"));
+        map.relocate(Some(affinities[0]), affinities[1]);
+        assert_eq!(map.get("x-one").unwrap(), "one");
+        assert_eq!(map.get("x-two").unwrap(), "two");
+    }
+
+    #[test]
+    fn request_relocate_propagates_to_body() {
+        let affinities = pinned_affinities(&[2]);
```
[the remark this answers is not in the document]
martin-kolinek: GHCP: After the latest design change, `Request<T>::relocate` actually does forward to the body only, so the test name `request_relocate_propagates_to_body` is accurate. The header insert in the test exists to verify the headers are preserved through the relocate call, not relocated themselves.

crates/thread_aware/src/third_party/http.rs
```diff
@@ -0,0 +1,115 @@
...
+        req.headers_mut()
+            .insert(HeaderName::from_static("x-trace"), HeaderValue::from_static("abc"));
+        req.relocate(Some(affinities[0]), affinities[1]);
+        assert_eq!(req.body().0, 1);
+        assert_eq!(req.headers().get("x-trace").unwrap(), "abc");
+    }
+
+    #[test]
+    fn response_relocate_propagates_to_body() {
+        let affinities = pinned_affinities(&[2]);
```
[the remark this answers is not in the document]
martin-kolinek: GHCP: Same as the Request thread — `Response<T>::relocate` is body-only after the design change, so `response_relocate_propagates_to_body` is accurate.

crates/thread_aware/src/third_party/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+//! crates additively: when, say, `bytes 2.0` ships we can add a `bytes_v2`
+//! feature without removing `bytes_v1`, avoiding a breaking release of this
+//! crate purely because of an upstream major bump.
+//!
+//! Tests in this module are compiled and run as part of `cargo test` without
+//! needing the features enabled — the wrapped crates are also listed as
+//! unconditional `dev-dependencies`, and the submodules below are gated on
+//! `any(test, feature = "...")`.
+//!
+//! See the workspace `Cargo.toml` for the exact versions used.
```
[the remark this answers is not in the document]
martin-kolinek: GHCP: Fixed in cf6d1f06 — updated the doc to point at the crate's own `Cargo.toml` (where the version pins now live since we inlined the deps).

crates/thread_aware/src/lib.rs
```diff
@@ -92,6 +92,25 @@
...
+//! crate brings in no extra dependencies.
+//!
+//! * **`bytes_v1`**: Impls for `bytes::Bytes`, `bytes::BytesMut`.
+//! * **`http_v1`**: Impls for `http::StatusCode`, `http::Method`, `http::Version`,
+//!   `http::HeaderName`, `http::HeaderValue`, `http::HeaderMap<HeaderValue>`,
+//!   `http::Request<T>`, `http::Response<T>`.
+//! * **`jiff_v0_2`**: Impls for `jiff::Timestamp`, `jiff::civil::DateTime`, etc.
+//! * **`uuid_v1`**: Impl for `uuid::Uuid`.
+//!
+//! For `Request<T>` and `Response<T>`, `relocate` is forwarded to the body
```
martin-kolinek: Document this on the ThreadAware impl for the type rather than at crate root.

martin-kolinek: GHCP: Done in e37cb9a8 — moved the body-only relocate note onto the `ThreadAware` impls for `Request<T>` and `Response<T>` in `http_v1.rs`, and trimmed the matching paragraph from the crate root. Also tightened the module-level docs so they only cover the no-op impls and point at the per-impl docs for Request/Response.

martintmk (approved):

crates/thread_aware/src/third_party/mod.rs
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! `ThreadAware` implementations for selected 3rd-party crate types.
```
sandersaares, line 4: Considering some comments in this PR already highlighted accidental inefficiencies in the implementations, I feel covering these `impl`s with a set of Criterion + Callgrind benchmarks would be good for some future-proofing, set a good example for future additions and to avoid accidental regressions in the future.

sandersaares (approved):

martin-kolinek:
GHCP: Benches landed in 06e64723 — paired Criterion (cross-platform) + gungraun (Linux-only) harnesses under `crates/thread_aware/benches/`, mirroring the multitude layout. Four groups cover the three design properties: `noop_value/*` (inert types stay no-op), `header_map/{empty,populated_8,populated_32}` (HeaderMap relocate is O(1) regardless of size), `request/*` and `response/*` (body-only forwarding — the `unit_body_*_headers` pair should be identical, and `counter_body` confirms the body's `relocate` is reached). Local smoke run on Windows shows all four header_map/request variants tied at ~1.7µs as expected.

crates/thread_aware/benches/criterion_third_party.rs
```diff
@@ -0,0 +1,246 @@
...
+// noop_value
+// =========================================================================
+
+/// Regression sentinels for the inert no-op impls. Any non-trivial cost
+/// here means a no-op `relocate` body grew real work.
+fn bench_noop_value(c: &mut Criterion) {
+    let (src, dst) = affinities();
+    let mut g = c.benchmark_group("noop_value");
+
+    macro_rules! noop {
```
martin-kolinek: Doesn't feel like this needs a macro a function should be more than sufficient

crates/thread_aware/benches/criterion_third_party.rs
```diff
@@ -0,0 +1,246 @@
...
+
+fn bench_header_map(c: &mut Criterion) {
+    let (src, dst) = affinities();
+    let mut g = c.benchmark_group("header_map");
+
+    for (name, count) in [("empty", 0_usize), ("populated_8", 8), ("populated_32", 32)] {
+        g.bench_function(name, |b| {
+            let mut m = build_header_map(count);
+            b.iter(|| {
+                for _ in 0..N {
```
martin-kolinek: Doesn't criterion already do its own iteration? Why do we need our own?

crates/thread_aware/benches/gungraun_third_party/linux.rs
```diff
@@ -0,0 +1,226 @@
...
+    ISOWeekDate::new(2026, 23, jiff_v0_2::civil::Weekday::Monday).expect("valid ISO week date")
+);
+
+// ===== header_map — cost must be identical across sizes =====
+
+#[library_benchmark]
+#[bench::run(header_map(0), affinity_pair())]
+fn header_map_empty(mut m: HeaderMap, sd: (Affinity, Affinity)) -> HeaderMap {
+    let (src, dst) = sd;
+    for _ in 0..N {
```
martin-kolinek, line 102: Any specific reason for N? Do instruction counts need longer runs?

crates/thread_aware/benches/criterion_third_party.rs
martin-kolinek: GHCP: Fixed in 3539217a — replaced the macro with a generic `bench_relocate<T: ThreadAware>` helper.

crates/thread_aware/benches/criterion_third_party.rs
martin-kolinek: GHCP: Good catch — fixed in 3539217a. Criterion does handle iteration on its own and the inner loop was just deflating per-op numbers (~1.7µs / 1000 ≈ ~1.7ns/op, which matches what criterion now reports directly). Kept the loop in the gungraun bench, since callgrind measures a single invocation deterministically and needs the loop to see meaningful instruction counts.

crates/thread_aware/benches/gungraun_third_party/linux.rs
martin-kolinek: GHCP: Inherited from the multitude bench convention (also `N = 1_000` there). The rationale for callgrind specifically: per-bench-function harness overhead (prologue, teardown, the gungraun shim) is constant per invocation, so for a single `relocate` call — especially a no-op — the harness instructions would dominate the reported count and small regressions in the impl itself would get lost in the noise. The loop amortises that to a per-iteration delta that callgrind can resolve cleanly. Happy to lower it if you have a preferred value; 1000 just matched the existing precedent.

martin-kolinek: Makes sense

sandersaares (approved):

ralfbiedert:
I have one more nit, can we rename `bytes_v1`,  `http_v1` and `uuid_v1` to just `bytes`, `http` and `uuid`? This is convention in the ecosystem (compare C-FEATURE), the `_vX` looks awkward, and the likelihood of them ever seeing another version are small given their pervasiveness (in which case `bytes2`, ...) would be better names.  

The other thing, I'm fine with us supporting some crates like that, but it raises the bigger question where we draw the line?

ralfbiedert (approved):

martin-kolinek:
GHCP: Renamed in 61e0184d per the scheme we agreed on in chat:

- `1.x` crates -> bare crate name (`bytes`, `http`, `uuid`)
- `0.x` crates -> `<crate>0<minor>` (`jiff02`)
- future major N -> `<crate>N` (additive, e.g. `bytes2`)

The convention is documented as a comment in the crate's `Cargo.toml`.

crates/thread_aware/Cargo.toml
```diff
@@ -27,17 +51,70 @@ default = ["derive"]
...
+#
+# A future breaking version of any of these crates can therefore be supported
+# additively (e.g. add a new `bytes2` feature alongside `bytes`) without forcing
+# a breaking release of this crate.
+#
+# See the `third_party` module for the list of types each feature covers.
+bytes = ["dep:bytes"]
+http = ["dep:http"]
+jiff02 = ["dep:jiff"]
+uuid = ["dep:uuid"]
```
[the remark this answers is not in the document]
martin-kolinek: GHCP: Updated the crate-level docs in ad7a0f90 to spell out the full rule (bare name for stable 1.x, `<crate><N>` for future major N>=2, `<crate>0<minor>` for 0.x). Also updated the PR description to use the actual feature names (`bytes`, `http`, `jiff02`, `uuid`).

crates/thread_aware/src/lib.rs
```diff
@@ -92,6 +92,21 @@
 //!   environments: `default-features = false`.
 //! * **`threads`**: Enables features mainly used by async runtimes for OS interactions.
 //!
+//! ## 3rd-party crate impls
+//!
+//! The following opt-in features provide [`ThreadAware`] implementations for
+//! inert value types from popular 3rd-party crates. Each feature is suffixed
+//! with the major (or 0.x minor) of the wrapped crate; enabling a feature
+//! pulls that crate in as a dependency. By default none are enabled and this
+//! crate brings in no extra dependencies.
```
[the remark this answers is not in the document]
martin-kolinek: GHCP: Updated the crate-level docs in ad7a0f90 to spell out the full rule (bare name for stable 1.x, `<crate><N>` for future major N>=2, `<crate>0<minor>` for 0.x). Also updated the PR description to use the actual feature names (`bytes`, `http`, `jiff02`, `uuid`).

ralfbiedert (approved):

crates/thread_aware/src/third_party/http.rs
```diff
@@ -0,0 +1,123 @@
...
+//! is itself no-op — iterating would be pure waste. See the per-impl docs
+//! on [`Request<T>`] and [`Response<T>`] for their `relocate` semantics.
+
+use ::http::header::{HeaderMap, HeaderName, HeaderValue};
+use ::http::{Method, Request, Response, StatusCode, Version};
+
+use crate::ThreadAware;
+use crate::affinity::Affinity;
+
+impl_noop_thread_aware!(StatusCode, Version, Method, HeaderName, HeaderValue, HeaderMap<HeaderValue>);
```
martintmk: imho, most of the types from http crate should be ThreadAware:

- Authority
- Port
- Scheme
- Uri
- PathAndQuery
- Error
- InvalidUri

martin-kolinek: GHCP: Added in 958f609b. The new no-op impls cover `Uri`, `uri::Authority`, `uri::Scheme`, `uri::PathAndQuery`, `uri::Port<T>` (with a `T: Send` bound so `ThreadAware`'s `Send` supertrait is satisfied for any representation), plus `http::Error` and `uri::InvalidUri`. Updated the crate docs, allowed_external_types, regenerated the README, and removed `http::Uri` from the PR description's "intentional omissions" list.

crates/thread_aware/src/third_party/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+//! `any(test, feature = "...")`.
+//!
+//! See this crate's `Cargo.toml` for the exact versions used.
+
+/// Generates a no-op [`ThreadAware`](crate::ThreadAware) impl for each listed type.
+///
+/// The bodies of the implementations are empty because the listed types are
+/// inert value types: they hold no thread-local state, perform no I/O, and
+/// participate in no cross-thread sharing that would benefit from relocation.
+#[cfg(any(test, feature = "bytes", feature = "http", feature = "jiff02", feature = "uuid",))]
```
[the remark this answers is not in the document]
martin-kolinek: Doesn't seem to fail to compile

ralfbiedert (approved):

PR #480, opened 2026-06-05
perf(layered): dynamic service does not allocate anymore

martintmk:
I got quite surprised how much allocations are happening in fetch with standard pipeline. This change leverages `BlindPool` that allows memory and future pooling, thus eliminating the memory allocations.

Before:

```
Allocation statistics:

| Operation                | Mean bytes | Mean count |
|--------------------------|------------|------------|
| custom_minimal_pipeline  |       3409 |         15 |
| custom_standard_pipeline |      11600 |         29 |
| minimal_pipeline         |       2721 |         14 |
| standard_pipeline        |      13667 |         31 |
```

After:

```
Allocation statistics:

| Operation                | Mean bytes | Mean count |
|--------------------------|------------|------------|
| custom_minimal_pipeline  |       2713 |         13 |
| custom_standard_pipeline |       3744 |         27 |
| minimal_pipeline         |       2713 |         13 |
| standard_pipeline        |       3771 |         29 |
```

no changes in public API


Vaiz:
considering that `infinity_pool` never releases memory, we should provide an ability to use user provided instance of a pool

@sandersaares, do you have any recommendations here?

martintmk:
Also worried about potential unsafety this might bring. How risky is this @sandersaares?

sandersaares:
> considering that infinity_pool never releases memory, we should provide an ability to use user provided instance of a pool

There is a [`shrink_to_fit()`](https://docs.rs/infinity_pool/latest/infinity_pool/struct.BlindPool.html#method.shrink_to_fit) if you want to release memory. It is possible some more advanced capacity-release policies could be implemented but what they are is not obvious to me.

When do we even need to release capacity? There are two reasons that immediately come to mind:

1. When load patterns change (e.g. less load during the night). This may benefit from some "shrink once every 30 minutes" logic (maybe). On the other hand, if load patterns change, will they also change for other services on the same PC or will we just complicate logic by releasing memory that other apps also do not need? In general, we should aim for a steady state, not ups-and-downs, even if load patterns change.
2. After a sudden load spike. However, load spikes should be mitigated by rate limiting and policy systems in the request path, so the actual complexity/value tradeoff here is not obvious to me.

It is an area that could benefit from exploration and data from real customer scenarios. It is not obvious to me that shrinking resource pools is at all necessary/appropriate.

> Also worried about potential unsafety this might bring. How risky is this @sandersaares?

There are no open bugs that I am aware of. More risky than safe Rust, less risky than skydiving. `infinity_pool` is heavily used in `oxidizer_executor` and `oxidizer_io`, for some reference point.

crates/layered/src/dynamic.rs
```diff
@@ -51,22 +54,52 @@ where
...
         T: Service<In, Out = Out> + Send + Sync + 'static,
     {
-        Self(DynService::new_arc(strategy))
+        let pool = BlindPool::new();
+        let service = Arc::new(strategy);
+        // define a delegate that wraps the service execution in a
+        // future and inserts it into the pool
+        let exec = move |input: In| {
+            let cloned = Arc::clone(&service);
+            let fut = async move { cloned.execute(input).await };
```
sandersaares, line 76: The generics are blinding my eyes today, so I cannot identify it myself but is it possible here to get a type for the thing being inserted, or at least a `Layout`? If so, switching to `OpaquePool` would eliminate some overhead. `OpaquePool` vs `BlindPool` is basically "one type/layout" versus "multiple types/layouts", with the `BlindPool` acting as a router over a pool of `OpaquePool` instances.

Vaiz:
> > considering that infinity_pool never releases memory, we should provide an ability to use user provided instance of a pool
> 
> There is a [`shrink_to_fit()`](https://docs.rs/infinity_pool/latest/infinity_pool/struct.BlindPool.html#method.shrink_to_fit) if you want to release memory. It is possible some more advanced capacity-release policies could be implemented but what they are is not obvious to me.

is it better to have one shared pool for all instances of `DynamicService`, or one pool per instance is fine?

sandersaares (approved):
How is the change in terms of processor time spent?

sandersaares:
> is it better to have one shared pool for all instances of DynamicService, or one pool per instance is fine?

Unclear - I do not have a good understanding of the actual usage patterns. Sharing a single pool across many callers improves resource utilization, of course.

What I have sometimes done is to have a `thread_local! { static POOL: BlindPool = BlindPool::new() }` so resources are shared but lock contention is kept minimal since accesses are always from the same thread (unless pooled objects are sent to other threads, of course, which is valid).

crates/layered/src/dynamic.rs
martintmk: I tried, but there is no way to extract the generic type definition. I tried some workarounds around using `OpaquePool`, but it was more verbose and it did not yield worthy perf differences.

martintmk:
> > is it better to have one shared pool for all instances of DynamicService, or one pool per instance is fine?
> 
> Unclear - I do not have a good understanding of the actual usage patterns. Sharing a single pool across many callers improves resource utilization, of course.
> 
> What I have sometimes done is to have a `thread_local! { static POOL: BlindPool = BlindPool::new() }` so resources are shared but lock contention is kept minimal since accesses are always from the same thread (unless pooled objects are sent to other threads, of course, which is valid).

At this point, it's better to not share the memory across many different dynamic service types. Currently, these services are long-lived and it's better that the memory they are using is isolated only to them. 

Later, if we find unnecessary memory consumption we can refine the behavior, or even reclaim memory periodically. (inside dynamic service)

ralfbiedert (approved):

PR #483, opened 2026-06-08
perf(fetch): reduce future sizes (up to 7x less)

martintmk:
Before:

```
running 1 test
fetch future size: 536 bytes
fetch_buffered future size: 2240 bytes
fetch_text future size: 3552 bytes
fetch_bytes future size: 2672 bytes
fetch_json future size: 3104 bytes
fetch_json_owned future size: 3536 bytes
test http_request_builder::tests::fetch_future_sizes ... ok
```

After:
```
running 1 test
fetch future size: 392 bytes
fetch_buffered future size: 400 bytes
fetch_text future size: 400 bytes
fetch_bytes future size: 400 bytes
fetch_json future size: 400 bytes
fetch_json_owned future size: 400 bytes
test http_request_builder::tests::fetch_future_sizes ... ok
```

### DIFF
--- a/crates/http_extensions/src/body/builder.rs
+++ b/crates/http_extensions/src/body/builder.rs
@@ -9,9 +9,9 @@ use http_body_util::BodyExt;
 use thread_aware::{ThreadAware, Unaware};
 use tick::Clock;
 
+use super::HttpBody;
 use super::options::HttpBodyOptions;
 use super::timeout_body::TimeoutBody;
-use super::{HttpBody, Kind};
 #[cfg(any(feature = "json", test))]
 use crate::json::JsonError;
 use crate::{HttpError, Result};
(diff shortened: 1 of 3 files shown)


sandersaares (approved):
This is purely "working around Rust's braindead `async fn` generators", right? No functional differences, no downside to the tradeoff?

I suggest also extracting the technique as an AI skill, so it can be applied automatically by the AI when generating such "simple" async functions. Happy to collaborate on that with you if you want to go down that path.

martintmk:
>This is purely "working around Rust's braindead async fn generators", right? No functional differences, no downside to the tradeoff?

True, the resulting behavior is the same. Messaged you offline, this could be indeed used as a stand-alone skill.

ralfbiedert (approved):

ralfbiedert:
> so it can be applied automatically by the AI when generating such "simple" async functions. 

I'd be reluctant using this trick too often, since it is changing (public) APIs to work around a compiler issue. I think it's fair game when there's a stack overflow or so and a function is particularly hot, but I fear that if this becomes a "best practice" all service API docs will look like crap.

sandersaares:
> service API docs will look like crap.

@ralfbiedert do you refer here to `async fn` becoming `impl Future<..>` or does it have some other damage that I am unaware of at first glance?

martintmk:
FYI, tried to unwrap the actual future into async block and it did still increase the future size 3x :/

```rust
    pub async fn fetch_buffered(self) -> Result<HttpResponse> {
        self.fetch()
            .and_then(|response| {
                let (parts, body) = response.into_parts();
                body.into_buffered().map_ok(move |body| HttpResponse::from_parts(parts, body))
            })
            .await
    }
```

(such a waste)

Vaiz (approved):

PR #486, opened 2026-06-08
docs(data_privacy): document derive macros

Vaiz:


martintmk (approved):

sandersaares (approved):

ralfbiedert (approved):

crates/data_privacy/src/lib.rs
```diff
@@ -185,9 +185,158 @@ extern crate self as data_privacy;
...
+/// [`classified`](macro@classified) attribute implement it automatically, so they can be nested
+/// directly inside a type that derives `RedactedDebug`.
+///
+/// # The `#[unredacted]` attribute
+///
+/// Mark a field with `#[unredacted]` to opt it out of redaction. Such a field is formatted with its
+/// regular [`Debug`](core::fmt::Debug) implementation, exactly as `#[derive(Debug)]` would do. Use
+/// this only for fields that are known to never carry sensitive data.
+///
+/// # Generated output
```
ralfbiedert: I would not document this, because it's not guaranteed (and if ever a nicer formatting helper is included in Rust we'd want to change to that). Id' either remove this (preferred), or add a note that this is just illustrative.

crates/data_privacy/src/lib.rs
```diff
@@ -185,9 +185,158 @@ extern crate self as data_privacy;
...
+/// example, an `#[unredacted]` string is printed without surrounding quotes here, unlike with
+/// [`RedactedDebug`](derive@RedactedDebug)).
+///
+/// # The `#[unredacted]` attribute
+///
+/// Mark a field with `#[unredacted]` to opt it out of redaction. Such a field is formatted with its
+/// regular [`Display`](core::fmt::Display) implementation. Use this only for fields that are known
+/// to never carry sensitive data.
+///
+/// # Generated output
```
ralfbiedert: Same

crates/data_privacy/src/lib.rs
```diff
@@ -185,9 +185,158 @@ extern crate self as data_privacy;
...
+///
+/// The shape of the output depends on the kind of struct:
+///
+/// - Named structs render as `Name { field: value, ... }`.
+/// - Tuple structs render as `Name(value, ...)`.
+/// - Unit structs render as `Name`.
+///
+/// Deriving on an enum or union is not supported and produces a compile-time error.
```
ralfbiedert: I also wouldn't print this. IIRC this is just a lack of support, but nothing we really want to document like this (it reads a bit as if this is a conscious choice of ours)

crates/data_privacy/src/lib.rs
Vaiz: done

crates/data_privacy/src/lib.rs
Vaiz: done

crates/data_privacy/src/lib.rs
Vaiz: I've removed it for now, but it still is a limitation that needs to be addressed in some way. Either we document it, or we implement missing functionality

ralfbiedert: We should implement this, not having this is just an oversight.

ralfbiedert (approved):

PR #489, opened 2026-06-11
feat(templated_uri): implement RedactedDisplay/RedactedDebug for BasePath and BaseUri

Vaiz:
Implement `RedactedDisplay` and `RedactedDebug` for both types. Consistent with the existing Uri impl (which renders its base components via Display, unredacted, and only routes the appended path-and-query through the Redactor), base URIs and base paths are treated as static configuration rather than request data: their redacted output matches their plain Display/Debug output.

### DIFF
--- a/crates/templated_uri/src/base_path.rs
+++ b/crates/templated_uri/src/base_path.rs
@@ -1,9 +1,10 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-use std::fmt::Display;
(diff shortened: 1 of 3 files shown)


ralfbiedert (approved):

martintmk:
Can we do the same for Origin?

kate-shine (approved):

sgalkin (approved):

Vaiz:
> Can we do the same for Origin?

done

sandersaares (approved):

PR #493, opened 2026-06-12
feat(workspace): enable and enforce unreachable_pub lint

martintmk:
## What

Enables and enforces the `unreachable_pub` rustc lint across the workspace, then fixes every offender.

- Adds `unreachable_pub = "warn"` to `[workspace.lints.rust]` in the root `Cargo.toml` (alphabetically placed between `trivial_numeric_casts` and `unsafe_op_in_unsafe_fn`). Member crates already inherit workspace lints via `[lints] workspace = true`.
- Resolves all offenders surfaced by `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` (the way CI runs clippy).

## How offenders were resolved (least-privilege)

- **Hand-written items** (fns, methods, structs, enums, traits, type aliases, `pub use` re-exports) across ~20 crates: `pub` -> `pub(crate)`. Applied mechanically with `cargo clippy --fix`; the few cases clippy emitted as `pub(super)` were normalized to `pub(crate)` for uniformity.
- **`cachet_tier::CacheTier`**: the trait is genuinely public (re-exported at the crate root via `pub use tier::CacheTier`), but the `#[dynosaur::dynosaur]` attribute macro makes `unreachable_pub` misfire on the generated definition. Kept `pub` and suppressed with an item-scoped `#[allow(clippy::allow_attributes, unreachable_pub, ...)]` on the trait (an item-level `#[expect]` is reported unfulfilled across the macro expansion, and the `clippy::allow_attributes` self-allow matches the existing pattern in `thread_aware/src/registry.rs`).
- **fundle bundle test component structs** (`Instance` / `Device` / `Vulkan`): must stay `pub` because `#[fundle::bundle]` generates a public interface over them (downgrading to `pub(crate)` yields `E0446`). Suppressed with a single module-level `#![expect(unreachable_pub, ...)]` per test file.

No `pub` struct fields were touched (`unreachable_pub` does not flag fields). There are no `include!`-generated sources or cfg-gated platform-placeholder modules among the offenders, so those suppression categories did not apply. All changes are visibility-only with no behavioral changes.

## Verification

- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` -- clean (exit 0).
- `cargo build --workspace --all-features --all-targets --locked` -- clean.
- `cargo test --workspace --all-features --locked` -- all pass, 0 failures.
- `cargo doc --no-deps --workspace --all-features` with `RUSTDOCFLAGS="-D warnings"` -- clean.
- `rustfmt --check` (stable, `max_width=140`) on all changed files -- clean.

Note: `just format-check` and the `--cfg docsrs` docs build use the pinned nightly toolchain (`nightly-2026-05-30`), which is validated in CI. The changes are import-free and do not touch doc code blocks, so the stable rustfmt check is a faithful proxy for the formatting gate.


sandersaares (approved):

crates/cachet_tier/src/tier.rs
```diff
@@ -7,6 +7,11 @@
 //! This trait is designed for composition: implement the storage operations,
 //! then use `cachet` to layer on telemetry, TTL, and multi-tier fallback.
 
+#![expect(
+    unreachable_pub,
+    reason = "CacheTier is re-exported at the crate root via `pub use tier::CacheTier`; the dynosaur attribute macro causes unreachable_pub to misfire on the generated definition"
```
Vaiz: if it's only one type, the `except` statement should not be global (with `!`)

crates/fundle/tests/bundle_all_features.rs
```diff
@@ -19,10 +19,22 @@ pub struct Telemetry {}
 
 mod gpu {
     #[derive(Clone, Default)]
+    #[expect(
```
Vaiz: for tests, it's ok to have a global `allow` to reduce boilerplate

Vaiz (approved):

crates/cachet_tier/src/tier.rs
martintmk: 🤖 Good call -- switched the crate-wide `#![expect(...)]` to an item-level attribute on the trait.

One nuance worth recording: an item-level `#[expect(unreachable_pub)]` is reported as *unfulfilled* here, because the `#[dynosaur::dynosaur]` attribute macro does not attribute the diagnostic to the trait's own HIR node. So I used an item-scoped `#[allow(clippy::allow_attributes, unreachable_pub, ...)]` instead -- the `clippy::allow_attributes` self-allow mirrors the existing pattern in `thread_aware/src/registry.rs`. The suppression is now scoped to the single trait rather than the whole module.

---

crates/fundle/tests/bundle_all_features.rs
martintmk: 🤖 Done -- replaced the three per-struct attributes with a single module-level `#![expect(unreachable_pub, ...)]` in both `bundle_getters.rs` and `bundle_all_features.rs`.

I used `#![expect]` rather than `#![allow]` so it stays consistent with the workspace `clippy::allow_attributes` policy (which prefers `expect`), but it is still the single global suppression you suggested.

---

Vaiz (approved):

PR #494, opened 2026-06-12
feat(fetch_azure): adapt fetch HttpClient to Azure's HttpClient

martintmk:
## Summary

Adds two crates that adapt Oxidizer building blocks to the Azure SDK abstractions in [`azure_core` 1.0.0](https://docs.rs/azure_core/1.0.0/azure_core/):

- **`fetch_azure`** — `HttpClient` implements `azure_core::http::HttpClient` on top of a `fetch::HttpClient` transport.
- **`anyspawn_azure`** — `Runtime` implements `azure_core::async_runtime::AsyncRuntime` on top of an `anyspawn::Spawner` (spawning) and a `tick::Clock` (sleeping). With the optional `azure-identity` feature it also implements `azure_identity::Executor`, running developer-credential commands on the spawner.

`fetch` is depended on without enabling any of its features.

## Details

- `HttpClient` converts an Azure SDK `Request` (method, URI, headers, and bytes or seekable-stream body) into a `fetch` request, executes it, and maps the `fetch` response back into an `AsyncRawResponse` with a streamed body. Empty bodies and byte bodies avoid copies; non-UTF-8 response headers are skipped (matching the built-in `reqwest` transport). Build errors map to `DataConversion`; transport errors to `Io`.
- Conversions: `From<fetch::HttpClient>` and `From<HttpClient>` into `Arc<dyn HttpClient>`; `From<Runtime>` into `Arc<dyn AsyncRuntime>` (and, under `azure-identity`, `Arc<dyn Executor>`).
- `Runtime` spawns tasks wrapped in `futures::future::Abortable` so cancellation wakes pending waiters; `sleep` uses `tick::Clock::delay`.
- Example: `blob_list` lists blobs in a Storage container using `fetch` as the transport and an `anyspawn_azure` runtime as the credential executor (on tokio).
- Integration tests + doctests; `client.rs` and `runtime.rs` are each at 100% line/region coverage.


crates/fetch_azure/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+/// An [`AsyncRuntime`] that spawns work on an [`anyspawn::Spawner`] and sleeps
+/// on a [`tick::Clock`].
+///
+/// Construct one from an existing [`Spawner`] and [`Clock`] with
+/// [`SpawnerRuntime::new`] (or via [`From`]) and install it as the Azure SDK
+/// runtime with [`azure_core::async_runtime::set_async_runtime`]. See
+/// [`new_async_runtime`] for a convenience that returns an
+/// `Arc<dyn AsyncRuntime>` directly.
+#[derive(Debug, Clone)]
+pub struct SpawnerRuntime {
```
martintmk: move to a separate module, rename to Runtime

crates/fetch_azure/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+use tick::Clock;
+
+/// An [`HttpClient`] that uses a [`fetch::HttpClient`] as its transport.
+///
+/// Construct one from an existing `fetch` client with [`AzureHttpClient::new`]
+/// (or via [`From`]) and pass it to the Azure SDK wherever a
+/// `dyn HttpClient` is expected. See [`new_http_client`] for a convenience that
+/// returns an `Arc<dyn HttpClient>` directly.
+#[derive(Debug, Clone)]
+pub struct AzureHttpClient {
```
martintmk: move to a separate client module

crates/fetch_azure/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+struct SpawnerTask {
+    handle: JoinHandle<()>,
+    aborted: AtomicBool,
+}
+
+impl SpawnerTask {
+    fn new(handle: JoinHandle<()>) -> Self {
+        Self {
+            handle,
+            aborted: AtomicBool::new(false),
```
martintmk: this canellation is wrong as pending waiters are not informated about the abortion, use cancellation feature from futures crate to implement proper cancellation

crates/fetch_azure/src/lib.rs
martintmk: @-

crates/fetch_azure/src/lib.rs
martintmk: @-

crates/fetch_azure/src/lib.rs
martintmk: @-

crates/fetch_azure/src/lib.rs
martintmk: @-

crates/fetch_azure/src/lib.rs
martintmk: @-

crates/fetch_azure/src/lib.rs
martintmk: @-

crates/fetch_azure/src/client.rs
```diff
@@ -0,0 +1,157 @@
...
+impl AzureHttpClient {
+    /// Creates a new adapter that forwards requests to the given `fetch` client.
+    #[must_use]
+    pub const fn new(client: fetch::HttpClient) -> Self {
+        Self { client }
+    }
+
+    /// Returns a reference to the wrapped [`fetch::HttpClient`].
+    #[must_use]
+    pub const fn inner(&self) -> &fetch::HttpClient {
```
martintmk: no need to expose this one

crates/fetch_azure/src/client.rs
```diff
@@ -0,0 +1,157 @@
...
+        &self.client
+    }
+
+    /// Consumes the adapter and returns the wrapped [`fetch::HttpClient`].
+    #[must_use]
+    pub fn into_inner(self) -> fetch::HttpClient {
```
martintmk: also drop this one

crates/fetch_azure/src/runtime.rs
```diff
@@ -0,0 +1,119 @@
...
+        Self::new(spawner, clock)
+    }
+}
+
+/// Wraps an [`anyspawn::Spawner`] and [`tick::Clock`] as an `Arc<dyn AsyncRuntime>`.
+///
+/// This is a convenience for installing a `fetch`-friendly runtime with
+/// [`azure_core::async_runtime::set_async_runtime`].
+#[must_use]
+pub fn new_async_runtime(spawner: Spawner, clock: Clock) -> Arc<dyn AsyncRuntime> {
```
martintmk: drop this method

crates/fetch_azure/tests/client.rs
```diff
@@ -0,0 +1,330 @@
+// Copyright (c) Microsoft Corporation.
```
martintmk, line 1: also split these tests into client and runtime related modules

crates/fetch_azure/src/client.rs
martintmk: @-

crates/fetch_azure/src/client.rs
martintmk: @-

crates/fetch_azure/src/runtime.rs
martintmk: @-

crates/fetch_azure/tests/client.rs
martintmk: @-

crates/fetch_azure/Cargo.toml
```diff
@@ -0,0 +1,59 @@
...
+
+[dependencies]
+# internal
+bytesbuf = { workspace = true, features = ["bytes-compat"] }
+fetch = { workspace = true }
+layered = { workspace = true }
+
+# external
+async-trait = { workspace = true }
+azure_core = { workspace = true }
```
martintmk, line 39: can we only depend on typespec_client_core?

martintmk: &#x1F916; Agreed - `azure_core` only re-exports the `HttpClient` / `AsyncRuntime` traits from `typespec_client_core` here (the `allowed_external_types` list already targets `typespec_client_core`). I'll depend on `typespec_client_core` directly for both crate libraries, keeping `azure_core` only as a dev-dependency for the blob example. Implementing now.

---

martintmk: &#x1F916; Fixed in 54223e0: both crate libraries now depend on `typespec_client_core` directly (`fetch_azure` enables its `http` feature), and the source/tests use `typespec_client_core::{http, async_runtime, error, stream, time}` instead of the `azure_core` re-exports. `azure_core` is retained only as a `fetch_azure` dev-dependency for the `blob_list` example, which needs the Azure-specific `ClientOptions`/`Transport`/credentials. Coverage stays at 100% and the docs build cleanly.

---

crates/fetch_azure/src/client.rs
```diff
@@ -0,0 +1,149 @@
...
+/// ```
+/// # use std::sync::Arc;
+/// # use azure_core::http::HttpClient;
+/// # use fetch_azure::AzureHttpClient;
+/// # fn wrap(client: fetch::HttpClient) -> Arc<dyn HttpClient> {
+/// AzureHttpClient::from(client).into()
+/// # }
+/// ```
+#[derive(Debug, Clone)]
+pub struct AzureHttpClient {
```
martintmk: rename to just `HttpClient`

crates/anyspawn_azure/src/runtime.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    /// Returns a reference to the wrapped [`Clock`].
+    #[must_use]
+    pub const fn clock(&self) -> &Clock {
+        &self.clock
+    }
+}
+
+impl From<Runtime> for Arc<dyn AsyncRuntime> {
```
martintmk, line 48: add similar conversion for azure_identity::Executor

crates/fetch_azure/src/client.rs
martintmk: &#x1F916; Agreed - renaming `AzureHttpClient` to `HttpClient` (so it reads as `fetch_azure::HttpClient`, mirroring `anyspawn_azure::Runtime`). Since the struct then shares its name with the `typespec_client_core::http::HttpClient` trait it implements, I'll alias the trait where it must be named. Implementing now.

---

martintmk: &#x1F916; Fixed in d3ff6f6: the transport adapter is now `fetch_azure::HttpClient`. Because the struct then shares its name with the `typespec_client_core::http::HttpClient` trait it implements, the trait is imported under the `HttpClientTrait` alias at the impl site and in the `Arc<dyn ...>` conversion. Docs, examples, and tests were updated; coverage stays at 100% and docs build cleanly.

---

crates/anyspawn_azure/src/runtime.rs
martintmk: &#x1F916; Agreed - I'll add a feature-gated `From<Runtime> for Arc<dyn azure_identity::Executor>` mirroring the existing `From<Runtime> for Arc<dyn AsyncRuntime>`, plus a test so the conversion stays covered. Implementing now.

---

martintmk: &#x1F916; Fixed in 4412e9e: added a feature-gated `From<Runtime> for Arc<dyn azure_identity::Executor>` next to the `AsyncRuntime` conversion, with a test that converts and runs a command through the boxed executor (coverage stays at 100%). I also had to enable the `http` feature on this crate's `typespec_client_core` dependency - its always-compiled `stream` module imports the http-gated `crate::http`, so the crate doesn't build without it.

---

crates/fetch_azure/src/lib.rs
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#![cfg_attr(all(coverage_nightly, test), feature(coverage_attribute))]
+#![cfg_attr(docsrs, feature(doc_cfg))]
```
martintmk, line 5: can you add this same line to fetch crate too?

#![cfg_attr(docsrs, feature(doc_cfg))]

(unrelated, but let's fix it here)

martintmk: &#x1F916; Sure - adding `#![cfg_attr(docsrs, feature(doc_cfg))]` to `crates/fetch/src/lib.rs` as well. Implementing now.

---

martintmk: &#x1F916; Done in 60af619: added `#![cfg_attr(docsrs, feature(doc_cfg))]` to `crates/fetch/src/lib.rs`. It is gated on the `docsrs` cfg, so it stays inert on stable builds; the crate documents cleanly under the docs.rs nightly config (`--cfg docsrs -D warnings`).

---

crates/fetch_azure/Cargo.toml
```diff
@@ -0,0 +1,59 @@
...
+# internal
+bytesbuf = { workspace = true, features = ["bytes-compat"] }
+fetch = { workspace = true }
+layered = { workspace = true }
+
+# external
+async-trait = { workspace = true }
+futures = { workspace = true, features = ["std"] }
+http = { workspace = true }
+typespec_client_core = { workspace = true, features = ["http"] }
```
heaths: Use `azure_core`. Forget `typespec_client_core` and below exist. Everything is re-exported from `azure_core`. It's an initiative we had to support but likely won't happen.

crates/fetch_azure/README.md
```diff
@@ -0,0 +1,50 @@
...
+[![Coverage](https://codecov.io/gh/microsoft/oxidizer/graph/badge.svg?token=FCUG0EL5TI)](https://codecov.io/gh/microsoft/oxidizer)
+[![License](https://img.shields.io/badge/license-MIT-blue.svg)](../../LICENSE)
+<a href="../.."><img src="../../logo.svg" alt="This crate was developed as part of the Oxidizer project" width="20"></a>
+
+</div>
+
+Adapt a [`fetch::HttpClient`][__link0] into an Azure SDK HTTP transport.
+
+The Azure SDK abstracts its HTTP transport behind the
+[`typespec_client_core::http::HttpClient`][__link1] trait. [`HttpClient`][__link2] implements that
```
heaths: `azure_core`. We don't talk about `typespec*` crates.

crates/fetch_azure/README.md
```diff
@@ -0,0 +1,50 @@
...
+trait on top of a [`fetch::HttpClient`][__link3], so Azure SDK pipelines run over
```
heaths: I know we devs are bad at naming variables, but is `__linkN` really the best convention? Why not something more descriptive that is easier to associate should you need to fix something. And if you only refer to them once, why even have links. In fact, `__link0` an `__link3` are dups.

crates/fetch_azure/Cargo.toml
```diff
@@ -0,0 +1,59 @@
...
+license = { workspace = true }
+homepage = { workspace = true }
+repository = "https://github.com/microsoft/oxidizer/tree/main/crates/fetch_azure"
+
+[package.metadata.cargo_check_external_types]
+allowed_external_types = [
+    # Workspace sibling crates
+    "fetch::*",
+    # External dependency that defines the HttpClient trait
+    "typespec_client_core::*",
```
heaths, line 25: Last time I'll mention it, but `azure_core`. That depends on - and re-exports all of - `typespec_client_core` and `typespec` so maybe they are still needed here as transitive dependencies, but `typespec*` should not appear in any docs or `Cargo.toml` files.

crates/fetch_azure/logo.png
heaths, line 1: You put your .ico in LFS but not you .png?

crates/fetch_azure/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+//! use std::sync::Arc;
+//!
+//! use fetch::HttpClient as FetchClient;
+//! use fetch_azure::HttpClient;
+//!
+//! // Adapt a `fetch` client into an Azure SDK transport.
+//! fn transport(client: FetchClient) -> Arc<dyn typespec_client_core::http::HttpClient> {
+//!     HttpClient::from(client).into()
+//! }
+//! # let _ = transport;
```
heaths: Probably should do something with it. This sample isn't inherently useful.

crates/fetch_azure/src/client.rs
```diff
@@ -0,0 +1,148 @@
...
+/// A [`typespec_client_core::http::HttpClient`] backed by a [`fetch::HttpClient`] transport.
+///
+/// Construct one from an existing `fetch` client with [`HttpClient::new`]
+/// (or via [`From`]), then convert it into an `Arc<dyn HttpClient>` via [`From`]
+/// / [`Into`] to hand to the Azure SDK:
+///
+/// ```
+/// # use std::sync::Arc;
+/// # use fetch_azure::HttpClient;
+/// # fn wrap(client: fetch::HttpClient) -> Arc<dyn typespec_client_core::http::HttpClient> {
```
heaths: Why does your elided sample func return an `Arc`? Seems unnecessary.

That said, elsewhere, you should state the reason the Azure SDK wants an `Arc` is to promote sharing. Same is true of `HttpClient`, `Policy`, and `TokenCredential` traits. Our pattern of having their concrete implementation constructors return an `Arc` isn't bad Rust code. It's intentional to promote sharing. Most `TokenCredential`s, for example, share an endpoint/token cache and reduce network calls. Same is true across all the Azure SDK languages - but all of them (besides C++) are ref-counted so get it for free.

heaths (changes requested):
The only place I should see "typespec*" is in `Cargo.lock`.

crates/fetch_azure/Cargo.toml
martintmk: &#x1F916; Agreed - reverting both crates to depend on `azure_core` and removing all `typespec*` mentions from docs and Cargo.toml. (The earlier switch to `typespec_client_core` was made per a different review comment; following your guidance now.) Implementing across both crates.
---

crates/fetch_azure/examples/azure_transport.rs
martintmk, line 1: drop this example, it has very little value

crates/fetch_azure/src/lib.rs
martintmk: let's improve it, the sample should show actual integration with APIs such as blovs, use blob_list example for reference

crates/fetch_azure/README.md
martintmk: We auto-generate the README.md from the lib.rs files. These links are just a way the generator does it. It will never be edited manually.

crates/fetch_azure/Cargo.toml
martintmk: &#x1F916; Fixed in 2fcf8ae: both crate libraries depend on `azure_core` again, and `typespec*` no longer appears in `[dependencies]`, the workspace manifest, or any docs/CHANGELOG. (The only remaining mention is the `allowed_external_types` linter directive - see the thread on line 24.)
---

crates/fetch_azure/README.md
martintmk: &#x1F916; Fixed in 2fcf8ae: the README is generated from the crate doc, which now references `azure_core`; regenerated so the rendered README no longer mentions `typespec*`.
---

crates/fetch_azure/README.md
martintmk: &#x1F916; The `__linkN` references are generated by `cargo-doc2readme` (the repo's README generator) from intra-doc links - that naming is the tool's own and is shared across every crate README, so I can't rename them just here. What I could fix is the count and the duplicates: the crate doc now produces only 3 unique links with no dups (the previous `__link0`/`__link3` duplicate of `fetch::HttpClient` is gone).
---

crates/fetch_azure/Cargo.toml
martintmk: &#x1F916; Fixed in 2fcf8ae: removed `typespec*` from `[dependencies]`, the workspace manifest, and all docs/CHANGELOG. The one remaining mention is the `allowed_external_types` entry: `cargo-check-external-types` reports re-exported types by their *defining* crate, so it sees `azure_core`'s `HttpClient`/`AsyncRuntime` as `typespec_client_core`. Listing `typespec_client_core::*` there is required for the `external-type-exposure` check to pass - it's a linter directive identifying where the exposed traits originate, not a dependency. Happy to revisit if you'd prefer a different approach.
---

crates/fetch_azure/logo.png
martintmk: &#x1F916; Both `logo.png` and `favicon.ico` are tracked in Git LFS - I verified the committed blobs are LFS pointer files (their content starts with `version https://git-lfs.github.com/spec/v1`). GitHub renders the LFS-backed PNG inline in the diff (it can display the image), which makes it look like a regular file, whereas the `.ico` shows as a binary/LFS entry. So they are consistent; no change needed.
---

crates/fetch_azure/src/lib.rs
martintmk: &#x1F916; Fixed in 2fcf8ae: the crate-doc sample now wires the adapter in as the transport for an Azure SDK client - `Transport::new(...)` plus `ClientOptions { transport: Some(...), ..Default::default() }` - which is the actual integration point, rather than just constructing and discarding a value.
---

crates/fetch_azure/src/client.rs
martintmk: &#x1F916; Fixed in 2fcf8ae. The sample no longer returns an `Arc` from a wrapper fn; it binds `let transport: Arc<dyn HttpClient> = HttpClient::from(client).into();`. I also documented the rationale on the type: the Azure SDK shares its transport as an `Arc` to promote reuse - one client (and its connection pool) is shared across the request pipeline rather than cloned per call - and noted the same intent underlies the `Policy` and `TokenCredential` traits.
---

crates/fetch_azure/examples/azure_transport.rs
martintmk: &#x1F916; Agreed - dropping `azure_transport.rs` (it overlaps the crate-doc example and the `blob_list` sample is the substantive one). Removing it now.
---

martintmk: &#x1F916; Done - removed `examples/azure_transport.rs`. The `blob_list` example remains as the end-to-end sample.
---

crates/anyspawn_azure/src/runtime.rs
```diff
@@ -0,0 +1,134 @@
...
+            // `time::Duration` can be negative; clamp such values to zero.
+            let duration = std::time::Duration::try_from(duration).unwrap_or_default();
+            clock.delay(duration).await;
+        })
+    }
+
+    fn yield_now(&self) -> TaskFuture {
+        std::thread::yield_now();
+        Box::pin(ready(()))
+    }
```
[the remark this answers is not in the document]
martintmk: &#x1F916; Good catch - fixed in c852204. `yield_now` now returns a `YieldNow` future that yields once (`Poll::Pending` plus a self-wake) before completing, instead of calling `std::thread::yield_now()` (which yields the OS thread) and returning a ready future. The awaiting task now actually yields to the executor.
---

crates/fetch_azure/src/client.rs
```diff
@@ -0,0 +1,154 @@
...
+        .filter_map(|(name, value)| {
+            value
+                .to_str()
+                .ok()
+                .map(|value| (HeaderName::from(name.as_str().to_owned()), HeaderValue::from(value.to_owned())))
+        })
+        .collect::<HashMap<_, _>>();
+
+    Headers::from(headers)
+}
```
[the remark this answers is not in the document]
martintmk: &#x1F916; I dug into this and don't think it's fixable at the adapter layer. `azure_core::http::headers::Headers` is defined as `pub struct Headers(HashMap<HeaderName, HeaderValue>)` - a single value per name, exposing `insert` (which overwrites) and `add`, but no `append`. Multi-value headers like `Set-Cookie` can't be represented in that type no matter how the adapter builds it (collecting into a `HashMap` is equivalent to inserting each pair, last-wins). Preserving duplicates would require a change to `azure_core`'s `Headers` type itself, so I'm leaving the conversion as-is.
---

crates/fetch_azure/examples/blob_list.rs
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Lists blobs in an Azure Storage container using `fetch` as the transport and
+//! an `anyspawn`/`tick`-backed [`Runtime`] as the credential executor.
+//!
+//! Set `AZURE_STORAGE_SERVICE_ENDPOINT` (and sign in with `az`/`azd`), then run:
+//! `cargo run --example blob_list --features azure-identity`
```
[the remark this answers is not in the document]
martintmk: &#x1F916; Fixed in c852204 - the example doc now says `cargo run -p fetch_azure --example blob_list` (dropping the non-existent `--features azure-identity`). `azure_identity` and `anyspawn_azure` (with its `azure-identity` feature) come in through `fetch_azure`'s dev-dependencies, so the example builds and runs without any extra features - the CI examples runner already runs it this way.
---

crates/anyspawn_azure/src/runtime.rs
```diff
@@ -0,0 +1,155 @@
...
+    fn sleep(&self, duration: Duration) -> TaskFuture {
+        let clock = self.clock.clone();
+        Box::pin(async move {
+            // `time::Duration` can be negative; clamp such values to zero.
+            let duration = std::time::Duration::try_from(duration).unwrap_or_default();
+            clock.delay(duration).await;
+        })
+    }
+
+    fn yield_now(&self) -> TaskFuture {
```
martintmk, line 84: why is azure SDK needing this API? Is the naive implementation we provide fine?

martintmk: &#x1F916; Fair question - here's what I found digging through `typespec_client_core`:

- `yield_now` is a required method on the `AsyncRuntime` trait (its doc says "Yield the current task back to the runtime scheduler"), so we have to provide *some* implementation.
- typespec's own `StandardRuntime` uses exactly the naive `std::thread::yield_now()` + ready future - so there's precedent it's "acceptable." But its `TokioRuntime` uses a real async yield (`tokio::task::yield_now().await`), because the naive form yields the OS thread, not the async task, and never returns control to the executor.
- I couldn't find a single call to `yield_now` inside `typespec_client_core` / `azure_core`, so in practice the SDK doesn't seem to invoke it today.

Since our `Runtime` drives real async tasks (on an `anyspawn::Spawner`, typically tokio), I went with the contract-correct version: a `Pending`-once future that actually yields the task. It's runtime-agnostic (works on any executor, doesn't tie us to tokio), matches the `TokioRuntime` behavior rather than the `StandardRuntime` shortcut, and is ~12 lines with full test coverage - so the cost is negligible.

That said, if you'd rather mirror `StandardRuntime` exactly and keep it naive, I'm happy to revert - just say the word.
---

crates/anyspawn_azure/src/runtime.rs
```diff
@@ -0,0 +1,155 @@
...
+/// A future that yields to the async scheduler exactly once before completing.
+///
+/// Unlike an immediately-ready future, this returns [`Poll::Pending`] on its
+/// first poll (waking itself), so the awaiting task actually yields control to
+/// the executor before resuming.
+struct YieldNow {
+    yielded: bool,
+}
+
+impl Future for YieldNow {
```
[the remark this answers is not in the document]
martintmk: &#x1F916; This actually compiles as-is. The workspace is on Rust edition 2024, where `Future` is part of the prelude (`std::prelude::rust_2024`), so the trait resolves without an explicit `use std::future::Future;`. The crate builds cleanly - clippy and the full test suite pass locally, and CI's `testing` / `static-analysis` jobs confirm it. Adding the import would in fact be reported as an unused import under the workspace's `-D warnings`, so I'm leaving it out. (`impl Future for RuntimeTask` has compiled this way since the first revision of the file.)
---

crates/anyspawn_azure/src/runtime.rs
```diff
@@ -0,0 +1,155 @@
...
+/// Holds the [`AbortHandle`] of the spawned [`Abortable`] task so [`abort`]
+/// can cancel work that is still pending and wake anyone awaiting it.
+///
+/// [`abort`]: AbortableTask::abort
+struct RuntimeTask {
+    handle: JoinHandle<()>,
+    abort_handle: AbortHandle,
+}
+
+impl Future for RuntimeTask {
```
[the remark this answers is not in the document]
martintmk: &#x1F916; This actually compiles as-is. The workspace is on Rust edition 2024, where `Future` is part of the prelude (`std::prelude::rust_2024`), so the trait resolves without an explicit `use std::future::Future;`. The crate builds cleanly - clippy and the full test suite pass locally, and CI's `testing` / `static-analysis` jobs confirm it. Adding the import would in fact be reported as an unused import under the workspace's `-D warnings`, so I'm leaving it out. (`impl Future for RuntimeTask` has compiled this way since the first revision of the file.)
---

ralfbiedert (approved):

ralfbiedert:
It's a bit unfortunate that the upstream APIs use a `static` singleton that will hammer the same instance. Someone, at some point, should probably try to redirect this to some thread_local, esp. when not running on tokio.

martintmk:
&#x1F916; Agreed - that global singleton is the `ASYNC_RUNTIME_IMPLEMENTATION: OnceLock<Arc<dyn AsyncRuntime>>` behind `azure_core::async_runtime::set_async_runtime`, so it lives upstream in `azure_core` / `typespec_client_core` rather than in this adapter. This crate's `Runtime` only implements the `AsyncRuntime` trait; it doesn't impose the singleton, so redirecting it to a thread_local would be an upstream change and is out of scope for this PR. Worth raising upstream if the single-instance behaviour becomes a problem off-tokio.
---

crates/anyspawn_azure/src/runtime.rs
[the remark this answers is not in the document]
martintmk: false report, future is in scope

crates/anyspawn_azure/src/runtime.rs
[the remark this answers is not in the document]
martintmk: false report, future is in scope

heaths (approved):

sgalkin (approved):

PR #500, opened 2026-06-15
fix(cachet): de-flake every_helper_emits_its_event test

martin-kolinek:
## Summary

Fixes the flaky `cachet::telemetry::cache::tests::every_helper_emits_its_event` reported in ADO #7485979.

## Root cause

A *different* test in the cachet unit-test binary -- `logging_enabled_without_subscriber_is_noop` -- exercises cachet's `tracing::error!` / `tracing::info!` paths with no `tracing` subscriber installed on the current thread. When it wins the race to be the first thread to register one of those callsites, it poisons the global `tracing-core` callsite interest cache, suppressing those events on every other thread for the rest of the binary's lifetime.

The mechanism:

1. `tracing-core::callsite::DefaultCallsite::register` -- called the first time a `tracing::*!` macro is hit -- invokes `rebuild_callsite_interest(self, &DISPATCHERS.rebuilder())`.
2. `Dispatchers::has_just_one` is `true` whenever the registered dispatcher count is `<= 1`, including 0.
3. In that case the `Rebuilder::JustOne` fast path calls `dispatcher::get_default()` **on the registering thread** instead of iterating registered dispatchers.
4. With no thread-local default and no global default, `get_default()` resolves to `NoSubscriber`, whose `register_callsite` returns `Interest::never()`.
5. That `Interest::never()` is cached process-wide. Every subsequent emission at that callsite -- from ANY thread, regardless of which subscriber that thread later installs -- is silently suppressed.

The flake surface in `every_helper_emits_its_event` was precisely the two rarest callsites in the binary (`cache.get_error` and `cache.insert_rejected`), because the no-subscriber test wins the registration race for them most often. The common callsites (`cache.hit` etc.) get registered first by tests that DO have subscribers, so they cache as enabled.

## Fix

Move the no-subscriber test to its own integration-test binary at `crates/cachet/tests/no_subscriber.rs`. Each integration test compiles to a separate executable with its own `tracing-core` callsite registry, so the scenario cannot poison any other test binary's cache. The test's intent is fully preserved -- it really does run cachet's tracing emission paths with no subscriber installed.

The new test drives the same scenarios through the public `Cache` API:

- Failing `MockCache` storage exercises the error events (`cache.get_error`, `cache.insert_error`, `cache.invalidate_error`, `cache.clear_error`).
- A working `MockCache` with `InsertPolicy::never()` exercises `cache.miss` / `cache.insert_rejected`.
- A working `MockCache` with the default policy exercises `cache.hit` / `cache.inserted` and the success path of `complete_operation`.

## Rejected alternatives

- **Install a discarding subscriber inside the unit test** (my earlier commit e2053013): rejected -- this defeats the test's intent. The test exists to verify cachet's emission paths tolerate the absence of a subscriber.
- **Invalidate the callsite interest cache after the test**: the corruption affects tests running CONCURRENTLY with the no-subscriber test, not just those running after it, so healing afterward is too late.
- **Serialize the test with all potential victims** (`serial_test` crate or a shared mutex): would require annotating many other tests in the binary and adds a new dependency for what is fundamentally an isolation problem.
- **Refactor `every_helper_emits_its_event` to share one subscriber** (my earliest commit b1365593): rejected -- stress testing showed it didn't actually fix anything (~1.4% post-fix vs ~1.3% pre-fix).

## Verification

Stress test (Windows, `--test-threads=32`):

| Scenario | Pre-fix | Post-fix |
|---|---|---|
| Full unit binary | 4/300 (~1.3%) | **0/300** |
| `telemetry::cache::tests::` filter | 25/100 (25%) | **0/200** |

The new integration test passes consistently and runs in complete isolation.


crates/cachet/tests/no_subscriber.rs
```diff
@@ -0,0 +1,91 @@
...
+//! That manifests as a flake in
+//! `telemetry::cache::tests::every_helper_emits_its_event` (assertion
+//! failures on `cache.get_error` / `cache.insert_rejected` with an empty
+//! capture buffer).
+//!
+//! Integration tests each compile to their own binary with their own
+//! `tracing-core` callsite registry, so running this scenario here cannot
+//! poison any other test binary's callsite cache.
+
+#![cfg(all(feature = "memory", feature = "logs", any(feature = "test-util", test)))]
```
martin-kolinek: Could this be done through required-features in Cargo.toml?

martin-kolinek: GHCP: Good catch — you're right, the `required-features` on the `[[test]]` entry in `Cargo.toml` already prevents the target from compiling without those features, so the file-level `#![cfg(...)]` was pure noise. Removed in f3127aac, with a short note in the module doc pointing at the Cargo.toml declaration. Verified `cargo test -p cachet --test no_subscriber --features memory,logs,test-util` still passes and `cargo build -p cachet --all-targets --no-default-features` cleanly skips the target.

martintmk (approved):

crates/cachet/src/telemetry/cache.rs
```diff
@@ -505,23 +505,12 @@ mod tests {
...
-                telemetry.record_hit("c", Duration::ZERO, false);
-                telemetry.record_get_error("c", Duration::ZERO, false);
-                telemetry.record_insert_rejected("c", false);
-                telemetry.complete_operation(request_id, "c", "cache.get", Duration::ZERO, true);
-            }
-            .with_request_id(request_id),
-        );
-        // No panic = all paths handled gracefully without a subscriber.
-    }
+    // NOTE: a previous `logging_enabled_without_subscriber_is_noop` unit test
```
Vaiz: there is no need to leave historical comments

crates/cachet/tests/no_subscriber.rs
```diff
@@ -0,0 +1,94 @@
...
+
+#[cfg_attr(miri, ignore)]
+#[tokio::test]
+async fn logging_enabled_without_subscriber_does_not_panic() {
+    // Deliberately: no `tracing::subscriber::set_default` / `set_global_default`
+    // call here, and no other test runs in this binary. The caches below have
+    // `enable_logs()` set, so each operation invokes `tracing::error!` /
+    // `tracing::info!` / `tracing::debug!` macros inside cachet. Those macros
+    // must degrade to a no-op without panicking.
+
```
Vaiz, line 52: could you add assert that subscriber is not set for easier debugging?

Vaiz (commented):
tbh, I don't see a point in such a test

by this logic we need to test every other piece of code that emits a log entry

crates/cachet/src/telemetry/cache.rs
martin-kolinek: GHCP: Done in 57e58197 — historical comment removed.

crates/cachet/tests/no_subscriber.rs
martin-kolinek: GHCP: Good idea. Added a precondition assert in 57e58197 that calls `tracing::dispatcher::get_default` and verifies the dispatch is `NoSubscriber`, with an error message explaining the test must run without a subscriber. If a future change accidentally installs a global default, the test will now fail loudly instead of silently losing its meaning.

crates/cachet/tests/no_subscriber.rs
```diff
@@ -0,0 +1,94 @@
...
+    failing.fail_when(|_| true);
+    let failing_cache: Cache<String, i32> = Cache::builder::<String, i32>(clock.clone())
+        .name("no_sub_failing")
+        .storage(failing)
+        .enable_logs()
+        .build();
+    let _ = failing_cache.get(&"k".to_string()).await;
+    let _ = failing_cache.insert("k".to_string(), CacheEntry::new(1)).await;
+    let _ = failing_cache.invalidate(&"k".to_string()).await;
+    let _ = failing_cache.clear().await;
```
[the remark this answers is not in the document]
martin-kolinek: GHCP: Done in 57e58197 — replaced the ignored `Result`s with `unwrap_err()` on all four failing-cache operations so the test fails fast if the failure-injection or API behavior changes and the error paths stop being exercised.

crates/cachet/tests/no_subscriber.rs
```diff
@@ -0,0 +1,94 @@
...
+    //     success/miss/insert-rejected tracing events: `cache.miss`,
+    //     `cache.insert_rejected`, plus `complete_operation`.
+    let rejecting_cache: Cache<String, i32> = Cache::builder::<String, i32>(clock.clone())
+        .name("no_sub_rejecting")
+        .storage(MockCache::<String, i32>::new())
+        .insert_policy(InsertPolicy::never())
+        .enable_logs()
+        .build();
+    let _ = rejecting_cache.get(&"k".to_string()).await;
+    let _ = rejecting_cache.insert("k".to_string(), CacheEntry::new(1)).await;
```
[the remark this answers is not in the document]
martin-kolinek: GHCP: Done in 57e58197 — the miss now asserts `.unwrap().is_none()`, the rejected insert asserts `.unwrap()` (since `InsertPolicy::never()` returns `Ok(())` after recording `cache.insert_rejected`), and a second `get` afterwards re-asserts `is_none()` to prove nothing was stored. Comment updated to call out that rejection is not an error.

crates/cachet/tests/no_subscriber.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    // (3) Working storage with default policy -> covers `cache.hit` and
+    //     `cache.inserted` (and the success path of `complete_operation`).
+    let working_cache: Cache<String, i32> = Cache::builder::<String, i32>(clock)
+        .name("no_sub_working")
+        .storage(MockCache::<String, i32>::new())
+        .enable_logs()
+        .build();
+    working_cache.insert("k".to_string(), CacheEntry::<i32>::new(1)).await.unwrap();
+    let _ = working_cache.get(&"k".to_string()).await; // hit
```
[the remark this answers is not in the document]
martin-kolinek: GHCP: Done in 57e58197 — the hit-path `get` now asserts `.unwrap().is_some()`.

crates/cachet/src/telemetry/cache.rs
```diff
@@ -524,110 +524,100 @@ mod tests {
...
-            });
-        });
-        assert_emits(attributes::EVENT_EVICTION, |t, request_id| {
-            futures::executor::block_on(async {
-                async { t.record_eviction("c") }.with_request_id(request_id).await;
-            });
-        });
+        #[expect(clippy::type_complexity, reason = "table-driven test cases")]
+        let cases: &[(&str, &dyn Fn(&CacheTelemetry, RequestId))] = &[
+            (attributes::EVENT_HIT, &|t, r| {
```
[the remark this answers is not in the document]
martin-kolinek: GHCP: This comment is stale — the `cases` table it refers to was part of an earlier refactor that has since been reverted (see commit 5e5f359c). The current fix moves the no-subscriber scenario into its own integration-test binary (`crates/cachet/tests/no_subscriber.rs`) and leaves the unit tests in this file unchanged from `main`, so there is no `type_complexity` suppression to remove.

sandersaares (approved):

ralfbiedert (approved):

crates/cachet/tests/no_subscriber.rs
```diff
@@ -0,0 +1,109 @@
...
+    assert!(rejecting_cache.get(&"k".to_string()).await.unwrap().is_none());
+
+    // (3) Working storage with default policy -> covers `cache.hit` and
+    //     `cache.inserted` (and the success path of `complete_operation`).
+    let working_cache: Cache<String, i32> = Cache::builder::<String, i32>(clock)
+        .name("no_sub_working")
+        .storage(MockCache::<String, i32>::new())
+        .enable_logs()
+        .build();
+    working_cache.insert("k".to_string(), CacheEntry::<i32>::new(1)).await.unwrap();
```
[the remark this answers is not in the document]
martin-kolinek: Clippy is happy

PR #501, opened 2026-06-15
test(multitude): gate over-aligned tests behind utc_backend cfg

martin-kolinek:
## Why

The UTC codegen backend (used by `ms-prod-1.95+` toolchains) caps type alignment at 8192 bytes. Six tests in `multitude` intentionally declare types with 32 KiB / 64 KiB / 128 KiB alignment to drive the allocator's over-alignment guards:

- `audit_repro::alloc_slice_ref_accepts_half_chunk_alignment_for_non_drop` (32 KiB ZST `Wide`)
- `pin_support::try_alloc_uninit_box_pin_rejects_over_alignment` (32 KiB `HalfChunk`)
- `arena::coverage_arena_gaps::*` — `HalfChunkAlign` (32 KiB), `ChunkAlign` (64 KiB)
- `coverage_extras::coverage::*` — `HugeAlign`/`HugeAlignBox` (128 KiB), `HalfChunkAlign{Drop,NoDrop}` (32 KiB), `OverAligned32K` (32 KiB)

These guards are gated on `align_of::<T>()`, which is intrinsic to `T` and cannot be exercised without a high-aligned type.

## What

- Register a new `utc_backend` cfg in the workspace `check-cfg` list (`Cargo.toml`).
- Gate every affected type declaration and test function with `#[cfg(not(utc_backend))]`. Existing `#[cfg(not(target_os = `"`windows`"`))]` gates were combined into `#[cfg(all(not(target_os = `"`windows`"`), not(utc_backend)))]`.
- LLVM builds are unaffected. Builders targeting UTC opt in via `RUSTFLAGS=--cfg utc_backend`.

## Companion PR (ox-sdk)

- ox-sdk PR 5303064: `ci(publish): set --cfg utc_backend in RUSTFLAGS for UTC toolchain` — https://o365exchange.visualstudio.com/O365%20Core/_git/ox-sdk/pullrequest/5303064

## Verification

- `cargo +ms-prod-1.95 build -p multitude --all-targets` with `RUSTFLAGS=--cfg utc_backend` — clean (was 6 errors).
- `cargo build -p multitude --all-targets` (default LLVM) — clean.
- `cargo clippy -p multitude --all-targets -- -D warnings` under both toolchains — clean, no `unexpected_cfgs` warnings.


crates/multitude/tests/coverage_extras.rs
```diff
@@ -154,9 +154,11 @@ mod coverage {
     // on `T`'s alignment, so the test runs on every platform — including
     // Windows, whose default 1 MiB stack can't accommodate the 128 KiB-
     // aligned frame the guarded body would otherwise require.
+    #[cfg(not(utc_backend))]
```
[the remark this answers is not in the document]
martin-kolinek: GHCP: Good catch — fixed in the next commit. Updated the comment block to make the LLVM-only scope explicit and added a sentence calling out that UTC builds (`--cfg utc_backend`) skip this test because UTC caps alignment at 8192 bytes.

crates/multitude/tests/coverage_extras.rs
```diff
@@ -1169,7 +1179,7 @@ mod coverage {
     }
 
     #[test]
-    #[cfg(not(target_os = "windows"))]
+    #[cfg(all(not(target_os = "windows"), not(utc_backend)))]
     fn try_alloc_slice_copy_arc_allows_half_chunk_align_for_copy_t() {
```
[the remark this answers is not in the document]
martin-kolinek: GHCP: This naming mismatch predates this PR (the test was added in commit `c0b75729` / PR #402 back in May; this PR only adds a cfg gate to its attribute line). Since it's not coupled to the UTC gating change, I'd prefer to leave it for a dedicated follow-up rather than fold an unrelated rename into a CI-fix PR. Tracking informally — happy to send a small follow-up PR if you'd like.

geeknoid (approved):

crates/multitude/tests/arena.rs
```diff
@@ -2234,6 +2234,7 @@ mod coverage_arena_gaps {
     /// Used to drive over-alignment rejection in `_with` family functions
     /// without ever instantiating the value on the test stack frame
     /// (the over-alignment guard fires before the closure is invoked).
+    #[cfg(not(utc_backend))]
```
sandersaares, line 2237: Have we reported this bug in the Developer Community feedback site? We should link it here if so.

sandersaares (approved):

ralfbiedert (approved):

crates/multitude/tests/arena.rs
martintmk: Let's merge to unblock the builds and report with Victor.

martin-kolinek: I did talk to Victor. The counterpart PR in the internal repo has a note. I'll add a devcom ticket there.

PR #507, opened 2026-06-17
perf(multitude): add arc_array allocation/timing benchmark

martintmk:
Add `criterion_arc_array` bench comparing construction of an `Arc<[Arc<[u8]>]>` (8 x 16B properties) via the global allocator vs the multitude arena, using the `alloc_tracker` crate for per-iteration allocation volume. Two strategies per path: build through a growable vec (`*`) and build directly from a pre-created slice (`*_from_slice`).

The bench surfaces a pathology in the arena vec-then-freeze path: under a warmed, reset-and-reused arena it re-grabs a full 64 KiB chunk every iteration and runs ~35x slower than the global allocator. The direct `alloc_slice_clone_arc` path reuses chunks correctly (0 bytes/op).

 | variant           | time      | bytes/op | allocs/op |
 |-------------------|-----------|----------|-----------|
 | global            | ~562 ns   | 528      | 10        |
 | arena             | ~20.0 µs  | 65535    | 1         |
 | global_from_slice | ~143 ns   | 144      | 1         |
 | arena_from_slice  | ~146 ns   | 0        | 0         |

### DIFF
--- a/crates/multitude/benches/criterion_arc_array.rs
+++ b/crates/multitude/benches/criterion_arc_array.rs
@@ -1,0 +1,142 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Builds an `Arc<[Arc<[u8]>]>` of `PROPERTIES` binary blobs two ways and
+//! compares them: `std::sync::Arc` (global allocator) vs `multitude::Arc`
+//! (arena, `reset` and reused between iterations).
+//!
+//! Each is benchmarked with two strategies:
+//!
+//! - `*` — push freshly allocated properties through a growable vec, then
+//!   freeze it into the `Arc`.
+//! - `*_from_slice` — build directly from a pre-created slice of properties,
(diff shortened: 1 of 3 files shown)


crates/multitude/benches/criterion_arc_array.rs
```diff
@@ -0,0 +1,142 @@
...
+// ---------------------------------------------------------------------------
+// Arena-backed array
+// ---------------------------------------------------------------------------
+
+fn build_arena(arena: &Arena, payload: &[u8]) -> ArenaArc<[ArenaArc<[u8]>]> {
+    let mut properties = arena.alloc_vec_with_capacity::<ArenaArc<[u8]>>(PROPERTIES);
+    for _ in 0..PROPERTIES {
+        properties.push(arena.alloc_slice_copy_arc(payload));
+    }
+    properties.try_into_arc().unwrap()
```
martintmk, line 67: this is ugly, we should have `into_arc` here

PR #510, opened 2026-06-18
feat(fetch): report fetch.runtime and fetch.transport telemetry attributes

sandersaares:
## Summary

Adds `fetch.runtime` and `fetch.transport` attributes identifying the async runtime and the transport handler used by each HTTP client. They are attached to the **instrumentation scope** of the client's `fetch` meter (both are client-scoped — a client always uses one runtime and one transport), so every metric the client emits inherits them without per-measurement overhead. This includes the hyper connection metrics, which are recorded against the same scoped meter handed to the transport via `CustomContext::meter`.

These started as a single combined `fetch.transport` attribute, but that conflated two independent dimensions (the runtime a client is built for and the transport it dispatches through, which today map 1:1 but need not in future, e.g. an h3-on-tokio transport). Splitting them keeps each value a single, unambiguous word.

## Runtime and transport names

| Client | `fetch.runtime` | `fetch.transport` |
|--------|-----------------|-------------------|
| Tokio + hyper (default `*_tokio()`) | `tokio` | `hyper` |
| Fake HTTP clients (`builder_fake`/`new_fake`) | `fake` | `fake` |
| Custom (`custom::create_builder`) | caller-supplied `runtime` | caller-supplied `transport` |

## Changes

- `telemetry.rs`: new `FETCH_RUNTIME_ATTRIBUTE`/`FETCH_TRANSPORT_ATTRIBUTE`, `client_scope`/`base_scope` helpers, and a reworked `Metering` enum (`Global(InstrumentationScope)` / `Custom(Meter)`) whose scope is supplied once at construction. The scope follows the existing `seatbelt::metrics::create_meter` pattern (`meter_with_scope` + `InstrumentationScope`).
- `custom::create_builder` gains two leading parameters, `runtime` and `transport` (each `impl Into<Cow<'static, str>>`), naming the two attributes (breaking change; no other workspace crate calls it). They are carried on the `Transport.runtime_name`/`Transport.transport_name` fields through the shared `builder_custom_internal` funnel.
- Built-in callers pass constants: `tokio.rs` -> runtime `tokio`, transport `hyper`; `fake.rs` -> `fake`/`fake`.
- `client_builder.rs` builds the meter with the client scope; the standalone `MetricsLayer` (which has no runtime/transport) keeps its existing scope-less behavior via `base_scope()`.

## Documentation

- New "Instrumentation scope" section in the telemetry reference; documented the new `runtime`/`transport` parameters and updated the `http_client_custom` example.

## Testing

- New `tests/telemetry_scope.rs` asserts both scope attributes on exported metrics for the fake, custom, and tokio transports, plus a custom transport recording its own instrument on `cx.meter`.
- Existing unit tests updated for the new `Metering` shape.

Verified with clippy (`-D warnings` across feature sets), the fetch test suite, doctests, `just format`, `just readme`, and `just spellcheck`.


crates/fetch/src/constants.rs
```diff
@@ -2,3 +2,11 @@
 // Licensed under the MIT License.
 
 pub(crate) const DEFAULT_HTTP_CLIENT_NAME: &str = "http_client";
+
+/// `fetch.transport` value for the bundled Tokio + hyper transport.
+#[cfg(all(feature = "tokio", any(feature = "rustls", feature = "native-tls")))]
+pub(crate) const HYPER_ON_TOKIO_TRANSPORT_NAME: &str = "hyper-on-tokio";
```
martintmk:
```suggestion
pub(crate) const HYPER_ON_TOKIO_TRANSPORT_NAME: &str = "tokio.hyper";
```

This should follow OTel naming conventions. (namespace based)

In oxidizer we would use: `oxidizer.hyper`

sandersaares: Can you point me to what conventions you refer to here? I am not aware of any OTel recommendations for namespace based values but maybe I have a gap in my knowledge. I am only aware of guidelines about keys, which are not applicable here.

Happy to adjust if there is some actual OTel guideline about value naming that seems preferential to align with. If not, I will stick to what is in the PR. The alternative example you provided seems markedly less clear due to obscuring the relationship between the two elements ("on" is a lot more meaningful than ".").

crates/fetch/src/custom.rs
```diff
@@ -153,15 +159,20 @@ pub struct CustomContext<Extras = ()> {
 /// # Ok(())
 /// # }
 /// ```
-pub fn create_builder<F, R, Extras>(factory: F, isolation: Isolation, deps: impl Into<CustomDeps<Extras>>) -> HttpClientBuilder
+pub fn create_builder<F, R, Extras>(
+    name: impl Into<Cow<'static, str>>,
+    factory: F,
+    isolation: Isolation,
+    deps: impl Into<CustomDeps<Extras>>,
+) -> HttpClientBuilder
```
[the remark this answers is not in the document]
sandersaares: Yes, intentional.

crates/fetch/src/constants.rs
martintmk: https://opentelemetry.io/docs/specs/semconv/general/naming/#general-naming-considerations

Also look at existing attributes we are reporting in fech, we want these to be aligned. To me the `hyper-on-tokio`, the `-on-` is arbitrary the same way we do not have any `HttpClientOnTokio` type but rather `tokio` module with `HttpClient` in it (theoretical). 

OTel tries to use the same principles. `tokio.hyper`, `tokio.winhttp`, etc. gives pretty clear meaning to me.

martintmk (changes requested):

crates/fetch/src/constants.rs
sandersaares: The page you linked applies to attribute names (see first sentence on page), which is not what we are dealing with here, so the linked guidelines seem to be not relevant.

Is there some specific prior art here you can link as reference for why we should use dots format? Or is this just personal preference?

martintmk: > Is there some specific prior art here you can link as reference for why we should use dots format? Or is this just personal preference?

Aforementioned OTel naming conventions. If possible, I try to apply the key naming conventions  to values too (where applicable),

> Use namespaces (and dot separator) whenever it makes sense. For example when introducing an attribute representing a property of some object, follow *{object}.{property} pattern. Avoid using underscore (*{object}_{property}) if this object could have other properties.

`tokio` is the parent namespace representing the runtime, `hyper` is the actual library used for transport. To me the grouping seems logical and aligned with how OTel represents attribute values.  

You can also review existing standardized semantic conventions for more context. 
https://opentelemetry.io/docs/specs/semconv/http/http-metrics/


FYI, i have never seen `-` in standard attributes nor values. At best it should be `hyper_on_tokio`, but that does not align with naming either as per this comment. 

edit: misread your comment, so clarified the comment

sandersaares: I am not convinced that dot-naming overall is sensible here nor that `tokio.hyper` specifically makes sense (for one, tokio is not a namespace we own, we cannot possibly namespace things under tokio).

However, this thread has revealed an underlying tension - this attribute really contains two pieces of data in one. That's a structural flaw - I will split it into separate `fetch.runtime` and `fetch.transport` attributes, each single-word.

martintmk: well, I am completely fine with that :D

martintmk (approved):

martintmk (approved):

crates/fetch/src/constants.rs
```diff
@@ -2,3 +2,19 @@
...
 
 pub(crate) const DEFAULT_HTTP_CLIENT_NAME: &str = "http_client";
+
+/// `fetch.runtime` value for the bundled Tokio runtime.
+#[cfg(all(feature = "tokio", any(feature = "rustls", feature = "native-tls")))]
+pub(crate) const TOKIO_RUNTIME_NAME: &str = "tokio";
+
+/// `fetch.transport` value for the bundled hyper transport.
+#[cfg(all(feature = "tokio", any(feature = "rustls", feature = "native-tls")))]
+pub(crate) const HYPER_TRANSPORT_NAME: &str = "hyper";
```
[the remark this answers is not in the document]
sandersaares: Aligned: the PR description now states the default Tokio+hyper client reports etch.runtime=tokio and etch.transport=hyper, matching the constants, docs, and tests.

crates/fetch/src/custom.rs
```diff
@@ -153,15 +161,27 @@ pub struct CustomContext<Extras = ()> {
...
 /// # }
 /// ```
-pub fn create_builder<F, R, Extras>(factory: F, isolation: Isolation, deps: impl Into<CustomDeps<Extras>>) -> HttpClientBuilder
+pub fn create_builder<F, R, Extras>(
+    runtime: impl Into<Cow<'static, str>>,
+    transport: impl Into<Cow<'static, str>>,
+    factory: F,
+    isolation: Isolation,
+    deps: impl Into<CustomDeps<Extras>>,
+) -> HttpClientBuilder
```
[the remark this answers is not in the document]
sandersaares: Aligned: the PR description now documents both leading parameters (
untime and 	ransport) and what each names.

crates/fetch/src/custom.rs
```diff
@@ -219,6 +254,14 @@ impl Transport {
         self.inner.as_ref()(options, meter, index)
     }
 
+    pub(crate) fn runtime(&self) -> &str {
```
Vaiz: would it make sense to safe an allocation and return `Cow` or `opentelemetry::Value` instead of `&str`?

since it's a private method, that won't be a problem

Vaiz (approved):

crates/fetch/src/custom.rs
sandersaares: Suggestion applied.

Vaiz (approved):

ralfbiedert (approved):

crates/fetch/src/telemetry.rs
```diff
@@ -61,28 +67,52 @@ impl Extend<KeyValue> for TelemetryAttributes {
...
 }
 
+/// Builds the `fetch` instrumentation scope carrying the `fetch.runtime` and
+/// `fetch.transport` attributes, attached to every metric a client records.
+pub(crate) fn client_scope(runtime: &str, transport: &str) -> InstrumentationScope {
+    InstrumentationScope::builder(METER_NAME)
+        .with_attributes([
+            KeyValue::new(FETCH_RUNTIME_ATTRIBUTE, runtime.to_string()),
+            KeyValue::new(FETCH_TRANSPORT_ATTRIBUTE, transport.to_string()),
+        ])
```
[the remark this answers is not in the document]
sandersaares: Will address as separate PR.

PR #514, opened 2026-06-22
perf(fetch): avoid allocations in client_scope for borrowed-'static names

sandersaares:
`client_scope` previously called `to_string()` on the runtime and transport names, forcing a heap allocation on every client build even though the transport already stores these names as `Cow<'static, str>`.

`client_scope` now takes `impl Into<Value>` for the runtime and transport names — the exact bound `KeyValue::new` requires — and hands each value straight into the instrumentation scope attributes with no extra clone inside the function. The production call sites in `client_builder.rs` pass `transport.runtime().clone()`/`transport.name().clone()`; cloning the transport's `Cow<'static, str>` is allocation-free for the common borrowed-`'static` case and simply hands over ownership. The in-file unit tests pass plain `&'static str` literals, which likewise convert into `Value` without allocating.

Net effect: borrowed-`'static` runtime/transport names no longer allocate when building the meter scope, and the signature stays idiomatic (no `clippy::ptr_arg` `#[expect]` or `&Cow` borrow needed).

Verified with `just package=fetch clippy` (clean) and `just package=fetch test` (153 passed).


geeknoid (approved):

crates/fetch/src/telemetry.rs
```diff
@@ -98,11 +100,14 @@ impl From<Metering> for Meter {
 
 /// Builds the `fetch` instrumentation scope carrying the `fetch.runtime` and
 /// `fetch.transport` attributes, attached to every metric a client records.
-pub(crate) fn client_scope(runtime: &str, transport: &str) -> InstrumentationScope {
+// Takes `&Cow<'static, str>` rather than `&str` so the borrowed-'static case is cloned without
+// allocating; lowering to `&str` would force a `to_string()` on every call.
+#[expect(clippy::ptr_arg, reason = "cloning the Cow avoids allocating for the common borrowed-'static case")]
+pub(crate) fn client_scope(runtime: &Cow<'static, str>, transport: &Cow<'static, str>) -> InstrumentationScope {
```
Vaiz:
```suggestion
pub(crate) fn client_scope(runtime: impl Into<Value>, transport: impl Into<Value>) -> InstrumentationScope {
```

Vaiz (approved):

ralfbiedert (approved):

crates/fetch/src/telemetry.rs
sandersaares: applied

crates/fetch/src/telemetry.rs
```diff
@@ -98,11 +98,11 @@ impl From<Metering> for Meter {
...
 /// `fetch.transport` attributes, attached to every metric a client records.
-pub(crate) fn client_scope(runtime: &str, transport: &str) -> InstrumentationScope {
+pub(crate) fn client_scope(runtime: impl Into<Value>, transport: impl Into<Value>) -> InstrumentationScope {
     InstrumentationScope::builder(METER_NAME)
         .with_attributes([
-            KeyValue::new(FETCH_RUNTIME_ATTRIBUTE, runtime.to_string()),
-            KeyValue::new(FETCH_TRANSPORT_ATTRIBUTE, transport.to_string()),
+            KeyValue::new(FETCH_RUNTIME_ATTRIBUTE, runtime),
+            KeyValue::new(FETCH_TRANSPORT_ATTRIBUTE, transport),
         ])
```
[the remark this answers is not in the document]
sandersaares: Updated the PR description (option b) to reflect the impl Into<Value> API and the call-site Cow::clone()s. The impl Into<Value> design is intentional — it's the bound KeyValue::new already requires, which avoids the clippy::ptr_arg #[expect] and keeps the borrowed-'static path allocation-free (cloning a Cow::Borrowed doesn't allocate).

Vaiz (approved):

PR #523, opened 2026-06-25
fix(justfiles): use cargo package name instead of directory name for -p selector

sgalkin:
The `loom` and `bolero` recipes in `extended.just` derived the cargo `-p` selector from the crate directory name under `crates/`. Cargo matches `-p` against the package name from `Cargo.toml`, not the directory name. When they differ (e.g. underscored dir vs hyphenated package), cargo errors with 'package ID specification did not match any packages'.

Fix by calling `cargo metadata` once at recipe start to build a manifest-path to package-name map, then using the real package name for `-p` and the directory path for filesystem operations.

AB#7524956

### DIFF
--- a/justfiles/extended.just
+++ b/justfiles/extended.just
@@ -22,11 +22,25 @@ loom FILTER="":
     $ErrorActionPreference = "Stop"
     $PSNativeCommandUseErrorActionPreference = $true
 
+    # Build a manifest-path → package-name map so we use the real Cargo
+    # package name for `-p` instead of the directory name (they can differ).
+    $metadata = cargo metadata --format-version=1 --no-deps --locked | ConvertFrom-Json
+    $packagesByManifest = @{}
+    foreach ($pkg in $metadata.packages) {
+        $norm = [System.IO.Path]::GetFullPath($pkg.manifest_path)
+        $packagesByManifest[$norm] = $pkg.name
+    }
+
     $crates = @()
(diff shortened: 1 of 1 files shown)


sandersaares (approved):

justfiles/extended.just
```diff
@@ -22,11 +22,25 @@ loom FILTER="":
...
+    # Build a manifest-path → package-name map so we use the real Cargo
+    # package name for `-p` instead of the directory name (they can differ).
+    $metadata = cargo metadata --format-version=1 --no-deps --locked | ConvertFrom-Json
+    $packagesByManifest = @{}
+    foreach ($pkg in $metadata.packages) {
+        $norm = [System.IO.Path]::GetFullPath($pkg.manifest_path)
+        $packagesByManifest[$norm] = $pkg.name
+    }
+
     $crates = @()
```
sandersaares, line 34: Shouldn't this say "pacakges" as well? We have some colloquial naming here - "crates" are the things inside packages, even though people often refer to packages as crates. But given the prevalence of the word "package" showing up now from Cargo metadata, maybe it is time to also fix this colloquialism in variable names and strings.

sandersaares (approved):

justfiles/extended.just
sgalkin: Will take as a follow up. Thank you for suggestion.

martin-kolinek (approved):

sandersaares (approved):

PR #526, opened 2026-06-25
fix: exclude non-source artifacts from published crates via include allowlist

sgalkin:
Published crates shipped Git LFS-tracked binaries (logo.png, favicon.ico, diagrams) which, depending on whether the publishing runner smudged LFS, were packaged as ~130-byte pointer stubs. That made the .crate non-reproducible and broke dependent crates' lockfile checksums on docs.rs (checksum changed between lock files), repeatedly taking down docs for most of the workspace.

Add an explicit include allowlist in [workspace.package] (opted into by every crate) so only source and build-required files are packaged. This keeps all LFS-tracked binaries out of the package by construction and also drops other non-source cruft (AGENTS.md, design docs, editable diagram sources). Logos are already referenced via absolute html_logo_url, so nothing consumes the packaged copies.

Validated by packaging and rebuilding all 38 crates from their .crate tarballs (38/38 OK) and confirming 0 crates ship any binary artifact.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,35 @@ authors = ["The Oxidizer Project Authors"]
 license = "MIT"
 homepage = "https://github.com/microsoft/oxidizer"
 
+# Allowlist of files packaged into published crates (`cargo package`/`publish`).
+# We use an explicit include (allowlist), not exclude, so that non-source
+# artifacts can never leak into the published `.crate` by construction.
+#
+# WARNING: never add a Git LFS-tracked file to a packaged path. `cargo` has no
+# LFS awareness: depending on whether the publishing runner smudged LFS, an
+# LFS-tracked binary is packaged either as its real bytes or as a ~130-byte
+# pointer STUB. That makes the `.crate` non-reproducible, so the checksum a
+# dependent records for this crate can differ from what crates.io ultimately
+# stores - which breaks the dependent's build on docs.rs ("checksum changed
+# between lock files"). The `.gitattributes` LFS globs (*.png, *.ico, *.jpg,
+# *.pdf, *.zip, *.dll, *.exe, ...) therefore must stay OUT of every packaged
+# path (src, examples, tests, benches, docs). Reference such assets by absolute
+# URL instead, e.g. `#![doc(html_logo_url = "https://.../logo.png")]`.
+#
+# The `docs/**/*.md` glob matches only Markdown, so compile-time `include_str!`
+# doc fragments are packaged while binary diagram assets beside them are not.
+include = [
+    "/src/**",
+    "/examples/**",
+    "/benches/**",
+    "/tests/**",
+    "/docs/**/*.md",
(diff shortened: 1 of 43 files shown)


Cargo.toml
```diff
@@ -12,6 +12,35 @@ authors = ["The Oxidizer Project Authors"]
...
+# dependent records for this crate can differ from what crates.io ultimately
+# stores - which breaks the dependent's build on docs.rs ("checksum changed
+# between lock files"). The `.gitattributes` LFS globs (*.png, *.ico, *.jpg,
+# *.pdf, *.zip, *.dll, *.exe, ...) therefore must stay OUT of every packaged
+# path (src, examples, tests, benches, docs). Reference such assets by absolute
+# URL instead, e.g. `#![doc(html_logo_url = "https://.../logo.png")]`.
+#
+# The `docs/**/*.md` glob matches only Markdown, so compile-time `include_str!`
+# doc fragments are packaged while binary diagram assets beside them are not.
+include = [
```
Vaiz, line 32: imo, there is not much value in keeping examples, tests, and CHANGELOG.md. we can exclude them as well

Vaiz (approved):

Cargo.toml
```diff
@@ -12,6 +12,35 @@ authors = ["The Oxidizer Project Authors"]
 license = "MIT"
 homepage = "https://github.com/microsoft/oxidizer"
 
+# Allowlist of files packaged into published crates (`cargo package`/`publish`).
```
sandersaares, line 15: Huh, I did not even realize all this random stuff is packed by default but I guess Cargo has no idea what rustc actually needs at compile time, so it makes sense.

I suggest also defining some `docs/packaging-guidelines.md` and referencing from `AGENTS.md` to inform future generations of agent-maintainers on how to do this properly.

sandersaares (approved):

martin-kolinek (approved):

Cargo.toml
sgalkin: done

Cargo.toml
sgalkin: Good call — dropped CHANGELOG.md (neither a declared target nor build-required, and crates.io doesn't render it).

I kept examples/tests/benches on the principle that anything declared in Cargo.toml should ship: several crates declare explicit [[example]]/[[test]]/[[bench]] targets, so excluding those dirs would leave the published manifest pointing at files not in the tarball. I checked the 20 most-downloaded crates and none ship such dangling declarations — they package the dirs they declare (or don't declare them at all).

Cargo does tolerate dangling declarations (verified: fetch/seatbelt/multitude/ohno still package + rebuild from their .crate with those dirs excluded), but keeping the manifest internally consistent — so cargo build --all-targets / cargo test work against the published tarball — seemed worth the few KB. Happy to revisit if you feel strongly.

sandersaares (approved):

PR #527, opened 2026-06-25
ci: run cargo udeps with and without --all-targets; remove unused dev-dependencies

sgalkin:
## Problem

The `Udeps` step in `.github/workflows/main.yml` ran `cargo udeps --all-features --workspace` **without `--all-targets`**. `cargo udeps` only analyzes dev-dependencies when `--all-targets` (or `--tests/--benches/--examples`) is passed — otherwise it checks lib/bins only. So CI never validated dev-dependencies, even though the local `just udeps` recipe (`justfiles/basic.just`) *does* pass `--all-targets`. Unused dev-deps accumulated undetected.

Demonstrated on `http_extensions`:

| invocation | result |
|---|---|
| `udeps --all-features --all-targets` (local `just udeps`) | flags `testing_aids` (exit 1) |
| `udeps --all-features` (CI) | "All deps seem to have been used" (exit 0) |

## Changes

- **CI:** add `--all-targets` to the `Udeps` step so it matches `just udeps` and validates test/bench/example dev-dependencies.
- **Cleanup:** remove 9 unused dev-dependencies surfaced by `cargo udeps --workspace --all-features --all-targets`:
  - `cachet`: `tracing-test`
  - `cachet_memory`: `futures`, `tick`
  - `data_privacy_macros_impl`: `mutants`
  - `fetch`: `rstest`
  - `http_extensions`: `testing_aids`
  - `testing_aids`: `mockall`, `mutants` (the `mutants` references were prose about *cargo-mutants*, not the crate)
  - `uniflight`: `tick`

## Validation

- `cargo udeps --workspace --all-features --all-targets` → clean (exit 0)
- `cargo build --workspace` ✓ &nbsp; `cargo build --workspace --release` ✓
- `cargo test --workspace --all-targets` ✓ &nbsp; `cargo test --workspace --doc` ✓

<sub>🤖 Prepared with AI assistance.</sub>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -282,7 +282,7 @@ jobs:
 
       # execute
       - name: Udeps
-        run: cargo +${{ env.RUST_NIGHTLY }} udeps --all-features --workspace ${{ needs.delta.outputs.exclude_not_modified }} --color always
+        run: cargo +${{ env.RUST_NIGHTLY }} udeps --all-features --all-targets --workspace ${{ needs.delta.outputs.exclude_not_modified }} --color always
       - name: Careful
         if: success() || failure()
         run: cargo +${{ env.RUST_NIGHTLY }} careful nextest run --all-features --workspace ${{ needs.delta.outputs.exclude_not_affected }} --color always --target ${{ matrix.target }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -282,7 +282,17 @@ jobs:
 
       # execute
       - name: Udeps
-        run: cargo +${{ env.RUST_NIGHTLY }} udeps --all-features --all-targets --workspace ${{ needs.delta.outputs.exclude_not_modified }} --color always
+        shell: pwsh
+        run: |
+          $ErrorActionPreference = "Stop"
+          $PSNativeCommandUseErrorActionPreference = $true
+          # Without --all-targets: enforce that every [dependencies] entry is used
+          # by the library/bins (catches normal deps used only by tests/benches/
+          # examples -- i.e. misplaced deps that should be [dev-dependencies]).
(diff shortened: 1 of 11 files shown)


.github/workflows/main.yml
```diff
@@ -282,7 +282,7 @@ jobs:
 
       # execute
       - name: Udeps
```
sandersaares, line 284: Does it still detect dependencies that are ONLY needed as dev-deps but present as regular deps? I fear we might need to have both `--all-targets` and "not all targets" runs of this to catch both variants. Or is the tool smart enough by design already to avoid this?

sandersaares (approved):

martin-kolinek (approved):

.github/workflows/main.yml
sgalkin: good catch

sgalkin: Great catch — confirmed, and you're right that we need both runs. I verified it empirically on `recoverable` by moving a test-only dep (`insta`) into `[dependencies]`:

- `udeps --all-features --all-targets` → *"All deps seem to have been used"* (misses it — the compiled tests make the normal dep look used)
- `udeps --all-features` (no `--all-targets`) → flags `recoverable → dependencies → "insta"` (catches it)

So they're mirror images:
- **`--all-targets`** enforces `[dev-dependencies]` (catches unused dev-deps),
- **no `--all-targets`** enforces `[dependencies]` (catches normal deps used only by dev targets — i.e. ones that should be dev-deps).

Updated this PR to run udeps **both** ways (pinned with `--locked`), and updated the `just udeps` recipe to match so local == CI. A full-workspace run passes both ways today, so no new ignores were needed.

🤖 _Authored by GitHub Copilot._

sandersaares (approved):

geeknoid (approved):

sandersaares (approved):

PR #528, opened 2026-06-25
fix(bytesbuf): size numeric reads by T::Bytes and read unaligned for FromBytes soundness

sandersaares:
## What

Two related soundness fixes for `BytesView::get_num_{le,be,ne}` and their buffered fallbacks, which read a `T: FromBytes` value out of a byte view.

### 1. Size reads by `T::Bytes`, not `T`

`FromBytes::from_*_bytes` consumes `&T::Bytes`, so every read must size, advance, and bound by `size_of::<T::Bytes>()` rather than `size_of::<T>()`. For a downstream `FromBytes` impl where the two differ, the old code read/advanced the wrong length and the buffered fallback left a `MaybeUninit<T::Bytes>` partially initialized (or overflowed it) before `assume_init`. For the standard integer/float types `Bytes = [u8; size_of::<T>()]`, so the two sizes are equal and this is behavior-neutral.

### 2. Read unaligned instead of building a misaligned reference

The fast path constructed a `&T::Bytes` directly from an unaligned byte-slice pointer (`bytes.as_ptr().cast::<T::Bytes>().as_ref()`). `FromBytes::Bytes` is only bounded by `NumBytes` (`AsRef<[u8]>`, `Borrow<[u8]>`, ...), which does not guarantee an alignment of one — a custom impl may use a `#[repr(align(N))]` byte wrapper. For such a type that reference is misaligned, which is undefined behavior. The value is now materialized with `ptr::read_unaligned` into a properly aligned local. For the standard types (`Bytes = [u8; N]`, alignment 1) codegen is unchanged.

## Tests

Added soundness coverage with custom `FromBytes` types whose `Bytes` associated type is wider than the value, narrower than the value, and over-aligned, exercising both the single-span fast path (including a guaranteed-misaligned source pointer) and the multi-span buffered path.

## Validation

- Negative check: reverting the fix makes 6/10 of the new tests fail under Miri (misaligned-reference UB, uninitialized-memory UB, and length-assertion panics), confirming the tests are meaningful.
- With the fix: full `bytesbuf` suite passes under Miri (186 passed, 6 skipped) and on stable; `cargo clippy --all-targets --all-features -- -D warnings` is clean; nightly rustfmt and spellcheck are clean.


crates/bytesbuf/src/view_get.rs
```diff
@@ -634,4 +631,225 @@ mod tests {
...
+            assert!(view.is_empty());
+        }
+
+        #[test]
+        fn over_aligned_le_misaligned_single_span() {
+            let memory = TransparentMemory::new();
+
+            // A single leading byte that we drop, leaving the eight value bytes at an odd offset
+            // from the allocation base. That address can never satisfy an eight-byte alignment, so
+            // the fast path reads from a guaranteed-misaligned pointer.
```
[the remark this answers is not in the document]
sandersaares: Good catch — reworded in 244fe099. The comment no longer claims the offset can *never* be eight-byte aligned; it now states that the value bytes are misaligned whenever the allocation base is itself eight-byte aligned (the common case), which is the condition the test relies on. The test's real guarantee comes from Miri, which flags the old reference-from-unaligned-pointer path as UB.

crates/bytesbuf/src/view_get.rs
```diff
@@ -634,4 +631,226 @@ mod tests {
...
+
+        use crate::BytesView;
+        use crate::mem::testing::TransparentMemory;
+
+        /// `Bytes` is wider than the value it produces: `size_of::<NarrowValue>() == 1` while the
+        /// serialized form is four bytes. Sizing reads by `size_of::<T>()` would read past a
+        /// one-byte first-span slice and leave the buffered path's `MaybeUninit<[u8; 4]>` only
+        /// partially initialized before `assume_init`.
+        #[derive(Debug, PartialEq, Eq, Clone, Copy)]
+        struct NarrowValue(u8);
```
Vaiz, line 656: `NarrowValue`, `WideValue` names don't really tell what's that code is doing

also, why 1 byte is serialized as 4, and 8 bytes as 2?

Vaiz: ok, I didn't realize that this is test code
it's really hard to see it when GH shows only a slice of code

Vaiz (approved):

martin-kolinek (approved):

PR #531, opened 2026-06-26
chore(release): re-release all packages with LFS-free tarballs

sgalkin:
Cascade patch release of all 36 publishable crates so crates.io receives deterministic, LFS-free .crate tarballs produced by the new package include allowlist (#526). This clears the recurring docs.rs "checksum changed between lock files" failures caused by Git LFS logo/favicon binaries being packaged as either real bytes or pointer stubs.

Driven by releasing the 8 foundational crates as patch; the dependent cascade propagates the patch bump to the remaining 28.


martin-kolinek (approved):

geeknoid (approved):

sandersaares (approved):
I am concerned by the logic in the PR description "Driven by releasing the 8 foundational crates as patch; the dependent cascade propagates the patch bump to the remaining 28." which does not actually mention anything about reviewing any of the crates for breaking changes.

However, upon my own review, I was not able to find breaking changes in scope of this release so perhaps that is just a vague PR description and not a real problem.

sandersaares (approved):

PR #532, opened 2026-06-27
feat: Introduce the plurality pool allocator

geeknoid:


crates/plurality/src/pool.rs
```diff
@@ -0,0 +1,830 @@
...
+        match self.alloc_slot() {
+            Some(slot) => {
+                let value = f();
+                // SAFETY: `slot` was just popped and is owned exclusively here.
+                unsafe { self.occupy(slot, value) };
+                Ok(Box::from_slot(slot))
+            }
+            None => Err(AllocError),
+        }
+    }
```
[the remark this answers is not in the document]
geeknoid: Fixed in 11a41235c0. `try_alloc_box_with` now reserves the slot via `try_alloc_uninit_box()` (an RAII handle that frees the slot on drop), then runs `f()`, writes, and `assume_init`s. If `f()` panics, the uninit handle's `Drop` returns the slot to the free list. Added the `closure_panic_returns_slot_to_pool` regression test (also checked under Miri).

crates/plurality/src/pool.rs
```diff
@@ -0,0 +1,830 @@
...
+        match self.alloc_slot() {
+            Some(slot) => {
+                let value = f();
+                // SAFETY: `slot` was just popped and is owned exclusively here.
+                unsafe { self.occupy(slot, value) };
+                Ok(Arc::from_slot(slot))
+            }
+            None => Err(AllocError),
+        }
+    }
```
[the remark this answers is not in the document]
geeknoid: Fixed in 11a41235c0. `try_alloc_arc_with` now reserves through `try_alloc_uninit_arc()` (RAII), runs `f()`, then initializes via `Arc::get_mut` + `assume_init`. A panic in `f()` frees the reserved slot through the uninit handle's `Drop` instead of leaking capacity. The `get_mut().expect()` is an unreachable invariant guard (the freshly reserved handle has refcount 1), annotated `#[expect(clippy::missing_panics_doc, reason = "only unreachable panics")]` per the workspace convention. Covered by the new regression test.

crates/plurality/src/pool.rs
```diff
@@ -0,0 +1,830 @@
...
+        match self.alloc_slot() {
+            Some(slot) => {
+                let value = f();
+                // SAFETY: `slot` was just popped and is owned exclusively here.
+                unsafe { self.occupy_local(slot, value) };
+                Ok(Alloc::from_slot(slot))
+            }
+            None => Err(AllocError),
+        }
+    }
```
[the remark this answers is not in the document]
geeknoid: Fixed in 11a41235c0. `try_alloc_with` now reserves via `try_alloc_uninit()` (RAII), runs `f()`, writes through the handle's `DerefMut`, then `assume_init`s, so a panic in `f()` frees the slot on drop. Covered by the new `closure_panic_returns_slot_to_pool` test.

crates/plurality/src/pool.rs
```diff
@@ -0,0 +1,830 @@
...
+        match self.alloc_slot() {
+            Some(slot) => {
+                let value = f();
+                // SAFETY: `slot` was just popped and is owned exclusively here.
+                unsafe { self.occupy(slot, value) };
+                Ok(Rc::from_slot(slot))
+            }
+            None => Err(AllocError),
+        }
+    }
```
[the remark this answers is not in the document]
geeknoid: Fixed in 11a41235c0. `try_alloc_rc_with` now reserves through `try_alloc_uninit_rc()` (RAII), runs `f()`, then initializes via `Rc::get_mut` + `assume_init`; a panic frees the reserved slot through the uninit handle's `Drop`. Same unreachable-panic annotation as the Arc path. Covered by the new regression test.

crates/plurality/src/pool.rs
```diff
@@ -0,0 +1,830 @@
...
+            Err(_) => pool_full(),
+        }
+    }
+
+    /// Fallible [`alloc_box`](Self::alloc_box). Returns the rejected value on a
+    /// full pool.
+    ///
+    /// # Errors
+    /// Returns [`AllocError`] if no slot is available; `value` is dropped.
+    pub fn try_alloc_box(&self, value: T) -> Result<Box<T, A>, AllocError> {
```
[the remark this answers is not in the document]
geeknoid: Fixed in 11a41235c0. Removed the inaccurate sentence; `try_alloc_box`'s summary is now just "Fallible \`alloc_box\`." and the `# Errors` section already documents that `value` is dropped on a full pool.

crates/plurality/src/slot.rs
```diff
@@ -0,0 +1,118 @@
...
+///
+/// Excluded from coverage: the overflow branch needs a `> i32::MAX` refcount,
+/// which a test cannot produce.
+#[inline]
+#[cfg_attr(coverage_nightly, coverage(off))]
+pub(crate) fn check_refcount_overflow(old: u32) {
+    if old > MAX_REFCOUNT {
+        abort();
+    }
+}
```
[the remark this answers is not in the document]
geeknoid: Fixed in 5b79c29fda. Tightened the guard to `old >= MAX_REFCOUNT`. Since `old` is the pre-increment value (from `fetch_add`), aborting once it reaches `MAX_REFCOUNT` keeps the surviving count from ever exceeding the bound — so the live refcount now stays `<= MAX_REFCOUNT`, matching the doc. (The original `>` mirrored `std::sync::Arc`, which tolerates a transient `MAX_REFCOUNT + 1`; the stricter form is unambiguous and free.) Updated the doc/coverage comments to match.

crates/plurality/src/pool.rs
```diff
@@ -0,0 +1,827 @@
...
+    pub(crate) allocator: A,
+}
+
+/// A growable, fixed-slot object pool.
+///
+/// See the [crate-level documentation](crate) for the concurrency model. The
+/// pool is `Send` (it can be moved between threads) but **not** `Sync` (only
+/// one thread allocates at a time). It produces four handle types — `Box`,
+/// `Alloc`, `Arc`, `Rc` — the `Send` ones (`Box`/`Arc`) may be dropped from any
+/// thread.
```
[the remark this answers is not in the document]
geeknoid: Fixed in 5b79c29fda. Reworded the `Pool` doc to make the conditional `Send` explicit: "`Box` and `Arc` are `Send` (when `T` and the allocator `A` are), so they may be dropped from any thread; `Alloc` and `Rc` are `!Send` and stay on the allocating thread." This matches the impl bounds (`Box<T, A>: Send` needs `T: Send, A: Send + Sync`; `Arc<T, A>: Send` needs `T: Send + Sync, A: Send + Sync`).

crates/plurality/src/builder.rs
```diff
@@ -0,0 +1,127 @@
...
+    #[cold]
+    pub fn build(self) -> Pool<T, A> {
+        let chunk_size = self.chunk_size;
+        if let Some(max) = self.max_chunks {
+            let total = u64::from(chunk_size) * u64::from(max);
+            assert!(
+                total <= u64::from(FREE_END),
+                "chunk_size * max_chunks exceeds the u32 slot-index ceiling"
+            );
+        }
```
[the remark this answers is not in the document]
geeknoid: Fixed in 1833d0019c. `build()` now validates `total` against a combined ceiling: `min(FREE_END, usize::MAX - 1)`. The second term ensures `1 + total` (the pool's own refcount plus every live refcounted allocation that `pool_refcount` can hold) fits in `usize`. On 64-bit the `u32` slot-index ceiling binds (unchanged); on 32-bit the `usize` term binds and rejects `total == u32::MAX`, closing the overflow-by-one window. Confirmed `clone` does **not** touch `pool_refcount` (only initial allocation does), so the max value is exactly `1 + total`. Folding it into the existing assert keeps the panic branch covered by `capacity_overflow_panics` on 64-bit; updated the doc and that test's expected message accordingly.

crates/plurality/src/chunk.rs
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use core::alloc::Layout;
+use core::ptr::NonNull;
+
+use crate::pool::PoolInner;
+use crate::slot::SlotCell;
```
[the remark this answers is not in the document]
geeknoid: This compiles as-is. The workspace is edition 2024 (`edition = "2024"` in the root `Cargo.toml`), and `size_of`/`align_of` have been in the prelude since Rust 1.80, so no import is needed. Verified: `cargo build -p plurality` finishes clean. Adding an explicit `use core::mem::{size_of, align_of}` would in fact trip the workspace's `redundant_imports` lint (enabled per M-STATIC-VERIFICATION), so it must stay unimported.

crates/plurality/benches/graph_churn.rs
```diff
@@ -0,0 +1,266 @@
...
+//! Both runs build identical node contents, in identical slab positions, in the
+//! identical order — verified by an end-to-end checksum that must match. The
+//! workload also chases graph edges on every insert, so it exercises memory
+//! locality (where the pool's contiguous chunks tend to win) as well as raw
+//! allocation throughput.
+//!
+//! Run with: `cargo bench --bench graph_churn`.
+
+use core::ops::Deref;
+use std::time::{Duration, Instant};
```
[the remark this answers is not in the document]
geeknoid: Same as the `chunk.rs` note: edition 2024 puts `size_of` in the prelude, so the bench compiles without an import. Verified with `cargo build -p plurality --benches` (clean). An explicit import would trip the workspace `redundant_imports` lint.

crates/plurality/README.md
```diff
@@ -0,0 +1,155 @@
+<div align="center">
+ <img src="./logo.png" alt="Plurality Logo" width="96">
+
+# Plurality
```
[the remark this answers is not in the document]
geeknoid: Good catch, and the absolute-URL approach is correct per `docs/packaging-guidelines.md`. This isn't plurality-specific though: the logo line comes from the shared `crates/README.j2` template (``), so every crate's generated README — multitude included — has the same relative reference, and `logo.png` is intentionally excluded from the packaging allowlist. Fixing it means updating the shared template and regenerating all crate READMEs (otherwise `readme-check` breaks for the rest), which is a workspace-wide change best done separately rather than bundled into this crate-introduction PR. The maintainer is tracking it separately.

crates/plurality/src/pool.rs
```diff
@@ -0,0 +1,869 @@
...
+/// The pool refcount must have just reached zero; the pool is quiescent.
+#[cold]
+#[inline(never)]
+unsafe fn teardown<T, A: Allocator>(pool: NonNull<PoolInner<T, A>>) {
+    // SAFETY: exclusive ownership; chunks were allocated with `chunk_layout`.
+    unsafe {
+        let inner = pool.as_ref();
+        let layout = inner.chunk_layout;
+        {
+            let dir = &*inner.directory.get();
```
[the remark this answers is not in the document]
geeknoid: Fixed in b929807c. `teardown` now performs an `Acquire` load of `chunks_allocated` before walking the directory, pairing with the `Release` store `grow()` already does right after each `directory.push`. That makes the directory writes (including any `Vec` reallocation) visible to a teardown running on a non-allocator thread, instead of relying on `pool_refcount` (whose increments are `Relaxed` and so do not publish growth on their own). It is on the `#[cold]` teardown path (once per pool lifetime), so there is no hot-path cost. Loom `teardown_on_worker_thread` and Miri (strict provenance) still pass.

crates/plurality/docs/DESIGN.md
```diff
@@ -0,0 +1,316 @@
...
+`T::drop`. Every value is dropped exactly once, on its own handle's drop. So
+`Pool::drop` is not synchronous w.r.t. outstanding handles: the backing memory
+lives until the last handle drops.
+
+Teardown can run on whichever thread drops the last handle (or the `Pool`),
+which may not be the allocator thread. That is sound: `pool_refcount == 0`
+implies the `Pool` handle is gone (no more allocation/growth) and no handles
+remain, so the directory is quiescent. The `Acquire` on the final
+`pool_refcount` decrement establishes happens-before with every prior allocation
+and growth, so the teardown thread sees a complete, frozen directory to walk.
```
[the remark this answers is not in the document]
geeknoid: Fixed in b929807c. Reworded this section: the final `pool_refcount` `Acquire` is now described as establishing happens-before with prior handle drops only, and directory publication is attributed to its real mechanism — `grow()` does `chunks_allocated.store(Release)` after each `directory.push`, and `teardown` does a matching `chunks_allocated.load(Acquire)`. The note also makes explicit that `pool_refcount` increments are `Relaxed` and therefore do not publish growth.

.cargo/mutants.toml
```diff
@@ -21,4 +21,10 @@ exclude_re = [
     "impl data_privacy::RedactedDebug",
     "impl data_privacy::RedactedDisplay",
     "impl data_privacy::RedactedToString",
+
+    # plurality: equivalent/unreachable mutants (documented in its DESIGN/TODO).
+    "::builder ->",
+    "check_refcount_overflow",
```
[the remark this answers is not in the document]
geeknoid: Fixed in b929807c. Narrowed the regex to `crates/plurality/.*::builder ->` so it is anchored to the plurality path. Verified via `cargo mutants --list`: `Pool<T, Global>::builder` stays excluded, while the 10 previously-suppressed `builder()` mutants in other crates (bytesbuf_io ×4, cachet, data_privacy, fetch_tls ×4) are now listed and mutation-tested again. The other plurality-specific excludes (`check_refcount_overflow`, `impl Drop for Bomb`, `chunk_align`) were checked and are unique to plurality, so they were left as-is.

crates/plurality/src/rc.rs
```diff
@@ -0,0 +1,184 @@
...
+    slot: NonNull<SlotCell<T>>,
+    _marker: PhantomData<(T, A)>,
+}
+
+// `Rc` is `!Send + !Sync` (enforced by its `NonNull` field), which is what
+// makes the non-atomic refcount sound: an occupied slot is never on the free
+// list, so its owning thread has exclusive access. The slot's `refcount` is an
+// `AtomicU32` (it doubles as the free-list link while free), so these RMWs use
+// plain integer ops; under `--cfg loom` (no `as_ptr`) they fall back to relaxed
+// atomics.
```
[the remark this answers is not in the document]
geeknoid: Addressed in aa43d954. To be precise: `Rc<T, A>` is already `!Send + !Sync` as written — its `slot: NonNull<SlotCell<T>>` field carries neither auto trait (raw pointers are `!Send`/`!Sync`), and nothing adds an `unsafe impl Send`, so the auto-derived traits are absent. I verified empirically that `is_send::<Rc<u32>>()` fails to compile (E0277), so the non-atomic refcount path is sound today; the comment's "enforced by its `NonNull` field" was accurate. Explicit negative impls (`impl !Send`) are not an option here because `negative_impls` is unstable and this crate builds on stable.

That said, your underlying concern is valid: the guarantee rested on a private implementation detail with nothing locking it in. So I added compile-time guards using the workspace's `static_assertions`:

```rust
assert_not_impl_any!(plurality::Rc<u32>: Send, Sync);
assert_not_impl_any!(plurality::Alloc<'static, u32>: Send, Sync);
```

(also covering `Alloc`, which shares the same single-thread soundness requirement). I confirmed the guard bites: temporarily adding `unsafe impl Send for Rc` makes the assertion fail to compile. I also reworded the `rc.rs` comment to spell out *why* the `NonNull` field is sufficient and to point at the new assertion.

crates/plurality/src/rc.rs
```diff
@@ -0,0 +1,187 @@
...
+    ///
+    /// # Safety
+    /// The value must have been fully initialized before calling.
+    #[must_use]
+    pub unsafe fn assume_init(self) -> Rc<T, A> {
+        let slot = self.slot.cast::<SlotCell<T>>();
+        // Don't run the uninit handle's destructor; transfer the slot as-is.
+        core::mem::forget(self);
+        Rc::from_slot(slot)
+    }
```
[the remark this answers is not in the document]
geeknoid: Fixed in 5be49461. Added the aliasing clause to mirror `Arc::assume_init`: "If other `Rc` clones to this slot exist, they must observe an initialized value."

crates/plurality/src/rc.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    /// Converts a pinned, uninitialized rc into a pinned, initialized one.
+    ///
+    /// # Safety
+    /// The value must have been fully initialized before calling.
+    #[must_use]
+    pub unsafe fn assume_init_pin(this: Pin<Self>) -> Pin<Rc<T, A>> {
```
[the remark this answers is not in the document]
geeknoid: Fixed in 5be49461. Same aliasing clause added to `assume_init_pin` (which delegates to `assume_init`).

README.md
```diff
@@ -44,6 +44,7 @@ These are the primary crates built out of this repo:
 - [`layered`](./crates/layered/README.md) - A foundational service abstraction for building composable, middleware-driven systems.
 - [`multitude`](./crates/multitude/README.md) - Fast and flexible arena allocator.
 - [`ohno`](./crates/ohno/README.md) - High-quality Rust error handling.
+- [`plurality`](./crates/plurality/README.md) - Pool allocator
```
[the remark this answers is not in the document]
geeknoid: Fixed in 5be49461. Added the trailing period for consistency with the surrounding crate-list entries.

crates/plurality/src/common.rs
```diff
@@ -0,0 +1,176 @@
...
+            /// fixed at allocation and the handle keeps the slot alive at that
+            /// same address until the value is dropped — exactly `Pin`'s
+            /// contract.
+            #[must_use]
+            #[inline]
+            pub fn into_pin(this: Self) -> core::pin::Pin<Self> {
+                // SAFETY: the value's address is fixed for the handle's lifetime
+                // (the pool never moves an occupied slot), satisfying `Pin`.
+                unsafe { core::pin::Pin::new_unchecked(this) }
+            }
```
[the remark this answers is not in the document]
geeknoid: Not changed — the suggestion would not compile. `Pin::new` requires the *pointee* `T: Unpin` (`impl<Ptr: Deref<Target: Unpin>> Pin<Ptr>`), not the handle. `into_pin` must accept `!Unpin` values (its whole purpose — the pool guarantees address stability so `!Unpin` types can be pinned), so `Pin::new(this)` fails with "T cannot be unpinned". The handle being `Unpin` is irrelevant to that bound; `Pin::new_unchecked` is required. I expanded the SAFETY comment to explain why.

crates/plurality/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+
+//! A growable, fixed-slot object pool that hands out thin smart pointers.
+//!
+//! A [`Pool<T>`] allocates `T` values from reusable slots and returns
+//! single-pointer-wide smart pointers that deref to `&T`. It grows on demand and
+//! never moves a value once allocated, so the pointers stay valid until they are
+//! dropped. There are four handle types, covering owned vs. shared and bound vs.
+//! `'static`:
+//!
+//! - [`Box<T>`] — unique owner, `Send`, may outlive the pool.
```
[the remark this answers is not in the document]
geeknoid: Fixed in 1f9ec310. The bullet now reads "`Send` when `T: Send` and `A: Send + Sync`", matching the `unsafe impl` bounds. README regenerated (`just readme`).

crates/plurality/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+//! - [`Alloc<'pool, T>`] — unique owner that borrows the pool; the cheapest
+//!   handle, but `!Send` and cannot outlive the pool.
+//! - [`Arc<T>`] — shared, atomically reference-counted, `Send + Sync`.
```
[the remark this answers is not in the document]
geeknoid: Fixed in 1f9ec310. The bullet now reads "`Send + Sync` when `T: Send + Sync` and `A: Send + Sync`", matching the impl bounds. README regenerated.

.cargo/mutants.toml
```diff
@@ -21,4 +21,12 @@ exclude_re = [
     "impl data_privacy::RedactedDebug",
     "impl data_privacy::RedactedDisplay",
     "impl data_privacy::RedactedToString",
+
+    # plurality: equivalent/unreachable mutants (documented in its DESIGN/TODO).
+    # Anchored to the plurality path so it does not suppress `builder()` mutants
+    # in other crates.
+    "crates/plurality/.*::builder ->",
+    "check_refcount_overflow",
+    "impl Drop for Bomb",
```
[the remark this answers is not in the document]
geeknoid: Undid changes to mutants.toml.

crates/plurality/docs/PERF.md
```diff
@@ -0,0 +1,62 @@
...
+| `Rc` — `alloc_uninit_rc` | 16.15 ns | 59 | 96 | 270 |
+
+### Clone + drop (shared handles)
+
+| Operation | Time / op | Instructions | Mem accesses | Est. cycles |
+|---|---:|---:|---:|---:|
+| `Arc` clone + drop | 4.89 ns | 25 | 41 | 177 |
+| `Rc` clone + drop | 1.29 ns | 26 | 41 | 143 |
+
+## Cross-crate comparison (allocate + free)
```
martin-kolinek, line 38: How does multitude do in comparison? I would expect that we should be able to allocate values of the same type from a multitude Arena as well, right? Is it worth it to have the additional implementation rather than reuse multitude underneath?

crates/plurality/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+mod boxed;
+mod builder;
+mod chunk;
+mod common;
+mod pool;
+mod rc;
+mod slot;
+mod sync;
+
+pub use allocator_api2::alloc::AllocError;
```
ralfbiedert: Why are you re-exporting this error type? Your own `try_` should use their own error type.

crates/plurality/Cargo.toml
```diff
@@ -0,0 +1,80 @@
...
+
+[dependencies]
+allocator-api2 = { workspace = true, features = ["alloc"] }
+
+# Loom is only used under `--cfg loom` for the model-checked tests.
+[target.'cfg(loom)'.dependencies]
+loom = { workspace = true }
+
+[dev-dependencies]
+allocator-api2 = { workspace = true, features = ["alloc"] }
```
ralfbiedert, line 39: Based on what we saw in multitude, it might make sense to add a test that pools only ever allocate once at creation for operations up to 'chunk' size.

crates/plurality/src/rc.rs
```diff
@@ -0,0 +1,189 @@
...
+}
+
+// `Rc` is `!Send + !Sync`: its `NonNull<SlotCell<T>>` field carries neither auto
+// trait (raw pointers are `!Send`/`!Sync`) and it adds no `unsafe impl Send`, so
+// no explicit negative impl is needed — and a compile-time `assert_not_impl_any!`
+// in `tests/smart_ptr.rs` locks the property in against future refactors. This is
+// what makes the non-atomic refcount sound: an occupied slot is never on the free
+// list, so its owning thread has exclusive access. The slot's `refcount` is an
+// `AtomicU32` (it doubles as the free-list link while free), so these RMWs use
+// plain integer ops; under `--cfg loom` (no `as_ptr`) they fall back to relaxed
```
ralfbiedert: I read the text, but I don't understand the loom reasoning, and the conditional atomic / non-atomic compilation flag looks very sus.

crates/plurality/src/builder.rs
```diff
@@ -0,0 +1,134 @@
...
+            mask: chunk_size - 1,
+            max_chunks: self.max_chunks,
+            chunks_allocated: AtomicU32::new(0),
+            chunk_layout: layout,
+            directory: UnsafeCell::new(Vec::new()),
+            allocator: self.allocator,
+        };
+        let raw = AllocBox::into_raw(AllocBox::new(inner));
+        // SAFETY: `into_raw` never returns null.
+        Pool::from_inner(unsafe { NonNull::new_unchecked(raw) })
```
ralfbiedert: I don't think this and similar should be unsafe unchecked operations, this doesn't look performance sensitive, and a regular .new().except() should work just fine.

ralfbiedert (approved):

crates/plurality/src/builder.rs
geeknoid: Done.

crates/plurality/src/rc.rs
geeknoid: Updated, should be more clear now.

crates/plurality/src/lib.rs
geeknoid: Now have a new AllocError type defined in the crate.

crates/plurality/Cargo.toml
geeknoid: Done.

crates/plurality/docs/PERF.md
geeknoid: Those are two very different allocators:

* plurality is designed for steady state use where you constantly allocate and free objects. Manipulating a graph is a good example of where you need to allocate and free a bunch of nodes over time. Plurality is implemented via a singly-linked list of objects held in chunks. Finally, a pool is homogeneous, all objects are the same type. 

* multitude on the other hand is meant to 'contextual allocations', where you build up a bunch of allocations which all get freed at the same time. Multitude is implemented by bump allocating through chunks. An Arena hold heterogeneous data.

They look superficially similar on the outside, but the lifetime of the memory is ultimately different. An Arc in multitude keeps a whole chunk alive and even if you drop the Arc, the underlying chunk will remain allocated until all Arcs on the chunk are freed. When you drop an Arc in plurality, it's memory becomes immediately available for reuse.

So I think they deserve separate APIs and implementaitons.

martin-kolinek (approved):

crates/plurality/src/pool.rs
```diff
@@ -0,0 +1,877 @@
...
+
+/// Pushes a freed slot back onto the MPSC free list and returns the owning
+/// pool. The load only needs a recent head value (a stale one just makes the CAS
+/// retry); the CAS `Release` publishes the link store to the consumer's
+/// `Acquire` load on pop.
+///
+/// # Safety
+/// `slot` must be an occupied slot whose value has already been dropped.
+#[inline]
+unsafe fn push_free<T, A: Allocator>(slot: NonNull<SlotCell<T>>) -> NonNull<PoolInner<T, A>> {
```
martin-kolinek, line 753: Benchmark gap: the crate's headline capability is concurrent cross-thread frees — one allocating thread while many `Send` handles are freed from other threads, all racing on this lock-free `push_free` Treiber stack (`Relaxed` load + `Release` CAS) against `alloc_slot`'s `Acquire`/`AcqRel` pop. Every bench in `benches/` (criterion_alloc, gungraun_alloc, pool_comparison, graph_churn) is single-threaded, so this MPSC free-list contention path — where the hand-tuned orderings, CAS-retry behavior, and any false sharing on `free_head`/slot `refcount` actually show up — is entirely unmeasured. A future ordering change or a field-layout tweak causing false sharing would regress silently. Consider one wall-clock bench that pre-warms a pool, then spawns K=2/4/8 threads draining `Send` `Box`/`Arc` handles produced by a single allocator thread, reporting frees/s vs. thread count (reusing the existing `Obj`/`setup_pool` helpers). Wall-clock only — Callgrind serializes threads, so gungraun isn't meaningful for contention.

crates/plurality/src/pool.rs
```diff
@@ -0,0 +1,877 @@
...
+    // SAFETY: exclusive ownership; chunks were allocated with `chunk_layout`.
+    unsafe {
+        let inner = pool.as_ref();
+        let layout = inner.chunk_layout;
+        // Synchronize with `grow()`'s `chunks_allocated.store(Release)`, which is
+        // sequenced after each `directory.push`. Without this acquire, a teardown
+        // running on a thread that never observed `grow()` directly could read a
+        // stale or partially published directory (its `pool_refcount` increments
+        // are `Relaxed`, so they do not publish the directory on their own).
+        let _ = inner.chunks_allocated.load(Acquire);
```
martin-kolinek, line 854: 🤖 Please add a Loom regression where a worker receives an early handle before the allocator thread grows several more chunks, then becomes the last owner after those later handles and the `Pool` are dropped. The current `teardown_on_worker_thread` grows before `thread::spawn`, so spawn synchronization publishes the directory and the test cannot fail if this Acquire is removed. Using only relaxed test coordination plus a custom allocator/deallocation counter would verify that worker-thread teardown observes and frees every later chunk.

crates/plurality/src/chunk.rs
```diff
@@ -0,0 +1,81 @@
...
+/// Alignment of a whole chunk allocation.
+#[cfg_attr(test, mutants::skip)] // `>` vs `>=` is equivalent here: when `ha == sa` both arms return the same value.
+const fn chunk_align<T, A>() -> usize {
+    let ha = align_of::<ChunkHeader<T, A>>();
+    let sa = align_of::<SlotCell<T>>();
+    if ha > sa { ha } else { sa }
+}
+
+/// Computes the [`Layout`] of a chunk holding `n` slots, or `None` on overflow.
+pub(crate) fn chunk_layout<T, A>(n: usize) -> Option<Layout> {
```
martin-kolinek, line 40: 🤖 Please add an integration/Miri test that allocates an over-aligned value (for example `#[repr(align(64))]`) across multiple slots and chunks, checks each exposed value pointer's alignment, then drops and reuses the slots. The current tests only use ordinarily aligned values, so a regression in the generic payload offset/alignment or header-recovery arithmetic could escape while breaking valid `Pool<T>` types.

crates/plurality/src/pool.rs
```diff
@@ -0,0 +1,877 @@
...
+/// Pushes a freed slot back onto the MPSC free list and returns the owning
+/// pool. The load only needs a recent head value (a stale one just makes the CAS
+/// retry); the CAS `Release` publishes the link store to the consumer's
+/// `Acquire` load on pop.
+///
+/// # Safety
+/// `slot` must be an occupied slot whose value has already been dropped.
+#[inline]
+unsafe fn push_free<T, A: Allocator>(slot: NonNull<SlotCell<T>>) -> NonNull<PoolInner<T, A>> {
+    // SAFETY: recovery is valid for any live slot from this crate.
```
martin-kolinek, line 754: 🤖 `SAFETY: recovery is valid for any live slot from this crate` is not trivially justified for handles from the uninit-placement APIs. `try_alloc_uninit_box`/`_arc`/`_rc`/`_local` cast the slot to `SlotCell<MaybeUninit<T>>`, so dropping such a handle *without* `assume_init` (a fully safe op, e.g. `let _ = pool.alloc_uninit_box();`) runs this free path monomorphized at `MaybeUninit<T>`: `header_of::<MaybeUninit<T>, A>` reads `(*header).pool` as `NonNull<PoolInner<MaybeUninit<T>, A>>` and then `pool.as_ref()` dereferences `free_head`/`pool_refcount` (and possibly `teardown`). But the allocation actually contains `PoolInner<T, A>`. `PoolInner` is `repr(Rust)`, so equal field offsets between `PoolInner<T, A>` and `PoolInner<MaybeUninit<T>, A>` are not language-guaranteed (the `MaybeUninit` same-layout guarantee covers the value itself, not a `repr(Rust)` container generic over it). Consider type-erasing the free/teardown operation, or normalizing the slot back to `SlotCell<T>` before freeing, so pool metadata is never re-typed as `PoolInner<MaybeUninit<T>, A>`.

crates/plurality/src/pool.rs
```diff
@@ -0,0 +1,877 @@
...
+    unsafe {
+        let index = (*slot.as_ptr()).index;
+        let header = header_of::<T, A>(slot, index);
+        let pool = (*header.as_ptr()).pool;
+        let global = (*header.as_ptr()).base_index + index;
+        let inner = pool.as_ref();
```
martin-kolinek, line 760: 🤖: Uninitialized-handle reclamation dereferences the pool as the wrong monomorphization — UB reachable from 100% safe code. `alloc_uninit_box`/`_arc`/`_alloc`/`_rc` hand out a handle over `SlotCell<MaybeUninit<T>>` (cast from `SlotCell<T>`). Dropping it *without* `assume_init` (supported and tested — `uninit_box_dropped_without_init`, `uninit_handle_outlives_pool`) runs `free_slot::<MaybeUninit<T>, A>` → `push_free::<MaybeUninit<T>, A>`, so `pool.as_ref()` here reads the `ChunkHeader` back-pointer — created as `PoolInner<T, A>` by `grow()`/`build()` — as `PoolInner<MaybeUninit<T>, A>`. `ChunkHeader`/`SlotCell` are `repr(C)` so the recovered pointer value is correct, but `PoolInner` (pool.rs:26) is `repr(Rust)`, and Rust does not guarantee equal field offsets across two monomorphizations even when every field is pairwise size/align-identical. Under `-Zrandomize-layout` the `free_head`/`pool_refcount` accesses (and `teardown` for an outliving handle) then read/write the wrong fields — corrupting `PoolInner`, or invalid-free / use-after-free on teardown. The `// SAFETY: recovery is valid for any live slot` comment proves liveness but silently assumes the generic parameter matches the allocation. Fix by type-erasing the back-pointer in `ChunkHeader` (a non-generic `#[repr(C)]` pool-core pointer + a concrete type-erased teardown callback) so reclamation never depends on `repr(Rust)` layout agreement across `T`/`MaybeUninit<T>`.

crates/plurality/src/pool.rs
```diff
@@ -0,0 +1,877 @@
...
+    /// `chunk_size` allocations.
+    #[cold]
+    #[inline(never)]
+    fn grow(&self) -> Result<NonNull<SlotCell<T>>, AllocError> {
+        let inner = self.inner();
+        let chunks = inner.chunks_allocated.load(Relaxed);
+        let n = inner.chunk_size;
+        // Cap = the user's `max_chunks`, or for an unbounded pool the chunk count
+        // that keeps every global index below the `FREE_END` sentinel.
+        let cap = inner.max_chunks.unwrap_or(FREE_END / n);
```
martin-kolinek, line 666: 🤖: This unbounded-pool cap only enforces the `FREE_END` (`u32::MAX`) slot-index ceiling, not the `pool_refcount` ceiling that `build()` applies for bounded pools (`min(FREE_END, usize::MAX - 1)`, builder.rs:110). `pool_refcount` (an `AtomicUsize`) holds `1 + live refcounted allocations`; on a 32-bit target an unbounded pool with `chunk_size == 1` permits up to `u32::MAX` live `Box`/`Arc`/`Rc`, so the final bump wraps the counter and breaks `len()`/teardown — the same overflow-by-one that was just closed on the bounded path. Derive this cap from the same `min(FREE_END, usize::MAX - 1)` bound so both paths agree.

crates/plurality/src/common.rs
```diff
@@ -0,0 +1,178 @@
...
+            /// same address until the value is dropped — exactly `Pin`'s
+            /// contract.
+            #[must_use]
+            #[inline]
+            pub fn into_pin(this: Self) -> core::pin::Pin<Self> {
+                // SAFETY: the value's address is fixed for the handle's lifetime
+                // (the pool never moves an occupied slot), satisfying `Pin` even
+                // for `!Unpin` `T`. `Pin::new` is unusable here: it requires the
+                // pointee `T: Unpin`, but this must accept `!Unpin` values.
+                unsafe { core::pin::Pin::new_unchecked(this) }
```
martin-kolinek, line 44: 🤖: `into_pin` is exposed for the shared `Arc`/`Rc` handles too, and there it is unsound. Because the handles are `Unpin`, this is a safe conversion from an arbitrary (possibly aliased) handle into `Pin<Self>`. For a `!Unpin` `T`: clone an `Arc`, `into_pin` one clone and establish a self-reference through the pinned value, drop the pinned clone, then call `Arc::get_mut` on the surviving now-unique clone and `mem::replace` the value — moving a value that was pinned (UB); the identical path exists for `Rc`. The SAFETY comment only argues the slot *address* is fixed; it does not prevent a sibling shared handle from moving the value once the pinned one is gone. `std` deliberately provides no safe `Arc<T> -> Pin<Arc<T>>` conversion for exactly this reason (only `Arc::pin` from a fresh, unique value). Keep `into_pin` for the unique `Box`/`Alloc`, but for `Arc`/`Rc` either drop it or require/verify uniqueness before constructing the `Pin`.

crates/plurality/tests/smart_ptr.rs
```diff
@@ -0,0 +1,321 @@
...
+    let mut l = pool.alloc(String::from("x"));
+    AsMut::<String>::as_mut(&mut l).push('y');
+    BorrowMut::<String>::borrow_mut(&mut l).push('z');
+    assert_eq!(&*l, "xyz");
+}
+
+// ── as_ptr / as_mut_ptr ──
+
+#[test]
+fn as_ptr_points_at_the_value_and_is_stable() {
```
martin-kolinek, line 78: 🤖: This test only compares two ways of obtaining the pointer *before any pool growth*, so a regression that relocated existing values while growing (new chunks / directory reallocation) would still pass. Please extend it (or add a sibling test) to exercise the crate's headline address-stability guarantee: keep this first handle/pointer, allocate enough additional live values to force several new chunks and directory growth, then assert the original pointer is unchanged and still reads its original value.

crates/plurality/src/slot.rs
```diff
@@ -0,0 +1,124 @@
...
+/// A single slot in a chunk: storage for one `T`, an atomic refcount that
+/// doubles as the free-list link when the slot is free, and the slot's
+/// immutable in-chunk index.
+#[repr(C)]
+pub(crate) struct SlotCell<T> {
+    /// The value. Initialized only while the slot is occupied.
+    pub(crate) value: UnsafeCell<MaybeUninit<T>>,
+    /// Dual-role:
+    /// - Occupied: the reference count (`>= 1`) for a shared handle (`Arc`/`Rc`);
+    ///   a `Box` sets it to `1` but never reads it. An `Alloc` neither writes
```
martin-kolinek, line 62: 🤖: This contract is inaccurate for `Box`: `occupy_box` (pool.rs:565) writes the value **without** initializing the slot refcount, so `Box` does *not* set it to `1` — it leaves the stale free-list link in place, exactly like `Alloc`, and `push_free` overwrites it on drop. Please reword so `Box` neither writes nor reads this field (only `Arc`/`Rc` maintain the `>= 1` count). DESIGN.md mirrors the same wrong claim (`Box` keeps it at 1 implicitly, line 105) and should be narrowed the same way.

crates/plurality/docs/DESIGN.md
```diff
@@ -0,0 +1,324 @@
...
+only the producers' pushes on `free_head`.
+
+**Pop** (allocator thread, single consumer):
+1. `g = free_head.load()`. If `g == FREE_END`, grow.
+2. Map `g -> slot`; `next = slot.refcount.load()` (read as link).
+3. `compare_exchange_weak(free_head, g -> next)`; retry on contention. ABA-free
+   because only this thread pops and a free slot is never re-pushed while free.
+
+**Grow** (allocator thread, rare, `#[cold]`): `grow()` returns
+`Option<NonNull<SlotCell<T>>>` — the reserved first slot of the new chunk, or
```
martin-kolinek, line 217: 🤖: `grow()` now returns `Result<NonNull<SlotCell<T>>, AllocError>`, not `Option<...>`, so capacity exhaustion and allocator failure stay distinguishable through the public fallible APIs. Please update this section (and the later `returns None` wording) to describe the `Result`/`AllocError` contract.

crates/plurality/docs/DESIGN.md
```diff
@@ -0,0 +1,324 @@
...
+`alloc` only — no `std`, no `Mutex`. Allocation/growth are single-threaded and
+the free list is a lock-free MPSC stack, so only `core::sync::atomic` is needed.
+Chunk allocations go through `allocator-api2` so custom allocators compose.
+
+## Verification strategy
+
+The crate is checked by several complementary tools (see `tests/` and
+`benches/`):
+
+- **Unit + integration tests** (`tests/integration.rs`, `tests/coverage.rs`) —
```
martin-kolinek, line 306: 🤖: These test files don't exist in this changeset — `tests/integration.rs` and `tests/coverage.rs` are not present. The integration coverage now lives in `tests/{alloc,arc,box,pool,rc,smart_ptr,alloc_tracking}.rs` (plus `loom.rs`/`bolero.rs`). Please update the verification map to point at the suites that actually implement the stated coverage.

crates/plurality/src/pool.rs
```diff
@@ -0,0 +1,877 @@
...
+        }
+        inner.chunks_allocated.store(chunks + 1, Release);
+
+        // Splice the free slots (base_index+1 .. base_index+n-1) onto the head;
+        // slot `base_index` is returned to the caller.
+        if n > 1 {
+            // SAFETY: `ptr` chunk is live; its last slot is index n-1.
+            let last = unsafe { slot_at::<T, A>(ptr, (n - 1) as usize) };
+            // SAFETY: `last` is the new chunk's (still-private) final slot.
+            unsafe { splice_chain(&inner.free_head, last, base_index + 1) };
```
martin-kolinek, line 705: 🤖: The `splice_chain` CAS-retry branch reached from here only fires when a concurrent free races this otherwise single-threaded splice (as its own coverage-off comment states), and no test exercises it: `contended_free_list_exercises_cas_retries` uses `chunk_size(1)`, so `grow()` never enters this `n > 1` splice. Consider a Loom regression that pauses second-chunk growth, drops an old handle on another thread to push onto `free_head`, then fills the pool — asserting no extra chunk was needed would catch a lost free-list head in this splice.

crates/plurality/src/pool.rs
```diff
@@ -0,0 +1,877 @@
...
+/// the `Box`/`Arc`/`Rc` drop paths.
+///
+/// # Safety
+/// `slot` must be an occupied, initialized slot whose last handle is being
+/// released; its value must not be accessed afterwards.
+#[inline]
+pub(crate) unsafe fn drop_and_free<T, A: Allocator>(slot: NonNull<SlotCell<T>>) {
+    // SAFETY: last handle to an occupied slot (caller contract).
+    unsafe {
+        SlotCell::drop_value(slot);
```
martin-kolinek, line 816: 🤖: The closure-panic path during *allocation* is carefully RAII-guarded (`try_alloc_*_with` frees the slot on unwind), but the *destructor*-panic path during release is not. If `SlotCell::drop_value(slot)` panics here, `free_slot` never runs, so the slot is never returned to the free list **and** `pool_refcount` is never decremented — permanently leaking that capacity for the pool's lifetime and preventing teardown. There is no functional regression for this: please add a bounded-capacity test (in `tests/pool.rs`) that drops a value whose `Drop` panics, catches the unwind, and asserts the slot can be re-allocated afterward — covering the last-owner `Box`/`Arc`/`Rc` path here and the analogous `Alloc` path in `drop_and_free_local`.

crates/plurality/docs/DESIGN.md
```diff
@@ -0,0 +1,324 @@
...
+- `Drop` → decrement refcount; on reaching 0, recover the header to reach
+  `free_head` (push the derived global index) and, for `Box`/`Arc`/`Rc`,
+  decrement `pool_refcount`. `Alloc` drop pushes the slot but does **not** touch
+  `pool_refcount`.
+
+### The four handle flavors
+
+| Handle | Outlives pool? | `Send`/`Sync` | Per-slot refcount | `pool_refcount` traffic |
+|---|---|---|---|---|
+| `Box<T>` | yes | `Send` (if `T: Send`) | none (unique) | bump on alloc, drop on free |
```
martin-kolinek, line 166: 🤖: This handle table omits the allocator bound on the `Send`/`Sync` columns. `Box<T, A>` is `Send` only when `T: Send` **and** `A: Send + Sync` (boxed.rs), and `Arc<T, A>` (line 168) is `Send + Sync` only when `T: Send + Sync` **and** `A: Send + Sync` (sync.rs). This is the same imprecision that was corrected in the crate rustdoc (lib.rs) and the `Pool` doc (pool.rs) during review — the DESIGN.md table was left behind and now contradicts them. Add the `A: Send + Sync` condition to the `Box` and `Arc` rows to keep the docs consistent.

crates/plurality/Cargo.toml
```diff
@@ -0,0 +1,75 @@
...
+alloc_tracker = "0.6"
+allocator-api2 = { workspace = true, features = ["alloc"] }
+bolero = { workspace = true, features = ["std"] }
+criterion = { workspace = true }
+mimalloc = { workspace = true }
+mutants = { workspace = true }
+static_assertions = { workspace = true }
+
+[target.'cfg(loom)'.dev-dependencies]
+loom = { workspace = true }
```
martin-kolinek, line 47: 🤖: `tests/loom.rs` ships here, but unlike the sibling `multitude` crate the loom target isn't wired for tooling discovery. On `main`, `multitude/Cargo.toml` declares a marker `loom = []` feature plus a `[[test]] name = "loom" path = "tests/loom.rs" required-features = ["loom"]` entry, and its comment notes this is precisely what lets `anvil-loom` discover the target authoritatively (the tests still build under `--cfg loom`). As written, plurality's loom suite is only reachable via ad-hoc `--cfg loom` and isn't advertised to that tooling. Consider adding the `loom = []` marker feature to `[features]` and the matching `[[test]]` entry to align with the established allocator-crate pattern.

crates/plurality/src/common.rs
martin-kolinek, line 44: 🤖: `into_pin` / `From<Self> for Pin<Self>` are emitted for the shared handles too (`Arc` via `sync.rs`, `Rc` via `rc.rs`), and there this blanket safe conversion is unsound. `Pin::new_unchecked` on a `!Unpin` pointee promises the value never moves for the rest of its life — across *every* alias, even after this `Pin` is dropped — but a shared handle can have an ordinary un-pinned clone that outlives the pin. Reachable safe UB:

```rust
let a = pool.alloc_arc(x);
let mut b = a.clone();
let p = Arc::into_pin(a);              // install self-refs relying on the pinned address
drop(p);
let m = Arc::get_mut(&mut b).unwrap(); // refcount==1, no "never pinned" gate
let _ = core::mem::replace(m, y);      // moves a previously pinned value -> dangling self-refs
```

`get_mut` (sync.rs / rc.rs) hands out `&mut T` at refcount 1 with no never-pinned gate, so self-references dangle → UB; `assume_init_pin` inherits the same hole. The conversion is only sound for the unique owners (`Box`/`Alloc`). Restrict the safe `into_pin`/`From` to unique owners and gate the shared conversion behind a path that proves a fresh, alias-free owner (e.g. pool-side pinned constructors). Worth an alias-sensitive Miri / compile-fail regression: clone an `Arc`/`Rc`, pin one clone, drop it, and confirm the survivor cannot move the value through `get_mut`.

crates/plurality/src/pool.rs
```diff
@@ -0,0 +1,877 @@
...
+use crate::builder::PoolBuilder;
+use crate::chunk::{ChunkHeader, header_of, slot_at};
+use crate::error::AllocError;
+use crate::rc::Rc;
+use crate::slot::{FREE_END, SlotCell};
+use crate::sync::Arc;
+
+/// Shared, refcounted state behind a [`Pool`]. Outlives the `Pool` handle when
+/// smart pointers are still alive.
+pub(crate) struct PoolInner<T, A> {
```
martin-kolinek, line 26: 🤖: `PoolInner` is `repr(Rust)` but the `alloc_uninit_*` paths type-pun it across monomorphizations. Handing out a `Box`/`Arc`/`Rc<MaybeUninit<T>, A>` casts the slot to `SlotCell<MaybeUninit<T>>`; dropping such a handle *without* `assume_init` — ordinary safe code, e.g. `drop(pool.alloc_uninit_box())`, plus the `try_alloc_*_with` panic path — runs `push_free::<MaybeUninit<T>, A>` (~L760), which recovers the pool via `header_of` as `&PoolInner<MaybeUninit<T>, A>` and reads `free_head`/`pool_refcount`/`directory`/`allocator`/`chunk_layout`, while the allocation was built as `PoolInner<T, A>`. `repr(Rust)` does **not** guarantee `PoolInner<T, A>` and `PoolInner<MaybeUninit<T>, A>` share field order even when every field'"'"'s (size, align) is `T`-independent; under `-Zrandomize-layout` the permutation seed depends on the generic args, so the offsets can differ → the free-list link / refcount read/written at the wrong offset → corrupted free list (double hand-out of a slot → aliasing/UAF) or a wrong `Layout` passed to `deallocate` at teardown. `SlotCell` and `ChunkHeader` are already `#[repr(C)]` for exactly this reason; no SAFETY comment establishes the cross-monomorphization layout equivalence. Fix: add `#[repr(C)]` to `PoolInner`, or recover the pool through a layout-stable, type-erased core so the uninit free/teardown path isn'"'"'t layout-dependent.

crates/plurality/src/pool.rs
martin-kolinek, line 816: 🤖: Reclamation runs only *after* `SlotCell::drop_value(slot)` returns, so a panicking `T::drop` skips `free_slot`/the pool-refcount release and permanently strands this slot (lost capacity, not just a leaked value). This is the free-path analogue of the closure-panic issue already fixed for `try_alloc_*_with`. Wrap the slot in an RAII guard before `drop_value` whose `Drop` pushes the slot and releases the pool claim (and the analogous local guard for `Alloc`), so both normal return and unwinding reclaim capacity. Please add a bounded-capacity regression that catches a destructor panic and then confirms the slot can be reallocated, covering `Box`, last-owner `Arc`/`Rc`, and `Alloc`.

crates/plurality/src/pool.rs
```diff
@@ -0,0 +1,877 @@
...
+    /// not counted; see [`len`](Self::len)).
+    #[must_use]
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    /// Free slots in already-allocated chunks. Approximate under concurrency;
+    /// like [`len`](Self::len), it does not account for live `Alloc` handles.
+    #[must_use]
+    pub fn available(&self) -> u64 {
```
martin-kolinek, line 163: 🤖: `available()` / `len()` / `is_empty()` derive occupancy from `pool_refcount`, but `Alloc` deliberately does not participate in that count. A one-slot bounded pool holding a live `pool.alloc(...)` therefore reports `available() == 1` and `is_empty() == true` even though the next allocation fails — the whole-pool accessors under-report occupancy whenever detached `Alloc` handles exist. Either track every occupied slot, or make the semantics explicit by renaming these to `refcounted_len`/`available_refcounted` (and dropping/renaming the whole-pool `is_empty`/`available`) so callers aren'"'"'t misled.

crates/plurality/src/pool.rs
martin-kolinek, line 666: 🤖: The unbounded path (`max_chunks == None`) caps growth only by `FREE_END / n`, bypassing the `min(FREE_END, usize::MAX - 1)` ceiling that `build()` applies to bounded pools. On 32-bit this can admit up to ~`FREE_END` slots, but `pool_refcount` must represent the pool'"'"'s own reference plus every live refcounted allocation (`1 + total`), so it can overflow by one at the extreme — the same overflow-by-one window the bounded `build()` fix closed, still open here. Fold the `usize::MAX - 1` term into the unbounded cap as well (e.g. `min(FREE_END / n, (usize::MAX - 1) / n)`).

crates/plurality/src/pool.rs
```diff
@@ -0,0 +1,877 @@
...
+        let index = (*slot.as_ptr()).index;
+        let header = header_of::<T, A>(slot, index);
+        let pool = (*header.as_ptr()).pool;
+        let global = (*header.as_ptr()).base_index + index;
+        let inner = pool.as_ref();
+
+        loop {
+            let head = inner.free_head.load(Relaxed);
+            (*slot.as_ptr()).refcount.store(head, Relaxed);
+            if inner.free_head.compare_exchange_weak(head, global, Release, Relaxed).is_ok() {
```
martin-kolinek, line 765: 🤖: This `compare_exchange_weak` retry on `free_head` is the only contended free path — it runs when `Send` handles (`Box`/`Arc`) are dropped concurrently on non-allocator threads, which is the crate'"'"'s defining SPMC feature. Yet all four bench suites (criterion_alloc, gungraun_alloc, graph_churn, pool_comparison) are single-threaded and only exercise the local `free_slot_local`/`drop_and_free_local` path, leaving this contended CAS — the reason this is lock-free rather than `Mutex`-guarded, and the most likely thing to regress under a future memory-ordering or `free_head` layout change — with zero throughput coverage (loom/Miri check correctness, not perf). Consider one multi-threaded "concurrent free" benchmark: an allocator thread hands `Send` handles to N workers (N = 1/2/4/8) that drop them (all frees hitting this loop), reporting frees/sec, reusing the existing `shared/ops.rs` `Obj` + `setup_pool`. It validates the SPMC scaling claim and acts as a regression tripwire for this hot atomic path.

crates/plurality/tests/bolero.rs
```diff
@@ -0,0 +1,112 @@
+// Copyright (c) Microsoft Corporation.
```
martin-kolinek, line 1: 🤖: `just bolero` only discovers fuzz targets in `tests/bolero_*.rs` and `just loom` only runs `tests/loom_*.rs` (see `justfiles/extended.just`, which globs `-Filter "bolero_*.rs"` / `"loom_*.rs"`). This crate ships its suites as `tests/bolero.rs` and `tests/loom.rs`, which don't match those globs, so both the advertised coverage-guided fuzz targets and the Loom model-checked concurrency tests are silently skipped by the project's own recipes — the crate's headline correctness verification never runs in the standard workflow. Rename to `tests/bolero_pool.rs` and `tests/loom_pool.rs` (the convention the recipes expect) and update any hard-coded references. This is distinct from the `anvil-loom` / `[[test]]`-discovery point already raised on `Cargo.toml`: even with that entry, the `just` recipes still won't pick up `loom.rs`/`bolero.rs`.

crates/plurality/CHANGELOG.md
```diff
@@ -0,0 +1,59 @@
...
+    clone/drop for single-threaded sharing (sound because an occupied slot is
+    never accessed atomically by another thread; validated with Miri).
+- Allocation API per flavor: `alloc*` / `*_with` closure variants / `*_uninit`
+  placement variants with `assume_init`, plus fallible `try_*` siblings.
+- `Arc::get_mut` / `Rc::get_mut` for uniqueness-checked mutable access.
+- `no_std` support (requires `alloc`); lock-free internals using only
+  `core::sync::atomic`.
+- `benches/graph_churn.rs`: a graph add/remove benchmark comparing the pool
+  against `std::Box` on mimalloc over an identical, checksum-verified op stream.
+- `benches/alloc_callgrind.rs`: deterministic instruction-count benchmarks of
```
martin-kolinek, line 31: 🤖: There is no `benches/alloc_callgrind.rs`; the implemented and Cargo-registered Callgrind benchmark is `gungraun_alloc` (see `benches/gungraun_alloc/` and the `[[bench]] name = "gungraun_alloc"` entry in `Cargo.toml`). Please use the shipped benchmark name so this release note doesn't point readers at a nonexistent target.

crates/plurality/src/sync.rs
```diff
@@ -0,0 +1,127 @@
...
+        core::mem::forget(self);
+        Arc::from_slot(slot)
+    }
+
+    /// Converts a pinned, uninitialized arc into a pinned, initialized one.
+    ///
+    /// # Safety
+    /// The value must have been fully initialized before calling.
+    #[must_use]
+    pub unsafe fn assume_init_pin(this: Pin<Self>) -> Pin<Arc<T, A>> {
```
martin-kolinek, line 83: 🤖: `assume_init_pin` delegates to `inner.assume_init()`, whose safety contract requires `If other `Arc` clones to this slot exist, they must observe an initialized value.` But this method's `# Safety` only documents that the value itself must be initialized, so it fails to pass that alias-wide obligation on to callers. Please mirror the alias clause from `Arc::assume_init` here (as the `Rc::assume_init_pin` docs already do), or otherwise prevent aliases.

PR #539, opened 2026-07-01
feat(bytesbuf)!: make MemoryShared thread-aware

sandersaares:
## Summary

Makes the `bytesbuf` `MemoryShared` abstraction thread-aware, closing a design gap where our thread-mobile (`Send`) memory providers were expected to be thread-aware but the abstraction did not require or express it. A provider now owns the decision of how and when to be thread-aware, and any adapter around it simply forwards relocation rather than imposing its own strategy.

## Changes

- `MemoryShared` now has `ThreadAware` as a supertrait, so the abstraction requires and expresses thread-awareness. This makes it possible to relocate erased/abstract memory providers correctly across threads.
- `MemoryShared` gains an object-safe `clone_boxed` method, blanket-provided for any `Clone` provider, so a type-erased provider can be duplicated without knowing its concrete type. Every `MemoryShared` provider is therefore required to be `Clone`.
- Redesigned `CallbackMemory` into a single `CallbackMemory<D>` provider that pairs thread-aware `data: D` with a bare `reserve_fn: fn(&D, usize) -> BytesBuf`. Because the reserve function is a bare function pointer it cannot capture anything, so all reservation state must live in `data`, which is `ThreadAware` and is relocated together with the provider. `CallbackMemory::new(data, reserve_fn)` constructs one; `data` is typically the wrapped inner memory provider, but can be any thread-aware value (e.g. a tuple pairing the provider with configuration).
- Reworked `OpaqueMemory` to own a `Box<dyn MemoryShared>`, forwarding both cloning (via `clone_boxed`) and `ThreadAware` relocation to the wrapped provider. It imposes no thread-awareness strategy of its own, leaving that decision entirely to the wrapped provider (e.g. `GlobalPool`, which already manages per-core state internally), and each clone is an independent provider.
- Gave the test providers `TransparentMemory` and `FixedBlockMemory` no-op `ThreadAware` impls, since they hold no thread-affine mutable state.
- Downstream: `http_extensions` drops its now-unnecessary `Unaware` wrapper around `OpaqueMemory`; `bytesbuf_io` test providers migrate to the new `CallbackMemory` constructor.

## Breaking changes

- `MemoryShared` gained `ThreadAware` as a supertrait and a new `clone_boxed` method, so implementors must now also be `ThreadAware` (and `Clone`, which the blanket impl requires).
- `CallbackMemory`'s constructor changed from taking a capturing `Fn(usize) -> BytesBuf` closure to taking thread-aware `data` plus a bare `fn(&D, usize) -> BytesBuf`. State that a closure would previously have captured must now be passed as `data`.

## Validation

- `just format`, `just readme`, and `just spellcheck` all clean.
- Tests pass for `bytesbuf`, `bytesbuf_io`, and `http_extensions` (unit + doctests), including new tests covering `CallbackMemory`/`OpaqueMemory` relocation forwarding, non-`Debug` callback data, and the no-op `relocate` on the test providers.
- Clippy `-D warnings` clean; all remaining dependents build.
- Mutation testing of the changed logic (`cargo mutants --in-diff`) leaves no surviving mutants.


crates/bytesbuf/src/mem/opaque_memory.rs
```diff
@@ -1,24 +1,33 @@
...
 pub struct OpaqueMemory {
-    inner: Arc<dyn Memory + Send + Sync + 'static>,
+    inner: thread_aware::Arc<dyn MemoryShared, PerCore>,
 }
 
 impl OpaqueMemory {
     /// Creates a new instance of the adapter.
+    ///
+    /// The wrapped provider must be [`Clone`] so that a thread-local instance can be materialized
+    /// per thread, preserving thread-awareness across relocations.
```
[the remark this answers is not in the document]
sandersaares: Fixed

martintmk (approved):

crates/bytesbuf/src/mem/global.rs
```diff
@@ -35,18 +35,11 @@ use crate::mem::{Block, BlockRef, BlockRefDynamic, BlockRefVTable, BlockSize, Me
 /// via the `thread_local!` macro.
 ///
 /// [thread-aware]: https://docs.rs/thread_aware
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, thread_aware::ThreadAware)]
```
ralfbiedert: Small nit, it's a bit weird this is used (or rather not `use`'ed) inconsistently.

crates/bytesbuf/src/mem/memory_shared.rs
```diff
@@ -1,13 +1,38 @@
...
 ///
 /// This is a narrowing of [`Memory`] that adds additional constraints that enable
 /// thread-safe shared access to the memory provider. If you do not need these extra
 /// constraints, just use [`Memory`] directly.
-pub trait MemoryShared: Memory + Send + Sync + 'static {}
+///
+/// # Thread awareness
+///
+/// A memory provider shared across threads is expected to be [thread-aware][ThreadAware], so it can
+/// relocate any thread-affine state when moved between threads and avoid contention on
```
ralfbiedert: What does that mean for me as a user, like what is thread-affine state I should relocate? It couldn't be any memory handed out?

crates/bytesbuf/src/mem/wrapping_memory.rs
```diff
@@ -0,0 +1,264 @@
...
+///     };
+///     pool.reserve(adjusted)
+/// });
+///
+/// let buf = memory.reserve(64);
+/// assert!(buf.capacity() >= 64);
+/// ```
+///
+/// For a complete implementation pattern, see `examples/bb_has_memory_optimizing.rs`.
+pub struct WrappingMemory<T, FReserve>
```
ralfbiedert: `WrappedMemory`?

crates/bytesbuf/src/mem/wrapping_memory.rs
```diff
@@ -0,0 +1,264 @@
...
+///
+/// This can be used to construct memory providers that add logic or configuration on top of an
+/// existing memory provider. The wrapped provider is owned by this type (not captured by the
+/// closure), so that its thread-affine state is relocated correctly when this type is moved between
+/// threads: [`ThreadAware`] relocation is forwarded to the inner provider.
+///
+/// # Inert closures
+///
+/// The closure must capture only inert configuration (e.g. flags, sizes) and never any thread-affine
+/// state such as another memory provider. Thread-affine state belongs in the wrapped `inner`
```
ralfbiedert: Would it be better to (also) say it should not capture any references or pointers? I find 'inert' configuration a bit of a weird term.

ralfbiedert: Alternatively, it could say it should only capture primitive or 'stack only' state?

crates/bytesbuf/src/mem/wrapping_memory.rs
```diff
@@ -0,0 +1,264 @@
...
+use std::sync::Arc;
+
+use thread_aware::ThreadAware;
+use thread_aware::affinity::Affinity;
+
+use crate::BytesBuf;
+use crate::mem::{Memory, MemoryShared};
+
+/// Wraps an inner [`MemoryShared`] provider, customizing reservations via a closure.
+///
```
ralfbiedert: Also, why would I need this? I feel the example below is a bit made up.

ralfbiedert (approved):

crates/bytesbuf/src/mem/memory_shared.rs
sandersaares: Legit questions. I would really like for us to have some documentation that I could link here for "what is a thread aware type and how to build one" but we really have nothing. `thread_aware` crate docs are impenetrable for a random reader, unfortunately. This chapter here is really just a stand-in for some link to such reference documentation.

I will try reword this if I can think of something but ultimately this needs a holistic solution. Opened https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7552151

crates/bytesbuf/src/mem/wrapping_memory.rs
sandersaares: Insufficient data - we have not yet implemented enough custom memory provider logic to fully understand the patterns here. It exists as a simple primitive that is likely to be useful but exact scope of usefulness is unclear.

crates/bytesbuf/src/mem/wrapping_memory.rs
sandersaares: Solved it by refactoring the API to follow example of `thread_aware::closure::Closure` so anyone familiar with that is going to know what's the deal (anyone not familiar with that will struggle but awareness of thread-awareness is a big picture problem we need to solve separately anyway).

crates/bytesbuf/src/mem/wrapping_memory.rs
sandersaares: I found a way to go back to the previous callback-oriented design (and just add thread-awareness to it a la `thread_aware::closure::Closure`), so reverting to `CallbackMemory` name as it was before.

crates/bytesbuf/src/mem/callback_memory.rs
```diff
@@ -79,114 +93,149 @@ where
...
+        (self.reserve_fn)(&self.data, min_bytes)
     }
 }
 
-impl<FReserve> Memory for CallbackMemory<FReserve>
+impl<D> Memory for CallbackMemory<D>
 where
-    FReserve: Fn(usize) -> BytesBuf + Send + Sync + 'static,
+    D: Debug + Send + Sync + 'static,
 {
```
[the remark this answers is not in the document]
sandersaares: Done. `Debug` is now implemented manually for `CallbackMemory<D>` (erasing `data` via `finish_non_exhaustive()`), so the `Memory` impl no longer requires `D: Debug`. This keeps thread-aware data that is intentionally not `Debug` usable (e.g. a tuple with a non-`Debug` field). Added a `works_with_non_debug_data` test. (commit d312b168)

crates/bytesbuf/src/mem/callback_memory.rs
```diff
@@ -1,64 +1,89 @@
...
 
 use crate::BytesBuf;
 use crate::mem::Memory;
 
-/// Implements [`MemoryShared`][crate::mem::MemoryShared] by delegating to a closure.
+/// A [`MemoryShared`][crate::mem::MemoryShared] provider that customizes reservations via a function.
+///
+/// This can be used to construct memory providers that add logic or configuration on top of an
+/// existing memory provider, without implementing a dedicated provider type.
 ///
```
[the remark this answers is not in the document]
sandersaares: Done. Rather than only rewording the doc, the constraints `D: ThreadAware + Clone + Send + Sync + 'static` are now declared on the `CallbackMemory<D>` struct itself, so it unconditionally is a `MemoryShared` provider and the doc statement is accurate. (commit 79ba5a38)

crates/bytesbuf/src/mem/opaque_memory.rs
```diff
@@ -81,4 +97,63 @@ mod tests {
...
+            fn reserve(&self, min_bytes: usize) -> crate::BytesBuf {
+                self.inner.reserve(min_bytes)
+            }
+        }
+
+        impl ThreadAware for TrackingMemory {
+            fn relocate(&mut self, _source: Option<Affinity>, _destination: Affinity) {
+                self.relocated.fetch_add(1, atomic::Ordering::SeqCst);
+            }
+        }
```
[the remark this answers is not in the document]
sandersaares: Done. `TrackingMemory::relocate` now forwards to its inner `GlobalPool` in addition to incrementing the counter, mirroring what real providers do. (commit 79ba5a38)

crates/bytesbuf/src/mem/callback_memory.rs
```diff
@@ -1,64 +1,89 @@
...
-pub struct CallbackMemory<FReserve>
-where
-    FReserve: Fn(usize) -> BytesBuf + Send + Sync + 'static,
-{
-    reserve_fn: Arc<FReserve>,
+#[derive(Clone)]
+pub struct CallbackMemory<D> {
+    data: D,
+    reserve_fn: fn(&D, usize) -> BytesBuf,
 }
```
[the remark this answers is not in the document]
sandersaares: Fixed. The rename to `WrappingMemory` was abandoned during review — the final design keeps the `CallbackMemory` name, redesigned to take thread-aware `data` plus a bare `fn(&D, usize) -> BytesBuf`. Updated the PR description to describe this accurately (no `WrappingMemory` symbol exists).

crates/bytesbuf/src/mem/opaque_memory.rs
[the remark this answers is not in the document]
sandersaares: Addressed by the redesign: `OpaqueMemory` no longer stores the provider in `thread_aware::Arc<..., PerCore>`. It now owns a `Box<dyn MemoryShared>` and imposes no thread-awareness strategy of its own, forwarding relocation to the wrapped provider. The docs now state that cloning clones the wrapped provider so each clone is independent, removing the misleading `per thread` wording.

crates/http_extensions/src/body/builder.rs
```diff
@@ -88,12 +88,13 @@ impl HttpBodyBuilder {
 
     /// Creates a new instance of [`HttpBodyBuilder`] with custom memory.
     ///
-    /// When using this method, the memory is shared across all threads as opposed
-    /// to the global per-thread memory used by [`HttpBodyBuilder::new`].
+    /// The provided memory provider is type-erased and used in place of the global per-thread
+    /// memory used by [`HttpBodyBuilder::new`]. It remains thread-aware: its thread-affine state is
+    /// relocated when the builder moves between threads.
```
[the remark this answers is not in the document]
sandersaares: Done. Reworded to state that relocation happens when the builder's `ThreadAware::relocate` is invoked (by a thread-aware transfer mechanism), not automatically on a plain move. (commit d312b168)

crates/bytesbuf/src/mem/callback_memory.rs
```diff
@@ -1,64 +1,89 @@
...
-/// # struct IoContext {
-/// #     pool: GlobalPool,
-/// # }
-/// # impl IoContext {
-/// #     fn reserve_with_config(&self, min_len: usize, _align: bool) -> BytesBuf {
-/// #         self.pool.reserve(min_len)
-/// #     }
-/// # }
+/// # use bytesbuf::mem::GlobalPool;
+/// # let inner = GlobalPool::new();
```
Vaiz, line 31: the example should not hide `inner` initialization because that would be the first thing I would look for in an example

crates/bytesbuf/src/mem/callback_memory.rs
```diff
@@ -79,114 +104,165 @@ where
...
-impl<FReserve> fmt::Debug for CallbackMemory<FReserve>
-where
-    FReserve: Fn(usize) -> BytesBuf + Send + Sync + 'static,
-{
-    #[cfg_attr(test, mutants::skip)] // We have no API contract for this.
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct(type_name::<Self>())
-            .field("reserve_fn", &"Fn(usize) -> BytesBuf")
-            .finish()
+impl<D: ThreadAware + Clone + Send + Sync + 'static> ThreadAware for CallbackMemory<D> {
```
Vaiz: I think it should work via derive macro

crates/bytesbuf/src/mem/opaque_memory.rs
```diff
@@ -1,24 +1,28 @@
...
 /// every memory reservation, so avoid this adapter if you can tolerate alternatives (generics).
-#[derive(Clone, Debug)]
+///
+/// The adapter is itself [`MemoryShared`]. It owns the wrapped provider and forwards [`ThreadAware`]
+/// relocation to it, leaving the decision of how to be thread-aware entirely with the wrapped
+/// provider. Cloning the adapter clones the wrapped provider, so each clone is independent.
+#[derive(Debug, ThreadAware)]
 pub struct OpaqueMemory {
-    inner: Arc<dyn Memory + Send + Sync + 'static>,
+    inner: Box<dyn MemoryShared>,
```
Vaiz, line 19: is it a good idea to do a deep copy on every clone?

Vaiz (approved):

crates/bytesbuf/src/mem/callback_memory.rs
sandersaares: Typically you would get the `inner` from e.g. the runtime. However, this is a runtime-agnostic crate so demonstrating realistic examples would be a bit weird because it would need to involve other crates. And I do not want to have examples that just `GlobalPool::new()` as that feels like encouraging an anti-patterns (memory pools should not be created by user code).

So I sort of agree but I could not think of a better option myself. Maybe you can offer something?

crates/bytesbuf/src/mem/opaque_memory.rs
sandersaares: All memory providers are required by API contract to be cheap to clone.

crates/bytesbuf/src/mem/callback_memory.rs
sandersaares: Fixed

crates/bytesbuf/src/mem/memory_shared.rs
```diff
@@ -1,13 +1,34 @@
...
 ///
 /// This is a narrowing of [`Memory`] that adds additional constraints that enable
 /// thread-safe shared access to the memory provider. If you do not need these extra
 /// constraints, just use [`Memory`] directly.
-pub trait MemoryShared: Memory + Send + Sync + 'static {}
+///
+/// # Thread awareness
+///
+/// Implementations must be [`ThreadAware`], which allows them to optimize their behavior when an
+/// instance is moved to a new thread.
```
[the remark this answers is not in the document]
sandersaares: Reworded: the docs now say behavior is optimized when an instance is *moved between threads via a thread-aware runtime mechanism*, rather than implying it happens on a plain move. (78259c12)

crates/bytesbuf/src/mem/memory_shared.rs
```diff
@@ -1,13 +1,34 @@
...
+pub trait MemoryShared: Memory + ThreadAware + Send + Sync + 'static {
+    /// Clones this provider into a boxed trait object.
+    ///
+    /// This lets [`OpaqueMemory`][crate::mem::OpaqueMemory] duplicate a type-erased provider without
+    /// knowing its concrete type. The returned provider is independent and owns its own
+    /// thread-affine state.
```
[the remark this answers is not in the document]
sandersaares: Dropped the explanatory paragraph entirely -- it was needless (cloning is a standard Rust feature) and over-promised independence/ownership. `clone_boxed` now simply documents that it clones the provider into a boxed trait object, leaving sharing semantics to the provider. (78259c12)

crates/bytesbuf/src/mem/opaque_memory.rs
```diff
@@ -1,24 +1,28 @@
...
 
 /// Adapter to erase the type of a [`MemoryShared`] implementation.
 ///
 /// This adapter adds some inefficiency due to additional indirection overhead for
 /// every memory reservation, so avoid this adapter if you can tolerate alternatives (generics).
-#[derive(Clone, Debug)]
+///
+/// The adapter is itself [`MemoryShared`]. It owns the wrapped provider and forwards [`ThreadAware`]
+/// relocation to it, leaving the decision of how to be thread-aware entirely with the wrapped
+/// provider. Cloning the adapter clones the wrapped provider, so each clone is independent.
```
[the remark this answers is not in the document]
sandersaares: Softened: the doc now says cloning clones the wrapped provider and *whether the clones then share any state is up to that provider*. (78259c12)

crates/bytesbuf/src/mem/opaque_memory.rs
```diff
@@ -81,4 +97,64 @@ mod tests {
...
+        assert_eq!(relocated.load(atomic::Ordering::SeqCst), 1);
+    }
+
+    #[test]
+    fn clone_is_usable_independently() {
+        let memory = OpaqueMemory::new(GlobalPool::new());
+        let mut clone = memory.clone();
+
+        // Relocating the clone must leave both the clone and the original usable, since each owns
+        // its own wrapped provider.
```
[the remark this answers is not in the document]
sandersaares: Reworded the comment to state only what the test checks -- relocating one handle leaves both the clone and the original usable -- dropping the ownership/independence claim. (78259c12)

crates/bytesbuf/src/lib.rs
```diff
@@ -339,29 +339,35 @@
 //!
 //! impl HasMemory for UdpConnection {
 //!     fn memory(&self) -> impl MemoryShared {
-//!         CallbackMemory::new({
-//!             // Cloning is cheap, as it is a service that shares resources between clones.
-//!             let io_context = self.io_context.clone();
+//!         // The I/O memory provider carries the thread-affine state and is relocated automatically
+//!         // when the returned provider moves between threads.
```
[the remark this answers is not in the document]
sandersaares: Reworded: the I/O memory provider carries thread-affine state that is relocated when the returned provider is *moved between threads via a thread-aware runtime mechanism* (no longer "automatically"). (78259c12)

crates/bytesbuf/examples/bb_optimal_path.rs
```diff
@@ -117,11 +118,12 @@ const CONNECTION_OPTIMAL_MEMORY_CONFIGURATION: MemoryConfiguration = MemoryConfi
 
 impl HasMemory for Connection {
     fn memory(&self) -> impl MemoryShared {
-        CallbackMemory::new({
-            // Cloning is cheap, as it is a service that shares resources between clones.
-            let io_context = self.io_context.clone();
+        // The I/O memory provider carries the thread-affine I/O resources and is relocated
+        // automatically when the returned provider moves between threads.
```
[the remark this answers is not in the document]
sandersaares: Reworded the example comment the same way: relocation happens when the returned provider is *moved between threads via a thread-aware runtime mechanism*. (78259c12)

crates/bytesbuf/examples/bb_has_memory_optimizing.rs
```diff
@@ -66,11 +67,14 @@ const UDP_CONNECTION_OPTIMAL_MEMORY_CONFIGURATION: MemoryConfiguration = MemoryC
 
 impl HasMemory for UdpConnection {
     fn memory(&self) -> impl MemoryShared {
-        CallbackMemory::new({
-            // Cloning is cheap, as it is a service that shares resources between clones.
-            let io_context = self.io_context.clone();
-
-            move |min_len| io_context.reserve_io_memory(min_len, UDP_CONNECTION_OPTIMAL_MEMORY_CONFIGURATION)
+        // The I/O memory provider carries the thread-affine state and is relocated automatically
+        // when the returned provider moves between threads.
```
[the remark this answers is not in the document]
sandersaares: Reworded the example comment the same way: relocation happens when the returned provider is *moved between threads via a thread-aware runtime mechanism*. (78259c12)

crates/bytesbuf/src/mem/callback_memory.rs
```diff
@@ -1,64 +1,90 @@
...
+///
+/// This can be used to construct memory providers that add logic or configuration on top of an
+/// existing memory provider, without implementing a dedicated provider type.
 ///
-/// This can be used to construct wrapping memory providers that add logic or configuration
-/// on top of an existing memory provider.
+/// Modeled on [`thread_aware::closure::Closure`]: because `reserve_fn` is a bare `fn` pointer it
+/// cannot capture anything, so all state the reservation logic needs must live in `data`. As `data`
+/// is [`ThreadAware`], it is relocated together with the provider when the provider moves between
+/// threads.
```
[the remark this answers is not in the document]
sandersaares: Reworded: `data` is relocated together with the provider when it is *moved between threads via a thread-aware runtime mechanism*. (78259c12)

crates/bytesbuf/src/mem/callback_memory.rs
sandersaares: Done -- `CallbackMemory` now uses `#[derive(ThreadAware)]`, with the bare-`fn`-pointer `reserve_fn` field marked `#[thread_aware(skip)]`, replacing the manual impl. (04047b01)

crates/bytesbuf/src/mem/callback_memory.rs
Vaiz: I guess it could work if you pass `()` and just allocate normally inside of the callback

crates/bytesbuf/src/mem/memory_shared.rs
ralfbiedert: I mean, `ThreadAware` just means "tell me when I move" and `relocated()` is called when that move happened. You must not break when you move, or when the move goes wrong or didn't happen, but you may exploit it for performance reasons.

crates/bytesbuf/src/mem/wrapping_memory.rs
ralfbiedert: I'm not very happy that we span up random API design space without knowing whether we need it or now. I get that we want to be somewhat defensive, but on the flipside we had / have a bunch of APIs that look rather convoluted and where the best argument also was "we might need this"; which makes it rather unpleasant for people to understand what's going on or contribute.

sandersaares: Well, sometimes we have evidence, other times we need to rely on vision and feeling. Both are valid - this particular one is more the latter than the former.

crates/bytesbuf/src/mem/memory_shared.rs
sandersaares: I do not disagree with the statements here but I also think these statements are not obvious and even if made will mean very little to 90% of readers without thorough guidelines and explanations we can link to.

crates/bytesbuf/src/mem/callback_memory.rs
sandersaares: In a real use case, there would always be an `inner` (just coming from a service framework), so that would be quite misleading IMO. I am also not quite happy but maybe this requires another day and inspiration from space to resolve. Will keep as-is for now.

Vaiz (approved):

DragosDanielBoia:

PR #541, opened 2026-07-01
feat(thread_aware): add ProcessorCount::AtMost for capped processor count

sandersaares:
## What

Adds a new `ProcessorCount::AtMost(NonZero<usize>)` variant that requests an upper bound on the number of processors, clamping down to every available processor when the hardware offers fewer than requested.

Also renames the existing `ProcessorCount::Manual` variant to `ProcessorCount::Exactly`. Reviewers noted that `Manual` was easy to confuse with the newly added `AtMost` variant; `Exactly` makes the "require exactly this many processors" semantics unambiguous versus `AtMost`'s upper-bound behavior.

## Why

`Exactly(N)` requires exactly `N` processors and panics via `ThreadRegistry::new`/`with_hardware` when the machine has fewer. There was no way to express "use up to `N`, but accept fewer if that is all the machine has". `AtMost` fills that gap for callers that want to cap resource usage while still running unchanged on smaller machines. For example, `AtMost(32)` uses 32 processors on a machine with at least 32, and all of them on a machine with fewer.

## How

`ThreadRegistry::with_hardware` clamps the ceiling to `hardware.processors().len()` (the quota-enforced available count) and calls the existing `ProcessorSetBuilder::take()`. No change to `many_cpus` is required. Unlike `Exactly`, `AtMost` never fails when the machine is smaller than the ceiling.

## Testing

Added three deterministic tests using the `many_cpus::fake` hardware harness: request below available, request equal to available, and request exceeding available (clamps to all, no panic). `cargo test -p thread_aware`, `just clippy`, `just format-check`, and the docs.rs doc build all pass.

### DIFF
--- a/crates/thread_aware/src/registry.rs
+++ b/crates/thread_aware/src/registry.rs
@@ -16,7 +16,7 @@ const POISONED_LOCK_MSG: &str = "poisoned lock means type invariants may not hol
 
 /// The number of processors to use for the registry.
 ///
-/// Use `Auto` for the default number of processors, `Manual` to require an exact
+/// Use `Auto` for the default number of processors, `Exactly` to require an exact
 /// number of processors, `AtMost` to request an upper bound that is satisfied by
 /// fewer processors when the hardware offers fewer, or `All` to use every processor.
 #[derive(Debug, Clone, Default, PartialEq, Eq)]
@@ -28,7 +28,7 @@ pub enum ProcessorCount {
     /// Use a specific number of processors.
     ///
     /// Registry construction panics if the hardware offers fewer processors than requested.
-    Manual(NonZero<usize>),
+    Exactly(NonZero<usize>),
     /// Use up to this many processors, or every available processor when the
     /// hardware offers fewer than requested.
     ///
@@ -66,7 +66,7 @@ impl ThreadRegistry {
     ///
     /// # Panics
     ///
-    /// This will panic if there are not enough processors available when using `Manual` or if no processors are available when using `Auto` or `All`.
+    /// This will panic if there are not enough processors available when using `Exactly` or if no processors are available when using `Auto` or `All`.
     /// If there are more than `u16::MAX` processors or memory regions.
     #[must_use]
     pub fn new(count: &ProcessorCount) -> Self {
@@ -80,11 +80,11 @@ impl ThreadRegistry {
 
         let processors = match count {
             ProcessorCount::Auto | ProcessorCount::All => builder.take_all(),
-            ProcessorCount::Manual(count) => builder.take(*count),
+            ProcessorCount::Exactly(count) => builder.take(*count),
             ProcessorCount::AtMost(max) => {
                 // A processor set is never empty, so taking every available processor
                 // always satisfies a ceiling that meets or exceeds what the machine
-                // offers, without failing the way `Manual` would.
+                // offers, without failing the way `Exactly` would.
                 if max.get() >= hardware.processors().len() {
                     builder.take_all()
                 } else {
@@ -238,8 +238,8 @@ mod tests {
 
     #[test]
     #[cfg_attr(miri, ignore)]
-    fn test_registry_manual() {
-        let registry = ThreadRegistry::new(&ProcessorCount::Manual(NonZero::new(1).unwrap()));
+    fn test_registry_exactly() {
+        let registry = ThreadRegistry::new(&ProcessorCount::Exactly(NonZero::new(1).unwrap()));
         assert_eq!(registry.num_affinities(), 1);
     }
 
@@ -251,12 +251,12 @@ mod tests {
         let iterator_count = registry.affinities().count();
         assert_eq!(registry.num_affinities(), iterator_count);
 
-        // Also test with manual processor count > 1 if available
+        // Also test with an exact processor count > 1 if available
         if iterator_count > 1 {
             let count = NonZero::new(2.min(iterator_count)).unwrap();
-            let registry_manual = ThreadRegistry::new(&ProcessorCount::Manual(count));
-            assert_eq!(registry_manual.num_affinities(), count.get());
-            assert_eq!(registry_manual.num_affinities(), registry_manual.affinities().count());
+            let registry_exactly = ThreadRegistry::new(&ProcessorCount::Exactly(count));
+            assert_eq!(registry_exactly.num_affinities(), count.get());
+            assert_eq!(registry_exactly.num_affinities(), registry_exactly.affinities().count());
         }
     }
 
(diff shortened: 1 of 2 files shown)


crates/thread_aware/src/registry.rs
```diff
@@ -72,6 +81,15 @@ impl ThreadRegistry {
...
             ProcessorCount::Manual(count) => builder.take(*count),
+            ProcessorCount::AtMost(max) => {
+                // Clamp the ceiling to what the hardware actually offers so a larger
+                // request than the machine provides is satisfied by taking every
+                // available processor instead of failing.
+                let available = hardware.processors().len();
+                let count =
+                    NonZero::new(max.get().min(available)).expect("a processor set is never empty, so the clamped count is at least one");
+                builder.take(count)
+            }
```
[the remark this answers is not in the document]
sandersaares: Addressed in 7127fffb, but not by adding a zero-length fallback. A processor set is never empty (many_cpus returns a NonEmpty<Processor>), so `len() == 0` is not a reachable state. The arm now relies on that invariant: when the ceiling meets or exceeds the available count it takes all available processors via take_all(), otherwise it takes exactly `max`. This removes the confusing panic site and the NonZero re-wrap entirely, and the shared .expect("Not enough processors available") still governs the genuinely-unsatisfiable (filtered) case.

crates/thread_aware/src/registry.rs
```diff
@@ -16,17 +16,26 @@ const POISONED_LOCK_MSG: &str = "poisoned lock means type invariants may not hol
...
 #[derive(Debug, Clone, Default, PartialEq, Eq)]
 pub enum ProcessorCount {
     /// Use the default number of processors. Right now this is equivalent to using
     /// all processors, but this default may change in the future.
     #[default]
     Auto,
     /// Use a specific number of processors.
+    ///
+    /// Construction fails if the hardware offers fewer processors than requested.
     Manual(NonZero<usize>),
```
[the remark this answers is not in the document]
sandersaares: Good catch — reworded in 7127fffb to "Registry construction panics if the hardware offers fewer processors than requested."

sandersaares:
Intentional. Adding `ProcessorCount::AtMost` to this exhaustive enum is a deliberate breaking change; the release should be cut as `thread_aware@breaking` (0.7.5 -> 0.8.0 under the repo's 0.x SemVer rules).

martintmk (approved):

crates/thread_aware/src/registry.rs
```diff
@@ -16,17 +16,26 @@ const POISONED_LOCK_MSG: &str = "poisoned lock means type invariants may not hol
...
+    ///
+    /// Registry construction panics if the hardware offers fewer processors than requested.
     Manual(NonZero<usize>),
+    /// Use up to this many processors, or every available processor when the
+    /// hardware offers fewer than requested.
+    ///
+    /// This suits callers that want to cap resource usage while still running on
+    /// smaller machines. For example, `AtMost(32)` uses 32 processors on a machine
+    /// with at least 32, and all of them on a machine with fewer.
+    AtMost(NonZero<usize>),
```
sgalkin, line 39: what will happen if one uses `Manual(32)` on machine with less than32 processors?

crates/thread_aware/src/registry.rs
```diff
@@ -16,17 +16,26 @@ const POISONED_LOCK_MSG: &str = "poisoned lock means type invariants may not hol
...
 pub enum ProcessorCount {
     /// Use the default number of processors. Right now this is equivalent to using
     /// all processors, but this default may change in the future.
     #[default]
     Auto,
     /// Use a specific number of processors.
+    ///
+    /// Registry construction panics if the hardware offers fewer processors than requested.
     Manual(NonZero<usize>),
+    /// Use up to this many processors, or every available processor when the
```
sgalkin, line 33: I don't have good name yet, but this line and `AtMost` doesn't compute together in my head. `AtMost` let me think about some number in between of 1 and value, not capped value. Also, it feels like `Manual` should cover the case.

crates/thread_aware/src/registry.rs
sandersaares: Panic. `Manual` really means `Exactly`

crates/thread_aware/src/registry.rs
sandersaares: > AtMost let me think about some number in between of 1 and value, not capped value

Well "some number between 1 and value" is kind of correct? At least, I would be happy with this interpretation. But alternative suggestions are welcome.

Perhaps `Manual` -> `Exactly` is a useful rename here for clarity. I will do that.

I suppose this API could use a UX review and adjustment pass in a bigger picture, I rather hesitate to scope creep this PR to make this API "nice" in existing regards. If you have more ideas, let me know though - we can collect them for the backlog and propose a future tidy-up feature.

sandersaares:
`AtMost` being a breaking change is annoying. I campaign for all y'all's upvotes at https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7546921/

crates/thread_aware/src/registry.rs
```diff
@@ -57,7 +66,7 @@ impl ThreadRegistry {
     ///
     /// # Panics
     ///
-    /// This will panic if there are not enough processors available when using `Manual` or if no processors are available when using `Auto` or `All`.
+    /// This will panic if there are not enough processors available when using `Exactly` or if no processors are available when using `Auto` or `All`.
```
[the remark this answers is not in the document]
sandersaares: Incorrect - AtMost can never fail - there is always at least 1 processor in a processor set, which it can use. There is no logic path that allows it to fail (unless we have a bug).

crates/thread_aware/src/registry.rs
```diff
@@ -71,7 +81,17 @@ impl ThreadRegistry {
 
         let processors = match count {
             ProcessorCount::Auto | ProcessorCount::All => builder.take_all(),
-            ProcessorCount::Manual(count) => builder.take(*count),
+            ProcessorCount::Exactly(count) => builder.take(*count),
+            ProcessorCount::AtMost(max) => {
+                // A processor set is never empty, so taking every available processor
+                // always satisfies a ceiling that meets or exceeds what the machine
+                // offers, without failing the way `Exactly` would.
+                if max.get() >= hardware.processors().len() {
```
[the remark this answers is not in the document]
sandersaares: Fixed

ralfbiedert (approved):

crates/thread_aware/src/registry.rs
```diff
@@ -57,7 +67,7 @@ impl ThreadRegistry {
     ///
     /// # Panics
     ///
-    /// This will panic if there are not enough processors available when using `Manual` or if no processors are available when using `Auto` or `All`.
+    /// This will panic if there are not enough processors available when using `Exactly` or if no processors are available when using `Auto` or `All`.
```
ralfbiedert: This feels a bit wrong. Panics equal programming errors, but not having enough HW capacity doesn't seem like a programming errors to me, but rather an Error?

sandersaares: Fair point. I think this was mostly motivated by a general principle you defined at some point that any runtime-creation error is a fatal error and should just panic, and this just picked up the pattern from the runtime (perhaps incorrectly). We can fine-tune as part of runtime-related API refinement, though - I certainly have no strong feelings on the subject.

crates/thread_aware/src/registry.rs
```diff
@@ -16,17 +16,27 @@ const POISONED_LOCK_MSG: &str = "poisoned lock means type invariants may not hol
...
 /// The number of processors to use for the registry.
 ///
-/// This can be set to `Auto` to use the default number of processors,
-/// or `Manual` to specify a specific number of processors.
-/// The `All` variant is used to specify that all processors should be used.
+/// Use `Auto` for the default number of processors, `Exactly` to require an exact
+/// number of processors, `AtMost` to request an upper bound that is satisfied by
+/// fewer processors when the hardware offers fewer, or `All` to use every processor.
 #[derive(Debug, Clone, Default, PartialEq, Eq)]
+#[non_exhaustive]
```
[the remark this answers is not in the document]
sandersaares: This is a language feature, it does not need documentation at every call site.

ralfbiedert (approved):

PR #546, opened 2026-07-05
feat: introduce REST-over-gRPC transcoding

geeknoid:
Expose an existing gRPC API as a traditional REST/JSON API in the same
process, driven by `google.api.http` annotations. The gRPC service is
implemented once; the REST layer is generated at build time and transcodes
requests and responses with minimal runtime overhead. The generated router is
built on a standalone, zero-allocation static router (`routerama`) that is also
useful on its own.

Crates added:

- `routerama`: dependency-free runtime for a compile-time static HTTP router —
  zero-allocation method + path matching returning the matched tag and captured
  path variables (with a SIMD segment scanner).
- `routerama_macros`: the `routes!` procedural macro that generates a static
  router inline from a named route enum, with no `build.rs` required.
- `routerama_build`: build-time code generator that lowers a route set into a
  `resolve` function (a nested `match` over path segments — a compile-time trie),
  reporting genuinely ambiguous routes as a `compile_error!`.
- `http_path_template`: dependency-free parser for the `google.api.http`
  path-template grammar (literals, `*`, `**`, `{field=sub}` variables, `:verb`),
  with an optional extended grammar for intra-segment prefix/suffix parameters.
- `rest_over_grpc`: framework-neutral transcoding crate spanning the runtime,
  the serving adapters, and the build-time code generator:
  - `transcoding` / `handling`: decode an HTTP request into a protobuf message,
    dispatch it to the gRPC handler, and encode the reply as JSON, mapping
    `Code` / `Status` to HTTP status codes.
  - `serving` (default, feature-gated): one-shot `serve_http*` helpers plus
    `RestService` / `StreamingRestService` adapters for the `tower` and
    `layered` ecosystems (hyper, axum, …). Request bodies are buffered with an
    optional `with_max_body_bytes` cap (`413 Payload Too Large` on overflow),
    and read failures surface as `400 Bad Request` rather than an empty body.
  - `streaming`: server-streaming responses forwarded frame by frame as a JSON
    array, NDJSON, or SSE.
  - `build` / `build-openapi`: a `build.rs` code generator that turns
    `google.api.http`-annotated gRPC services (read from compiled descriptors)
    into a static REST router, service traits, a dispatcher, an optional `tonic`
    bridge, and an optional OpenAPI 3.1 specification.
- `rest_over_grpc_examples`: worked examples organized by architecture layer,
  covering both the generated `tonic` bridge and a hand-written service that
  implements the generated trait directly.
- `rest_over_grpc_tests`: end-to-end tests and benchmarks for the generated
  router and dispatcher.


crates/rest_over_grpc/src/transcode/error.rs
```diff
@@ -0,0 +1,201 @@
...
+    }
+}
+
+impl From<TranscodeError> for Status {
+    /// Renders the error as a client-facing [`Status`] (see
+    /// [`TranscodeError::into_status`]).
+    fn from(error: TranscodeError) -> Self {
+        error.into_status()
+    }
+}
```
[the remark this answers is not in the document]
geeknoid: Good catch — fixed. I removed the doc comment on the `from` trait-impl method; the documentation already lives on `TranscodeError::into_status` (rustdoc does not render docs placed on trait-impl methods).

crates/http_path_template/README.md
```diff
@@ -0,0 +1,73 @@
+<div align="center">
+ <img src="./logo.png" alt="Http Path Template Logo" width="96">
+
```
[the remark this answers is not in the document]
geeknoid: This `./logo.png` reference is produced by the shared `crates/README.j2` template and is byte-identical across every published crate in the workspace (e.g. `bytesbuf`, `ohno`, `fetch`), so it reflects an established repo-wide convention rather than something specific to these crates. The logos are intentionally Git-LFS-tracked and kept out of the packaged `include` allowlist workspace-wide. For the rendered docs that matter on docs.rs, the crate root already sets `#![doc(html_logo_url = "https://media.githubusercontent.com/.../logo.png")]` (an absolute URL), so rustdoc renders correctly. Changing only these crates would diverge from the convention and be overwritten by `just readme`; making the README logo absolute everywhere would be a separate workspace-wide change.

crates/rest_over_grpc_example/Cargo.toml
```diff
@@ -0,0 +1,39 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+[package]
+name = "rest_over_grpc_example"
+description = "Worked example wiring a tonic-generated gRPC server into a REST API via the rest_over_grpc_build tonic bridge"
```
ralfbiedert: Why does this have an `_example` and `_custom_example`?

crates/rest_over_grpc/src/lib.rs
```diff
@@ -0,0 +1,363 @@
...
+pub use context::Context;
+#[cfg(feature = "streaming")]
+#[doc(inline)]
+pub use dispatch_response::{DispatchResponse, FrameStream, ResponseStream, StreamingResponse};
+/// Re-exports of the [`http`] header types used by the neutral value types
+/// ([`Context`], [`HttpResponse`]) and by `rest_over_grpc_build`-generated code,
+/// so consumers can name them through `rest_over_grpc` without taking a direct
+/// `http` dependency.
+#[doc(no_inline)]
+pub use http::{HeaderMap, HeaderName, HeaderValue};
```
ralfbiedert: M-FOREIGN-REEXPORTS. Also, double check some of the examples here, e.g., in `Context`, they are inconsistent w.r.t. pulling in `HeaderMap` and friends.

crates/rest_over_grpc/src/lib.rs
```diff
@@ -0,0 +1,363 @@
...
+#[cfg(any(feature = "tower", feature = "layered"))]
+mod adapter;
+#[cfg(feature = "axum")]
+mod axum_support;
+#[cfg(feature = "streaming")]
+mod stream;
+
+#[cfg(all(feature = "streaming", any(feature = "tower", feature = "layered")))]
+#[doc(inline)]
+pub use adapter::{BoxError, StreamingBody, StreamingRestService, transcode_http_streaming};
```
ralfbiedert: Also, half of the items here violate M-FIRST-DOC-SENTENCE

crates/rest_over_grpc/tests/streaming_serving.rs
```diff
@@ -0,0 +1,131 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Integration tests for the streaming HTTP adapter (`transcode_http_streaming`
+//! and `StreamingRestService`), which forward server-streaming frames to the
+//! wire incrementally.
+
+#![cfg(all(feature = "streaming", feature = "tower"))]
```
ralfbiedert: If you read the docs it's quite a bit of feature-logic going on, are these really needed; like what's the downside of just adding streaming w/o features?

crates/rest_over_grpc/src/http_response.rs
```diff
@@ -0,0 +1,372 @@
...
+    /// use http::StatusCode;
+    /// use rest_over_grpc::HttpResponse;
+    ///
+    /// let response = HttpResponse::new(StatusCode::ACCEPTED, "text/plain", b"queued".to_vec());
+    /// assert_eq!(response.status(), StatusCode::ACCEPTED);
+    /// assert_eq!(response.content_type(), "text/plain");
+    /// assert_eq!(response.body(), b"queued");
+    /// ```
+    #[must_use]
+    pub fn new(status: StatusCode, content_type: &'static str, body: Vec<u8>) -> Self {
```
ralfbiedert, line 62: Is the fact that this uses `Vec<u8>` instead of `Bytes`, `BytesBuf` or some `&mut ...` at various places caused by having to be compatible with a particular Response signature?

crates/rest_over_grpc/src/http_response.rs
```diff
@@ -0,0 +1,372 @@
...
+    /// ```
+    /// use http::StatusCode;
+    /// use rest_over_grpc::HttpResponse;
+    ///
+    /// let response = HttpResponse::ok_json(br#"{"ok":true}"#.to_vec());
+    /// assert_eq!(response.status(), StatusCode::OK);
+    /// assert_eq!(response.content_type(), "application/json");
+    /// ```
+    #[must_use]
+    pub fn ok_json(body: Vec<u8>) -> Self {
```
ralfbiedert, line 84: ~Speaking of small APIs, should this be removed and just be `::json` below?~

Hm, I take it back, doing a `HttpResponse::not_found()` via `json()` would feel weird.

crates/rest_over_grpc/src/lib.rs
```diff
@@ -0,0 +1,363 @@
...
+//! A gRPC service is defined once in `.proto` and speaks an efficient binary
+//! protocol over HTTP/2 — ideal for service-to-service calls, but not something
+//! a browser `fetch`, a `curl` one-liner, a webhook sender, or a third-party
+//! integrator can call directly. Those clients want plain REST/JSON over
+//! ordinary HTTP. The [`google.api.http`](https://github.com/googleapis/googleapis/blob/master/google/api/http.proto)
+//! annotation standard already expresses that mapping (e.g. `GET
+//! /v1/shelves/{shelf}` `⇄` `GetShelf`) inline in the `.proto`, so the REST
+//! surface is fully specified — it just needs something to serve it.
+//!
+//! The usual answer is a **separate transcoding gateway** in front of the gRPC
```
ralfbiedert: I'm not particularly fond of this LLM-generated introduction; it's using quite a bit of jargon and its own weird style, but having read all of it I'm still not sure if this is actually the crate I should be looking at (as a user), over some of its sibling ones. In other words, it's very technical but not very helpful.

crates/routerama_build/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+// Licensed under the MIT License.
+
+#![cfg_attr(coverage_nightly, feature(coverage_attribute))]
+#![cfg_attr(docsrs, feature(doc_cfg))]
+#![doc(html_logo_url = "https://media.githubusercontent.com/media/microsoft/oxidizer/refs/heads/main/crates/routerama_build/logo.png")]
+#![doc(
+    html_favicon_url = "https://media.githubusercontent.com/media/microsoft/oxidizer/refs/heads/main/crates/routerama_build/favicon.ico"
+)]
+
+//! Build-time code generator for a compile-time static HTTP router.
```
ralfbiedert: Neither this crate nor `routerama` does a good job explaining what's going on, what parameters are supported, how the resulting function looks like, ... It's a bit like a black box, the only thing I can see is an example below where I specify some strings and apparently get a `TokenStream` out ...

crates/rest_over_grpc_example/Cargo.toml
geeknoid: There are two ways to define the shape of the REST service. The simple way is to point to a file descriptor set. The more complicated way is to manually describe the service's shape explicitly. So two examples.

crates/rest_over_grpc/src/lib.rs
geeknoid: Yes, I did mention that I wasn't done with the docs for that crate :-)

crates/rest_over_grpc/tests/streaming_serving.rs
geeknoid: It's the usual: extra dependencies and thus increased compile time. I could remove the feature if you'd like.

crates/rest_over_grpc/src/http_response.rs
geeknoid: Yep. I tried to use something else but it doesn't compose.

crates/rest_over_grpc/src/lib.rs
geeknoid: Yes, I know. I've gone through the whole API surface and wrote most of the prose by hand. But I haven't finished with this specific crate.

crates/http_path_template/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+#![doc(html_logo_url = "https://media.githubusercontent.com/media/microsoft/oxidizer/refs/heads/main/crates/http_path_template/logo.png")]
+#![doc(
+    html_favicon_url = "https://media.githubusercontent.com/media/microsoft/oxidizer/refs/heads/main/crates/http_path_template/favicon.ico"
+)]
+
+//! A parser for the [`google.api.http`] path-template grammar.
+//!
+//! A path template is the pattern that appears in a `google.api.http`
+//! annotation, for example `shelves/{shelf}/books/{book=**}:archive`. This crate
+//! turns such a string into a validated, structured [`PathTemplate`] — an
```
sandersaares, line 15: Would be helpful for understanding if this example actually showed the result of parsing the example string given here - what is the resulting abstract syntax tree it yields?

crates/http_path_template/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+//! abstract syntax tree of [`Segment`]s (literals, `*`, `**`, and
+//! `{field.path=sub-template}` [`Variable`] bindings) plus an optional custom
+//! `:verb`.
+//!
+//! The grammar mirrors the reference [`google.api.HttpRule`] path syntax:
+//!
+//! - a **literal** segment (`shelves`) must match verbatim;
```
sandersaares, line 27: How is case-sensitivity handled?

crates/http_path_template/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+#![doc(
+    html_favicon_url = "https://media.githubusercontent.com/media/microsoft/oxidizer/refs/heads/main/crates/http_path_template/favicon.ico"
+)]
+
+//! A parser for the [`google.api.http`] path-template grammar.
+//!
+//! A path template is the pattern that appears in a `google.api.http`
+//! annotation, for example `shelves/{shelf}/books/{book=**}:archive`. This crate
+//! turns such a string into a validated, structured [`PathTemplate`] — an
+//! abstract syntax tree of [`Segment`]s (literals, `*`, `**`, and
```
sandersaares, line 16: It seems implicit here that segments are separated by `'/'` but this feels important enough to explicitly state.

crates/http_path_template/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+//! - **`**`** ([`Segment::Rest`]) matches the remaining segments and may only
+//!   appear as the final element;
+//! - **`{field.path=sub-template}`** ([`Segment::Variable`]) captures the portion
+//!   of the path matched by its sub-template into a dotted message field; the
+//!   shorthand `{field}` is `{field=*}` and nested variables are rejected;
+//! - a trailing **`:verb`** declares a custom method verb.
+//!
+//! # Extended grammar
+//!
+//! [`PathTemplate::parse_extended`] accepts a superset of the strict grammar
```
sandersaares: I generally find that "extended" functions/types are just a "I did not want to bother thinking of a proper API" escape hatch. Can we instead model this as an explicit choice made at some point? What if we need a 3rd extended grammar in 2029 - would be nice to support that without breaking changes.

```rust
let template = PathTemplate::parse(s, Grammar::default();

const FANCY: Grammar = Grammar::builder().allow_segment_affixes().build();
let template = PathTemplate::parse(s, FANCY);
```

crates/http_path_template/src/error.rs
```diff
@@ -0,0 +1,163 @@
...
+/// use http_path_template::PathTemplate;
+///
+/// let error = PathTemplate::parse("/a/**/b").expect_err("rest is not last");
+/// assert_eq!(
+///     error.to_string(),
+///     "'**' may only appear as the final segment"
+/// );
+/// ```
+#[derive(Debug)]
+pub struct ParseError {
```
sandersaares, line 67: Why not `ohno`? This appears to be hand-rolling part of `ohno` functionality for no obvious reason. We have an error library, let's use it.

crates/rest_over_grpc/examples/generate_service.rs
```diff
@@ -0,0 +1,56 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Generating a REST service from `google.api.http` rules.
+//!
+//! Builds [`HttpRule`]s (normally read straight from proto annotations by
```
sandersaares, line 6: I am confused by this example. Why is it saying how things are done "normally"? Is this an "abnormal" example? To an uninformed reader the context of this example is hard to identity: "generating a REST service from 'google.api.http' rules" feels like a "normal" thing to do but this statement suggests to me I am doing it "abnormally" somehow. More clarity on what is special in this example would be desirable.

crates/rest_over_grpc/examples/generate_service.rs
```diff
@@ -0,0 +1,56 @@
...
+    library
+        .add_method(get_shelf, "crate::pb::GetShelfRequest", "crate::pb::Shelf", None)
+        .add_method(create_shelf, "crate::pb::CreateShelfRequest", "crate::pb::Shelf", None);
+
+    let (transcoder, generated) = Generator::new().add(library).generate();
+
+    // Render the generated Rust for display; a build script would instead
+    // write the code to files under `OUT_DIR` with `Generator::write`.
+    for service in generated {
+        let mut code = service.r#trait().clone();
```
sandersaares, line 45: Having to write `r#trait()` suggests some rather suboptimal naming. Might be better to avoid reserved keywords in public APIs.

crates/rest_over_grpc/examples/layered_service.rs
```diff
@@ -0,0 +1,68 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Serving REST/JSON through the `layered` adapter.
+//!
+//! The repository's `layered::Service` is an `async fn`-with-`&self` alternative
```
sandersaares, line 6: "The repository's" is cumbersome wording. Let's speak of the `layered` package, not of "the repository" - packages can move between repositories and readers might not always have context to understand what "the repository" refers to.

crates/rest_over_grpc/examples/layered_service.rs
```diff
@@ -0,0 +1,68 @@
...
+        } else {
+            HttpResponse::from_status(&Status::not_found("no such route"))
+        };
+        async move { Some(response.into()) }
+    }
+}
+
+#[tokio::main(flavor = "current_thread")]
+async fn main() {
+    // `layered::Service` needs no `Clone` and takes `&self`, so the service is
```
sandersaares, line 50: Unclear what the impact of this comment is. Why are we document things that are not? Same for the comment a few lines below.

It almost sounds as if this was made as a "comparison" but no comparand is in sight.

crates/rest_over_grpc/src/lib.rs
```diff
@@ -0,0 +1,450 @@
...
+// Licensed under the MIT License.
+
+#![cfg_attr(coverage_nightly, feature(coverage_attribute))]
+#![cfg_attr(docsrs, feature(doc_cfg))]
+#![doc(html_logo_url = "https://media.githubusercontent.com/media/microsoft/oxidizer/refs/heads/main/crates/rest_over_grpc/logo.png")]
+#![doc(html_favicon_url = "https://media.githubusercontent.com/media/microsoft/oxidizer/refs/heads/main/crates/rest_over_grpc/favicon.ico")]
+
+//! Automatically transcode gRPC services to REST/JSON endpoints.
+//!
+//! Services that expose a gRPC API often need to also expose a parallel REST API
```
ralfbiedert, line 11: Documentation is _so_ much better than the previous one!

crates/rest_over_grpc/src/lib.rs
```diff
@@ -0,0 +1,450 @@
...
+//! impl<S> Transcoder<S> {
+//!     pub fn new(service: S) -> Self { /* ... */ }
+//! }
+//!
+//! impl<S: Greeter> rest_over_grpc::transcoding::Transcode for Transcoder<S> {
+//!     // Routes `GET /v1/greet/{name}` to `Greeter::say_hello`, decoding the JSON
+//!     // request and encoding the reply. Returns a `TranscodeResponse` (a buffered
+//!     // unary reply or a server-streaming frame stream); `None` when no route
+//!     // matches.
+//!     async fn try_transcode(&self, method: &str, target: &str,
```
ralfbiedert, line 254: I'm not sure about the stringiness here (over URI parts or similar), but maybe that comes from having to be compatible with a range of crates?

crates/rest_over_grpc/src/lib.rs
```diff
@@ -0,0 +1,450 @@
...
+//!
+//! - **Client-streaming and bidirectional RPCs are unsupported.** They have no
+//!   `google.api.http` mapping (an HTTP request is one message), so the
+//!   [`build`] module rejects them at build time. Handle such an API —
+//!   e.g. a chunked upload — with a dedicated HTTP handler that reads the body
+//!   incrementally and bridges it to your native gRPC client-streaming call,
+//!   composed alongside the transcoded JSON routes. See
+//!   `rest_over_grpc_examples`'s `examples/handling/client_streaming_upload.rs`.
+//!
+//! - **The request body is buffered as one JSON `&[u8]`.** The generated
```
ralfbiedert, line 278: Nit, it's not the `&[u8]` that does buffering, but rather what it's exposed as?

crates/rest_over_grpc/src/serving.rs
```diff
@@ -0,0 +1,594 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! The serving layer: bridges the `http` / `http-body` ecosystem to the neutral
+//! transcoder signature. Gated on the `serving` feature (on by default, and
```
ralfbiedert, line 5: This is what the `docrs` attributes are for, this shouldn't be documented as text.

crates/rest_over_grpc_examples/proto/google/api/http.proto
```diff
@@ -0,0 +1,84 @@
+// Copyright 2024 Google LLC
```
ralfbiedert, line 1: Is the idea that every user has to place exactly these files besides their project for the main annotations to work?

crates/rest_over_grpc_examples/build.rs
```diff
@@ -0,0 +1,124 @@
+// Copyright (c) Microsoft Corporation.
```
ralfbiedert, line 1: I'm not very fond of most examples. There is _a lot_ going on in both the library and the examples, and it's hard to know where to even start looking, everything is sort of related to everything else somehow, concept wise. Basic transcode is small enough to start wrapping my head around things, but even then I have to dive into `LibraryServce`, the generated `"/tonic_bridge/library.rs"` ...

Like is there a way to have basic examples of fundamental concepts only, say, 1/3 as large?

crates/rest_over_grpc/src/lib.rs
geeknoid: Yeah, we can't have AI write API documentation.

crates/rest_over_grpc/src/lib.rs
geeknoid: Correct, this is the only lingua franca there is at this level of the stack.

crates/http_path_template/src/lib.rs
geeknoid: Good call. The module-level `# Examples` now spells out the AST that `/shelves/{shelf}/books/{book=**}:archive` parses into — four top-level segments (literal `shelves`, `Variable{shelf=*}`, literal `books`, `Variable{book=**}`) plus the `archive` verb — and the runnable doctest asserts each one, so the shown AST stays verified.

crates/http_path_template/src/lib.rs
geeknoid: Added an explicit note to the module docs: literal segments, variable field names, and the custom `:verb` are stored and compared verbatim, so the grammar is case-sensitive and the parser performs no case folding (matching the `google.api.http` reference).

crates/http_path_template/src/lib.rs
geeknoid: Made it explicit in the intro: a template must begin with `/` and is a `/`-separated sequence of segments, with each `/` delimiting one segment.

crates/http_path_template/src/lib.rs
geeknoid: Done — replaced the `parse_extended` escape hatch with an explicit `Grammar` argument. There's now a single `PathTemplate::parse(template, grammar)`; `Grammar` implements `Default` (the strict `google.api.http` grammar), and non-standard extensions are opted into explicitly, e.g. `Grammar::default().with_segment_affixes()`. New `with_*` options can be added later without breaking existing callers. All call sites (routerama_macros, routerama_build, rest_over_grpc) and tests are updated.

PR #549, opened 2026-07-07
feat(fetch)!: use SimpleClock and add total-duration + http.client.name metrics

martintmk:
## Summary

Improves the `fetch` crate's metrics and logging handlers.

## Clock -> SimpleClock

`Logging` and `Metrics` now use `tick::SimpleClock` instead of `tick::Clock`. Since `SimpleClock::new_system()` needs no runtime or driver, the clock is optional:

- Removed the clock argument from `Logging::layer()` and `Metrics::layer()`.
- Added a `.clock(impl Into<SimpleClock>)` setter on both layers, defaulting to a system clock (useful for injecting a controlled clock in tests).

## Total request-duration metric

The standard pipeline records total request duration in addition to per-attempt duration:

- New `MetricsLayer::report_total_duration(bool)` records the histogram under `http.client.request.total_duration` instead of `http.client.request.duration`.
- The standard pipeline gains a `total_metrics` layer (configurable via `StandardRequestPipeline::total_metrics`), placed as the outermost layer so it measures the whole logical request including all retries and hedged attempts.

## `http.client.name` metric attribute

- The configured client name (`HttpClientBuilder::name`, default `http_client`) is reported as the `http.client.name` instrumentation-scope attribute alongside `fetch.runtime` / `fetch.transport`, so all of a client's metrics can be correlated with a specific named client. The name is applied whether it is set before or after the meter provider.

## Documentation

- The `telemetry` documentation module documents the `http.client.request.total_duration` instrument and the `http.client.name` attribute.

## Breaking changes

- `Logging::layer` and `Metrics::layer` no longer take a clock argument (use `.clock(..)`).
- `HttpClientBuilder::meter_provider` now takes the provider by value (`impl MeterProvider + Send + Sync + 'static`) instead of `&dyn MeterProvider`.


martintmk:
can you also update the "telemetry" documentation module with the information about the metrics and new attribtues being reported

martintmk:
&#x1F916; Done -- updated the `telemetry` documentation module (`crates/fetch/src/_documentation/telemetry.rs`):

- Added `http.client.request.overall_duration` to the metrics table plus a dedicated section describing it as the whole-request duration (all retries and hedged attempts) recorded by the outermost `overall_metrics` layer in the standard pipeline.
- Documented the new `resilience.attempt.index` / `resilience.attempt.is_last` attributes, reported when `MetricsLayer::include_attempt` is enabled (with the `0` / `is_last = true` fallback when no attempt is present).
- Clarified that `http.client.request.duration` is a per-attempt measurement.

Verified with `cargo doc -p fetch --no-deps --all-features` under `-D warnings` (all intra-doc links resolve). Pushed as 4d371483.

---

crates/fetch/src/_documentation/telemetry.rs
```diff
@@ -30,20 +30,27 @@
 //!
 //! | Metric | Instrument | Unit | Emitted when |
 //! |--------|-----------|------|--------------|
-//! | [`http.client.request.duration`](#httpclientrequestduration) | `Histogram<f64>` | `s` | Every HTTP request completes (success **or** failure) |
+//! | [`http.client.request.duration`](#httpclientrequestduration) | `Histogram<f64>` | `s` | Every HTTP request attempt completes (success **or** failure) |
+//! | [`http.client.request.overall_duration`](#httpclientrequestoverall_duration) | `Histogram<f64>` | `s` | An entire logical request completes, including all retries and hedged attempts (opt-in; enabled in the standard pipeline) |
```
sandersaares: Just looking at the names, "overall" duration feels a bit vague because it does not really distinguish from "request" duration, it is very up to interpretation; can we come up with a better term?

Transaction duration? Outer duration? Operation duration? Client duration? Group duration?

I can't say I particularly love any of the alternatives but perhaps you or other reviewers or AIs have better ideas.

martintmk: I would like that this new metrics would be close to `http.client.request.duration`. Maybe `http.client.request.total_duration` or `http.client.overall_request.total_duration` 

@copilot Can you make some suggestions?

crates/fetch/src/_documentation/telemetry.rs
```diff
@@ -59,11 +66,47 @@
 //! | `network.protocol.version` | on success | Negotiated HTTP version of the response | `"1.1"` |
```
martintmk, line 67: I realized we are missing `http.client.name` attribute that can be assigned to HttpClient to correlate it with particular named client. Add this.

martintmk: the name should be required when creating metrics and logging layers `name: impl INto<Cow<'static, str>>` and filled by default for standard pipeline. Client name should be accessible on pipeline context

crates/fetch/src/_documentation/telemetry.rs
martintmk: &#x1F916; Agreed -- renamed the metric to `http.client.request.total_duration` so it stays close to `http.client.request.duration` and drops the vaguer `overall`. To keep the API coherent, this also renamed `MetricsLayer::report_total_duration` and the standard pipeline's `total_metrics` layer. Pushed in d07e710f.

---

crates/fetch/src/_documentation/telemetry.rs
martintmk: &#x1F916; Done -- added an `http.client.name` metric attribute sourced from the client's configured name (`HttpClientBuilder::name`, default `http_client`). It flows through the standard pipeline's per-attempt and total request metrics, and is also settable on the standalone handler via `MetricsLayer::client_name`. Documented in the telemetry module and covered by unit tests (added `ResilienceContext::get_name` to read the configured name). Pushed in d07e710f.

---

crates/seatbelt/src/context.rs
```diff
@@ -71,6 +71,12 @@ impl<In, Out> ResilienceContext<In, Out> {
         self
     }
 
+    /// Returns the configured pipeline name used for telemetry correlation.
+    #[must_use]
+    pub fn get_name(&self) -> &str {
```
martintmk: rather than exposing this field here, can we just do it in fetch by introducing PipelineContext::name that returns Cow<'static, str>

martintmk: &#x1F916; Agreed -- I'll drop the `ResilienceContext::get_name` getter and thread the client name through fetch instead, exposing it via `PipelineContext::name() -> Cow<'static, str>`. Working on it now.

---

martintmk: &#x1F916; Done -- dropped the `ResilienceContext::get_name` getter and now thread the client name through fetch instead. Added `PipelineContext::name() -> Cow<'static, str>`, which the standard pipeline reads in `ConfigureStandardPipeline::create` to set `http.client.name` on the per-attempt and total request metrics. Also added an end-to-end integration test that a named client emits the attribute. Pushed in 3d81663d.

---

crates/fetch/src/handlers/metrics.rs
```diff
@@ -71,14 +88,21 @@ impl Debug for ResponseEnricher {
...
 pub struct Metrics<T> {
     inner: T,
-    clock: Clock,
+    clock: SimpleClock,
     request_duration: Histogram<f64>,
     on_record: Option<OnRecordCallback>,
     enrich_from_request: Option<RequestEnricher>,
     enrich_from_response: Option<ResponseEnricher>,
+    include_attempt: bool,
+    client_name: Option<Cow<'static, str>>,
```
martintmk: rather than re-defining the name here, can we rather use InstrumentationScope that is apssed with each Meter? (similar to how transport attributes are done).

The idea is to update instrumentation scope when http client name is being set in the http client builder.

martintmk: &#x1F916; Agreed -- switching `http.client.name` to an instrumentation-scope attribute (alongside `fetch.runtime` / `fetch.transport`), updated when the name is set in the builder. Reworking now.

---

martintmk: &#x1F916; Done -- `http.client.name` is now an instrumentation-scope attribute (next to `fetch.runtime` / `fetch.transport`), so it lands on all of a client's metrics (request and connection). The client scope is rebuilt with the configured name in `HttpClientBuilder::name` for the global meter, and applied when a custom meter provider is supplied. Dropped the per-request `MetricsLayer::client_name` / `PipelineContext::name` plumbing in favor of this. Added a scope-attribute integration test and updated the telemetry docs. Pushed in 1e1875ec.

---

crates/fetch/src/client_builder.rs
```diff
@@ -67,6 +73,11 @@ impl HttpClientBuilder {
     /// The name is used in logging and metrics to identify the HTTP client instance. The name should
     /// follow the `snake_case` convention. By default, the client is named "`http_client`".
     pub fn name(mut self, name: impl Into<Cow<'static, str>>) -> Self {
```
martintmk, line 74: we should have a test that sets a meter provider then name and ensures the custom name is correctly reported. Possibly requires some changes to the design. Come up with the plan.

martintmk: &#x1F916; Good catch -- this is a real gap. Here's the plan.

**Why it happens today:** `meter_provider(&dyn MeterProvider)` materializes the scoped `Meter` eagerly (the provider is borrowed, not retained), so `http.client.name` is baked into the scope at that call. A later `.name(...)` can rebuild the *global* scope but cannot rebuild an already-materialized *custom* meter -- so `.meter_provider(p).name("x")` still reports the default name.

**Proposed fix -- defer custom-meter creation to `build()`:**
1. Retain the provider instead of materializing eagerly: change `Metering::Custom` to hold `{ provider: Arc<dyn MeterProvider>, scope }` and create the `Meter` only at `build()` time.
2. This needs `meter_provider` to take an owned provider so it can be stored -- e.g. `meter_provider<P: MeterProvider + 'static>(self, provider: P)` (wrapped internally in `Arc<dyn MeterProvider>`). Callers change from `.meter_provider(&p)` to `.meter_provider(p)` / `.meter_provider(p.clone())` (`SdkMeterProvider` is `Clone`).
3. `name()` then rebuilds the stored scope for both global and custom, so the final name is always applied at build time regardless of call order.
4. Add the requested test: `.meter_provider(p).name("x")` asserts the recorded scope carries `http.client.name = "x"`.

**Tradeoff:** step 2 is a breaking change to `meter_provider` (`&dyn` -> owned/Arc). Given this PR already ships intentional breaking changes and the crate is 0.x, I think it's acceptable, but flagging it.

**Non-breaking alternative:** keep `&dyn` and require `.name()` before `.meter_provider()` for custom providers (documented + a debug-assert to catch misuse). Less robust.

I'd recommend option 1. Want me to proceed with it, or prefer the non-breaking alternative?

---

martintmk: let's go with 1.

martintmk: &#x1F916; Done -- implemented option 1. `meter_provider` now takes an owned provider (`P: MeterProvider + Send + Sync + 'static`) and retains it; the scoped meter is materialized at `build()` time, so the client name is applied regardless of order. `Metering::Custom` stores `{ provider, scope }` and `with_client_name` rebuilds the custom scope too. Added the requested test (`.meter_provider(p).name("x")` -> `http.client.name = "x"`) plus the before-order variant. Pushed in a31cf8c4.

Note: as discussed, this makes `meter_provider` take the provider by value instead of `&dyn MeterProvider` (breaking).

---

crates/fetch/tests/client_name_metric.rs
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) Microsoft Corporation.
```
martintmk: this tests can be merged with telemetry_scope.rs

crates/fetch/src/telemetry.rs
```diff
@@ -67,10 +73,22 @@ impl Extend<KeyValue> for TelemetryAttributes {
...
 }
 
-#[derive(Debug, Clone)]
+#[derive(Clone)]
 pub(crate) enum Metering {
     Global(InstrumentationScope),
-    Custom(Meter),
+    Custom {
+        provider: Arc<dyn MeterProvider + Send + Sync>,
+        scope: InstrumentationScope,
```
martintmk: rather than storing  InstrumentationScope, can't we just store individual properties and then just materialize it when Meter is created?

crates/fetch/tests/client_name_metric.rs
martintmk: &#x1F916; Done -- merged both client-name tests into `telemetry_scope.rs` (reusing its `scope_attribute_for_metric` helper) and removed the standalone `client_name_metric.rs`. Pushed in 775eb675.

---

crates/fetch/src/telemetry.rs
martintmk: &#x1F916; Done -- `Metering` now stores the individual scope properties (runtime, transport, client_name) as `ScopeProperties` and only materializes the `InstrumentationScope` when the `Meter` is created. Renaming just updates the stored `client_name`, and `into_custom` carries the current properties onto a custom provider, so the name is applied regardless of order. Also removed the now-redundant `client_name` field from `HttpClientBuilder`. Pushed in 5cc8c2f1.

---

crates/fetch/src/telemetry.rs
```diff
@@ -67,42 +74,109 @@ impl Extend<KeyValue> for TelemetryAttributes {
     }
 }
 
+/// The `fetch` scope-defining properties of a client's meter.
+///
+/// The [`InstrumentationScope`] is only materialized when the [`Meter`] is
+/// created (see [`From<Metering>`][Metering]), so the client name can still be
+/// updated after the meter provider has been chosen.
 #[derive(Debug, Clone)]
+pub(crate) struct ScopeProperties {
```
martintmk: Can you merge ScopeProperties with Metering. Just metering remains, no longer enum, stores Option<Arc<dyn MeterProvider + Send + Sync>> if custom provider is set, otherwise global is used.

martintmk: &#x1F916; Done -- `Metering` is now a single struct (no enum) holding the scope properties plus `Option<Arc<dyn MeterProvider + Send + Sync>>` (`Some` = custom provider, `None` = global). `with_provider` sets a custom provider and `with_client_name` updates the name; the scope is still materialized only when the meter is created, so the client name stays order-independent. Pushed in 20b11b44.

---

crates/fetch/src/handlers/metrics.rs
```diff
@@ -114,7 +146,7 @@ impl MetricsLayer {
     /// When no meter is configured, the global meter provider is used.
     #[must_use]
     pub fn meter_provider(mut self, meter_provider: &dyn MeterProvider) -> Self {
-        self.meter = Some(Metering::custom(meter_provider, base_scope()).into());
+        self.meter = Some(meter_provider.meter_with_scope(base_scope()));
```
martintmk, line 144: can we use Telemetry::default().with_provider(meter_provider).into() here?

martintmk: &#x1F916; Not cleanly here, for two reasons:

1. `Metering` always materializes the *client* scope (`fetch.runtime` / `fetch.transport` / `http.client.name`), whereas this standalone `MetricsLayer` path intentionally uses the attribute-less `base_scope()`. A bare `Metrics` handler isn't tied to a client/transport, so it shouldn't carry those attributes (nor a default/empty client name).
2. `MetricsLayer::meter_provider` takes `&dyn MeterProvider` (borrowed), while `Metering::with_provider` needs an owned `Arc<dyn MeterProvider + Send + Sync>`.

So `meter_provider.meter_with_scope(base_scope())` is the right fit here. When a `MetricsLayer` is used inside the standard pipeline it instead receives the client's full-scope meter via `.meter(...)`, so the client attributes come from there.

Happy to reconsider if you'd actually like the standalone layer to carry client-scope attributes too.

---

crates/fetch/src/telemetry.rs
```diff
@@ -67,42 +74,85 @@ impl Extend<KeyValue> for TelemetryAttributes {
...
     }
 }
 
-/// Builds the `fetch` instrumentation scope carrying the `fetch.runtime` and
-/// `fetch.transport` attributes, attached to every metric a client records.
-pub(crate) fn client_scope(runtime: impl Into<Value>, transport: impl Into<Value>) -> InstrumentationScope {
+/// Builds the `fetch` instrumentation scope carrying the `fetch.runtime`,
+/// `fetch.transport`, and `http.client.name` attributes, attached to every
+/// metric a client records.
+pub(crate) fn client_scope(runtime: impl Into<Value>, transport: impl Into<Value>, client_name: impl Into<Value>) -> InstrumentationScope {
```
martintmk: can this be private?

martintmk:
update the PR title and PR description to align with latest changes

crates/fetch/src/telemetry.rs
martintmk: &#x1F916; Yes -- made it private. `client_scope` is only used within the telemetry module now that the builder constructs scopes through `Metering`. Pushed in 4df0c2c9.

---

martintmk:
&#x1F916; Done -- updated the PR title and description to reflect the current changes: `SimpleClock` (optional, with `.clock(..)`), the `http.client.request.total_duration` metric plus `report_total_duration` / `include_attempt` toggles, the `http.client.name` instrumentation-scope attribute, and the breaking `layer()` / `meter_provider` signature changes.

---

crates/fetch/src/client_builder.rs
```diff
@@ -231,23 +238,22 @@ impl HttpClientBuilder {
...
-            ),
-            resilience_context: self.resilience_context.use_metrics(meter_provider),
-            ..self
-        }
+    pub fn meter_provider<P: MeterProvider + Send + Sync + 'static>(mut self, meter_provider: P) -> Self {
+        // Update the metering at all relevant places.
+        self.resilience_context = self.resilience_context.use_metrics(&meter_provider);
+        self.metering = self.metering.with_provider(Arc::new(meter_provider));
+        self
     }
```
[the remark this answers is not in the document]
martintmk: &#x1F916; Good catch -- added `HttpClientBuilder::meter_provider_arc(Arc<dyn MeterProvider + Send + Sync>)` for callers sharing a type-erased provider across clients, and the generic `meter_provider` now delegates to it. Pushed in 220402f0.

---

crates/fetch/src/handlers/metrics.rs
```diff
@@ -114,7 +146,7 @@ impl MetricsLayer {
...
         self
```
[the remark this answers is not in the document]
martintmk: &#x1F916; The PR description no longer claims a `MetricsLayer::client_name` setter -- I updated it just as this review landed. The standalone `MetricsLayer` intentionally uses the attribute-less `base_scope()`; the `http.client.name` scope attribute is applied only for full clients built via `HttpClientBuilder`. So the description now matches the code.

---

crates/fetch/src/pipeline/standard.rs
```diff
@@ -344,6 +361,7 @@ mod tests {
 
         let debug_str = format!("{pipeline:?}");
         assert!(debug_str.contains("StandardRequestPipeline"));
+        assert!(debug_str.contains("total_metrics"));
         assert!(debug_str.contains("total_timeout"));
```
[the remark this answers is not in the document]
martintmk: &#x1F916; Good idea -- hardened `test_debug_implementation` to assert that `total_metrics`'s debug output contains `report_total_duration: true`, so it verifies the layer is actually configured to emit the total-duration instrument rather than merely being present. Pushed in b41b50ab.

---

crates/fetch/src/client_builder.rs
```diff
@@ -231,23 +238,34 @@ impl HttpClientBuilder {
...
-        // Update the metering at all relevant places
-        Self {
-            metering: Metering::custom(
-                meter_provider,
-                client_scope(self.transport.runtime().clone(), self.transport.name().clone()),
-            ),
-            resilience_context: self.resilience_context.use_metrics(meter_provider),
-            ..self
-        }
+    pub fn meter_provider_arc(mut self, meter_provider: Arc<dyn MeterProvider + Send + Sync>) -> Self {
```
martintmk: I am not fan of this change. Can you drop it.

martintmk: &#x1F916; Agreed - dropped the `meter_provider_arc` overload. `meter_provider` is back to a single generic method taking `P: MeterProvider + Send + Sync + 'static`.

---

crates/fetch/benches/pipelines.rs
```diff
@@ -78,8 +78,10 @@ fn entry(c: &mut Criterion) {
                 Timeout::layer("attempt_timeout", context.resilience_context())
                     .timeout(Duration::from_secs(10))
                     .http_timeout_error(),
-                Logging::layer(context.clock(), context.redaction_engine()),
-                Metrics::layer(context.clock()).meter_provider(opentelemetry::global::meter_provider().as_ref()),
+                Logging::layer(context.redaction_engine()).clock(context.clock()),
```
martintmk: can you also make the redaction engine optional, use default instance if not specified

martintmk: &#x1F916; Done - the redaction engine is now optional. `Logging::layer()` takes no arguments and defaults to `RedactionEngine::default()`; a new `LoggingLayer::redaction_engine(&RedactionEngine)` setter supplies a custom engine (the standard pipeline still threads the client's configured engine through it). Mirrors the optional clock. Pushed in 45a566c7.

---

crates/fetch/src/client_builder.rs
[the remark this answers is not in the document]
martintmk: &#x1F916; Not implementing this - the `meter_provider_arc(Arc<dyn MeterProvider + Send + Sync>)` overload was added earlier in this PR and the maintainer explicitly asked to drop it (they were not a fan of the extra API surface). `meter_provider` intentionally stays a single generic method taking a concrete provider by value. Deferring to the maintainer's decision here.

---

crates/fetch/src/client_builder.rs
```diff
@@ -231,23 +240,22 @@ impl HttpClientBuilder {
...
-                client_scope(self.transport.runtime().clone(), self.transport.name().clone()),
-            ),
-            resilience_context: self.resilience_context.use_metrics(meter_provider),
-            ..self
-        }
+    pub fn meter_provider<P: MeterProvider + Send + Sync + 'static>(mut self, meter_provider: P) -> Self {
+        // Update the metering at all relevant places.
+        self.resilience_context = self.resilience_context.use_metrics(&meter_provider);
+        self.metering = self.metering.with_provider(Arc::new(meter_provider));
+        self
```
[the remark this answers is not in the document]
martintmk: &#x1F916; Not implementing this. The `meter_provider_arc(Arc<dyn MeterProvider + Send + Sync>)` overload was added earlier in this PR and the maintainer explicitly asked to drop it, preferring `meter_provider` to remain a single generic method taking a concrete provider by value. The internal `Arc<dyn ...>` storage is an implementation detail and does not change that API decision. Deferring to the maintainer.

---

crates/fetch/src/pipeline/standard.rs
```diff
@@ -88,6 +92,7 @@ pub struct StandardRequestPipeline {
 impl StandardRequestPipeline {
     pub(crate) fn new(options: &HttpResilienceContext, redaction: &RedactionEngine, clock: &Clock, meter: &Meter, router: &Router) -> Self {
         Self {
+            total_metrics: Metrics::layer().clock(clock).meter(meter.clone()).report_total_duration(true),
```
martintmk, line 95: can you include attempts for total_metrics and attempt_metrics?

crates/fetch/src/_documentation/telemetry.rs
```diff
@@ -30,20 +31,27 @@
 //!
 //! | Metric | Instrument | Unit | Emitted when |
 //! |--------|-----------|------|--------------|
-//! | [`http.client.request.duration`](#httpclientrequestduration) | `Histogram<f64>` | `s` | Every HTTP request completes (success **or** failure) |
+//! | [`http.client.request.duration`](#httpclientrequestduration) | `Histogram<f64>` | `s` | Every HTTP request attempt completes (success **or** failure) |
+//! | [`http.client.request.total_duration`](#httpclientrequesttotal_duration) | `Histogram<f64>` | `s` | An entire logical request completes, including all retries and hedged attempts (opt-in; enabled in the standard pipeline) |
```
sandersaares: Ambiguous: do hedged attempts duplicate time counted or is it merely "duration ends when last hedged attempt ends"? (My expectation as user: the latter - no duplication of duration just because two concurrent requests were happening)

sandersaares: It is not obvious to me why this should be opt-in. We should not offer unnecessary configurability to users (configurability just adds breakage opportunity for both us and customers). Unless there is significant necessity to filter, should just emit all metrics.

crates/fetch/src/_documentation/telemetry.rs
```diff
@@ -59,11 +67,47 @@
...
+//! | `resilience.attempt.is_last` | when attempt tracking enabled | Whether this is the final attempt that will be performed | `true` |
 //!
 //! > **Custom attributes.** Any [`TelemetryAttributes`](crate::telemetry::TelemetryAttributes)
 //! > attached to the request or response extensions are merged into the
 //! > attribute set. This allows callers to inject domain-specific dimensions.
 //!
+//! > **Attempt attributes.** `resilience.attempt.index` and
+//! > `resilience.attempt.is_last` are only reported when
+//! > [`MetricsLayer::include_attempt`](crate::handlers::MetricsLayer::include_attempt)
+//! > is enabled (disabled by default to keep metric cardinality low). When no
```
sandersaares: I have several remarks on this arrangement:

1. Making things opt-in mostly just guarantees that nobody will use them and makes logic more likely to be brittle (accidental misconfiguration or accidental wrong logic decisions due to untested combinations). The ideal number of configurable options is zero. We should strive for that.
2. While it is true that `attempt.index` having many values could contribute to cardinality issues, this is a motivation to **drop this attribute** rather than to make it opt-in. "The 4th attempt succeeded" is very unlikely to be useful information for metrics purposes. It might be useful for logging purposes, to backtrace specific histories of specific requests.
3. Opting in to cardinality problems still leaves you with cardinality problems, so an opt-in does not make it a valid metrics attribute to have, it just causes confusion in how the thing is meant to be used.
4. Even if this were somehow opt-in, it is not clear to me how one would conditionally enable this opt-in via ECS on the fly (as would be necessary for any opt-in investigations where you only enable high cardinality metrics for a specific subset of instances).

I recommend therefore to have all the attributes always-on, and to move a specific index to logs and not metrics.

The `is_last` flag is fine as always-on under all conditions, this does not contribute to cardinality issues in any way as it has low cardinality.

crates/fetch/src/_documentation/telemetry.rs
```diff
@@ -59,11 +67,47 @@
...
+//! > [`Attempt`](seatbelt::Attempt) is present in the request extensions — for
+//! > example on the first attempt or outside a retry stack — the fallback is
+//! > index `0` with `is_last = true`.
+//!
+//! ---
+//!
+//! ## `http.client.request.total_duration`
+//!
+//! An Oxidizer-specific extension to the OpenTelemetry HTTP semantic
```
sandersaares: Framing this as somehow related to OTel semantic conventions feels wrong. There is no official concept of "extensions" in the semantic conventions. We are intentionally creating vendor-specific metrics in namespaces officially owned by OTel, so to speak. We are violating OTel territory.

Personally, I am fine with us doing this as I have low respect for OTel but this paragraph somehow feels like it is trying to legalize our violation, which is just not true - we should simply be open and honest that we are adding proprietary extensions here. (Alternatively, we should properly namespace them with `oxidizer.http.client...` if we do not want to violate OTel territory).

ralfbiedert (approved):

crates/fetch/src/_documentation/telemetry.rs
martintmk: Yes, the latter, for hedging the duration of first accepted response is counted. There is no magic in time tracking, just  execution duration of inner service.

crates/fetch/src/_documentation/telemetry.rs
martintmk: let me just drop the resilience attributes for now, it does pollute this PR and can be added as a separate feature

crates/fetch/src/_documentation/telemetry.rs
martintmk: indeed

crates/fetch/src/pipeline/standard.rs
martintmk: &#x1F916; Superseded by the decision below to drop the resilience attempt attributes for now. Rather than wiring `include_attempt` into `total_metrics` / `attempt_metrics`, the entire opt-in attempt-attributes feature (`resilience.attempt.index` / `resilience.attempt.is_last`) has been removed from this PR and will be reintroduced separately. Resolving as no longer applicable.

---

crates/fetch/benches/pipelines.rs
```diff
@@ -78,8 +78,10 @@ fn entry(c: &mut Criterion) {
                 Timeout::layer("attempt_timeout", context.resilience_context())
                     .timeout(Duration::from_secs(10))
                     .http_timeout_error(),
-                Logging::layer(context.clock(), context.redaction_engine()),
-                Metrics::layer(context.clock()).meter_provider(opentelemetry::global::meter_provider().as_ref()),
+                Logging::layer().clock(context.clock()),
```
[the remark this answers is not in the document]
martintmk: &#x1F916; Good catch — fixed. The `custom_standard_pipeline` benchmark now passes the context-provided redaction engine to its logging layer (`Logging::layer().redaction_engine(context.redaction_engine()).clock(context.clock())`), matching how the standard pipeline wires its `attempt_logs`. Pushed in the latest commit.

---

crates/fetch/src/_documentation/telemetry.rs
martintmk: &#x1F916; Confirmed: the latter. `total_duration` is simply the execution duration of the inner service — timing ends when the first accepted response (or terminal error) is returned, and concurrent hedged attempts do not double-count the measured time. I've clarified this explicitly in the `http.client.request.total_duration` docs to remove the ambiguity.

---

martintmk: &#x1F916; Agreed — the opt-in high-cardinality attempt attributes (`resilience.attempt.*`) have been removed from this PR entirely, so there is no longer an opt-in metrics dimension here. The `total_duration` metric itself is not user opt-in; it is enabled by the standard pipeline's outermost metrics layer. `report_total_duration` is only an internal toggle that lets the single `MetricsLayer` type serve both the per-attempt and total instruments — not a user-facing metrics filter. I also removed the misleading "opt-in" wording from the docs.

---

crates/fetch/src/_documentation/telemetry.rs
martintmk: &#x1F916; Done — as decided, the resilience attempt attributes have been dropped from this PR. Removed the `include_attempt` API, the `resilience.attempt.index` / `resilience.attempt.is_last` attributes, and their docs and tests. They can be reintroduced as a separate feature, with a high-cardinality attempt index better suited to logs than metrics per your point (2).

---

crates/fetch/src/_documentation/telemetry.rs
martintmk: &#x1F916; Reworded. The `http.client.request.total_duration` docs no longer frame this as an OpenTelemetry-sanctioned extension. They now state plainly that it is a proprietary Oxidizer metric, not part of the official OpenTelemetry semantic conventions, added inside the OTel-owned `http.client.*` namespace as a vendor-specific instrument.

---

crates/fetch/src/telemetry.rs
```diff
@@ -67,42 +74,85 @@ impl Extend<KeyValue> for TelemetryAttributes {
...
-pub(crate) enum Metering {
-    Global(InstrumentationScope),
-    Custom(Meter),
+/// How a client's `fetch` metrics are metered.
+///
+/// Holds the scope-defining properties (runtime, transport, client name) and an
+/// optional custom meter [provider][MeterProvider]; when no provider is set, the
+/// global meter provider is used. The [`InstrumentationScope`] is only
+/// materialized when the [`Meter`] is created (see [`From<Metering>`][Metering]),
+/// so the client name can still be updated after the provider has been chosen.
```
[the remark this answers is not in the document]
martintmk: &#x1F916; Adjusted. For the record, the original `[`From<Metering>`][Metering]` did resolve (to the `Metering` struct) and `cargo doc --no-deps -D warnings` passed locally with no `broken_intra_doc_links` / `invalid_html_tags` warning — the `From<Metering>` text was inside a code span — so it wasn't actually breaking the docs job. That said, your suggested form is clearer, so the reference now uses plain code for `From<Metering>` plus a resolvable `[`Meter`]` link (the impl is `From<Metering> for Meter`). Pushed.

---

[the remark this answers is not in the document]
martintmk: &#x1F916; Already addressed — this comment is on an outdated revision. In the latest commit the reference was reworded to use plain code for `From<Metering>` plus a resolvable `[`Meter`]` intra-doc link (the impl is `From<Metering> for Meter`). Resolving.

---

crates/fetch/src/handlers/metrics.rs
```diff
@@ -89,14 +103,27 @@ pub struct Metrics<T> {
...
     enrich_from_response: Option<ResponseEnricher>,
 }
 
 impl MetricsLayer {
+    /// Sets the clock used to measure request duration.
+    ///
+    /// When no clock is configured, a system clock ([`SimpleClock::new_system`])
+    /// is used. Supplying a controlled clock (for example
+    /// [`SimpleClock::new_frozen`]) is primarily useful for deterministic timing
+    /// in tests.
```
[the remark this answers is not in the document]
martintmk: &#x1F916; This link is fine as-is. `fetch`'s `test-util` feature forwards to `tick/test-util`, and this crate is only ever documented with all features enabled — CI runs `cargo doc --all-features` and docs.rs uses `[package.metadata.docs.rs] all-features = true`. Under those features `SimpleClock::new_frozen` is compiled, so `[`SimpleClock::new_frozen`]` resolves. Verified locally: `cargo doc -p fetch --all-features --no-deps` with `-D warnings` passes with no broken-intra-doc-link warning. Keeping the resolvable link. Resolving.

---

crates/fetch/src/handlers/logging.rs
```diff
@@ -27,39 +27,71 @@ use crate::{HttpRequest, HttpResponse, RequestExt, RequestHandler, Result};
...
+    redaction_engine: Option<RedactionEngine>,
+}
+
+impl LoggingLayer {
+    /// Sets the clock used to measure request duration.
+    ///
+    /// When no clock is configured, a system clock ([`SimpleClock::new_system`])
+    /// is used. Supplying a controlled clock (for example
+    /// [`SimpleClock::new_frozen`]) is primarily useful for deterministic timing
+    /// in tests.
```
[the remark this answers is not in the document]
martintmk: &#x1F916; Same as the metrics.rs thread. `SimpleClock::new_frozen` is behind `tick/test-util`, but `fetch`'s `test-util` feature enables it and this crate is always documented with all features (CI `cargo doc --all-features`; docs.rs `all-features = true`), so the link resolves. Verified `cargo doc -p fetch --all-features --no-deps -D warnings` passes with no broken-link warning. Keeping the link. Resolving.

---

sandersaares (approved):

PR #552, opened 2026-07-08
feat(cachet_memory)!: surface evicted key and value to on_eviction listeners

schgoo:
## Summary

The in-memory tier's eviction listener previously received only a `RemovalCause`, discarding the key and value of the entry that was removed. This makes it impossible for callers to react to *which* entry was evicted (e.g. cascading invalidation, secondary-index cleanup, logging). This PR passes the evicted key and value through to `on_eviction`.

## Changes

- `InMemoryCacheBuilder::on_eviction` now takes `Fn(Arc<K>, V, RemovalCause)` instead of `Fn(RemovalCause)`. The listener receives the evicted entry's key as a shared `Arc<K>` and its owned value `V`.
- The `cachet` eviction-telemetry hook is unchanged in behavior - it continues to emit `cache.eviction` / `cache.expired` from `RemovalCause` only and ignores the key/value.
- Adds a new doc hidden `with_removal_observer` to enable removal cause telemetry events in cachet. Setting it up as an eviction listener requires cloning `V` or passing it by reference.

## Breaking change

- Callers of `on_eviction` must update their closure from `|cause|` to `|key, value, cause|`. Telemetry consumers (`with_eviction_telemetry`) are unaffected.
- At most one listener is held. Calling `on_eviction` more than once replaces the previously registered listener.

### DIFF
--- a/crates/cachet_memory/src/builder.rs
+++ b/crates/cachet_memory/src/builder.rs
@@ -26,6 +26,14 @@ use crate::tier::InMemoryCache;
 /// along with the [`RemovalCause`].
 pub(crate) type EvictionListener<K, V> = Arc<dyn Fn(Arc<K>, V, RemovalCause) + Send + Sync + 'static>;
 
+/// Type-erased, cause-only removal observer.
+///
+/// Unlike [`EvictionListener`], observers receive only the [`RemovalCause`] —
+/// never the key or value — so registering one never forces a clone of the
+/// evicted entry. Used internally by host crates (such as `cachet`) to bridge
+/// removals into their own telemetry.
+pub(crate) type RemovalObserver = Arc<dyn Fn(RemovalCause) + Send + Sync + 'static>;
+
 /// Builder for configuring an `InMemoryCache`.
 ///
 /// This builder provides a stable API for common cache configuration
@@ -54,6 +62,7 @@ pub struct InMemoryCacheBuilder<K, V, H = RandomState> {
     pub(crate) name: Option<&'static str>,
     pub(crate) eviction_policy: EvictionPolicy,
     pub(crate) eviction_listener: Option<EvictionListener<K, V>>,
+    pub(crate) removal_observers: Vec<RemovalObserver>,
     pub(crate) eviction_telemetry: bool,
     pub(crate) hasher: H,
     _phantom: PhantomData<(K, V)>,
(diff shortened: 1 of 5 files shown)


afoxman:
Another useful scenario which I plan on using this for: V isn't just a value, it is a composed value w/metadata Metadata<V>. On eviction, I can act on that metadata (e.g. log it, trigger other actions).

crates/cachet_memory/src/builder.rs
```diff
@@ -292,12 +297,14 @@ impl<K, V, H> InMemoryCacheBuilder<K, V, H> {
...
+        F: Fn(Arc<K>, V, RemovalCause) + Send + Sync + 'static,
+        K: 'static,
+        V: Clone + 'static,
     {
         self.eviction_listener = Some(match self.eviction_listener.take() {
-            Some(previous) => Arc::new(move |cause| {
-                previous(cause);
-                listener(cause);
+            Some(previous) => Arc::new(move |key: Arc<K>, value: V, cause| {
+                previous(Arc::clone(&key), value.clone(), cause);
```
afoxman: What do you think about supporting only 1 listener, to avoid cloning the value?

afoxman (approved):

crates/cachet_memory/src/builder.rs
schgoo: That's probably the better choice for now, until we get a legitimate feature request for >1

schgoo: It's complicated, though, since cachet's telemetry layer is also an event listener

geeknoid (approved):

crates/cachet_memory/src/builder.rs
```diff
@@ -291,25 +305,44 @@ impl<K, V, H> InMemoryCacheBuilder<K, V, H> {
     /// ```
     #[must_use]
     pub fn on_eviction<F>(mut self, listener: F) -> Self
+    where
+        F: Fn(Arc<K>, V, RemovalCause) + Send + Sync + 'static,
```
ralfbiedert, line 309: Why couldn't this be a `&K`?

ralfbiedert (approved):

crates/cachet_memory/src/builder.rs
schgoo: It definitely could. I guess not knowing what the user is doing with the key, and acknowledging that some keys may be large, it made sense to optimize for cloning in case they need ownership. Moka itself {which is where we get this notification) exposes the key as `Arc<K>` in the eviction listener this way to begin with, so I didn't have a reason to change the pattern.

afoxman (approved):

PR #554, opened 2026-07-09
fix(release): classify releases via cargo-semver-checks vs previous version-bump commit

Vaiz:
The release cascade decided each crate's change type from a [package.metadata.cargo_check_external_types] allowlist heuristic to guess whether a dependent exposed a changed dependency. That allowlist drifts from the real public API, so an exposed dependency's breaking change could be mis-cascaded as a patch (AB#7577159).

Replace the heuristic with real API diffing: run cargo-semver-checks against each crate's previous version-bump commit in git history — the most recent commit that changed the crate's [package] version, supplied as --baseline-rev <sha> so the baseline rustdoc is rebuilt from that commit's source. Every release in the plan — user-source roots and cascade-pulled dependents — is classified from its own working-tree API diff. Dependents are floored at patch then raised to their own semver verdict. No fallback: cargo-semver-checks is required and an unanalysable crate is a hard error; a crate with no prior version-bump commit imposes no floor.

The baseline is source-level, not registry-based: whatever the repo last declared, regardless of publish status — no crates.io access, identical for OSS and enterprise/offline. A version committed but never published (an aborted 4.0.0) is the baseline, not a stale published 3.3.3.

Also rework the CI semver job: baseline each publishing crate against its previous version-bump commit (--baseline-rev, no registry), run only on the publishing set, non-failing — posts a fresh per-crate table (baseline + resolved commit vs this-PR vs minimum required). Outputs: publishing=true|false, status=pass|warn|fail.

Example of a CI check: https://github.com/microsoft/oxidizer/pull/565#issuecomment-4955258561

image


scripts/lib/release-flow.ps1
```diff
@@ -246,6 +249,61 @@ function Get-TransitivePublishedDependentsFromBaseline {
...
+        [Parameter(Mandatory = $true)][string]$RequirementLabel,
+        [Parameter(Mandatory = $true)][string]$RequirementDetail,
+        [switch]$Force
+    )
+
+    $existingRank = $script:ChangeTypeRank[$Entry.EffectiveChangeType]
+    $newRank      = $script:ChangeTypeRank[$RequiredChangeType]
+    if ($null -eq $existingRank) { $existingRank = 0 }
+    if ($null -eq $newRank) { $newRank = 0 }
+    if ($newRank -le $existingRank) { return }
```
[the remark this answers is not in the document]
Vaiz: Incorrect as a live bug — both files dot-source into one scope, so it resolved. Proven by the full suite (pin-rejection/auto-upgrade/cascade paths would all no-op if null) and a targeted probe (patch→breaking upgraded correctly). But the cross-file $script: access is genuinely fragile. 

Hardened anyway: Update-EntryForRequiredChangeType now uses Get-StrongerChangeType (ranking encapsulated next to the table). Behaviour identical; 24/24 scenarios + 120 unit tests still pass.

scripts/lib/release-flow.ps1
```diff
@@ -307,14 +365,22 @@ function Get-TransitivePublishedDependentsFromBaseline {
...
+# allowed_external_types exposure heuristic.
 function Resolve-ReleaseSet {
     param(
         [Parameter(Mandatory = $true)][AllowEmptyCollection()][object[]]$ParsedTokens,
         [Parameter(Mandatory = $true)][AllowEmptyCollection()][object[]]$WorkspaceBaseline,
+        # Classifier scriptblock: (folder, cargoName) -> 'breaking'|'non-breaking'|
+        # 'patch'|'none'. Decides each crate's minimum change type from its real
+        # API diff (cargo-semver-checks) vs its last published version. Production
+        # passes New-SemverChangeTypeClassifier; the default is a no-op ('none')
+        # so callers that only exercise version/pin/BFS math need not supply one.
```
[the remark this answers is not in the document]
Vaiz: Correct — violated the no-fallback contract.

Removed the generic match; only true crate-absent signals map to none. A transient failure now throws. Added a regression test.

.github/workflows/main.yml
```diff
@@ -196,6 +196,13 @@ jobs:
       - name: Check Spelling
         run: just spellcheck
 
+  # Runs cargo-semver-checks against each crate's LAST PUBLISHED version on
```
sandersaares: This change of base does not seem desirable to me, for two reasons:

1. The technical reason - in enterprise environments, crates.io is simply not accessible and even when accessible, content only becomes accessible after a delay. This workflow therefore cannot be generalized to cover private half of the universe. We need to create workflows that can operate both in the OSS world and the enterprise world, as we do not have the capacity/funding to maintain two separate workflows for each world.
2. The semantic reason - it does not seem justified to ignore the fact that the repo declared a previous release, even if it did not get published. If I commit a breaking change as `4.0.0` but for some reason refuse to publish it to crates.io, this does not make mean my next attempt to publish a `4.0.1` is suddenly a `5.0.0` breaking change (which it would be because the latest on crates.io might be `3.3.3` compared to which our main branch is breaking).

We should instead evaluate breakage against the most recent commit that changed the last commit that changed the `version` attribute. [I think the script also looks at the `publish` attribute today for other reasons but that seems not relevant for breakage analysis]

docs/releasing.md
```diff
@@ -203,31 +203,59 @@ Examples:
...
+crates and the cascade-pulled dependents — is derived by running
+[`cargo semver-checks`](https://crates.io/crates/cargo-semver-checks)
+against each crate's **last published version on crates.io** (the tool's
+default baseline). The current working-tree API is analysed, so a
+coordinated release's in-progress edits — including a dependency whose
+public types a dependent re-exports — are reflected in the dependent's
+own API diff. This replaces the former
+`[package.metadata.cargo_check_external_types]` allowlist heuristic,
+which could misfire when the allowlist drifted from the real public API
+(see the bug that motivated this change: a breaking change in an exposed
```
sandersaares: This "see the bug" here is ambiguous - where is the link? Actually, I am also curious what actually went wrong there. It is not obvious to me what was broken with the previous logic (though I can certainly believe it is not a comprehensive check).

docs/releasing.md
```diff
@@ -203,31 +203,59 @@ Examples:
...
+dependency was cascaded as `patch` instead of `breaking`).
+
+Mapping from `cargo semver-checks` verdicts to change types:
```
sandersaares: It is unclear what this "verdicts" refers to. Is there any documentation we can link?

I found this https://docs.rs/cargo-semver-checks/latest/cargo_semver_checks/struct.CrateReport.html which seems to sort of match by intent (if not by name) via its `required_bump() -> Option<ReleaseType>` getter.

Do we use cargo-semver-checks as an app or as a lib? I suppose the app has some other contract we use to detect the verdicts? What exactly is that? (Might not be relevant to document exactly here but maybe wherever the detection is in a script we can ensure the mapping is well documented)

docs/releasing.md
```diff
@@ -203,31 +203,59 @@ Examples:
...
+- minor API change required → `non-breaking`
+- compatible / no update required → `patch`
+- crate not yet on crates.io (never published) → no constraint
+
+Cascade dependents are floored at `patch` (they must re-release to pick
+up the new dependency version even when their own public API is
+unchanged), then raised to whatever their own `cargo semver-checks`
+verdict requires.
+
+> **cargo-semver-checks is a hard dependency of the release scripts.**
```
sandersaares: This paragraph seems unnecessary. We can assume that all our required tools from `just install-tools` are always installed.

docs/releasing.md
```diff
@@ -203,31 +203,59 @@ Examples:
...
-the planner auto-upgrades it and notes the upgrade in the review output.
-The caller's `-Packages` token is therefore a *lower bound*, not a
-guarantee — the caller can always elevate further on the next iteration
-of the review, but cannot suppress a cascade-imposed change type.
+change type for a package is *weaker* than `cargo semver-checks`
+requires (for that package or via a cascade), the planner auto-upgrades
+it and notes the upgrade in the review output. The caller's `-Packages`
+token is therefore a *lower bound*, not a guarantee — the caller can
+always elevate further on the next iteration of the review, but cannot
+suppress a semver-imposed change type.
```
sandersaares: `semver-imposed` seems linguistically wrong. Semver is a specification, it does not force us to do things. It is a tool we use to signal things. Perhaps "change type analysis imposed" would be better here.

sandersaares (changes requested):

scripts/ci/semver-report.ps1
```diff
@@ -0,0 +1,198 @@
...
+}
+
+# --- 3. Render the Markdown report. -------------------------------------------
+$underBumped = @($rows | Where-Object { -not $_.Sufficient })
+$overallFail = $underBumped.Count -gt 0
+
+$bt = [char]0x60   # backtick, kept in a variable to avoid PowerShell escaping.
+$sb = New-Object System.Text.StringBuilder
+if ($overallFail) {
+    [void]$sb.AppendLine('## 🛑 Semver: version bump required')
```
sandersaares: "Semver" here seems to be a bit of a parasite word that just adds noise. I feel this word is not necessary and can be removed - the title is fine without this prefix.

sandersaares: I suggest "additional version increments required" for two reasons:

1. "bump" seems very colloquial, feels not professional language (applies to all usages of this word).
2. If we trigger this check only when something is getting published, it is presumably always the case that versions increments already exist in this PR, so "additional" clarifies that they are not enough (as opposed to "they were not detected" or "they were done wrong")

.github/workflows/main.yml
Vaiz: for AI agent: instead of relying on crate.io, use `cargo info` and local registry to access the changes. Add optional `Registry` parameter so script can be used on a private registry.

regarding second case, verify behavior of Sander's example (maybe even add a test), but don't try to fix it. For such cases we still have a workaround of manual version changes

Vaiz: Done in 565f013. The baseline is now discovered with `cargo info <crate> --registry <reg>` (run in a scratch dir outside the workspace so it returns the published version, not the local path package), then pinned with `cargo semver-checks --baseline-version`. Added an optional `-Registry` parameter (default crates-io) threaded through `release-packages.ps1`, the classifier in `releasing.ps1`, and the CI report; the workflow exposes it as SEMVER_REGISTRY so a fork publishing to a private registry can repoint it. Verified locally against private feed.

Vaiz: > The semantic reason - it does not seem justified to ignore the fact that the repo declared a previous release, even if it did not get published. If I commit a breaking change as 4.0.0 but for some reason refuse to publish it to crates.io, this does not make mean my next attempt to publish a 4.0.1 is suddenly a 5.0.0 breaking change (which it would be because the latest on crates.io might be 3.3.3 compared to which our main branch is breaking).

Verified and left intentionally unfixed per the discussion. The baseline is always the last published version, so an aborted release that committed 4.0.0 without publishing is invisible — semver-checks diffs against the older registry version instead. Documented this limitation and the workaround (explicit <name>@<x.y.z> pin, plus -Force if the pin is below the computed floor) in docs/releasing.md, and added a covering test in PureFunctions.Tests.ps1 (ConvertFrom-CargoInfoOutput — "reports the last PUBLISHED version…").

scripts/ci/semver-report.ps1
Vaiz: Fixed. Replaced the colloquial "bump" with "additional version increment"/"increase" throughout the report body and the table status column.

Vaiz: Fixed. Titles are now ## 🛑 Additional version increments required and ## ✅ Version increments look sufficient — dropped the redundant "Semver:" prefix. The table's registry column is now "Published" rather than "crates.io" so it's registry-agnostic.

docs/releasing.md
Vaiz: Reworded. That section now talks about the change type "cargo semver-checks requires" rather than "semver-imposed".

docs/releasing.md
Vaiz: Removed. The standalone install/hard-dependency paragraph is gone; the tools are assumed present (CI installs the pinned CARGO_SEMVER_CHECKS_VERSION, and the local script fails fast with an actionable message if it's missing).

docs/releasing.md
Vaiz: Done. Added a "How the change type is determined" subsection with a result→change-type mapping table (major → breaking, minor → non-breaking, compatible → patch, not-in-registry → no constraint), linked the cargo-semver-checks crate and its required_bump docs, and pointed at ConvertFrom-SemverChecksOutput for the exact parsing.

docs/releasing.md
Vaiz: Clarified. It now explains the concrete defect: the old cargo_check_external_types allowlist was repurposed as a proxy for what a crate actually re-exports, and when it drifted a breaking change in bytesbuf wasn't propagated to bytesbuf_io (which re-exports bytesbuf types) — so it cascaded as patch instead of breaking.

.github/workflows/main.yml
sandersaares: My concern remains - I do not see any justification provided for using the published version over the actual history. It just seems to be an error in the design as far as I can tell. If there is justification, let's discuss it.

Whether/when something gets published to feeds does not have any impact on versioning in my book. Versioning is purely a source-level concern. Please educate me if you disagree.

scripts/ci/semver-report.ps1
```diff
@@ -0,0 +1,255 @@
...
+        [void]$sb.AppendLine("${bt}cargo semver-checks${bt} compared the crate(s) this PR publishes against their latest published release. **$($realUnder.Count) of $($rows.Count)** need a higher version than this PR sets — the increment already applied is not enough for the API changes:")
+    }
+    if ($unknownRows.Count -gt 0) {
+        [void]$sb.AppendLine("⚠️ The baseline (last published version) could not be determined for **$($unknownRows.Count)** crate(s); their version increment was **not** verified — check them manually.")
+    }
+} else {
+    [void]$sb.AppendLine('## ✅ Version increments look sufficient')
+    [void]$sb.AppendLine()
+    [void]$sb.AppendLine("${bt}cargo semver-checks${bt} compared the **$($rows.Count)** crate(s) this PR publishes against their latest published release. Every version increment is sufficient for the detected API changes.")
+}
```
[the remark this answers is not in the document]
Vaiz: fixed in 1f52c05. The headline and the status step output are now computed from genuinely under-incremented rows only (ChangeType -ne 'unknown'), and there's a distinct third state for the unknown-baseline case:

fail (🛑 Additional version increments required) — at least one crate is truly under-incremented.
warn (⚠️ Semver baseline could not be determined) — no crate is under-incremented, but at least one baseline lookup failed, so the check is incomplete.
pass (✅) — every crate is sufficiently incremented.

If both conditions occur, the real under-increment wins the headline (fail), with the unknown crates called out as a secondary note. The docstring's status contract is updated to pass | warn | fail accordingly, and the per-crate table / collapsible detail already flagged unknown-baseline rows distinctly.

sandersaares (changes requested):
Reactivated a comment thread from my previous review - my concern remains unaddressed.

.github/workflows/main.yml
Vaiz: > Whether/when something gets published to feeds does not have any impact on versioning in my book. Versioning is purely a source-level concern

I disagree, versioning is all about publishing. You can do whatever you want with versions in the source code, don't publish anywhere, and it wouldn't matter

how about a middle ground, prev version gets extracted from git diff specifically for current branch vs master, and then it tries to find that previously published version. If it was never published, automation spits stop signs at developers, and they have to decide what they to do on their own. The hope is that those would be some rare situations, and we can live with it.

ralfbiedert (approved):

.github/workflows/main.yml
sandersaares: Please explain to me why publishing matters. I am not going to agree to some middle ground without a reason - this is just wrong as far as I can tell and no rationale has been provided for why it is right.

Vaiz: Done — switched the semver baseline from the crates.io published version to the crate's previous version-bump commit in git history (the most recent commit that changed its [package] version), fed to cargo-semver-checks via --baseline-rev. It's now purely source-level and needs no registry access, so the same workflow works for OSS and enterprise/offline, and a declared-but-unpublished version (your 4.0.0 example) is the baseline rather than a stale published 3.3.3. Applied to both the CI report and the release planner; -Registry/SEMVER_REGISTRY and the cargo info lookup are gone.

scripts/lib/releasing.ps1
```diff
@@ -439,53 +559,111 @@ function Get-WorkspacePackages {
...
+
+    return ConvertFrom-SemverChecksOutput -Output $output -ExitCode $exitCode -PackageName $PackageName
+}
+
+# Parses `cargo semver-checks` combined output into a change type. Pure (no I/O)
+# so it can be unit-tested against captured tool output. Mapping:
+#   * "N major and M minor checks failed" -> major>0 breaking; minor>0 non-breaking
+#   * "no semver update required"                              -> patch (compatible)
+#   * baseline not found on the registry (never published)    -> none
+#   * anything else (tool/network/build failure)              -> throw (no silent fallback)
```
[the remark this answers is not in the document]
Vaiz: cleaned up

.github/workflows/main.yml
```diff
@@ -215,41 +226,37 @@ jobs:
...
-
-      - name: Remove Semver Comment on Success
-        if: steps.semver.outcome == 'success' && github.event.pull_request.head.repo.full_name == github.repository
-        uses: marocchino/sticky-pull-request-comment@v3.0.4
-        with:
-          header: semver-check
-          delete: true
+          $ErrorActionPreference = 'Stop'
+          git fetch --no-tags origin "+refs/heads/$env:BASE_REF`:refs/remotes/origin/$env:BASE_REF"
+          ./scripts/ci/semver-report.ps1 -BaseRef "origin/$env:BASE_REF" -ReportPath semver-comment.md -RunUrl "$env:RUN_URL"
```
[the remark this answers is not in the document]
Vaiz: switched to the explicit ${env:BASE_REF} form and dropped the backtick, so the refspec is +refs/heads/main:refs/remotes/origin/main with no escaping. Braced the other env references in the step too

scripts/lib/releasing.ps1
```diff
@@ -347,6 +365,80 @@ function Get-PackageVersionFromRef {
...
+        $script:PreviousVersionBumpCommitCache = @{}
+    }
+    $cacheKey = "$RepoRoot`u{2402}$BaseRef`u{2402}$PackageFolder"
+    if ($script:PreviousVersionBumpCommitCache.ContainsKey($cacheKey)) {
+        return $script:PreviousVersionBumpCommitCache[$cacheKey]
+    }
+
+    $relPath = "crates/$PackageFolder/Cargo.toml"
+    $commits = Invoke-Git -Arguments @('log', '--format=%H', $BaseRef, '--', $relPath) -RepoRoot $RepoRoot -AllowFailure
+
```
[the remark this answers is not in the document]
Vaiz: fixed

.github/workflows/main.yml
```diff
@@ -215,41 +226,37 @@ jobs:
...
+          ./scripts/ci/semver-report.ps1 -BaseRef "origin/${env:BASE_REF}" -ReportPath semver-comment.md -RunUrl "${env:RUN_URL}"
+
+      # Post a NEW comment on every push (not a sticky update) so each run's
+      # result is visible in the PR timeline. Only fires when the PR actually
+      # publishes something and is not from a fork (forks lack a write token).
+      - name: Post Semver Comment
+        if: steps.report.outputs.publishing == 'true' && github.event.pull_request.head.repo.full_name == github.repository
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: gh pr comment ${{ github.event.pull_request.number }} --body-file semver-comment.md
```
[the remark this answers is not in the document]
Vaiz: no, we want that part to succeed

sandersaares (approved):

PR #558, opened 2026-07-09
feat(cachet): add `encrypt` feature for authenticated value encryption

schgoo:
## Summary

Adds an optional `encrypt` feature to `cachet` that protects cache **values**
before they reach a fallback/remote tier, binding each value to its storage key so
it cannot be read back under a different key.

The feature ships only the protection *mechanism* and carries **no cryptographic
dependency of its own** — callers plug in a protector backed by their approved
cryptographic library via `.protect_with(protector)`:

```rust
let cache = Cache::builder::<String, String>(clock)
    .memory()
    .serialize()
    .protect_with(my_protector)   // any `ValueProtector` implementation
    .fallback(remote)
    .build();
```

## Motivation

When a cache tiers down to an untrusted store (Redis, S3, etc.), values are
exposed at rest. This feature keeps the ergonomic typed cache API while
transparently protecting values on the way out and recovering them on the way
back, with no hand-rolled serialization/encryption pipeline. Shipping the
mechanism without any bundled crypto lets each consumer satisfy its own
cryptographic-library compliance requirements and keeps the crate dependency-free
and portable across all CI targets.

## What it does

- **`.protect_with(protector)`** — available after `.serialize()` (once values are
  `BytesView`). Protects each value with a caller-supplied `ValueProtector`.
- **`ValueProtector` trait** — the pluggable `protect(context, plaintext)` /
  `unprotect(context, protected)` contract. The verb pair mirrors OS
  data-protection APIs (Windows DPAPI `CryptProtectData`, .NET `IDataProtector`);
  `context` is the AEAD-associated-data / DPAPI-entropy role.
- **Key binding** — the storage key is passed as the `context` and must be bound,
  so a value protected for one key fails to recover under any other key. This
  prevents an attacker with write access to the backing store from relocating or
  swapping values between keys.
- **Keys are not protected** — they stay serialized-but-plaintext to remain
  deterministic and lookupable, so secrets/PII must not be placed in cache keys.
- **Unrecoverable entries read as a miss** — a value that fails authentication
  (corrupt, truncated, wrong key, tampered, or relocated) is treated as a cache
  miss (`Ok(None)`) and emits a `cache.unprotect_failed` telemetry event (tagged
  `fallback = true`, since the protected tier always sits on the fallback side) so
  tampering is observable, consistent with the serialization codec's soft-failure
  behavior.

## Design

- Protection is applied by an internal `ProtectedTier` installed at the storage
  boundary, where both the key and value are in scope — this is what makes the
  key-as-context binding possible.
- `.protect_with()` returns a dedicated `ProtectedTransformBuilder` (storage types
  fixed to `BytesView`) supporting `.fallback()` and `.build()`, mirroring
  `TransformBuilder`. It lives in its own `builder/encrypt.rs`, matching the
  `serialize` feature's file layout.
- The public `ValueProtector` trait is the pluggable seam; the crate provides no
  cryptographic implementation. A complete reference `ValueProtector` backed by
  SymCrypt (FIPS-certifiable AES-256-GCM) is documented in the crate-level docs as
  a copy-paste example, since SymCrypt requires a native library at build/run time
  and pulling it into the workspace would force it onto all `--all-features` CI.
- Naming follows a capability-vs-mechanism split: the `encrypt` **feature** names
  the capability (discoverable; avoids overloading the existing
  `stampede_protection`), while the API speaks `protect`/`unprotect` — the same way
  .NET's `DataProtection` package exposes an `IDataProtector.Protect` method.

## Testability

Per the Microsoft Pragmatic Rust Guidelines (M-MOCKABLE-SYSCALLS,
M-DESIGN-FOR-AI, M-TEST-UTIL), the crate abstracts the non-deterministic
crypto/entropy work behind the `ValueProtector` seam and ships a mock so
downstream consumers can test their protected-cache pipelines without a crypto
dependency:

- **`MockValueProtector`** — a deterministic, crypto-free `ValueProtector` gated
  behind `test-util` and re-exported at the crate root next to `MockCache`. It
  binds `context` and soft-fails on mismatch/corruption, but provides **no
  confidentiality** (documented as test-only).

## Dependencies

None. The `encrypt` feature adds no cryptographic dependency; consumers bring
their own approved library.

## Testing

- Unit tests for the `ProtectedTier` mechanism and the `MockValueProtector`.
- Integration tests drive the full `.serialize().protect_with().fallback()`
  pipeline via `MockValueProtector`: stored bytes are ciphertext with the
  plaintext never appearing verbatim, values round-trip through the protected tier,
  a fresh nonce is used per insert, the boundary is reachable on a `FallbackBuilder`
  and through chained post-transform fallbacks, and a value relocated to a
  different key reads as a miss.
- 100% line and region coverage on the added `transform/encrypt.rs` and
  `builder/encrypt.rs`.


crates/cachet/src/transform/encrypt.rs
```diff
@@ -0,0 +1,407 @@
...
+//! Authenticated encryption of cache values stored in an untrusted tier.
+//!
+//! [`AeadCipher`] is the pluggable encryption contract; [`Aes256GcmCipher`] is
+//! the built-in AES-256-GCM implementation. [`EncryptedTier`] installs a cipher
+//! at the storage boundary, where both the key and value are available, and
+//! authenticates each value against its storage key.
+
+use std::borrow::Cow;
+
+use aes_gcm::aead::{Aead, AeadInPlace, KeyInit, Payload};
```
ralfbiedert: From a compliance perspective, should this code really come from that crate? Would love to hear Sergey's input here.

ralfbiedert: @sgalkin

ralfbiedert (dismissed):
(Adding a blocker for now until we know more about that)

crates/cachet/src/transform/encrypt.rs
schgoo: I've converted it to take an AES cipher implementation as input, so we don't _have_ to be opinionated about the crypto library. But I _also_ added a symcrypt implementation behind a feature gate, rather than use aes_gcm.

crates/cachet/src/transform/symcrypt_cipher.rs
```diff
@@ -0,0 +1,209 @@
...
+        // copy), encrypt it in place, then assemble `nonce || ciphertext || tag`.
+        let mut buffer = to_contiguous(plaintext).into_owned();
+        let mut tag = [0u8; TAG_SIZE];
+        self.key.encrypt_in_place(&nonce, aad, &mut buffer, &mut tag);
+
+        let mut result = Vec::with_capacity(NONCE_SIZE + buffer.len() + TAG_SIZE);
+        result.extend_from_slice(&nonce);
+        result.extend_from_slice(&buffer);
+        result.extend_from_slice(&tag);
+        Ok(result.into())
```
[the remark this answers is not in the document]
schgoo: Deleting this for now anyways

crates/cachet/tests/encrypt.rs
```diff
@@ -0,0 +1,304 @@
...
+//! These exercise the encryption *pipeline* (builder wiring, key-as-AAD binding,
+//! relocation defense, fallback chaining) using a crypto-free mock [`AeadCipher`],
+//! so they run under the base `encrypt` feature with no cryptographic dependency.
+
+#![cfg(all(feature = "encrypt", feature = "serialize", feature = "test-util"))]
+
+use std::sync::atomic::{AtomicU32, Ordering};
+
+use bytesbuf::BytesView;
+use cachet::{AeadCipher, Cache, CacheEntry, CacheOp, CacheTier, DecodeOutcome, Error, MockCache};
```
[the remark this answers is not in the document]
schgoo: Copilot is wrong.

crates/cachet/src/transform/encrypt.rs
```diff
@@ -0,0 +1,517 @@
...
+impl MockValueProtector {
+    /// Creates a new mock protector.
+    #[must_use]
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Derives a nonce from the counter (counter in the first 4 bytes, then filler).
+    fn nonce_bytes(counter: u32) -> [u8; MOCK_NONCE_SIZE] {
+        let mut nonce = [0xA5u8; MOCK_NONCE_SIZE];
```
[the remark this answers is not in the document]
schgoo: This is part of a non-cryptographic mock helper, but sure ok

crates/cachet/src/builder/encrypt.rs
```diff
@@ -0,0 +1,248 @@
...
+        crate::Cache::new(type_name::<Self::TierOutput>(None), tier, clock, telemetry, stampede_protection)
+    }
+
+    fn build_tier(self, clock: Clock, telemetry: CacheTelemetry, fallback: bool) -> Self::TierOutput {
+        let pre_tier = self.pre.build_tier(clock.clone(), telemetry.clone(), fallback);
+
+        // Build the post-transform tier chain and wrap it so values are protected
+        // (and key-bound) before reaching it.
+        let post_tier = self.post.build_tier(clock.clone(), telemetry.clone(), true);
+        let protected = ProtectedTier::new(
```
martin-kolinek: 🤖: Wrapping the already-built `post_tier` in a single `ProtectedTier` means unprotect soft-failures happen *outside* the post chain's own `FallbackCache`. With chained post tiers, if the first post tier returns tampered/corrupt bytes that inner fallback treats it as a hit and never consults the next post tier; only afterward does the outer `ProtectedTier` turn the result into `None`. That breaks the promised "soft failure = miss, fall through" behavior for chained post tiers. Consider applying the protection decorator to each post storage tier before composing the fallback chain (e.g. `FallbackCache<ProtectedTier<L2>, ProtectedTier<L3>>`, sharing the protector via `Arc`), or add validated-hit handling that lets the chain continue after an unprotect soft-failure. Worth covering with an integration test that nests an unbuilt `ProtectedTransformBuilder` as a `.fallback(...)` across multiple post tiers and forces a tampered read on the first store — the current chained-fallback test only adds tiers inside one protected builder, so it would not catch this.

crates/cachet/src/builder/encrypt.rs
```diff
@@ -0,0 +1,248 @@
...
+use crate::{Codec, Encoder};
+
+/// The builder produced by [`TransformBuilder::protect_with`].
+///
+/// It mirrors [`TransformBuilder`] but fixes the storage types to
+/// [`BytesView`] and carries a value protector. At build time the post-transform tier
+/// chain is wrapped in an internal `ProtectedTier`, which protects values and binds
+/// each value to its storage key. Add post tiers with [`fallback`](Self::fallback) and
+/// finish with [`build`](Self::build), exactly as with `TransformBuilder`.
+pub struct ProtectedTransformBuilder<K, V, Pre, Post = ()> {
```
martin-kolinek: 🤖: Design/maintainability: `ProtectedTransformBuilder` re-implements almost the entire `TransformBuilder` surface (both `fallback` impls, `Sealed`, `CacheTierBuilder`, `Buildable`, `build`, `Debug`, and the field set). After `.protect_with`, the storage types are already fixed to `BytesView, BytesView` and `TransformBuilder::fallback` already constrains post tiers to that same pair, so the parallel builder buys little extra type-state safety — it mainly carries the `Box<dyn ValueProtector>`. Consider instead giving `TransformBuilder` a feature-gated `Option<Box<dyn ValueProtector>>` field that `.protect_with` sets, and branching in `build_tier` to wrap the post tier in `ProtectedTier` when present. That removes ~200 lines of parallel surface that otherwise must be kept in lockstep as fallback/refresh/stampede wiring evolves. Tradeoff: it threads an optional crypto concern through the general builder and adds some cfg-gated field noise — a maintainability-vs-clarity call for a maintainer, not a correctness issue.

crates/cachet/src/telemetry/cache.rs
```diff
@@ -372,6 +372,33 @@ impl CacheTelemetry {
...
+    /// relocated value. The protected tier always sits on the post-transform
+    /// (fallback) side of the hierarchy, so the event is tagged
+    /// `fallback = true` to match the tier's other events. Like other validation
+    /// outcomes (e.g. `insert_rejected`), it is not a timed operation and so omits
+    /// `cache.duration_ns`.
+    #[cfg(feature = "encrypt")]
+    pub(crate) fn record_unprotect_failure(&self, cache_name: CacheName) {
+        #[cfg(any(feature = "logs", test))]
+        if self.logging_enabled {
+            tracing::warn!(cache.name = cache_name, cache.event = attributes::EVENT_UNPROTECT_FAILED);
```
martin-kolinek, line 390: 🤖: This emits `cache.unprotect_failed` at WARN, but the public event-level table in `src/lib.rs` (the `### Event types` table, ~lines 370-374) has only ERROR/INFO/DEBUG rows — no WARN row — and omits this stable event, and its generated `README.md` mirror is likewise missing it. Please add a WARN row for `cache.unprotect_failed` to the table and regenerate the README so the documented telemetry inventory matches what this feature emits.

crates/cachet/src/telemetry/cache.rs
```diff
@@ -372,6 +372,33 @@ impl CacheTelemetry {
...
+        }
+
+        self.emit_tier_event(
```
martin-kolinek, line 393: 🤖: Consider a builder-level test that drives an unprotect soft-failure (malformed or relocated ciphertext) through a `CacheEventHandler` and asserts the `cache.unprotect_failed` tier callback carries `fallback = true` and the same nonzero request ID as the completed `get`. The existing focused test only greps tracing text for the event name, so regressions in the public structured telemetry fields / request-id correlation would still pass.

crates/cachet/src/transform/encrypt/mock.rs
```diff
@@ -0,0 +1,194 @@
...
+use bytesbuf::BytesView;
+
+use super::{Rejection, Unprotected, ValueProtector};
+use crate::Error;
+
+/// Length of the mock protector's nonce prefix, in bytes.
+const MOCK_NONCE_SIZE: usize = 12;
+
+/// Width of the little-endian context-length field stored after the nonce.
+const CONTEXT_LEN_SIZE: usize = size_of::<u64>();
```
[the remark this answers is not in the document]
schgoo: `size_of` is in prelude. False positive

crates/cachet/src/transform/codec.rs
```diff
@@ -65,29 +111,42 @@ pub trait Encoder<From, To>: Send + Sync {
 pub enum DecodeOutcome<T> {
     /// The value was successfully decoded.
     Value(T),
-    /// The stored data is undecodable and should be treated as a cache miss.
-    ///
-    /// The string describes the reason (e.g., "version mismatch", "empty payload").
-    SoftFailure(&'static str),
+    /// The stored data is undecodable and should be treated as a cache miss (as opposed
+    /// to a hard [`Error`], which propagates). Why it was undecodable is not part of the
```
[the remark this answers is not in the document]
schgoo: semver bump will happen in a separate PR

crates/cachet/src/transform/encrypt/protector.rs
```diff
@@ -0,0 +1,89 @@
...
+/// as the context, so a value is cryptographically bound to the key it was stored under.
+///
+/// This trait supplies no implementation of its own: implement it with your
+/// organization's approved cryptographic library and register it via
+/// [`protect_with`](crate::SerializeBuilder::protect_with). See the crate-level
+/// "Encryption Boundary" docs for a reference `SymCrypt`-backed implementation.
+///
+/// # Security contract
+///
+/// Implementors **must** bind `context`: [`unprotect`](Self::unprotect) must return
```
geeknoid: **The security contract does not require confidentiality or ciphertext integrity** — *Security · Medium · High*

The least-privileged relevant actor is a reader or writer of the untrusted backing tier, after a consumer supplies a protector that follows this documented `must` literally. The only normative requirement is context mismatch rejection. An implementation can therefore store plaintext plus the context and accept arbitrary modifications to the body while still satisfying the stated contract; `ProtectorCodec` stores that output verbatim and trusts `Recovered` on read. That leaves both the advertised encryption and authenticated-integrity properties unenforced at the pluggable boundary.

**Direction:** Make confidentiality and authentication of the complete protected representation explicit normative requirements, including rejection after any protected-byte modification; distinguish deliberately supported non-confidential protectors if those are intended.

**Done when:** An implementation that exposes plaintext or recovers any modified protected bytes is unambiguously contract-violating, and contract tests/examples demonstrate both properties.

crates/cachet/src/transform/encrypt/codec.rs
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    fn decode(&self, ctx: &CodecContext<'_>, value: BytesView) -> Result<DecodeOutcome<BytesView>, Error> {
+        match self.protector.unprotect(ctx.key(), &value)? {
+            Unprotected::Recovered(recovered) => Ok(DecodeOutcome::Value(recovered)),
+            Unprotected::Rejected(Rejection::AuthenticationFailed) => {
+                self.telemetry.record_unprotect_failure(std::any::type_name::<Self>());
+                Ok(DecodeOutcome::SoftFailure)
+            }
+            Unprotected::Rejected(Rejection::Malformed) => Ok(DecodeOutcome::SoftFailure),
```
geeknoid: **Malformed envelopes let backing-store tampering evade the advertised security signal** — *Security · Medium · High*

An actor with write access to the untrusted tier controls these stored bytes. Truncating the envelope or changing its framing/version follows the public `ValueProtector` contract into `Rejection::Malformed`, and this arm converts it directly to a miss without `cache.unprotect_failed`. The operation therefore succeeds as `Ok(None)` while the integrity/availability attack is indistinguishable from an ordinary cache miss, contrary to the PR description's claim that corrupt, truncated, or tampered entries are observable.

**Direction:** Emit structured, rate-limitable telemetry for every unrecoverable protected value, retaining the rejection category so expected format migration can be distinguished from authentication failure.

**Done when:** Corrupt, truncated, wrong-version, wrong-key, relocated, and modified protected values all produce a correlated security event with their category before becoming a miss.

crates/cachet/src/transform/encrypt/codec.rs
```diff
@@ -0,0 +1,52 @@
...
+impl Codec<BytesView, BytesView> for ProtectorCodec {
+    fn encode(&self, ctx: &CodecContext<'_>, value: &BytesView) -> Result<BytesView, Error> {
+        self.protector.protect(ctx.key(), value)
+    }
+
+    fn decode(&self, ctx: &CodecContext<'_>, value: BytesView) -> Result<DecodeOutcome<BytesView>, Error> {
+        match self.protector.unprotect(ctx.key(), &value)? {
+            Unprotected::Recovered(recovered) => Ok(DecodeOutcome::Value(recovered)),
+            Unprotected::Rejected(Rejection::AuthenticationFailed) => {
+                self.telemetry.record_unprotect_failure(std::any::type_name::<Self>());
```
geeknoid: **The unprotect event is attributed to the codec type instead of the affected tier** — *Correctness · Medium · High*

`record_unprotect_failure` writes its argument into both `cache.name` and `CacheTierEvent::tier_name`, but this always passes `cachet::transform::encrypt::codec::ProtectorCodec`. In an L2/L3 hierarchy every failure is consequently reported under the same implementation type rather than the configured storage-tier name, so operators cannot identify which backing tier contains the bad value.

**Direction:** Emit the event at a layer that already knows the tier name and fallback position, or inject that identity when the protected tier is built.

**Done when:** A multi-tier test causes failures in two named protected tiers and observes each `cache.unprotect_failed` event carrying the correct configured tier name.

crates/cachet/src/transform/encrypt/codec.rs
```diff
@@ -0,0 +1,52 @@
...
+    }
+}
+
+impl Codec<BytesView, BytesView> for ProtectorCodec {
+    fn encode(&self, ctx: &CodecContext<'_>, value: &BytesView) -> Result<BytesView, Error> {
+        self.protector.protect(ctx.key(), value)
+    }
+
+    fn decode(&self, ctx: &CodecContext<'_>, value: BytesView) -> Result<DecodeOutcome<BytesView>, Error> {
+        match self.protector.unprotect(ctx.key(), &value)? {
```
geeknoid: **Hard protector failures are not tested through the cache boundary** — *Testing · Medium · High*

All added tests use `MockValueProtector`, whose `protect` and `unprotect` paths return `Ok`. A mutation that turns either propagated `Err` into `SoftFailure` or a successful empty value would therefore survive, even though the public contract says operational failures must reach `Cache::insert`/`Cache::get`. The encode-side propagation is at [line 39](https://github.com/microsoft/oxidizer/blob/ec21d739b7da30188a964d4ba5cba9ab68922b13/crates/cachet/src/transform/encrypt/codec.rs#L39).

**Direction:** Add an integration-test protector that independently fails `protect` and `unprotect`, and assert the exact cache-operation result plus the absence of misleading authentication telemetry.

**Done when:** Tests fail if either `?` propagation is replaced by a miss or success, and pass only when the original hard error reaches the caller.

geeknoid (commented):
Completed a static-only five-pass review of the full PR and unchanged blast radius: **150 security, 162 correctness, 182 testing, 183 performance, and 143 conformance classes**. Nothing was built, executed, benchmarked, profiled, fuzzed, or dynamically tested.

The key-as-context data flow is coherent, protected-tier soft failures fall through to later tiers, and the type-state prevents applying protection twice. The main concerns are the pluggable security contract and whether tampering telemetry is complete and operationally attributable.

Body-only documentation defect: `crates/cachet/src/transform/mod.rs:22` and `crates/cachet/src/serialize/mod.rs:10` still describe `DecodeOutcome::SoftFailure(reason)`, although this PR changes it to a unit variant.

Prior feedback was deduplicated, including the public API/semver change, key-copy allocations, telemetry duration/fallback fields, feature-gated doctests, and earlier mock implementation issues.

crates/cachet/src/builder/encrypt.rs
```diff
@@ -0,0 +1,51 @@
...
+    /// ```ignore
+    /// let cache = Cache::builder::<String, String>(clock)
+    ///     .memory()
+    ///     .serialize()
+    ///     .protect_with(my_protector) // any `ValueProtector` implementation
+    ///     .fallback(remote)
+    ///     .build();
+    /// ```
+    #[must_use]
+    pub fn protect_with(self, protector: impl ValueProtector + 'static) -> SerializeBuilder<K, V, Pre, true> {
```
geeknoid: **The new per-access protection stage has no benchmark or allocation guard** — *Performance · Low · Structural*

Every insert reaching a protected fallback and every protected fallback hit executes serialization, context construction, an additional dynamic codec stage, and the caller's protector. The existing `cachet` benchmarks cover plain wrapper/fallback operations but no serialize/protect path, so this hot-path cost and its allocations per operation can regress without evidence. No magnitude is claimed here; the path frequency is one protection call per affected remote-tier access. This partially misses Pragmatic Rust guideline [M-HOTPATH](https://microsoft.github.io/rust-guidelines/guidelines/performance/#M-HOTPATH), which requires identifying, profiling, and optimizing the hot path early.

**Direction:** Add paired benchmarks for protected insert and fallback hit across representative payload sizes and single-/multi-span keys, recording wall time, instruction count, and allocations per operation against an unprotected serialization baseline.

**Done when:** The path has a reproducible baseline and regression gate, and any optimization is resolved by measurement—landed with evidence or closed as not worthwhile.

PR #560, opened 2026-07-10
perf(seatbelt): cache default breaker engine and drop metrics allocation

geeknoid:
Add gungraun (Callgrind) instruction-count benchmarks for the timeout, retry, breaker, hedging, and observability middleware, and use them to drive two hot-path optimizations. No public API change.

- breaker: pre-create a cached default engine so the common single-breaker path (no id_provider) skips the RwLock + HashMap + SipHash lookup in Engines::get_engine.
- retry: emit the recovery-kind metric attribute as a &'static str (RecoveryKind::as_str) instead of allocating a String via to_string.

| Benchmark (instructions)      | Before | After  | Change |
| ----------------------------- | -----: | -----: | -----: |
| breaker / with_breaker        |  4,938 |  3,263 | -33.9% |
| observability / retry_metrics | 60,063 | 56,209 |  -6.4% |

All other benchmark scenarios are unchanged.

Copilot-Session: 0edbf54e-9f2c-4ad4-bd94-88aa25d68eda

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3354,6 +3354,7 @@ dependencies = [
  "fastrand",
  "futures",
  "futures-util",
+ "gungraun",
  "http",
  "insta",
  "jiff",
--- a/crates/seatbelt/Cargo.toml
+++ b/crates/seatbelt/Cargo.toml
@@ -91,6 +91,9 @@ tower-service = { workspace = true }
 tracing = { workspace = true }
 tracing-subscriber = { workspace = true, features = ["fmt", "std"] }
 
+[target.'cfg(target_os = "linux")'.dev-dependencies]
+gungraun = { workspace = true, features = ["default"] }
+
 [[example]]
 name = "timeout"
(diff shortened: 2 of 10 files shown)


martintmk (approved):

Vaiz (approved):

ralfbiedert (approved):

martintmk:
Note that circuit breaker partitioning is a common scenario, another optimalization that can be done is for cases where partitioning is set, but only ever partition is created. It could use similar optimalization. 

Typical scenario, HTTP client that uses authority-based partitioning but all requests send to the same authority. 

- `https://example.com/hello-1`
- `https://example.com/hello-2`

both of these will re-use the same circuit and it's pretty common that all requests coming from HTTP client will keep sending the request to one partition. (unless user opts-in into hedging or fallback routing strategies)

geeknoid:

sandersaares (approved):

PR #562, opened 2026-07-10
perf(seatbelt): use a BTreeMap for circuit breaker partition lookup

geeknoid:
Partitioned circuit breakers keep their per-partition engines in a map keyed by `BreakerId`. Switching that map from `HashMap` to `BTreeMap` resolves engines by ordered key comparison instead of hashing the `BreakerId` on every call. Partition sets are low-cardinality by design, so a handful of comparisons is cheaper than a SipHash across every size, and — unlike a hash map — the lookup is not exposed to hash-flooding via request-derived IDs. `BreakerId` gains `PartialOrd`/`Ord` (a compatible addition).

Adds breaker Callgrind and Criterion benchmarks covering the default, single-partition, many-partition, high-cardinality, and miss/insert cases so lookup regressions are visible across the full range.

| Benchmark (instructions)          | HashMap | BTreeMap | Change |
| --------------------------------- | ------: | -------: | -----: |
| breaker / with_partitioned (n=1)  |   2,927 |    2,713 |  -7.3% |
| breaker / with_partitioned (n=16) |   2,927 |    2,741 |  -6.4% |
| breaker / with_partitioned (n=256)|   3,050 |    2,769 |  -9.2% |
| breaker / with_partitioned_miss   |   4,291 |    3,994 |  -6.9% |

### DIFF
--- a/crates/seatbelt/benches/breaker.rs
+++ b/crates/seatbelt/benches/breaker.rs
@@ -5,7 +5,7 @@ use alloc_tracker::{Allocator, Session};
 use criterion::{Criterion, criterion_group, criterion_main};
 use futures::executor::block_on;
 use layered::{Execute, Service, Stack};
-use seatbelt::breaker::Breaker;
+use seatbelt::breaker::{Breaker, BreakerId};
 use seatbelt::{RecoveryInfo, ResilienceContext};
 use tick::Clock;
 
@@ -22,7 +22,7 @@ fn entry(c: &mut Criterion) {
     group.bench_function("no-breaker", |b| {
         b.iter(|| {
             let _span = operation.measure_thread();
-            _ = block_on(service.execute(Input));
+            _ = block_on(service.execute(Input(0)));
         });
     });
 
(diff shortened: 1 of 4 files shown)


geeknoid:

crates/seatbelt/benches/breaker.rs
```diff
@@ -46,6 +46,29 @@ fn entry(c: &mut Criterion) {
...
+            .recovery_with(|_, _| RecoveryInfo::never())
+            .rejected_input_error(|_input, _args| Output)
+            .min_throughput(1000),
+        Execute::new(|_input: Input| async move { Ok(Output) }),
+    )
+        .into_service();
+
+    let operation = session.operation("with-partitioned");
+    group.bench_function("with-partitioned", |b| {
+        b.iter(|| {
```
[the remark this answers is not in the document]
geeknoid: Good catch. The Criterion `with-partitioned` scenario now warms the service with one `execute` call before the measured loop (mirroring the Callgrind `setup`), so the loop only exercises the steady-state single-partition fast path.

Vaiz (commented):
- what if we use `BTreeMap` instead? wouldn't it give us the same effect for free? 
- benchmark doesn't capture the negative impact in case where there are more than one engine
- it ignores the case when the map is empty

geeknoid:

Vaiz (approved):

crates/seatbelt/src/breaker/engine/engines.rs
```diff
@@ -35,12 +35,15 @@ impl Engines {
...
         // Fast path: the default breaker (the common single-breaker configuration, used
         // whenever no ID provider is configured) is served by a pre-created engine. This
-        // avoids a lock, a hash of the key, and a map probe on every call.
+        // avoids a lock and the map lookup entirely on every call.
         if key.is_default() {
             return Arc::clone(&self.default_engine);
         }
 
-        // Fast path: read lock for existing engines (common case).
+        // Read-lock path for existing engines (common case). The map is a `BTreeMap`, so a
```
martintmk, line 43: alternatively, how about just using cheaper hash such as:

https://docs.rs/rustc-hash/latest/rustc_hash/index.html

Vaiz: cheaper hash still hashes. if your main case is 0-2 items, comparison is cheaper

crates/seatbelt/src/breaker/engine/engines.rs
```diff
@@ -35,12 +35,15 @@ impl Engines {
...
+        // lookup is a handful of key comparisons rather than a hash of `key`; partitioned
+        // breakers are low-cardinality by design, so this stays cheap and, unlike a hash
```
sandersaares, line 45: I would like to see a design document reference for this "low-cardinality by design" claim.

sandersaares (approved):

ralfbiedert (approved):

martintmk (approved):

crates/seatbelt/src/breaker/engine/engines.rs
geeknoid: I think this is inherent in how this stuff works. High cardinality would led to massive inefficiency.

PR #564, opened 2026-07-11
feat(http_path_template): add parser for google.api.http path templates

geeknoid:
Introduce `http_path_template`, a dependency-free, zero-copy parser for the `google.api.http` path-template grammar: literal segments, `*`, `**`, `{field.path=sub-template}` variables, and a trailing `:verb`. A parse borrows from the input string and allocates only the top-level segment list.

The accepted syntax is selected with a `Grammar` argument: the default is the strict `google.api.http` grammar, and `with_segment_affixes` opts into an extended grammar that also allows intra-segment prefix/suffix parameters (e.g. `/files/{name}.json`).

Copilot-Session: 867c37d0-9e7a-46b0-b5a9-dc9f551ed128

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ These are the primary crates built out of this repo:
 - [`fetch_options`](./crates/fetch_options/README.md) - Options types for 'fetch' crate.
 - [`fundle`](./crates/fundle/README.md) - Compile-time safe dependency injection for Rust.
 - [`http_extensions`](./crates/http_extensions/README.md) - Shared HTTP types and extension traits for clients and servers.
+- [`http_path_template`](./crates/http_path_template/README.md) - Parser for the google.api.http path-template grammar (literals, `*`, `**`, `{field=sub}` variables, `:verb`).
 - [`layered`](./crates/layered/README.md) - A foundational service abstraction for building composable, middleware-driven systems.
 - [`multitude`](./crates/multitude/README.md) - Fast and flexible arena allocator.
 - [`ohno`](./crates/ohno/README.md) - High-quality Rust error handling.
--- a/crates/http_path_template/README.md
+++ b/crates/http_path_template/README.md
@@ -1,0 +1,121 @@
+<div align="center">
+ <img src="./logo.png" alt="Http Path Template Logo" width="96">
+
+# HTTP Path Template
+
+[![crate.io](https://img.shields.io/crates/v/http_path_template.svg)](https://crates.io/crates/http_path_template)
+[![docs.rs](https://docs.rs/http_path_template/badge.svg)](https://docs.rs/http_path_template)
+[![MSRV](https://img.shields.io/crates/msrv/http_path_template)](https://crates.io/crates/http_path_template)
+[![CI](https://github.com/microsoft/oxidizer/actions/workflows/main.yml/badge.svg?event=push)](https://github.com/microsoft/oxidizer/actions/workflows/main.yml)
+[![Coverage](https://codecov.io/gh/microsoft/oxidizer/graph/badge.svg?token=FCUG0EL5TI)](https://codecov.io/gh/microsoft/oxidizer)
+[![License](https://img.shields.io/badge/license-MIT-blue.svg)](../../LICENSE)
+<a href="../.."><img src="../../logo.svg" alt="This crate was developed as part of the Oxidizer project" width="20"></a>
+
+</div>
+
+A parser for the [`google.api.http`][__link0] path-template grammar.
+
+A path template is the pattern that appears in a `google.api.http`
+annotation, for example `/shelves/{shelf}/books/{book=**}:archive`. This crate
+turns such a string into a validated, structured [`PathTemplate`][__link1] — an
+abstract syntax tree of [`Segment`][__link2]s (literals, `*`, `**`, and
+`{field.path=sub-template}` [`Variable`][__link3] bindings) plus an optional custom
+`:verb`.
+
+Parsing is zero-copy: the returned [`PathTemplate`][__link4] borrows from the input
+string (every literal, field name, and verb is a slice into it), so a parse
+copies no text and allocates only the top-level segment list.
+
+A template must begin with `/` and is a `/`-separated sequence of segments —
+each `/` delimits one segment. Literal segments, variable field names, and the
+custom `:verb` are preserved and compared **verbatim**, so the grammar is
+case-sensitive; the parser performs no case folding.
+
+The grammar mirrors the reference [`google.api.HttpRule`][__link5] path syntax:
+
+* a **literal** segment (`shelves`) must match verbatim;
+* **`*`** ([`Segment::Single`][__link6]) matches exactly one non-empty segment;
+* **`**`** ([`Segment::Rest`][__link7]) matches the remaining segments and may only
+  appear as the final element;
+* **`{field.path=sub-template}`** ([`Segment::Variable`][__link8]) captures the portion
+  of the path matched by its sub-template into a dotted message field; the
+  shorthand `{field}` is `{field=*}` and nested variables are rejected;
+* a trailing **`:verb`** declares a custom method verb.
+
+## Extended grammar
+
+[`PathTemplate::parse`][__link9] takes a [`Grammar`][__link10] argument. The default grammar is
+the strict `google.api.http` syntax above; passing a [`Grammar`][__link11] with
+[`Grammar::with_segment_affixes`][__link12] enabled additionally allows **intra-segment
+prefix/suffix parameters**: a single segment may wrap one `{field.path}`
+variable in literal text, for example `/files/{name}.json`, `/v{version}/x`,
+or `/img-{id}.png`. Such a segment parses to a [`Segment::Affix`][__link13]. The strict
+grammar rejects this syntax.
+
+## Examples
+
+Parsing `/shelves/{shelf}/books/{book=**}:archive` yields four top-level
+[`Segment`][__link14]s plus the custom verb `archive`:
+
+* `shelves` — a [`Segment::Literal`][__link15];
+* `{shelf}` — a [`Segment::Variable`][__link16] binding field `shelf` to a single
+  segment (`*`, i.e. [`Segment::Single`][__link17]);
+* `books` — a [`Segment::Literal`][__link18];
+* `{book=**}` — a [`Segment::Variable`][__link19] binding field `book` to the remaining
+  segments (`**`, i.e. [`Segment::Rest`][__link20]).
+
+```rust
+use http_path_template::{Grammar, PathTemplate, Segment};
+
+let template = PathTemplate::parse(
+    "/shelves/{shelf}/books/{book=**}:archive",
+    Grammar::default(),
+)?;
+
+assert_eq!(template.segments().len(), 4);
+assert_eq!(template.verb(), Some("archive"));
+
+assert_eq!(template.segments()[0], Segment::Literal("shelves"));
+assert_eq!(template.segments()[2], Segment::Literal("books"));
+
+let Segment::Variable(shelf) = template.segments()[1] else {
+    panic!("expected variable")
+};
+assert_eq!(shelf.field_path(), "shelf");
+assert!(shelf.segments().eq([Segment::Single]));
+
+let Segment::Variable(book) = template.segments()[3] else {
+    panic!("expected variable")
+};
+assert_eq!(book.field_path(), "book");
+assert!(book.segments().eq([Segment::Rest]));
+```
+
+
+<hr/>
+<sub>
+This crate was developed as part of <a href="../..">The Oxidizer Project</a>. Browse this crate's <a href="https://github.com/microsoft/oxidizer/tree/main/crates/http_path_template">source code</a>.
+</sub>
+
+ [__cargo_doc2readme_dependencies_info]: ggGmYW0CYXZlMC43LjJhdIQbLiTyV0MU86EbZU15e0PmecoboQ9jo59bnAEbyDXw04U13GlhYvRhcoQbdehWW1wo3-MbQZ4f-tMBwgsbFQlyCp62fwcbNkxL2v1MviZhZIGCcmh0dHBfcGF0aF90ZW1wbGF0ZWUwLjEuMA
+ [__link0]: https://github.com/googleapis/googleapis/blob/master/google/api/http.proto
+ [__link1]: https://docs.rs/http_path_template/0.1.0/http_path_template/?search=PathTemplate
+ [__link10]: https://docs.rs/http_path_template/0.1.0/http_path_template/?search=Grammar
+ [__link11]: https://docs.rs/http_path_template/0.1.0/http_path_template/?search=Grammar
+ [__link12]: https://docs.rs/http_path_template/0.1.0/http_path_template/?search=Grammar::with_segment_affixes
+ [__link13]: https://docs.rs/http_path_template/0.1.0/http_path_template/?search=Segment::Affix
+ [__link14]: https://docs.rs/http_path_template/0.1.0/http_path_template/?search=Segment
+ [__link15]: https://docs.rs/http_path_template/0.1.0/http_path_template/?search=Segment::Literal
+ [__link16]: https://docs.rs/http_path_template/0.1.0/http_path_template/?search=Segment::Variable
+ [__link17]: https://docs.rs/http_path_template/0.1.0/http_path_template/?search=Segment::Single
+ [__link18]: https://docs.rs/http_path_template/0.1.0/http_path_template/?search=Segment::Literal
+ [__link19]: https://docs.rs/http_path_template/0.1.0/http_path_template/?search=Segment::Variable
+ [__link2]: https://docs.rs/http_path_template/0.1.0/http_path_template/?search=Segment
+ [__link20]: https://docs.rs/http_path_template/0.1.0/http_path_template/?search=Segment::Rest
+ [__link3]: https://docs.rs/http_path_template/0.1.0/http_path_template/?search=Variable
+ [__link4]: https://docs.rs/http_path_template/0.1.0/http_path_template/?search=PathTemplate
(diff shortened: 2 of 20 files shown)


crates/http_path_template/src/error.rs
```diff
@@ -0,0 +1,164 @@
+// Copyright (c) Microsoft Corporation.
```
geeknoid, line 1: @ralfbiedert So should this be using ohno?

geeknoid: @sandersaares Based on your question in my other PR.

crates/http_path_template/README.md
```diff
@@ -0,0 +1,119 @@
+<div align="center">
+ <img src="./logo.png" alt="Http Path Template Logo" width="96">
```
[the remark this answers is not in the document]
geeknoid: This `./logo.png` reference is produced by the shared `crates/README.j2` template and is byte-identical across every published crate in the workspace (e.g. `bytesbuf`, `ohno`, `fetch`), so it reflects an established repo-wide convention rather than something specific to this crate. The logos are intentionally Git-LFS-tracked and kept out of the packaged `include` allowlist workspace-wide. For the rendered docs that matter on docs.rs, the crate root already sets `#![doc(html_logo_url = "https://media.githubusercontent.com/.../logo.png")]` (an absolute URL), so rustdoc renders correctly. Changing only this crate would diverge from the convention and be overwritten by `just readme`; making the README logo absolute everywhere would be a separate workspace-wide change.

crates/http_path_template/src/path_template.rs
```diff
@@ -0,0 +1,482 @@
...
+                if verb_at.is_some() {
+                    return Err(ParseError::new(ParseErrorKind::MultipleVerbs));
+                }
+                verb_at = Some(idx);
+            }
+            _ => {}
+        }
+    }
+
+    match verb_at {
```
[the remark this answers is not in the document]
geeknoid: Good catch — fixed. `split_verb` no longer uses `saturating_sub`: a `}` with no matching `{` before the verb separator is now reported immediately as `UnbalancedBraces`, so `/a}:` returns `UnbalancedBraces` (the leftmost structural error) instead of `EmptyVerb`. This also made the depth-0 `}` guard in `split_and_parse_segments` redundant (the body reaching it can no longer contain a stray `}`), so it was simplified. Added regression cases (`/a}:`, `/a}:b`, and `/a:b}` for a trailing brace inside the verb).

crates/http_path_template/src/path_template.rs
```diff
@@ -0,0 +1,489 @@
...
+        Self::parse_inner(template, grammar)
+    }
+
+    fn parse_inner(template: &'a str, grammar: Grammar) -> Result<Self, ParseError> {
+        let (body, verb) = split_verb(template)?;
+
+        let rest = match body.as_bytes().first() {
+            Some(&b'/') => &body[1..],
+            _ => return Err(ParseError::new(ParseErrorKind::MissingLeadingSlash)),
+        };
```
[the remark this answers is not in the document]
geeknoid: Fixed. `parse_inner` now checks for the leading `/` before calling `split_verb`, so a template without one (e.g. `a}`, `}`, `a:b`) returns `MissingLeadingSlash` — the most fundamental error — rather than a later `UnbalancedBraces`/`EmptyVerb`. Added regression cases.

crates/http_path_template/src/path_template.rs
```diff
@@ -0,0 +1,489 @@
...
+            }
+            _ => {}
+        }
+    }
+    if braces != 2 {
+        return Err(ParseError::new(ParseErrorKind::UnbalancedBraces));
+    }
+    // With exactly one `{` and one `}`, and `split_and_parse_segments` having
+    // rejected any `}` that precedes its `{` (a closing brace at brace-depth 0),
+    // the open brace always comes first and `open < close` holds.
```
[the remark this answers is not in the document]
geeknoid: Good catch — the comment was misattributed. Updated it to reference `split_verb`, whose top-level scan rejects any `}` with no matching `{` (preventing the depth underflow) and thereby guarantees `open < close` here.

crates/http_path_template/src/path_template.rs
```diff
@@ -0,0 +1,497 @@
...
+/// # fn main() -> Result<(), http_path_template::ParseError> {
+/// let template = PathTemplate::parse("/v1/shelves/{shelf}:get", Grammar::default())?;
+/// assert_eq!(template.verb(), Some("get"));
+/// assert_eq!(template.segments().len(), 3);
+/// # Ok(())
+/// # }
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct PathTemplate<'a> {
+    segments: Box<[Segment<'a>]>,
```
martintmk, line 28: Given that amount of segments in URL is generally low we could use something such as https://docs.rs/slimvec/latest/slimvec/ to avoid allocations entirely for small number of elements (<= 4)

sandersaares: Do you mean `smallvec`? AFAICT `slimvec` is meant to minimize the inline part by moving more metadata to the heap.

crates/http_path_template/src/grammar.rs
```diff
@@ -0,0 +1,94 @@
...
+    /// # Examples
+    ///
+    /// ```
+    /// use http_path_template::Grammar;
+    ///
+    /// assert!(!Grammar::default().segment_affixes());
+    /// assert!(Grammar::default().with_segment_affixes().segment_affixes());
+    /// ```
+    #[must_use]
+    pub const fn segment_affixes(self) -> bool {
```
ralfbiedert, line 91: Purely from a MVP perspective, is this needed?

crates/http_path_template/src/error.rs
ralfbiedert: This is one of the optional cases. The crate has zero dependencies (and is probably mostly used during compile / initialization time), so most ohno benefits don't come into play. 

That said we could just make our life easy and say _every_ `oxidizer` error is `ohno`, full stop.

crates/http_path_template/src/error.rs
```diff
@@ -0,0 +1,165 @@
...
+    ///
+    /// ```
+    /// use http_path_template::{Grammar, PathTemplate};
+    ///
+    /// let error = PathTemplate::parse("shelves/{shelf}", Grammar::default())
+    ///     .expect_err("missing leading slash");
+    /// assert!(error.is_missing_leading_slash());
+    /// ```
+    #[must_use]
+    pub const fn is_missing_leading_slash(&self) -> bool {
```
ralfbiedert, line 137: Is anyone asking for / needing these? Like what would I actually do instead of printing the error and then fixing the issue?

crates/http_path_template/src/variable.rs
```diff
@@ -0,0 +1,132 @@
...
+        self.sub
+    }
+}
+
+/// A lazy iterator over a [`Variable`]'s sub-template [`Segment`]s.
+///
+/// Created by [`Variable::segments`]. Because the sub-template was validated at
+/// parse time, iteration is infallible and allocation-free.
+#[derive(Debug, Clone)]
+pub struct SubSegments<'a> {
```
ralfbiedert: `SegmentIter`?

ralfbiedert (approved):

crates/http_path_template/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+//! [`google.api.http`]: https://github.com/googleapis/googleapis/blob/master/google/api/http.proto
+//! [`google.api.HttpRule`]: https://github.com/googleapis/googleapis/blob/master/google/api/http.proto
+
+mod error;
+mod grammar;
+mod path_template;
+mod segment;
+mod variable;
+
+#[doc(inline)]
```
sandersaares: These `doc(inline)` do not do anything - they are wasted lines of code.

crates/http_path_template/src/path_template.rs
martintmk: Yes, I meant `smallvec`, thanks for correction

crates/http_path_template/src/error.rs
martintmk: `ohno` is not exactly cheap. It brings `ohno_macros` that are very dependency heavy. I am myself hesitant to bring it into `tick` because of such heavy dependency tree. I think it's ok for heavier dependencies such as fetch that have complex error types.

sandersaares (approved):

crates/http_path_template/src/error.rs
sandersaares: Is that something we can fix in `ohno` or something inevitable? It feels a little strange to go to the effort of building a fancy error type and then to not use it everywhere. We need to eat our own dog food. If it tastes bad, we need to fix it.

crates/http_path_template/src/grammar.rs
geeknoid: This enables scenarios in rest_over_grpc that are otherwise not implementable. So yes, I think it is needed.

crates/http_path_template/src/error.rs
geeknoid: This is a pattern recommended by the guidelines: C-GOOD-ERR and M-ERRORS-CANONICAL-STRUCTS. So what's the right choice then?

crates/http_path_template/src/path_template.rs
geeknoid: This would bloat PathTemplate to hundreds of bytes. Given the use case, I don't think this would be a good tradeoff.

Note that PathTemplate is primarily going to be used at compile time in the rest_over_grpc crate as part of the code gen step.

crates/http_path_template/src/variable.rs
geeknoid: Done.

crates/http_path_template/src/lib.rs
geeknoid: Removed.

martintmk (approved):

crates/http_path_template/src/error.rs
geeknoid: I've decided I don't want to use `ohno` for this crate. Rationale:

- `http_path_template` is a deliberately zero-dependency, low-level parser, and parsing runs at build/init time, not on any request hot path — so ohno's runtime strengths (cause chaining, enrichment, labels) don't come into play here.
- `ParseError` is a pure leaf validation error: there's no source error to wrap and no context to enrich, which are exactly the two things ohno exists for. An `OhnoCore` here would just carry a permanently-empty cause chain.
- ohno also can't model the shape we want: its derive rejects enums, and this error is a discriminated `ParseErrorKind` exposed via `is_*` predicates (per C-GOOD-ERR / M-ERRORS-CANONICAL). We'd keep all of that regardless, so ohno would replace ~15 lines of trivial trait impls while pulling in the heavy `ohno_macros` tree that @martintmk flagged.
- This matches existing precedent: foundational primitives like `multitude` and `bytesbuf` hand-roll their leaf errors, while ohno is reserved for heavier crates like `fetch` with complex, wrapping error types.

On the dogfooding point: to me, eating our own dog food means using ohno where it earns its keep (wrapping/enrichment/labels), not mandating it on zero-dep leaf primitives where it would be pure overhead.

crates/http_path_template/src/error.rs
```diff
@@ -0,0 +1,165 @@
...
+    backtrace: Box<Backtrace>,
+}
+
+impl ParseError {
+    pub(crate) fn new(kind: ParseErrorKind) -> Self {
+        Self {
+            kind,
+            backtrace: Box::new(Backtrace::capture()),
+        }
+    }
```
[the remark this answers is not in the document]
geeknoid: Fixed. `ParseError` no longer boxes a backtrace unconditionally — it now holds a small `enum MaybeBacktrace { Captured(Box<Backtrace>), Disabled }`, and `Backtrace::capture()` returns `Disabled` (carrying no frames) when `RUST_BACKTRACE` is off, so we skip the allocation entirely in that common case. A full backtrace is still boxed and surfaced in `Debug` when capture is enabled. Net: a failing parse now does **zero** heap allocations by default (verified), while `ParseError` stays 16 bytes so the returned `Result` isn't inflated.

crates/http_path_template/src/error.rs
geeknoid: As discussed offline, keeping these since they're very useful in this crate's own test suite.

crates/http_path_template/src/error.rs
```diff
@@ -0,0 +1,222 @@
...
+    InvalidFieldName,
+    /// A literal segment contained a misplaced `*` (`*`/`**` are only valid as
+    /// whole segments).
+    InvalidLiteral,
+    /// A `**` appeared somewhere other than the final position.
+    RestNotLast,
+    /// A `:` verb separator was present but the verb was empty.
+    EmptyVerb,
+    /// A `:` verb contained a `/` (a verb must be a single trailing literal).
+    InvalidVerb,
```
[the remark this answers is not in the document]
geeknoid: Good catch — updated the `InvalidVerb` doc comment to list all four rejected characters (`/`, `*`, `{`, `}`), matching `split_verb`'s validation and the `describe()` message.

ralfbiedert (approved):

crates/http_path_template/src/segment.rs
```diff
@@ -0,0 +1,48 @@
...
+/// # }
+/// ```
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[non_exhaustive]
+pub enum Segment<'a> {
+    /// A literal path segment that must match verbatim, e.g. `shelves`.
+    Literal(&'a str),
+    /// `*` — matches exactly one (non-empty) path segment.
+    Single,
+    /// `**` — matches the remaining path segments (zero or more). Only valid as
```
schgoo: My local code review agent is flagging this comment as not matching the README

crates/http_path_template/src/error.rs
```diff
@@ -0,0 +1,235 @@
...
+///     "'**' may only appear as the final segment"
+/// );
+/// ```
+#[derive(Debug)]
+pub struct ParseError {
+    kind: ParseErrorKind,
+    #[expect(dead_code, reason = "captured for Debug output and RUST_BACKTRACE diagnostics")]
+    backtrace: MaybeBacktrace,
+}
+
```
psandana, line 116: **[design]** `ParseError` offers eleven `is_*` predicates but no way to `match` on the failure. A `#[non_exhaustive] pub enum ParseErrorKind` returned by a `pub fn kind(&self)` would let callers branch with a single `match`, reads more idiomatically (spirit of M-ERRORS-CANONICAL-STRUCTS), and stays extensible. Trade-off: it makes the kind enum public. Was the predicate-only API a deliberate encapsulation choice?

crates/http_path_template/src/path_template.rs
```diff
@@ -0,0 +1,497 @@
...
+fn split_and_parse_segments(body: &str, extended: bool) -> Result<Box<[Segment<'_>]>, ParseError> {
+    // Pre-size to the segment count so pushing never reallocates (a hint only;
+    // see `segment_count_hint`).
+    let mut out = Vec::with_capacity(segment_count_hint(body));
+
+    let mut depth = 0_usize;
+    let mut start = 0_usize;
+    for (idx, &b) in body.as_bytes().iter().enumerate() {
+        match b {
+            b'{' => depth += 1,
```
psandana, line 290: **[perf]** `segment_count_hint` walks every byte of `body` to pre-size `out`, then `split_and_parse_segments` walks it again to split. For typical short templates the extra full pass likely costs more than the one or two `Vec` reallocations it avoids. Consider dropping the hint (or using a cheaper estimate) and confirming the effect with the existing Callgrind bench.

crates/http_path_template/src/error.rs
```diff
@@ -0,0 +1,235 @@
...
+    pub(crate) fn new(kind: ParseErrorKind) -> Self {
+        Self {
+            kind,
+            backtrace: MaybeBacktrace::capture(),
+        }
+    }
+
+    /// Whether the template did not begin with `/`.
+    ///
+    /// # Examples
```
psandana, line 127: **[design]** Capturing a `Backtrace` on every `ParseError` couples this crate to `std`. Capture is guarded against allocation when backtraces are disabled, but attaching a backtrace to a small structural parse error is fairly heavyweight for a 'dependency-free, zero-copy' crate and precludes `no_std`. Consider feature-gating the backtrace, or whether it belongs on a grammar error at all.

schgoo (approved):

psandana (approved):
Non-critical comments; Approving

crates/http_path_template/src/error.rs
geeknoid: This is fine, this crate is used during built time.

crates/http_path_template/src/segment.rs
geeknoid: Good catch — the `Segment::Rest` doc said "matches the remaining path segments (zero or more)" while the README (generated from the crate-level docs) describes `**` as simply "matches the remaining segments". Dropped the `(zero or more)` parenthetical so the two are consistent.

psandana (approved):

README.md
```diff
@@ -41,6 +41,7 @@ These are the primary crates built out of this repo:
 - [`fetch_options`](./crates/fetch_options/README.md) - Options types for 'fetch' crate.
 - [`fundle`](./crates/fundle/README.md) - Compile-time safe dependency injection for Rust.
 - [`http_extensions`](./crates/http_extensions/README.md) - Shared HTTP types and extension traits for clients and servers.
+- [`http_path_template`](./crates/http_path_template/README.md) - Parser for the google.api.http path-template grammar (literals, `*`, `**`, `{field=sub}` variables, `:verb`).
```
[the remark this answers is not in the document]
geeknoid: We'll fix this next time.

PR #569, opened 2026-07-13
feat: Implement ThreadAware for more more types

AdomasBekeras:
Implement ThreadAware for std::net::SocketAddr


sandersaares (changes requested):
We need to define clear acceptance criteria in `docs/DESIGN.md` for why/when it is OK to add such out-of-box `ThreadAware` implementations. Today this feels like we are just randomly adding stuff here and I do not really see a rationale or a way to review whether something is OK to include or not in this way today. Certainly, including one random piece of Tokio feels weird (why not other pieces?).

@ralfbiedert can you help create these acceptance criteria so we can review this PR and other similar future PRs with some grounding?

AdomasBekeras:
> We need to define clear acceptance criteria in `docs/DESIGN.md` for why/when it is OK to add such out-of-box `ThreadAware` implementations. Today this feels like we are just randomly adding stuff here and I do not really see a rationale or a way to review whether something is OK to include or not in this way today. Certainly, including one random piece of Tokio feels weird (why not other pieces?).
> 
> @ralfbiedert can you help create these acceptance criteria so we can review this PR and other similar future PRs with some grounding?

I'm adding it because Ralf asked to add it so I assume it's ok to add it here. I don't have a comment on the acceptance criteria but that's not really in the scope of this PR.

ralfbiedert:
> I'm adding it because Ralf asked to add it so I assume it's ok to add it here. I don't have a comment on the acceptance criteria but that's not really in the scope of this PR.

Sorry there was a misunderstanding. Adding `ThreadAware` for `std` items is generally fine, adding extra crates not so much. In fact, I'm a bit unhappy we have the existing ones already and we should consider removing them for 1.0 as they must live in the core crate and will be perma stable.

ralfbiedert (approved):

AdomasBekeras:
Removed the impl for the channel now

martintmk (approved):

sandersaares (changes requested):
> I'm adding it because Ralf asked to add it so I assume it's ok to add it here. I don't have a comment on the acceptance criteria but that's not really in the scope of this PR.

This, combined with Ralf's reply, highlights the importance of including acceptance criteria. Let's please add at least a statement into AGENTS.md or DESIGN.md or similar that we can only add implementations for `std` without special exception here. If you think this is an egregious violation of this PR's scope, I am OK if you just file a work item to add this line later.

Vaiz (approved):

crates/thread_aware/src/impls.rs
```diff
@@ -332,6 +334,27 @@ mod tests {
         assert_eq!(err_string, Err("error".to_string()));
     }
 
+    #[test]
```
ralfbiedert: These tests are mostly useless and just noise. The only thing that should be asserted for most types is that the upstream type in fact impls it, which in and by itself can be a simple `const fn assert_impl_threadware<T>() {}`, which in turn violates  M-TAUTOLOGICAL-TESTS.

ralfbiedert (approved):

crates/thread_aware/src/impls.rs
AdomasBekeras: Makes sense, removed

crates/thread_aware/src/impls.rs
```diff
@@ -36,6 +37,7 @@ impl_transfer!(String);
 impl_transfer!(PathBuf);
 impl_transfer!(Duration);
 impl_transfer!(&Path);
+impl_transfer!(SocketAddr);
```
geeknoid: **Rebase this impl into the trait-owning crate** — *Correctness · Medium · High confidence*

This branch is currently unmergeable and 123 commits behind `main`. Since #721, `ThreadAware` is defined and re-exported from `thread_aware_core`, and this `impls.rs` was deleted. Retaining this impl in the `thread_aware` facade while resolving conflicts cannot work under Rust’s orphan rules because both the trait and `SocketAddr` are foreign there. The destination crate is also `#![no_std]`; its existing standard-library imports and impls are explicitly gated ([current `impls.rs` lines 15–22 and 67–72](https://github.com/microsoft/oxidizer/blob/7a00180f9911c3158f8837bbd3f157f87f7f17e5/crates/thread_aware_core/src/impls.rs#L15-L22)).

**Direction:** Rebase, move the import and no-op impl into `thread_aware_core`, and gate both with `cfg(any(test, feature = "std"))`.

**Done when:** the rebased PR is mergeable, the impl is owned by `thread_aware_core`, and the repository’s std and no-default-features compile gates cover it.

geeknoid (commented):
Static five-pass review found one integration blocker: the semantic choice of a no-op `ThreadAware` implementation for `SocketAddr` is sound, but this branch is now 123 commits behind `main` and GitHub reports it as `dirty`/unmergeable. Since #721, the trait and standard-library implementations moved to `thread_aware_core`, so the change must be rebased and relocated there with its `std` gate preserved.

The final two-line diff introduces no unsafe/FFI surface, shared state, allocation, I/O, parsing, authorization, or runtime behavior. `SocketAddr` is a portable value representation of an IP address and port, so an empty relocation callback does not weaken the trait’s contention-avoidance contract. The earlier runtime test was removed after prior review correctly identified it as tautological; I did not duplicate that point.

All 820 classes were assessed statically (150 security, 162 correctness, 182 testing, 183 performance, 143 conformance). Nothing was executed, cloned, edited in the source checkout, or posted.

PR #572, opened 2026-07-14
feat: introduce observed crate

Vaiz:


crates/observed/src/emit.rs
```diff
@@ -0,0 +1,24 @@
...
+/// 2. **Expression**: `emit!(sink, my_event_variable)`
+///
+/// The macro automatically captures [`SourceLocation`](crate::metadata::SourceLocation) at the call site and
+/// dispatches through the explicitly-passed sink plus any scoped emitters.
+#[macro_export]
+macro_rules! emit {
+    ($sink:expr, $event:expr $(,)?) => {{
+        ::core::convert::AsRef::<$crate::Sink>::as_ref(&$sink).emit::<_, _>(
+            || $event,
+            $crate::metadata::SourceLocation::new(::core::module_path!(), ::core::file!(), ::core::line!()),
```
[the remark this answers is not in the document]
Vaiz: fixed

geeknoid:
I just got this code and did a "cargo doc --all-features" and the resulting docs don't show most of your macros. These are the only macros avaiable:

image

In addition, the Copilot-generated summary above mentions crates which are not currently in the PR (observed_helpers) which presumably means you removed them from the PR at some point.

So is this ready to review?

geeknoid:
As we discussed during the architecture meeting a few weeks back, I find the syntax is overly verbose which I think will hamper its adoption as it makes your code difficult to read. Below are several suggestions to improve this.

* The docs for the crate aren't being generated correctly, the various macros aren't being surfaced.

* Could we remove the requirement to type "name =" when declaring an event?

    ```
    #[event("foo")]
    ```

* Could we make it so the name for the #[event] macro is not required, and just assume the struct name in that case?

* For #[log], could we remove the requirement to type "severity ="?

    ```
    #log(warn, name = "Foo")]
    ```

  or perhaps even shorter and consistent with my suggestion below on #[counter] and friends:

    ```
    #[warn(name = "foo")]
    ```

* In the architecture meeting, we discussed the "target" value for the #[log] attribute. It's not there anymore, so how does one categorize log entries?

* As we discussed in the arch meeting, #[metrics(kind = Counter)] is just overly verbose for no gain to our customers. I'd like to see dedicated attributes:

    ```
    #[counter(...)]
    #[gauge(...)]
    #[histogram(...)]
    ```

* Instead of a distinct #[disabled] attribute, could we just add a boolean to the #[event] attribute instead?

    ```
    #[event("foo", disabled)]
    ```

* What is the 'description' value for a metric used for? Shouldn't logs also have a description? And if both have description, maybe the description should be on #[event] instead?

@ralfbiedert, @martin-kolinek

Vaiz:
> I just got this code and did a "cargo doc --all-features" and the resulting docs don't show most of your macros. These are the only macros avaiable:

Those are only 3 macros that crate provides. I assume that other macros you refer to are part of Event, Enrichment macros.

> In addition, the Copilot-generated summary above mentions crates which are not currently in the PR (observed_helpers) which presumably means you removed them from the PR at some point.
>
> So is this ready to review?

yes, it's ready. Copilot didn't update its overview. I decided to push other crates in separate PR.


> Could we make it so the name for the #[event] macro is not required, and just assume the struct name in that case?

__ *We can do it later if we agree on that*

that question has been discussed and we explicitly decided that this should not be allowed to avoid various issues with unexpected event name that don't follow the schema.
Good news, adding this won't be a breaking change, so we can do it later.

> Could we remove the requirement to type "name =" when declaring an event?

> For #[log], could we remove the requirement to type "severity ="?

> As we discussed in the arch meeting, #[metrics(kind = Counter)] is just overly verbose for no gain to our customers. I'd like to see dedicated attributes

for first version it's better to keep more verbose syntax, we can always add this shortcut later without introducing breaking changes.

> In the architecture meeting, we discussed the "target" value for the #[log] attribute. It's not there anymore, so how does one categorize log entries?

this will be introduced later after we do a proper design for that part


> Instead of a distinct #[disabled] attribute, could we just add a boolean to the #[event] attribute instead?

that one I actually missed, will fix
**UPD: FIXED**

> What is the 'description' value for a metric used for? Shouldn't logs also have a description? And if both have description, maybe the description should be on #[event] instead?

metric instruments have a bigger set of properties that can be set, we will add support for more of them in the future. I don't think we should merge description for metrics and logs in one bucket, but if we introduce description for logs, we can have a shortcut that sets the same description for logs, though first we need to decide what to do with it.


```
name: Cow<'static, str>
  Name of the Histogram.
description: Option<Cow<'static, str>>
  Description of the Histogram.
unit: Option<Cow<'static, str>>
  Unit of the Histogram.
boundaries: Option<Vec<f64>>
  Bucket boundaries for the histogram.
```
https://docs.rs/opentelemetry/latest/opentelemetry/metrics/struct.HistogramBuilder.html

crates/observed/src/enrichment/slot.rs
```diff
@@ -0,0 +1,199 @@
...
+//!
+//! See the [Enrichment](crate#enrichment) section for a high-level overview.
+
+use std::any::type_name;
+use std::cell::RefCell;
+use std::sync::Arc;
+
+use smallvec::{SmallVec, smallvec};
+use thread_aware::ThreadAware;
+use thread_local::ThreadLocal;
```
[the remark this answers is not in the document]
Vaiz: as you mentioned, CI can check it so there is no need for you comment here

crates/observed/src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+//! # #[event(name = "my.event")]
+//! # #[log(severity = info, message = "body")]
+//! # struct MyEvent;
+//! # let sink = Sink::noop();
+//! let transfer = sink.transfer_context();
+//!
+//! let sink = sink.clone();
+//! let handle = std::thread::spawn(move || {
+//!     // Restore the captured enrichment on this thread for the guard's lifetime.
+//!     let _guard = transfer.apply();
```
martin-kolinek, line 116: The transfer semantics aren't completely clear - even after looking at the docs of transfer_context/apply. This could probably be rephrased to better describe the enrichment model and how the transferring works within that model.

Also, what happens in the following case?
```
let transfer = sink.transfer_context();

let sink = sink.clone();
let handle = tokio::spawn(async move || {
    // Restore the captured enrichment on this thread for the guard's lifetime.
    let _guard = transfer.apply();
    some_io_operation().await;
    emit!(sink, MyEvent);
});
handle.join().unwrap();
```
During the await of some_io_operation, will other tasks running on the same tokio thread see enrichment even though they are unrelated, or is there logic in place to prevent it?

crates/observed/src/context/mod.rs
```diff
@@ -0,0 +1,1288 @@
...
+use std::any::type_name;
+use std::pin::Pin;
+use std::task::Poll;
+
+use crate::enrichment::{Enrichment, EnrichmentTransfer};
+
+/// Captured sink context (enrichment) for cross-thread transfer.
+///
+/// Created by [`Sink::transfer_context`](crate::Sink::transfer_context). Apply it on the target
+/// thread by calling [`Transfer::apply`].
```
martin-kolinek, line 16: As mentioned in the other comment, it would be good to specify the exact semantics - at least mention that it affects the original sink, so that emissions using that sink work with the enrichment.

crates/observed/src/enrichment/enrichment_trait.rs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Enrichment derive trait for typed enrichment structs.
+
+use crate::enrichment::EnrichmentEntry;
+
+/// A typed enrichment struct that converts into enrichment entries.
```
martin-kolinek: Nit: the line mentions struct, but this is a trait.

crates/observed/src/enrichment/enrich_ext.rs
```diff
@@ -0,0 +1,111 @@
...
+        }
+    }
+
+    /// Wraps this future so that a captured
+    /// [`Transfer`] is restored on every poll.
+    ///
+    /// Chain with [`.enrich()`](EnrichFutureExt::enrich) to also push
+    /// per-scope entries on top of the transferred context.
+    ///
+    /// WARN: calling `enrich/enrich_for` after attach won't work because attach replaces
```
martin-kolinek, line 76: Is this unavoidable? This sounds like a pretty unintuitive behavior.

crates/observed/src/processing/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+
+/// Visitor callback for lazily iterating over event fields or enrichment entries.
+///
+/// Each invocation receives a [`FieldDescriptor`] and a [`FieldValueFn`] getter.
+/// The getter takes a [`data_privacy::RedactionEngine`] reference and returns the
+/// redacted [`Value`] - it is only called if the processor needs the value.
+///
+/// Return [`ControlFlow::Continue`] to keep iterating, or
+/// [`ControlFlow::Break`] to stop early.
+pub type FieldVisitorFn<'a> = dyn FnMut(&FieldDescriptor, &FieldValueFn<'_>) -> ControlFlow<()> + 'a;
```
martin-kolinek, line 38: Does this need to be dyn? Would we benefit from visit_field accepting a generic function rather than a dyn one?

crates/observed/src/metadata/event.rs
```diff
@@ -0,0 +1,125 @@
...
+
+/// Static description of a telemetry event type.
+///
+/// Available as a `const` on every type that implements [`crate::Event`],
+/// providing compile-time metadata about the event's shape.
+///
+/// The event name is shared across all signals; per-signal metadata lives
+/// in `log` / `metric`.
+#[derive(Debug, Clone, Copy)]
+pub struct EventDescription {
```
martin-kolinek, line 19: I'm not completely sure, but I feel like EventDescription giving access to field descriptions would be a good thing - right now it's impossible to figure out the fields from just the type, you need an instance (as the only way to get the field descriptions is through visit).

crates/observed/src/metadata/event.rs
```diff
@@ -0,0 +1,125 @@
...
+    ///
+    /// Convenience for building lookup tables keyed by event type.
+    #[must_use]
+    pub fn type_id_of<T: crate::Event>() -> Option<TypeId> {
+        T::DESCRIPTION.type_id
+    }
+
+    /// Returns the per-signal log description, if the event produces logs.
+    #[must_use]
+    pub const fn log(&self) -> Option<&LogDescription> {
```
martin-kolinek, line 74: Should we support multiple logs?

martin-kolinek, line 74: I'm a bit unsure about the naming - this is just the default log configuration, right? Since the logging instructions can be changed, I'd consider calling this default_log?

crates/observed/src/metadata/field.rs
```diff
@@ -0,0 +1,156 @@
...
+    /// Returns the log key for this field.
+    #[must_use]
+    pub const fn key(&self) -> &'static str {
+        self.key
+    }
+}
+
+/// Per-field metric routing entry.
+///
+/// If [`instrument`](Self::instrument) is `Some`, the field is the
```
martin-kolinek: Wrong link - should be Self::instrument_description instead of Self::instrument

crates/observed/src/processing/event_view.rs
```diff
@@ -0,0 +1,241 @@
...
+/// Processors receive this from [`EventProcessor::process`](super::EventProcessor::process)
+/// and pull only the fields/enrichments they need. Skipped fields never
+/// invoke their redaction closure.
+///
+/// The enrichment context is captured (as an `Arc` snapshot) at view
+/// construction. Concurrent enrichment pushes after that point are not
+/// visible to this view - same semantics as the previous live-walk because
+/// the chain is an immutable linked list (`Arc<EnrichmentNode>`).
+///
+/// This is a concrete struct (not a trait) - `EventProcessor::process` takes
```
martin-kolinek: Nit: useless sentence

crates/observed/src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+/// | `#[if_none(drop)]` / `#[if_none("...")]` | Control how a `None` `Option<T>` is recorded. The default is `#[if_none("n/a")]`. |
+///
+/// `#[unredacted]` and `#[data_class(...)]` are mutually exclusive.
+///
+/// ## Optional fields
+///
+/// A field of type `Option<T>` is captured like a `T` when it is `Some(_)`. When
+/// it is `None`, `#[if_none(...)]` decides the outcome (default
+/// `#[if_none("n/a")]`, or `drop` to omit it) - the same behavior as in
+/// [`Event`].
```
martin-kolinek: Nit: this links the trait instead of macro - not sure if it's possible to link the derive macro instead.

martin-kolinek (approved):
Note that most of my comments can be deferred to a separate ticket/pull request.

crates/observed/src/metadata/event.rs
Vaiz: I don't see a use case for that

Vaiz: the whole struct is immutable, the only way to change something is to construct a new instance. I'm not sure why it should be reflected in the struct's method names. By that logic everything should have `default` prefix

crates/observed/src/metadata/event.rs
Vaiz: I deffer it to later as it won't be a breaking change

crates/observed/src/enrichment/enrichment_trait.rs
Vaiz: fixed

crates/observed/src/metadata/field.rs
Vaiz: fixed

crates/observed/src/processing/event_view.rs
Vaiz: fixed

crates/observed/src/lib.rs
Vaiz: fixed

sgalkin (approved):

PR #578, opened 2026-07-15
ci: add cargo-machete for workspace-wide unused-dependency detection

sandersaares:
[Copilot speaking]

Strengthens unused-dependency detection by adding `cargo-machete` alongside `cargo-udeps`, and closes a scoping gap that let unused deps outside a PR's diff go undetected on `main`.

## Why

Two independent shortcomings:

- **Detection coverage.** `cargo-udeps`, by design, ignores a needless direct dependency when it is still pulled in transitively, so some genuinely-unused deps were never flagged. `cargo-machete` uses a different algorithm and catches these, covering udeps' blind spot.
- **Scoping.** The `Udeps` step lived in the `extended-analysis` job, which runs only on `pull_request` and was scoped to `exclude_not_modified` (crates whose own files changed). Any crate that was not modified was skipped, even when its dependency usage could have changed - so unused deps in such crates were never re-checked.

## Changes

- Add `cargo-machete` to the toolset (`constants.env`, `setup.just`, and the CI setup step). It runs workspace-wide (no package selection), covering udeps' transitive-use blind spot and acting as the backstop for crates outside a PR's diff.
- Add `just unused-dependencies-check` combining `cargo machete --with-metadata` + the existing `cargo udeps` pair.
- The CI step (renamed to "Unused dependencies") now runs workspace-wide `cargo machete --with-metadata` and scopes `udeps` by the affected/impacted set (`exclude_not_affected`) rather than only modified crates.
- Rename `scripts/crate-template/Cargo.toml` to `Cargo.toml.template` so its `{{...}}` placeholders (invalid TOML) no longer break machete's scan; `add-crate.ps1` restores the name when generating a crate.
- Remove the unused dependencies flagged by machete (cachet, data_privacy, data_privacy_core, fetch, http_extensions, ohno). Suppress the sole false positive - `http_extensions`' `uuid`, used only by a doctest - via `[package.metadata.cargo-machete]`.

## Validation

`just unused-dependencies-check` (machete + udeps) is green workspace-wide; the affected crates build with `--all-features --all-targets` and their tests and doctests pass.


Vaiz (approved):

psandana (approved):
LGTM. Nice, well-documented change. Verified the risky bits: `uuid` is genuinely doctest-only in `http_extensions/src/_documentation/recipes.rs` so the `[package.metadata.cargo-machete] ignored` suppression is correct; the `Cargo.toml.template` rename leaves no stale references (add-crate.ps1 restores it, and packaging-guidelines.md points at the directory); machete runs before udeps and gates hard on findings. Good call using machete as the workspace-wide backstop for udeps' transitive-use blind spot.

PR #585, opened 2026-07-17
feat: support type erasure of pooled handles to unsized types

geeknoid:
- Onboard bytesbuf and layered to plurality


crates/plurality/tests/unsize.rs
```diff
@@ -0,0 +1,395 @@
...
+
+#[test]
+fn erased_debug_forwards_to_value() {
+    let pool = Pool::<u32>::new();
+    let d: Box<dyn Debug> = Box::unsize::<dyn Debug>(pool.alloc_box(7u32), Coercion::to_debug());
+    assert_eq!(format!("{d:?}"), "7");
+}
+
+struct ImmediateReady(u32);
+impl Future for ImmediateReady {
```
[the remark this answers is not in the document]
geeknoid: No change is needed here: this crate uses Rust edition 2024, whose prelude includes `core::future::Future`. The integration test compiles on the workspace MSRV (Rust 1.93) and passes.

crates/plurality/src/rc.rs
```diff
@@ -22,17 +24,19 @@ use crate::slot::{SlotCell, check_refcount_overflow};
...
 
-// `Rc` is `!Send + !Sync` automatically: its only field is a
-// `NonNull<SlotCell<T>>`, and raw pointers are neither `Send` nor `Sync`. We add
-// no `unsafe impl Send`/`Sync`, so no explicit negative impl is needed; a
-// compile-time `assert_not_impl_any!` in `tests/smart_ptr.rs` locks the property
-// in against future refactors.
+// `Rc` is `!Send + !Sync` automatically: its `slot` field is a `NonNull<T>`, and
+// raw pointers are neither `Send` nor `Sync`. We add no `unsafe impl Send`/`Sync`,
+// so no explicit negative impl is needed; a compile-time `assert_not_impl_any!` in
+// `tests/smart_ptr.rs` locks the property in against future refactors.
```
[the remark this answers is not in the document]
geeknoid: Made the non-thread-safe guarantee explicit with a zero-sized `PhantomData<alloc::rc::Rc<()>>` marker, so `plurality::Rc` remains `!Send + !Sync` independently of `NonNull` auto-trait behavior. The existing compile-time `assert_not_impl_any!` continues to lock this invariant in. Stable Rust does not permit explicit negative impls for user-defined types.

crates/bytesbuf/src/mem/global.rs
```diff
@@ -317,34 +334,35 @@ fn in_place_initialize_block<const SIZE: usize>(block: &mut MaybeUninit<NeutralB
...
 {
-    // SAFETY: We are required to fully initialize the object. We "do" because the entire
-    // object `T` is wrapped in MaybeUninit, so we are not required to do anything at all.
-    // We do this purely to get the handle, because we need the handle to do the real
-    // initialization.
-    let handle = unsafe { pool.insert_with(|_| {}) };
+    // Reserve a slot without initializing it, then leak the box to a raw pointer:
+    // the slot stays occupied and no destructor runs until the block is released.
+    let uninit = pool.alloc_uninit_box();
+    let uninit_ptr = plurality::Box::into_raw(uninit);
```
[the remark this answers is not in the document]
geeknoid: Fixed. The uninitialized pooled box now remains owned while the initializer runs and is converted to raw only after successful initialization. If initialization unwinds, RAII returns the slot and decrements the pool refcount. Added a single-slot regression test proving the slot is reusable after panic, and ran that test under Miri.

crates/plurality/tests/unsize.rs
```diff
@@ -0,0 +1,395 @@
...
+
+#[test]
+fn unsize_to_dyn_future_and_poll_via_as_pin_mut() {
+    let pool = Pool::<ImmediateReady>::new();
+    let b = pool.alloc_box(ImmediateReady(42));
+    let mut fut: Box<dyn Future<Output = u32>> = Box::unsize::<dyn Future<Output = u32>>(b, Coercion::to_future());
+
+    let waker = Waker::noop();
+    let mut cx = Context::from_waker(waker);
+    match fut.as_pin_mut().poll(&mut cx) {
```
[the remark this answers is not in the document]
geeknoid: `Waker::noop()` returns `&'static Waker`, not an owned `Waker`, so the `waker` binding already has the exact reference type required by `Context::from_waker`. This test compiles and passes on the workspace MSRV.

crates/plurality/benches/gungraun_dyn_box/linux.rs
```diff
@@ -0,0 +1,77 @@
...
+    reason = "benchmark code"
+)]
+
+use std::hint::black_box;
+
+use gungraun::prelude::*;
+use infinity_pool::{BlindPool, LocalBlindPool, LocalPinnedPool, PinnedPool};
+use plurality::Pool;
+
+#[path = "../shared/dyn_box_ops.rs"]
```
sandersaares: `#[path]` is an abomination as it breaks standard Cargo project management patterns and causes IDE confusion as well as issues with linting the same code in different compiles. Please factor shared helpers out to a separate internal package instead of path-referencing them like this.

crates/bytesbuf/src/mem/global.rs
```diff
@@ -94,7 +94,33 @@ impl Memory for GlobalPool {
...
+struct BlockHandle<const SIZE: usize>(NonNull<NeutralBlock<SIZE>>);
+
+// SAFETY: the handle only ever reconstructs the owning box to free the slot,
+// which plurality supports from any thread (the free list is atomic MPSC), and
+// the block's lifecycle is governed by the atomic `ref_count`. It carries no
+// thread-affinity, so it is safe to move and share across threads — mirroring
+// the `Send + Sync` raw pool handle it replaces.
+unsafe impl<const SIZE: usize> Send for BlockHandle<SIZE> {}
+// SAFETY: see the `Send` justification above.
+unsafe impl<const SIZE: usize> Sync for BlockHandle<SIZE> {}
```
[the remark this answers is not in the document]
geeknoid: These impls are required: `core::ptr::NonNull<T>` has explicit negative `Send` and `Sync` impls, regardless of `T`, so `BlockHandle` does not derive either auto trait. This handle is intentionally thread-mobile because plurality supports cross-thread drops through its atomic free list and `NeutralBlock` lifecycle is protected by an atomic reference count; the safety comments document those invariants. Removing the impls fails the existing `GlobalPool: MemoryShared` compile-time assertion.

crates/plurality/docs/DESIGN.md
```diff
@@ -20,7 +20,7 @@ so the pointers stay valid until dropped.
 | `lib.rs` | Crate root, feature gates, re-exports, crate-level docs. |
```
sandersaares: I would expect the design document to actually say something about type erasure of pooled handles to unsized types. But it seems to mostly be about implementation details? This document that lists things like field names does not really sound like a design document. I would like to see a document that describes what features exist, what design tenets are used to implement those features, how features combine, etc.

crates/plurality/src/lib.rs
```diff
@@ -42,6 +42,30 @@
...
+//! token:
+//!
+//! ```
+//! use core::fmt::Debug;
+//!
+//! use plurality::{Box, Coercion, Pool};
+//!
+//! let pool = Pool::<u32>::new();
+//! let b = pool.alloc_box(7u32);
+//! let erased: Box<dyn Debug> = Box::unsize::<dyn Debug>(b, Coercion!(to dyn Debug));
```
sandersaares: The "to" here does not seem like Rust syntax, I would expect `Coercion!(dyn Debug)` or something like that. Can the other generics on the same line be inferred from that? The same generic parameter is stated three times here, which feels two times too many.

crates/plurality/src/lib.rs
```diff
@@ -168,3 +192,5 @@ pub use pool::Pool;
 pub use rc::Rc;
 #[doc(inline)]
 pub use sync::Arc;
+#[doc(inline)]
```
sandersaares: There are a lot of `#[doc(inline)]` here that do not do anything - this is dead code.

crates/plurality/src/lib.rs
```diff
@@ -42,6 +42,30 @@
 //! All four deref to `&T`; [`Box`] and [`Alloc`] also give `&mut T`. Dropping a
```
sandersaares, line 27: Unrelated to this PR but I would expect some description here of how/when the pool releases memory, so I know what to expect as a user.

crates/plurality/src/pool.rs
```diff
@@ -18,19 +19,58 @@ use crate::builder::PoolBuilder;
...
-    /// `1` for the live `Pool` handle plus `1` per live *refcounted* allocation
-    /// (`Box`/`Arc`/`Rc`). `Alloc` handles are lifetime-bound and do **not**
-    /// contribute — their borrow already keeps the inner alive.
+    /// `1` for the live `Pool` handle plus one per live refcounted allocation.
     pub(crate) pool_refcount: AtomicUsize,
+    /// Returns the core to its concrete `PoolInner<T, A>` type for teardown.
+    pub(crate) teardown: unsafe fn(NonNull<Self>),
+}
+
+struct ChunkAllocationGuard<'a, T, A: Allocator> {
```
sandersaares, line 50: This type could use some comments to explain what it is guarding and from whom

crates/layered/src/dynamic.rs
```diff
@@ -55,26 +55,43 @@ where
...
-        // future and inserts it into the pool
+        // Wrap each service execution in a future, pool it, and hand back a
+        // type-erased handle so `DynamicService` need not carry the future type.
         let exec = move |input: In| {
             let cloned = Arc::clone(&service);
             let fut = async move { cloned.execute(input).await };
-            pool.insert(fut).cast_send_future()
+            let boxed = pool
+                .lock()
+                .expect("dynamic-service pool mutex poisoned by a prior panic while allocating")
```
sandersaares: This is just an `unwrap()` is disguise - the message describes what went wrong. However, panicking under the mutex is presumably something we consider impossible, so we should state why we believe that to be the case (typically that we do not call any user code, which may panic, while holding the guard).

If we do call user code and it really may panic, the expect-message should clearly state that and reference what user code the guilty party might be, so maintainers can backtrack from "expect" to what poisoned it.

crates/plurality/src/lib.rs
geeknoid: Good catch — removed. Rustdoc already inlines re-exports from a crate's own private modules by default, so those `#[doc(inline)]` attributes were no-ops. Kept it only on `unsize::Coercion`, where it is meaningful because `unsize` is an external crate whose docs are not inlined by default, and added a comment explaining why.

crates/plurality/src/pool.rs
geeknoid: Added a doc comment on `ChunkAllocationGuard` explaining what it guards and against what: it owns a freshly allocated, fully-initialized-but-unpublished chunk during the window between allocation and its `directory.push`. That push can reallocate and panic (allocator failure); without the guard the half-installed chunk — not yet reachable from the directory or free list — would leak on unwind. On success `grow` transfers ownership to the pool and `mem::forget`s the guard. The note also documents the loom-only per-slot refcount teardown in its `Drop`.

crates/plurality/src/pool.rs
```diff
@@ -770,51 +820,148 @@ unsafe fn push_free<T, A: Allocator>(slot: NonNull<SlotCell<T>>) -> NonNull<Pool
...
 /// # Safety
-/// `slot` must be an occupied, initialized slot whose last handle is being
-/// released; its value must not be accessed afterwards.
+/// `value` must point at the (valid, occupied) value of a live slot.
 #[inline]
-pub(crate) unsafe fn drop_and_free<T, A: Allocator>(slot: NonNull<SlotCell<T>>) {
-    // SAFETY: last handle to an occupied slot (caller contract).
+pub(crate) unsafe fn refcount_ptr<T: ?Sized>(value: NonNull<T>) -> *mut AtomicU32 {
+    // SAFETY: `value` is field 0 of the slot; the refcount follows it. Reading
+    // the value's metadata (for `size_of_val`) never dereferences the pointee.
```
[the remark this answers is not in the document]
geeknoid: Fixed in 7cd398d2. Reworded the inline SAFETY comment to drop the incorrect "never dereferences the pointee" claim; it now states that `value.as_ref()` forms a `&T` the caller guarantees is a valid, live value, and that `size_of_val` only reads the pointer metadata (length/vtable), not the value's bytes.

crates/bytesbuf/src/mem/global.rs
```diff
@@ -389,20 +428,20 @@ unsafe impl<const SIZE: usize> BlockRefDynamic for BlockMeta<SIZE> {
...
-
-        // SAFETY: We must promise that it is no longer references and is still in the pool.
-        // Sure, we can promise that because tracking that is the entire purpose of this type.
-        unsafe {
-            pool.remove(handle);
-        }
+        // Reconstruct the owning box from the block's self-handle and drop it. This
+        // runs the block's destructor (dropping its `block_pool` handle) and returns
+        // the slot to the pool via plurality's lock-free reclaim path — no lock is
+        // required to free, only to allocate.
```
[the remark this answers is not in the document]
geeknoid: Fixed in 7cd398d2. Removed the stale "dropping its `block_pool` handle" phrasing (`BlockMeta` no longer has that field). The comment now states the real effect: dropping the box runs the block's destructor and returns the slot to the pool via plurality's lock-free reclaim path, releasing the block's plurality pool reference count.

crates/layered/src/dynamic.rs
geeknoid: Fixed in 7cd398d2. No user code runs under this guard (`alloc_box` only moves the already-constructed future into a pool slot; the service future is polled later, outside the lock), so the only possible poisoner is a prior `alloc_box` panicking on allocation failure (OOM/pool exhaustion). The expect message and a new comment now state this so maintainers can backtrack a poisoned guard to that cause.

crates/plurality/docs/DESIGN.md
geeknoid: Done.

crates/plurality/src/lib.rs
geeknoid: Done.

crates/plurality/src/pool.rs
```diff
@@ -5,8 +5,9 @@ use alloc::boxed::Box as AllocBox;
 use alloc::vec::Vec;
 use core::alloc::Layout;
 use core::cell::UnsafeCell;
-use core::mem::MaybeUninit;
-use core::ptr::NonNull;
+use core::marker::PhantomData;
+use core::mem::{MaybeUninit, needs_drop};
```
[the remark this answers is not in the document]
geeknoid: This is a false positive. `plurality` is edition 2024, and `size_of`/`align_of`/`size_of_val`/`align_of_val` were added to the standard library prelude in the Rust 2024 edition (via the `std::mem::{size_of, align_of, size_of_val, align_of_val}` prelude additions). They therefore need no `use` import here. The crate compiles cleanly (`cargo build -p plurality --all-features` succeeds), so no change is required.

crates/plurality/src/chunk.rs
```diff
@@ -4,16 +4,16 @@
 use core::alloc::Layout;
 use core::ptr::NonNull;
 
-use crate::pool::PoolInner;
+use crate::pool::PoolCore;
 use crate::slot::SlotCell;
```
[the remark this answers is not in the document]
geeknoid: This is a false positive. `plurality` is edition 2024, and `size_of`/`align_of`/`size_of_val`/`align_of_val` were added to the standard library prelude in the Rust 2024 edition (via the `std::mem::{size_of, align_of, size_of_val, align_of_val}` prelude additions). They therefore need no `use` import here. The crate compiles cleanly (`cargo build -p plurality --all-features` succeeds), so no change is required.

crates/plurality/src/lib.rs
geeknoid: Fixed the overly-specific generics.

The macro syntax comes from the unsized crate, so it is what it is. It should also technically be a lowercase "c"...

crates/plurality/benches/gungraun_dyn_box/linux.rs
geeknoid: Fixed in ee8bfb27 — all `#[path]` includes are gone.

Rather than introduce a new internal crate, I eliminated the shared code by duplicating it, which is acceptable here because it's benchmark-only code. Concretely:

- Merged the two Criterion benches into a single `criterion` target (`benches/criterion/`) and the two Gungraun benches into a single `gungraun` target (`benches/gungraun/`). Bench count drops from 6 to 4.
- Merged the two shared helper files (`shared/ops.rs` + `shared/dyn_box_ops.rs`) into one helper module, of which each harness now keeps its own plain-`mod` copy (`benches/criterion/ops.rs`, `benches/gungraun/ops.rs`) — no `#[path]`, resolved through normal Cargo module paths.
- `tests/alloc_tracking.rs` also `#[path]`-included the shared helper; it now carries a small self-contained inline copy instead.
- `scripts/perf_report.rs` and `docs/PERF.md` updated for the new single `criterion`/`gungraun` bench names. Both harnesses verified to run (Callgrind headers `gungraun::<group>::<fn>` and criterion `group/name` still parse correctly).

geeknoid: Fixed.

crates/plurality/docs/DESIGN.md
```diff
@@ -1,324 +1,320 @@
...
+lock-free stack, so only plain atomics are required. Chunk acquisition goes
+through the standard allocator abstraction, so custom and instrumented allocators
+compose naturally.
+
+## Design invariants at a glance
+
+The safety and correctness of the whole system rest on a short list of
+invariants:
+
+1. **Single allocator thread.** Only one thread ever grows the pool or pops
```
sandersaares: Single at a time, right? But we can still move it to another thread, right? Might be useful clarification as it can be read two ways right now

crates/plurality/tests/alloc.rs
```diff
@@ -65,3 +65,24 @@ fn alloc_uninit_placement() {
...
     assert_eq!(*a, 7);
 }
+
+struct PanicOnce(StdArc<AtomicBool>);
+
+impl Drop for PanicOnce {
+    fn drop(&mut self) {
+        assert!(self.0.swap(true, Ordering::SeqCst), "first destructor panics");
+    }
+}
```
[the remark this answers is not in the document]
geeknoid: Fixed in 4ee3fd94. Rewrote `PanicOnce::drop` to make the intent explicit: a named `dropped_before` binding plus a comment spelling out that `swap` returns the previous flag (`false` on the first drop → assert fails and unwinds; `true` afterwards → passes). Kept `assert!` as the panic mechanism because `clippy::panic` and `clippy::manual_assert` (both `-D warnings` in this crate's tests) forbid a bare `if … { panic!() }` in test code.

crates/plurality/tests/pool.rs
```diff
@@ -237,6 +237,60 @@ fn closure_panic_returns_slot_to_pool() {
     });
 }
 
+struct PanicOnce(StdArc<AtomicBool>);
+
+impl Drop for PanicOnce {
+    fn drop(&mut self) {
+        assert!(self.0.swap(true, Ordering::SeqCst), "first destructor panics");
+    }
+}
```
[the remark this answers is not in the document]
geeknoid: Fixed in 4ee3fd94. Same clarification as the `tests/alloc.rs` copy: named `dropped_before` binding + a comment explaining the `swap` semantics, keeping `assert!` to satisfy `clippy::panic`/`clippy::manual_assert`.

crates/layered/src/dynamic.rs
```diff
@@ -55,26 +55,49 @@ where
...
 pub struct DynamicService<In, Out> {
-    exec: Arc<dyn Fn(In) -> BlindPooledMut<dyn SendFuture<Out>> + Send + Sync>,
+    exec: Arc<dyn Fn(In) -> plurality::Box<dyn SendFuture<Out>> + Send + Sync>,
 }
 
 pub(crate) trait SendFuture<Out>: Future<Output = Out> + Send {}
 impl<Out: Send + 'static, F: Future<Output = Out> + Send> SendFuture<Out> for F {}
-define_pooled_dyn_cast!(SendFuture<Out>);
+
+fn send_future_coercion<Out: 'static, F: SendFuture<Out> + 'static>() -> Coercion<F, dyn SendFuture<Out>> {
```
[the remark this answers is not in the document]
geeknoid: This compiles as written — it's a false positive. `unsize::Coercion` declares its third parameter with a default: `pub struct Coercion<T, U: ?Sized, F: FnOnce(*const T) -> *const U = fn(*const T) -> *const U>`. So `Coercion<F, dyn SendFuture<Out>>` resolves the third arg to the `fn(...)` default, and the function-item `coerce::<Out, F>` coerces to that `fn` pointer at the annotated return site. `cargo build -p layered` (and CI) pass.

crates/layered/src/dynamic.rs
```diff
@@ -55,26 +55,49 @@ where
 /// }
 /// ```
 pub struct DynamicService<In, Out> {
-    exec: Arc<dyn Fn(In) -> BlindPooledMut<dyn SendFuture<Out>> + Send + Sync>,
+    exec: Arc<dyn Fn(In) -> plurality::Box<dyn SendFuture<Out>> + Send + Sync>,
 }
```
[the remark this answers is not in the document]
geeknoid: Verified this is not needed. `plurality::Box` does not appear in layered's public API: the `SendFuture` trait is `pub(crate)`, and `DynamicService.exec` / `ServiceFuture` are private, so `plurality::Box<dyn SendFuture<Out>>` is never reachable through a public signature. `cargo check-external-types` (nightly-2026-03-20, allowlist = tower types only) passes with no violations, so no `plurality::*` entry is required.

crates/plurality/src/lib.rs
```diff
@@ -4,20 +4,9 @@
...
-#![allow(
-    clippy::cast_possible_truncation,
-    clippy::cast_ptr_alignment,
-    clippy::cast_precision_loss,
-    reason = "slot-index and chunk-layout arithmetic casts are bounded and correctly aligned by the pool's build-time capacity and `chunk_layout` invariants"
-)]
-#![allow(
-    clippy::elidable_lifetime_names,
-    reason = "the handle macros emit an explicit `'pool` lifetime, which reads more clearly than `'_` and is shared across all four handle types"
+    reason = "the pool's pointer-recovery and slot-lifecycle paths group tightly-coupled unsafe operations under a single documented safety invariant; splitting them into one block per operation would duplicate that invariant and obscure it"
```
sandersaares: I can theoretically envision this being a legit reason but is it really true that it cannot be scoped more narrowly to specific files/functions/modules?

crates/bytesbuf/src/mem/global.rs
```diff
@@ -94,7 +94,33 @@ impl Memory for GlobalPool {
...
+unsafe impl<const SIZE: usize> Sync for BlockHandle<SIZE> {}
+
+/// Creates a sub-pool for `SIZE`-byte blocks with a chunk size that keeps each
+/// chunk near a uniform footprint regardless of block size, so growth
+/// granularity is consistent across the four sub-pools.
+fn new_block_pool<const SIZE: usize>() -> Pool<NeutralBlock<SIZE>> {
+    const TARGET_CHUNK_BYTES: usize = 64 * 1024;
+    #[expect(clippy::cast_possible_truncation, reason = "slot count is bounded by TARGET / SIZE, always small")]
+    let slots_per_chunk = (TARGET_CHUNK_BYTES / SIZE).max(1) as u32;
+    Pool::builder().chunk_size(slots_per_chunk).build()
```
sandersaares, line 122: I recall you saying there is some threshold of 64k after which allocation becomes special but I lack an understanding of the specifics to be aware of there. For `SIZE = 64k` (== an object slightly more than 64K due to metadata), are we good here?

crates/bytesbuf/src/mem/global.rs
```diff
@@ -389,20 +428,21 @@ unsafe impl<const SIZE: usize> BlockRefDynamic for BlockMeta<SIZE> {
...
+        // Dropping the box runs the block's destructor and returns the slot to the
+        // pool via plurality's lock-free reclaim path, releasing the block's
+        // plurality pool reference count — no lock is required to free, only to
+        // allocate.
+        //
+        // SAFETY: this is the block's own handle, produced by `into_raw` at
+        // allocation; the last-reference check above guarantees it is reconstructed
+        // exactly once and that no references to the block remain.
+        let block: plurality::Box<NeutralBlock<SIZE>> = unsafe { plurality::Box::from_raw(handle.0) };
+        drop(block);
```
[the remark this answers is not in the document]
geeknoid: Good call — updated the drop path to reconstruct as `plurality::Box<MaybeUninit<NeutralBlock<SIZE>>>` (casting the self-handle back to the `MaybeUninit` pointer produced by `into_raw`) and then `assume_init()` before dropping, so the type transition is explicit and matches `from_raw`'s contract.

crates/layered/src/dynamic.rs
```diff
@@ -55,26 +55,49 @@ where
...
+            let boxed = pool
+                .lock()
+                // No user code runs under this guard: `alloc_box` only moves the
+                // already-constructed `fut` into a pool slot (the service future
+                // is polled later, outside the lock). The only way to poison this
+                // mutex is therefore a prior `alloc_box` panicking on allocation
+                // failure (OOM / pool exhaustion) — a fatal, non-recoverable
+                // condition — so maintainers can backtrack a poisoned guard to it.
+                .expect("dynamic-service pool mutex poisoned by a prior alloc_box allocation failure (OOM/exhaustion); no user code runs under this guard")
+                .alloc_box(fut);
```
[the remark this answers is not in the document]
geeknoid: Fixed — shortened the message to "dynamic-service pool mutex poisoned by a prior panic under this guard" and kept a brief comment noting no user code runs under the guard, without over-claiming OOM as the sole cause.

crates/plurality/docs/DESIGN.md
geeknoid: Right — it's "one at a time," not thread-affinity. The pool is `Send` but not `Sync`, so it can be moved to and resumed on another thread; what's forbidden is *concurrent* allocation. I clarified the invariant to say so explicitly:

> **Single allocator thread.** At most one thread at a time grows the pool or pops slots... This is a "no concurrent allocation" rule, not a thread-affinity rule: the pool may be moved to and resumed on a different thread, so long as allocations never overlap in time.

crates/plurality/src/lib.rs
geeknoid: Good push — it can be narrower. I removed the crate-level blanket suppression and moved a module-scoped `#![expect(...)]` into just the seven low-level unsafe implementation modules that actually group these ops (alloced, boxed, chunk, pool, rc, slot, sync). The lint now stays enforced everywhere else (lib.rs, builder, error, etc.). Per-block annotation would mean ~30 near-identical `#[expect]`s duplicating the same invariant, so module scope is the sweet spot.

crates/plurality/benches/gungraun/ops.rs
```diff
@@ -0,0 +1,324 @@
...
+    clippy::cast_precision_loss,
+    reason = "test and benchmark code"
+)]
+
+//! Single-operation benchmark bodies for the alloc and fat-pointer suites.
+//!
+//! Each `fn` performs **exactly one** hot operation. The gungraun suite runs
+//! each one once — Callgrind counts instructions exactly — while the criterion
+//! suite loops each one `N` times for a wall-clock signal. Both suites call
+//! these same functions, so the two measurements exercise identical code.
```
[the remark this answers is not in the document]
geeknoid: Fixed the wording. It now says this file holds the gungraun (Callgrind) bodies and notes the criterion suite has its own copy at `benches/criterion/ops.rs` — reflecting the copy relationship rather than implying a shared module.

crates/plurality/benches/criterion/ops.rs
```diff
@@ -0,0 +1,310 @@
...
+    clippy::cast_precision_loss,
+    reason = "test and benchmark code"
+)]
+
+//! Single-operation benchmark bodies for the alloc and fat-pointer suites.
+//!
+//! Each `fn` performs **exactly one** hot operation. The gungraun suite runs
+//! each one once — Callgrind counts instructions exactly — while the criterion
+//! suite loops each one `N` times for a wall-clock signal. Both suites call
+//! these same functions, so the two measurements exercise identical code.
```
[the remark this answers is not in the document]
geeknoid: Fixed. The comment now describes this file as the criterion (wall-clock) copy and points at the gungraun copy (`benches/gungraun/ops.rs`), so it no longer implies a single shared module.

crates/bytesbuf/src/mem/global.rs
geeknoid: The funny 64K thing is for multitude, plurality doesn't have this kind of thing.

crates/layered/src/dynamic.rs
```diff
@@ -55,26 +55,47 @@ where
 /// }
 /// ```
 pub struct DynamicService<In, Out> {
```
martintmk, line 57: FYI: I ran the benchmarks against main and plurality performs better compared to infinity_pool:

```
     Running benches\dynamic.rs (D:\dev\oxidizer\target\release\deps\dynamic-40ab873d66694f6e.exe)
typed-vs-dynamic/typed  time:   [19.868 ns 20.073 ns 20.255 ns]
                        change: [+0.4885% +2.4961% +4.4913%] (p = 0.01 < 0.05)
                        Change within noise threshold.
Found 16 outliers among 100 measurements (16.00%)
  4 (4.00%) low severe
  11 (11.00%) low mild
  1 (1.00%) high mild
typed-vs-dynamic/dynamic
                        time:   [54.678 ns 55.471 ns 56.265 ns]
                        change: [−18.751% −16.913% −15.089%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 18 outliers among 100 measurements (18.00%)
  14 (14.00%) low severe
  4 (4.00%) low mild
typed-vs-dynamic/wrapped_typed
                        time:   [20.580 ns 20.786 ns 20.966 ns]
                        change: [−2.7566% −0.8031% +1.2580%] (p = 0.44 > 0.05)
                        No change in performance detected.
Found 8 outliers among 100 measurements (8.00%)
  3 (3.00%) low severe
  5 (5.00%) low mild
typed-vs-dynamic/wrapped_dynamic
                        time:   [59.805 ns 60.392 ns 60.909 ns]
                        change: [−18.922% −17.205% −15.384%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 15 outliers among 100 measurements (15.00%)
  7 (7.00%) low severe
  8 (8.00%) low mild
Allocation statistics:

| Operation       | Mean bytes | Mean count |
|-----------------|------------|------------|
| dynamic         |          0 |          0 |
| typed           |          0 |          0 |
| wrapped_dynamic |          0 |          0 |
| wrapped_typed   |          0 |          0 |
```

martintmk (approved):

crates/bytesbuf/src/mem/global.rs
```diff
@@ -317,36 +334,58 @@ fn in_place_initialize_block<const SIZE: usize>(block: &mut MaybeUninit<NeutralB
...
 ///
-/// After this method returns, the object in the pool is guaranteed to be initialized.
-/// Correspondingly, the handle it is provided is stripped of the `MaybeUninit` wrapper.
+/// After this function returns, the object in the slot is guaranteed to be
+/// initialized, and the returned handle points at the initialized value. The
+/// slot stays occupied — ownership is held by the raw handle until the block's
+/// last reference reconstructs the box (via [`plurality::Box::from_raw`]) and
+/// drops it.
 ///
 /// # Safety
```
sandersaares: This safety chapter appears to be on a type but the content appears to be talking about the safety contract of a function. Is it misplaced or obsolete or what happened?

The text above also seems to be talking about a function ("After this function returns") but this is not a function. Something is misaligned.

Is this comment block meant to be on insert_with_handle_to_self?

crates/bytesbuf/src/mem/global.rs
```diff
@@ -389,20 +428,28 @@ unsafe impl<const SIZE: usize> BlockRefDynamic for BlockMeta<SIZE> {
...
-        let mut pool = pool.lock().expect(ERR_POISONED_LOCK);
-
-        // SAFETY: We must promise that it is no longer references and is still in the pool.
-        // Sure, we can promise that because tracking that is the entire purpose of this type.
-        unsafe {
-            pool.remove(handle);
-        }
+        // Reconstruct the owning box from the block's self-handle and drop it.
+        // Dropping the box runs the block's destructor and returns the slot to the
+        // pool via plurality's lock-free reclaim path, releasing the block's
```
sandersaares: Any locks are presumably internal to plurality, so I am unsure what the talk about lock-freeness is meant to say to the user. It seems unnecessary detail here, unless I miss some obvious point of it.

sandersaares (approved):

crates/layered/src/dynamic.rs
geeknoid: Thanks for running these — great to see the ~17-19% improvement on the dynamic paths and zero allocations across the board. Appreciate the confirmation.

crates/bytesbuf/src/mem/global.rs
geeknoid: Good catch — the block was misplaced. That function-contract doc + `# Safety` chapter belongs on `insert_with_handle_to_self`, so I moved it there. `UninitBoxGuard` now has its own short doc describing it as the RAII guard that frees the slot on the unwinding path if initialization doesn't complete.

crates/bytesbuf/src/mem/global.rs
geeknoid: Agreed — that's a plurality-internal detail with no bearing on this call site. Trimmed it to just: dropping the box runs the block's destructor and returns the slot to the pool.

sandersaares (approved):

ralfbiedert:
Hang on, this is going a bit too fast for me. In particular, this is re-exporting without a feature gate items from a crate that hasn't been updated in 5 years for unsized coercion, which has a trouble soundness history in core Rust. 

I don't know whether this here has the same or not, but I'm not comfortable pushing this in without a thorough look and comparison what the issue in `std` is, and why this here differs.

crates/plurality/Cargo.toml
```diff
@@ -18,7 +18,7 @@ include.workspace = true
 repository = "https://github.com/microsoft/oxidizer/tree/main/crates/plurality"
 
 [package.metadata.cargo_check_external_types]
-allowed_external_types = ["allocator_api2::*"]
+allowed_external_types = ["allocator_api2::*", "unsize::*"]
```
ralfbiedert: Nit, I'm not a fan of these globs, it makes it hard to know which parts are actually used or not. I'd hope that we never reexport so many items that they couldn't be named.

ralfbiedert:
See for example this here [The Pin unsoundness and `CoerceUnsized`](https://internals.rust-lang.org/t/the-pin-unsoundness-and-coerceunsized/11593)

ralfbiedert (changes requested):
Resolve possible soundness questions

crates/plurality/Cargo.toml
geeknoid: Done — replaced the globs with the two concrete types the public API actually exposes: `allocator_api2::alloc::Allocator` and `allocator_api2::alloc::global::Global`.

geeknoid:
@ralfbiedert Good call — rather than feature-gate it, I removed the `unsize` dependency entirely and vendored the small piece we need into `plurality::coerce` (one module). It no longer appears anywhere in the public API, and `Cargo.lock`/workspace deps drop it too.

The vendored version is also stronger than the original with respect to the soundness history you linked:

- **No fat-pointer layout assumptions / no byte-write.** plurality only ever coerces its own `slot: NonNull<T>`. `unsize()` now just applies the compiler-generated unsizing closure directly — `NonNull::new_unchecked((coercion.coerce)(ptr.as_ptr()).cast_mut())`. That closure has body `{ ptr }`, i.e. a genuine `T -> U` unsizing coercion checked by the compiler. So the fat-pointer metadata is compiler-produced (never forged), and we make **zero** assumptions about fat-pointer internals — we dropped the `unsize` crate’s `set_ptr_value` byte-write entirely.

- **The `Pin` + `CoerceUnsized` unsoundness class does not apply here.** That issue requires coercing a `Pin<P>` while a malicious `Deref`/`DerefMut` observes the projection. plurality never coerces `Pin<…>` — it coerces a raw `NonNull<T>` pointing at a fixed-address pooled value, and there is no user-overridable `Deref` on the coercion path. The metadata comes from a real coercion, not from reconstructing a `Pin`.

- `Coercion` is still re-exported (`plurality::Coercion` struct + `Coercion!` macro), but it is now backed by first-party code, so there is no unmaintained-crate exposure to reason about.

Happy to walk through `crates/plurality/src/coerce.rs` if useful — it is intentionally tiny.

crates/plurality/src/pool.rs
```diff
@@ -770,51 +871,159 @@ unsafe fn push_free<T, A: Allocator>(slot: NonNull<SlotCell<T>>) -> NonNull<Pool
...
+    // guarantees is a valid, live value for the duration of this call;
+    // `size_of_val` then reads only the pointer metadata (length or vtable),
+    // not the value's bytes.
     unsafe {
-        SlotCell::drop_value(slot);
-        free_slot::<T, A>(slot);
+        let refcount_off = round_up(size_of_val(value.as_ref()), align_of::<AtomicU32>());
+        value.as_ptr().cast::<u8>().add(refcount_off).cast::<AtomicU32>()
     }
 }
```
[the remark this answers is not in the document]
geeknoid: Fixed in 88536d31. Reworded `drop_and_free_local` directly instead of referencing the removed helper.

crates/plurality/src/pool.rs
```diff
@@ -770,51 +871,159 @@ unsafe fn push_free<T, A: Allocator>(slot: NonNull<SlotCell<T>>) -> NonNull<Pool
...
+    }
+}
+
+/// Reclaims a slot from a pointer to its **value** (field 0 of `SlotCell<T>`),
+/// for a possibly-unsized `T`, by reconstructing the slot and chunk layout from
+/// the value's runtime size and alignment.
+///
+/// This is the erased counterpart of [`drop_and_free`]: it never names
+/// `SlotCell<T>` (which is illegal for unsized `T`), so it can drive the drop
+/// path for an unsized handle. For a `Sized` `T`, `size_of_val`/`align_of_val`
```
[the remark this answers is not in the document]
geeknoid: Fixed in 88536d31. Reworded the comment around `drop_and_free_val` so it describes the erased reclamation path directly and no longer links to the removed `drop_and_free` helper.

crates/plurality/README.md
```diff
@@ -126,33 +166,43 @@ assert!(pool.try_alloc_box(2).is_err());
...
+ [__link17]: https://crates.io/crates/allocator-api2
+ [__link18]: https://doc.rust-lang.org/stable/alloc
+ [__link19]: https://docs.rs/plurality/0.1.1/plurality/?search=Box
+ [__link2]: https://docs.rs/plurality/0.1.1/plurality/?search=Alloc
+ [__link20]: https://docs.rs/plurality/0.1.1/plurality/?search=Arc
+ [__link21]: https://docs.rs/plurality/0.1.1/plurality/?search=Rc
+ [__link22]: https://docs.rs/plurality/0.1.1/plurality/?search=Box::unsize
+ [__link23]: https://docs.rs/plurality/0.1.1/plurality/?search=Arc::unsize
+ [__link24]: https://docs.rs/plurality/0.1.1/plurality/?search=Rc::unsize
+ [__link25]: struct@Coercion
```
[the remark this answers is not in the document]
geeknoid: Fixed in 6ee8619f. The crate-level source now uses an absolute docs.rs URL, and regenerating the README produces `https://docs.rs/plurality/latest/plurality/struct.Coercion.html` instead of the rustdoc-only `struct@Coercion` target.

ralfbiedert:
>@ralfbiedert Good call — rather than feature-gate it, I removed the unsize dependency entirely and vendored the small piece we need into plurality::coerce (one module). It no longer appears anywhere in the public API, and Cargo.lock/workspace deps drop it too.

Could you-as-a-human please also look at and double check this? I can produce UB:

```rust
use core::marker::PhantomPinned;
use core::pin::Pin;
use core::sync::atomic::{AtomicPtr, Ordering};
use core::{mem, ptr};

use plurality::{Arc, Pool, coerce};

trait AddressSensitive {
    fn initialize(self: Pin<&Self>);
}

struct Value {
    self_ptr: AtomicPtr<Value>,
    byte: u8,
    _pin: PhantomPinned,
}

impl Value {
    fn new() -> Self {
        Self {
            self_ptr: AtomicPtr::new(ptr::null_mut()),
            byte: 42,
            _pin: PhantomPinned,
        }
    }

    /// This is "safe" because a value initialized through `Pin<&Self>` normally
    /// remains at that address until it is dropped.
    fn read_pinned_byte(&self) -> u8 {
        let self_ptr = self.self_ptr.load(Ordering::Relaxed);
        assert!(!self_ptr.is_null());
        // SAFETY: `initialize` records this value's pinned address.
        unsafe { ptr::read_volatile(ptr::addr_of!((*self_ptr).byte)) }
    }
}

impl AddressSensitive for Value {
    fn initialize(self: Pin<&Self>) {
        self.self_ptr.store(ptr::from_ref(self.get_ref()).cast_mut(), Ordering::Relaxed);
    }
}

fn main() {
    let pool = Pool::<Value>::new();
    let value = pool.alloc_arc(Value::new());
    let mut unpinned_alias = value.clone();

    let erased: Arc<dyn AddressSensitive> = Arc::unsize(value, coerce!(dyn AddressSensitive));
    let pinned = Arc::into_pin(erased);
    pinned.as_ref().initialize();
    drop(pinned);

    let moved = mem::replace(Arc::get_mut(&mut unpinned_alias).unwrap(), Value::new());
    drop(unpinned_alias);
    drop(pool);

    let _ = moved.read_pinned_byte();
}
```

This fails on miri with 

```
error: Undefined Behavior: in-bounds pointer arithmetic failed: alloc393 has been freed, so this pointer is dangling
    --> crates\plurality\examples\pin_alias_move.rs:40:37
     |
  40 |         unsafe { ptr::read_volatile(ptr::addr_of!((*self_ptr).byte)) }
     |                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Undefined Behavior occurred here
```

However, this is not directly related to `Coerce`, but a deeper flaw in the `Arc::into_pin` machinery introduced earlier, the pinning / coercion combo just brought it up the radar as a well-known minefield. 

The Pin contract says

> From the moment a value is pinned by constructing a [Pin](https://doc.rust-lang.org/stable/std/pin/struct.Pin.html)ning pointer to it, that value must remain, valid, at that same address in memory, until its [drop](https://doc.rust-lang.org/stable/std/ops/trait.Drop.html#method.drop) handler is called.

I wasn't able to make this fully safe, but I think the `unsafe` here is valid. `Value::read_pinned_byte()` only ever tries to access its self-referential pointer after it has been set, and it can only be set once pinned via a `Pin<&Self>`. Once that pin happens, `Value` must stay where it was no-matter-what. In a real world this can probably be triggered by some `Future` combinator / wrapper. 

I think it would be prudent to remove at least the pinning API surface for now unless it has seen more human expert eyes.
I also think we are moving too fast where `unsafe` is  involved in too short of a time to absorb them. This is the second time in a few weeks where we had LLMs falsely insist `unsafe` constructs are sound (including miri tests they created) in foundational abstractions. It would be unfortunate if in a few months we'd have to recall layers of APIs built on top of these if it turns out that some composition primitives didn't actually compose.

crates/layered/src/dynamic.rs
```diff
@@ -55,26 +55,47 @@ where
...
+
+fn send_future_coercion<Out: 'static, F: SendFuture<Out> + 'static>() -> Coercion<F, dyn SendFuture<Out>> {
+    fn coerce<Out, F: SendFuture<Out> + 'static>(future: *const F) -> *const dyn SendFuture<Out> {
+        future
+    }
+
+    // SAFETY: this function performs only the compiler-checked pointer
+    // unsizing from `F` to the trait object, preserving its address.
+    unsafe { Coercion::new(coerce::<Out, F>) }
+}
```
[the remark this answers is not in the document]
geeknoid: Fixed in 078d15d3. Layered now uses the safe `coerce!` path. Because this trait object references the enclosing generic `Out`, I added explicit generic binding syntax (`coerce!(<Out> dyn SendFuture<Out>)`) and an integration test covering generic trait-object coercion.

ralfbiedert (approved):

PR #587, opened 2026-07-17
feat(multitude): add arena-aware Serde and Cow

geeknoid:
## Summary

- Add a generic arena-aware `Cow` with borrowed and arena-owned forms, fallible and infallible copy-on-write mutation, and explicit allocator-aware cloning.
- Add allocator-aware deserialization under `multitude::de`, including derive support, recursive containers, arena-backed dynamic values, resource limits, JSON helpers, and ordinary-Serde interoperability.
- Add local and escape-capable root ownership plus inherent reusable deserialization for Multitude strings and vectors.
- Harden chunk ownership and cross-thread teardown, tighten `Send`/`Sync` bounds, and fix ZST drain and `Rc` overflow handling.
- Add matched Criterion and Callgrind benchmarks for Serde and allocation teardown, plus updated examples, architecture documentation, and generated performance reporting.

## Validation

- Full unit, integration, property, doctest, Miri, and Loom coverage.
- Regression coverage for panic-safe cleanup, deserialization-limit enforcement, numeric replay, and derive hygiene.
- 100% LLVM line coverage for the new deserialization runtime, generic `Cow`, and derive implementation.


crates/multitude/src/internal/chunk.rs
```diff
@@ -150,13 +150,17 @@ impl<A: Allocator + Clone> Chunk<A> {
     // Mutation testing is suppressed: `> → >=` only differs at the
     // unreachable exact-`isize::MAX` boundary.
     #[cfg_attr(test, mutants::skip)]
+    #[allow(
+        clippy::cast_ptr_alignment,
+        reason = "alloc_chunk_raw returns storage aligned to struct_align(), which is at least Chunk's value alignment"
+    )]
     pub(in crate::internal) fn allocate(
-        allocator: A,
+        allocator: Arc<A>,
```
Vaiz, line 154: might a good place to start using more efficient [`triomphe`](https://crates.io/crates/triomphe) crate

Vaiz (approved):

crates/multitude/src/internal/chunk.rs
```diff
@@ -422,19 +426,19 @@ mod tests {
...
-    /// is appended (at offset 32) for 36 bytes total.
+    /// is appended (at offset 40) for 44 bytes total.
     #[test]
     fn header_size_for_global_matches_layout() {
         #[cfg(not(feature = "stats"))]
-        assert_eq!(Chunk::<Global>::header_size(), 32);
+        assert_eq!(Chunk::<Global>::header_size(), 40);
         #[cfg(feature = "stats")]
-        assert_eq!(Chunk::<Global>::header_size(), 36);
+        assert_eq!(Chunk::<Global>::header_size(), 44);
```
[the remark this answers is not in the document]
geeknoid: Good catch. Reworked the test to derive the expected header size from `mem::size_of` of the actual field types (`Arc` handle + `Weak` + `capacity` + `ref_count` + `next`, plus `AtomicU32` under `stats`) instead of hard-coding 8-byte widths, so it stays valid on non-64-bit targets. Fixed in 0bba6f1e.

crates/multitude/src/internal/chunk.rs
geeknoid: Thanks for the pointer. The `Arc<A>` here only wraps the backing allocator and is cloned once per chunk allocation (not on any hot allocation path), so the atomic-refcount overhead of `std::sync::Arc` is negligible in this spot. I would rather not pull in an extra external dependency for it right now, but `triomphe` is a reasonable option to revisit if allocator-handle refcounting ever shows up in a profile. Leaving as-is for this PR.

crates/multitude/tests/send_sync.rs
```diff
@@ -57,6 +77,11 @@ fn arena_is_send() {
     });
 }
 
+#[test]
+fn arena_is_not_send_when_allocator_is_not_sync() {
+    assert_not_send::<Arena<SendOnlyAllocator>>();
+}
```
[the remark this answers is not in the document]
geeknoid: Added compile-time negative `Send` assertions for both `Arc<u64, SendOnlyAllocator>` and `Box<u64, SendOnlyAllocator>`, alongside the existing `Arena` assertion. This now locks the `A: Send + Sync` requirement for cross-thread smart-pointer drops.

crates/multitude/src/rc.rs
```diff
@@ -210,15 +219,13 @@ impl<T: ?Sized + Pointee, A: Allocator + Clone> Clone for Rc<T, A> {
...
-            // `>=`: aborting when the count is *at* the threshold guarantees the
-            // post-increment value never exceeds `MAX_STRONG_REFCOUNT`. (Unlike
-            // `Arc`, whose atomic `fetch_add` post-increments before the check,
-            // `Rc` reads first and can refuse the increment outright.)
-            if prev >= MAX_STRONG_REFCOUNT {
+            let next = ptr::read_unaligned(strong).wrapping_add(1);
+            ptr::write_unaligned(strong, next);
+            // The process aborts before the temporarily over-limit count is
+            // observable, matching `Arc`'s post-increment saturation check.
+            if next > MAX_STRONG_REFCOUNT {
```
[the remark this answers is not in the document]
geeknoid: Fixed in bdb8f65f. `Rc::clone` now checks the current strong count before writing the increment, so the test-only overflow panic leaves the allocation state unchanged. The regression test catches the unwind, asserts that the count remains at the threshold, and only then restores the synthetic test count for normal teardown.

crates/multitude/src/serialization.rs
```diff
@@ -0,0 +1,562 @@
...
+//! standard `String` and `Vec<T>` visitors therefore allocate through their
+//! normal global storage. [`DeserializeIn`] is the allocator-aware counterpart:
+//! it receives an [`Arena`] and recursively passes that arena through
+//! [`DeserializeInSeed`].
+//!
+//! Derive [`DeserializeIn`](derive@DeserializeIn) for application types whose
+//! owned fields should live in an arena:
+//!
+//! ```
+//! # #[cfg(feature = "serde")] {
```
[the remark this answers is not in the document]
geeknoid: Fixed in dba58c19. The doctest now uses the repository-standard hidden `fn main` plus `#[cfg(feature = "serde_json")]` shim, so it compiles with `serde` alone and still exercises the JSON example when `serde_json` is enabled.

crates/multitude/src/serialization/reuse.rs
```diff
@@ -0,0 +1,96 @@
...
+            }
+
+            fn visit_borrowed_str<E: serde::de::Error>(self, v: &str) -> Result<Self::Value, E> {
+                self.visit_str(v)
+            }
+
+            fn visit_string<E: serde::de::Error>(self, v: alloc::string::String) -> Result<Self::Value, E> {
+                self.visit_str(&v)
+            }
+        }
```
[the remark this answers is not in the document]
geeknoid: Fixed in dba58c19. The visitor implementation now names the deserialization lifetime explicitly, and visit_borrowed_str uses that same lifetime, matching the Serde visitor contract.

crates/multitude/src/rc.rs
```diff
@@ -235,11 +240,18 @@ impl<T: ?Sized + Pointee, A: Allocator + Clone> Drop for Rc<T, A> {
...
         // SAFETY: non-atomic, single-thread unaligned read-modify-write.
-        let prev = unsafe { ptr::read_unaligned(strong) };
-        if prev != 1 {
-            // SAFETY: `prev >= 2`, so decrementing keeps the count positive.
-            unsafe { ptr::write_unaligned(strong, prev - 1) };
+        let next = unsafe { ptr::read_unaligned(strong).wrapping_sub(1) };
+        // SAFETY: same slot; the count stays valid (positive until the final
+        // drop, then `0`).
+        unsafe { ptr::write_unaligned(strong, next) };
+        if next != 0 {
```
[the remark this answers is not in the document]
geeknoid: Fixed in b3b17447. `Rc::drop` now reads the strong count once, debug-asserts that a live handle has a nonzero count, and uses normal subtraction.

sandersaares (approved):

crates/multitude/Cargo.toml
```diff
@@ -62,8 +65,10 @@ bytemuck = { workspace = true, features = ["derive"], optional = true }
 bytes = { workspace = true, optional = true }
 bytesbuf = { workspace = true, optional = true }
 hashbrown = { workspace = true, features = ["allocator-api2", "default-hasher"], optional = true }
+multitude_macros = { workspace = true, optional = true }
 ptr_meta = { workspace = true }
-serde = { workspace = true, optional = true }
+serde = { workspace = true, features = ["alloc"], optional = true }
```
martintmk, line 70: Few notes:

- What is the performance of arena-aware serialization and deserialization vs using plain serde
- would it make sense to introduce multitude_serde crate that adds serde support? (given the number of new serde-related code required)

geeknoid: I'll add a benchmark comparing normal serde and multitude serde.

I'm not sure if I can put the serde stuff in a separate crate, I think I'm using some of multitude's internal types. But is it worth splitting, given the cargo feature?

crates/multitude/src/serialization/reuse/deserialize_reuse.rs
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use serde::de::Deserializer;
+
+/// Deserialize into an existing value while reusing its arena allocation.
```
ralfbiedert: I don't get this trait here and the examples, do I as a user need this?

crates/multitude/src/de/mod.rs
```diff
@@ -0,0 +1,631 @@
...
+//! The derive supports named, tuple, and unit structs; transparent structs;
+//! and externally tagged enums. It follows Serde's deserialization behavior for
+//! `rename`, `rename_all`, `alias`, `default`, `skip`, `deny_unknown_fields`,
+//! `expecting`, explicit deserialize bounds, and `deserialize_with`/`with`.
+//! Named structs and struct variants accept both map and ordered-sequence
+//! representations.
+//!
+//! Arena-specific attributes are:
+//!
+//! - `#[multitude(via_serde)]` to use ordinary Serde for one field.
```
ralfbiedert, line 129: This needs more clarification / documentation. It took me a while playing with this crate to figure out how things belong together, in particular how `Deserialize` and `DeserializeIn` (not) work together, and what that means for my own or 3P types (which might impl `Deserialize` already).

crates/multitude/src/serialization/limits.rs
```diff
@@ -0,0 +1,450 @@
...
+/// let value = deserialize_seed_with_limits(
+///     U64Deserializer::new(11),
+///     PhantomData::<u64>,
+///     DeserializationLimits::unlimited(),
+/// )?;
+/// assert_eq!(value, 11);
+/// # Ok(())
+/// # }
+/// ```
+pub fn deserialize_seed_with_limits<'de, D, S>(deserializer: D, seed: S, limits: DeserializationLimits) -> Result<S::Value, D::Error>
```
ralfbiedert: Same here, do I as a user need to call this?

crates/multitude/src/serialization/cow_str.rs
```diff
@@ -0,0 +1,161 @@
...
+/// use [`Owned`](Self::Owned), while unescaped strings read from a slice can use
+/// [`Borrowed`](Self::Borrowed).
+///
+/// ```
+/// use multitude::serialization::CowStr;
+///
+/// let text: CowStr<'_> = CowStr::Borrowed("borrowed");
+/// assert!(text.is_borrowed());
+/// ```
+pub enum CowStr<'de, A: Allocator + Clone = Global> {
```
ralfbiedert: This feels a bit misplaced. Why would that live inside `serialization`?

ralfbiedert (approved):

crates/multitude/src/de/mod.rs
```diff
@@ -0,0 +1,631 @@
+// Copyright (c) Microsoft Corporation.
```
ralfbiedert, line 1: Nit, it's kind of weird this is called `serialization` while it's only about `deserialization`.

ralfbiedert: (That said, both are too long for my taste. Maybe `de`, `deser`, ... `serde` maybe?)

crates/multitude/src/serialization/reuse/deserialize_reuse.rs
geeknoid: No longer required for normal use. The public `DeserializeReuse` trait and seed have been removed; Multitude `String` and `Vec` now expose inherent `deserialize_reusing` methods directly. The implementation abstraction is private.

crates/multitude/src/de/mod.rs
geeknoid: Expanded the new `de` module documentation around this distinction. It now leads with the normal derive-and-`Arena` path, states that `Deserialize` and `DeserializeIn` are independent and may both be derived, includes a compatibility table, demonstrates `#[multitude(via_serde)]`, and explains the options and allocator behavior for third-party types.

crates/multitude/src/serialization/cow_str.rs
geeknoid: Agreed. The Serde-specific `CowStr` has been replaced by a general root-level `multitude::Cow<T, A>`. It supports sized values, strings, and slices with arena-aware `into_owned`, `to_mut`, and `clone_in` APIs; Serde now only supplies `DeserializeIn` integrations for strings and bytes.

crates/multitude/src/serialization/limits.rs
geeknoid: No. The generic seed-level limits adapter is now internal. Applications use `DeserializationLimits` with the `Arena::*_with_limits` convenience methods; only `de::seed::DeserializeInSeed`, needed by custom `DeserializeIn` implementations, remains public.

crates/multitude/src/de/mod.rs
geeknoid: Renamed the public module from `serialization` to `de`, with no compatibility alias since this API has not shipped. The common dynamic types are flattened under `multitude::de`, while advanced custom integration lives under `de::seed`.

crates/multitude/Cargo.toml
geeknoid: Follow-up: the PR now includes matched Criterion and Callgrind comparisons against ordinary Serde, including initial typed/dynamic deserialization, reused-allocator lifecycles, Bumpalo, and a separate allocation-teardown suite; `docs/PERF.md` is generated from those runs. I kept the integration feature-gated in `multitude`: Serde impls for Multitude types and inherent `Arena::deserialize*` methods must live with the defining types, so an extension crate would require wrappers or a much deeper crate split without reducing the user-facing dependency surface.

ralfbiedert (approved):

ralfbiedert (approved):

sandersaares (approved):

PR #590, opened 2026-07-20
fix(bytesbuf): make BytesView hash independent of span segmentation

sandersaares:
[Copilot speaking]

## Problem

`BytesView` equality (`PartialEq`/`Eq`) compares only the logical byte sequence and is independent of how the bytes are segmented into spans. The `Hash` impl, however, called `Hasher::write` once per span:

```rust
self.len.hash(state);
let mut span_idx = self.spans_reversed.len();
while span_idx > 0 {
    span_idx -= 1;
    state.write(&self.spans_reversed[span_idx]);
}
```

The `Hasher` contract does **not** guarantee that `write(a); write(b)` hashes identically to `write(a ++ b)`. So two **equal** views with different span segmentation could produce **different** hashes under a boundary-sensitive hasher, violating the `Hash`/`Eq` contract (equal values must hash equally) and potentially breaking `HashMap` lookups.

`DefaultHasher` (SipHash) happens to be boundary-insensitive, which is why the existing tests never caught this.

Fixes AB#7622952.

## Fix

`Hash` now re-chunks the logical byte sequence into fixed-size blocks (`HASH_BLOCK_BYTES = 64`) using a small **stack** buffer, then feeds those blocks to the hasher. Block boundaries depend only on the byte content (a full block every 64 bytes, then a final partial block), never on span boundaries, so equal views always drive the hasher with an identical sequence of `write` calls.

Key properties:

- **Allocation-free** — uses only a `[u8; 64]` stack buffer; no `to_vec()`/heap, as hashing may be on a hot path.
- **Fast path** — when the buffer is empty and a span holds at least a full block, that block is written directly from the span, avoiding a copy of large contiguous views through the buffer. It carries the exact same bytes, so the canonical call sequence is preserved.
- The length prefix (`self.len.hash(state)`) is retained to keep the hash prefix-free, matching how the standard library hashes byte slices.

## Tests

- `hash_is_independent_of_span_segmentation` and `..._across_block_boundary`: use a **boundary-sensitive** hasher (mixes each `write` call's length into its state) to prove equal views with 1/2/3 spans, and with span splits straddling the 64-byte block boundary, hash identically. These fail against the old per-span implementation.
- `hash_write_sequence_is_identical_regardless_of_segmentation`: a **recording** hasher captures every `write` call and asserts the exact call sequence is identical across many segmentations for lengths `{0, 1, 63, 64, 65, 128, 130, 200}`, locking in the canonical behavior and exercising the fast path.

## Validation

- `bytesbuf` lib: 205 tests + doctests pass.
- `cargo clippy -p bytesbuf --lib --all-features`: clean.
- `just spellcheck`: clean. Formatted.

### DIFF
--- a/crates/bytesbuf/src/view.rs
+++ b/crates/bytesbuf/src/view.rs
@@ -928,20 +928,66 @@ impl<const LEN: usize> PartialEq<BytesView> for &[u8; LEN] {
 
 impl Eq for BytesView {}
 
+/// Block size used when hashing a [`BytesView`], chosen as one cache line so that small views
+/// (the common case) are hashed with a single `write` call while larger views still drive the
+/// hasher with a bounded number of bulk writes. The exact value only affects performance; any
+/// fixed constant yields a correct, segmentation-independent hash.
+const HASH_BLOCK_BYTES: usize = 64;
+
 impl Hash for BytesView {
     fn hash<H: Hasher>(&self, state: &mut H) {
-        // Hash all bytes in logical order, consistent with PartialEq.
-        // We also hash the cached length as an additional input, but equality and hashing are
-        // both defined by the logical byte sequence rather than how it is segmented into spans.
+        // Equality is defined by the logical byte sequence, independent of how it is segmented
+        // into spans, so hashing must be independent of segmentation too. The `Hasher` contract
+        // does not guarantee that `write(a); write(b)` produces the same result as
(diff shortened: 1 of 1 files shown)


martin-kolinek (approved):
Nice catch

geeknoid (approved):

PR #595, opened 2026-07-22
docs(fetch_winhttp): add design document for WinHTTP custom transport

sandersaares:
[Copilot speaking]

Adds a placeholder `fetch_winhttp` crate whose sole content is a design document: `crates/fetch_winhttp/docs/DESIGN.md`. No transport code yet - this PR is to review the design before implementation.

`fetch_winhttp` is a Windows-only custom `fetch` transport that drives the OS WinHTTP API in asynchronous mode as a Tokio-free alternative to `fetch_hyper`. OS calls go through a bindings facade (real `windows`-crate impl plus a mock for unit tests), mirroring the pattern used in `oxidizer_io`.

The design covers the threading model (one shared OS session per built client, one transport instance per core, callbacks completing inline for processor affinity), the cancellation and FFI-ownership model (pooled `RequestContext` reclaimed on the final `HANDLE_CLOSING`), error handling, HTTP protocol negotiation, TLS (WinHTTP/Schannel-managed, with an insecure mode), compression, and request/response body streaming modelled as `bytesbuf_io` reads/writes. It also includes a test plan (mock-bindings unit tests plus localhost integration tests) and a chapter of feedback on the `fetch` API where its Hyper/Tokio-shaped assumptions force workarounds.

The document has been through several rounds of rubber-duck review; opening now for wider human review of the design.


crates/fetch_winhttp/README.md
```diff
@@ -0,0 +1,39 @@
+<div align="center">
+ <img src="./logo.png" alt="Fetch Winhttp Logo" width="96">
+
+# Fetch Winhttp
```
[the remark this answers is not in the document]
sandersaares: This is generated content - we cannot/shouldnot touch it. Make a note in agents.md not to touch generated readme.md files or complain about their formatting in reviews.

crates/fetch_winhttp/docs/DESIGN.md
```diff
@@ -0,0 +1,1381 @@
...
+block (§3.1).
+
+### 1.2 TLS is configured on the transport, not through `fetch`'s `TlsOptions`
+
+`fetch`'s generic `TlsOptions`/`TlsBackend` carries rustls/native-tls material
+(crypto providers, verifiers, client-cert resolvers) that is meaningless to
+Schannel. WinHTTP does TLS itself and accepts only a small set of knobs, so
+`fetch_winhttp` ignores `CustomContext::tls` and takes its own `WinHttpTlsConfig`
+through `Extras` (§9). That TLS configuration is inherently transport-specific is,
+in our view, a `fetch` API shortcoming; see §17.
```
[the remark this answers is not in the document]
sandersaares: [Copilot speaking] Fixed in db5cb490 - reworded to `TLS configuration being inherently transport-specific is, in our view, a fetch API shortcoming` to drop the doubled verb.

crates/fetch_winhttp/src/lib.rs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#![cfg_attr(docsrs, feature(doc_cfg))]
```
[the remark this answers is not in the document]
sandersaares: [Copilot speaking] Deferring this. The crate is currently a design-only placeholder whose sole test is a cross-platform trivial `placeholder` test; gating the crate root with `#![cfg(windows)]` now would compile that test out on non-Windows targets and make `cargo nextest` fail with `no tests to run` (which already broke the pr-test job once). The `#![cfg(windows)]` gate described in DESIGN.md S2.1 will land together with the implementation and its Windows-only test suite, at which point the non-Windows no-tests case will be handled deliberately.

crates/fetch_winhttp/README.md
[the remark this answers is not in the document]
sandersaares: [Copilot speaking] Agreed - this README is generated by `just readme` from the crate doc comments and should not be hand-edited or flagged for casing/formatting. Per @sandersaares's note, I documented this in AGENTS.md (db5cb490) so future reviews skip generated READMEs.

crates/fetch_winhttp/docs/DESIGN.md
```diff
@@ -0,0 +1,1381 @@
+# `fetch_winhttp` design
```
martintmk: nit: 1400 lines of design, can we shorten it?

crates/fetch_winhttp/docs/DESIGN.md
```diff
@@ -0,0 +1,1381 @@
...
+the *transport* is the component that actually performs network communication.
+Configuration should be split along that line - anything that governs how bytes
+travel over the network (TLS, the network-phase timeouts, connection lifetime and
+reuse) is a transport concern and belongs on the transport, while end-to-end pipeline
+concerns (overall response deadline, retry, telemetry) stay in `fetch`. Several of
+the items below are instances of that split being in the wrong place today.
+
+- **No first-class way to plug in a downstream transport.** `HttpClient` and its
+  builder live in `fetch`, so a transport crate cannot add an inherent
+  `HttpClient::winhttp(..)` constructor (orphan rule). We expose free functions
```
martintmk: it's better to expose these as extensions so you get a nice HttpClient::builder_win_http experience

crates/fetch_winhttp/docs/DESIGN.md
```diff
@@ -0,0 +1,1381 @@
...
+  (`fetch_winhttp::builder`/`new`) instead, which is fine but asymmetric: transports
+  that happen to live in `fetch` get first-class methods, downstream ones get
+  bolted-on free functions. A transport-agnostic API where the caller always
+  explicitly constructs and plugs in a transport - e.g.
+  `HttpClient::build(fetch_winhttp::transport().tls(cfg))` - would be more
+  predictable, avoid a Hyper-colored default, and treat every transport uniformly.
+  The slightly more verbose hello-world is worth the consistency.
+
+- **TLS configuration lives at the wrong layer.** `fetch`'s generic `TlsOptions`
+  carries rustls/native-tls material that only the Hyper transport understands;
```
martintmk: TlsOptions are agnostic to hyper, the plumbing that consumes TlsOptions and adapts them to hypers lives in fetch_hyper

crates/fetch_winhttp/docs/implementation.md
```diff
@@ -0,0 +1,1381 @@
...
+in our view, a `fetch` API shortcoming; see §17.
+
+### 1.3 Architecture at a glance
+
+The whole design in one picture, so later sections can be read as elaborations of
+this model rather than assembled from details:
+
+- **One shared OS session per built client**, opened at build time, `Arc`-shared by
+  all per-core instances, and immutable after setup (§4.2).
+- **One transport instance per core.** `fetch` clones and relocates the transport
```
martintmk, line 13: we might leave this configurable, for clients with low-traffic, there is no need to create per-core instances

crates/fetch_winhttp/docs/DESIGN.md
sandersaares: Do you have specific suggestions for which sections are unnecessary or overly verbose? Hard to take action based on line count alone.

crates/fetch_winhttp/docs/implementation.md
sandersaares: Might be worth noting down as a low-traffic client optimization pass (possibly deferred), though needs measurements/benchmarks to evaluate impact. Which of our current known customer scenarios would you classify as the most representative of this situation and how would you characterize their usage of fetch (i.e. what aspects would be important to replicate in a benchmark)?

crates/fetch_winhttp/docs/DESIGN.md
ralfbiedert: Since I had the same issue, I used an agent to help me outline my thoughts here. Please take some time go through them. If you think anything is unnecessary or verbose, I'd love to hear specific suggestions how to improve it!

[fetch_winhttp_design_analysis.md](https://github.com/user-attachments/files/30302889/fetch_winhttp_design_analysis.md)

martintmk: I read Ralf's analysis and more or less agree with that assessment pages. I would love to see a shortened section, max 3-4 pages with core tenets and requirements.

Many parts of this document read more as an implementation plan for agents.

martintmk: What would really help is to see actual APIs and how these are used in fetch.

``` rust
// do we need some arguments for win http?
HttpClient::builder_winhttp(...)
   .winhttp(...) // is there win-http specific configuration
   .tls(TlsOptions::default()) // !!! THESE ARE IGNORED, we need to address this in fetch
   .build()
```
Maybe even highlight fetch knobs that are not applied.

ralfbiedert (approved):
Approved. The design is internally coherent, including FFI ownership, callback handoff, cancellation, and timeout reasoning. For implementation follow-up: keep the cross-crate design critique out of public rustdoc, verify docs.rs/fetch links, and hold the eventual unsafe FFI layer to the proposed mock/Miri validation.

crates/fetch_winhttp/docs/DESIGN.md
sandersaares: Factored document into 3 separate documents to simplify reviewing from different viewpoints. Added API example for usage in fetch. The API shape is not something I am particularly happy with but I think this needs fetch-level design work to make working with transport-specific concerns more ergonomic, not something specific to winhttp integration.

@ralfbiedert I found it quite hard to extract actionable feedback from the attached analysis. There is a lot of just description in there and a hilarious amount of "the key problem with this plan is that it is a plan so we don't really know for sure yet", which obscures the more valuable points. I was able to extract some specific items from there but it feels like it was missing a "robot, condense this analysis to actionable suggestions" turn at the end.

crates/fetch_winhttp/docs/DESIGN.md
sandersaares: Integrated into doc.

crates/fetch_winhttp/docs/DESIGN.md
sandersaares: Clarified in text that "hyper" here refers to "fetch_hyper" (the fetch feedback is that TLS is a property of transports like `fetch_hyper` not of the pipeline composer `fetch`).

crates/fetch_winhttp/docs/DESIGN.md
martintmk: Thanks Sander, the shortened design is now much more digestible. Will go through it.

crates/fetch/docs/stabilization.md
```diff
@@ -0,0 +1,100 @@
...
+
+There is no first-class way to plug in a downstream transport. `HttpClient` and its
+builder live in `fetch`, and the orphan rule forbids a downstream crate from adding
+inherent constructors, so `fetch_winhttp` ships an extension trait
+(`HttpClientWinHttpExt`) the caller must import to get `HttpClient::builder_winhttp`/
+`new_winhttp`. That works, but it is asymmetric: transports bundled in `fetch` get
+first-class inherent methods (`HttpClient::builder_tokio`/`new_tokio`), while a
+downstream transport is reachable only through an imported trait. The extension trait is a
+workaround, not a fix - it papers over the call syntax but leaves the asymmetry, and the
+asymmetry itself is the wart: all HTTP-client composition options should be on equal
```
martintmk, line 35: The convenience for common workloads is intentional. We don't want to go through extra loops for 90% of users that just want to use tokio with whatever default transport we choose.

Also, let's think about the current setup more as a stopgap while we work on fine-tuning and stabilizing other parts of our stack (runtime, security). Once that's in, we will also have built-in `builder_oxidizer` + `new_oxidizer` APIs.

crates/fetch/docs/stabilization.md
```diff
@@ -0,0 +1,100 @@
...
+footing for good UX. A transport-agnostic
+API where the caller always explicitly constructs and plugs in a transport - e.g.
+`HttpClient::build(fetch_winhttp::transport().tls(cfg))` - would be more predictable,
+avoid a Hyper-colored default, and treat every transport uniformly. The slightly more
+verbose hello-world is worth the consistency.
+
+## 2. Transport-specific TLS configuration
```
martintmk, line 70: Current bundling to fetch crate might be unfortunate because indeed the fetch_tls might not be useable in current form with winhttp.

However, my thinking is that fetch_tls might server as a set of descriptors that can be interpreted by different TLS technologies.

One thing we need to address is how the different capabilities should be represented. (for example, certificate validation can inspected by callbacks for rustls but not for native-tls.

crates/fetch/docs/stabilization.md
```diff
@@ -0,0 +1,100 @@
...
+
+TLS configuration is over-abstracted. `fetch`'s generic `TlsOptions`
+carries rustls/native-tls material that only `fetch_hyper` - the Hyper-based `fetch`
+transport, which configures its TLS through exactly that material - can consume;
+Schannel-based `fetch_winhttp` cannot use it and needs its own knobs (see fetch_winhttp design.md §1.2 and §4).
+The TLS model is a property of each transport, not of `fetch`: `fetch_hyper` and
+`fetch_winhttp` have fundamentally different, non-interchangeable TLS configuration, so
+it cannot be handled at the `fetch` layer at all. Per the split above, TLS should be
+configured per transport, on the transport being plugged in; the
+`HttpClient::build(transport().tls(cfg))` shape would express this cleanly.
```
martintmk, line 80:
```suggestion
`HttpClient::builder_tokio(deps, tls)`
```

this is also possible alternative, do not tie tls options to builder but rather explicitly require them on construction

crates/fetch/docs/stabilization.md
```diff
@@ -0,0 +1,100 @@
...
+concurrency differently again. The other options are the same story: WinHTTP exposes no
+per-connection age control, so `connection_lifetime` cannot be implemented here at all
+(see fetch_winhttp design.md §2.2), and it applies its own per-protocol keep-alive with
+its own floors, so `ConnectionKeepAlive` and `connection_idle_timeout` map only
+approximately or not at all. Per the split above, this configuration belongs on the
+transport.
+
+## 4. Scope-based timeout configuration
+
+Timeouts are over-abstracted, not under-modeled. `fetch` models a connect
```
martintmk, line 131: If we ignore all technical details of timeouts across different timeouts, one timeout that the end consumer of fetch cares about is "connect timeout". 

While I believe configuration knobs are useful, on the level of common API, connect timeout is something we should keep and let transports interpret it.

Last resort for transport is to leverage drop safety, if the future is cancelled, transport should handle it properly.

crates/fetch/docs/stabilization.md
```diff
@@ -0,0 +1,100 @@
...
+## 5. Transport construction ergonomics
+
+Constructing a client from a transport is clumsy. Because a transport's configuration
+arrives as a single deps struct, the caller writes a struct literal -
+`HttpClient::builder_winhttp(WinHttpDeps { tls: ..., options: ..., sink: ... })` - rather
+than a fluent chain. A more natural surface would configure the transport through closures
+on the builder, e.g.
+`HttpClient::builder_winhttp().tls(|tls| tls.accept_invalid_certs()).timeouts(|t| t.read(3000)).custom_pipeline(..).build()`,
+or a compositional form where the transport is itself a builder plugged into the client:
+`HttpClient::builder().transport(WinHttpTransport::builder().tls(|tls| ..).build()).build()`.
```
martintmk, line 159: Maybe `HttpClient::builder()` could return `HttpClientBuilder<NeedsTransport>`. APIs for `HttpClientBuilder<NeedsTransport>` could expose configuration for pipeline, i.e. things that are not transport agnostic.

crates/fetch/docs/stabilization.md
```diff
@@ -0,0 +1,100 @@
+# `fetch` API stabilization feedback
```
martintmk, line 1: TLDR: many good points but one major point I want to highlight. fetch is `batteries includes` so we **DO** play favorites. We will pick our pick for default transport for each supporter runtime and support for custom transports should not impact the ergonomics for our default choices. 

This can be done ergonomically where `HttpClient::builder(...)` could return `HttpClientBuilder<NeedsTransport>` while   `HttpClient::builder_tokio(...)` returns something like `HttpClientBuilder<DefaultTransport>` where we plug-in our recommended transport.

martintmk: also the generic parameter should be defaulted, i.e. `HttpClientBuilder<T = DefaultTransport>`

(we can live with generics on `HttpClientBuilder`, generics on `HttpClient` are probably a no-go)

martintmk: I also think that is ok for us to expose lowest common denominator for our default transport and then also enable deeper customization through specific transport knobs:

```
HttpClient::builder(...).hyper(hyper_transport)
```

crates/fetch_winhttp/docs/design.md
```diff
@@ -0,0 +1,318 @@
...
+
+```rust,ignore
+use fetch::HttpClient;
+use fetch_winhttp::HttpClientWinHttpExt; // brings the constructors into scope
+
+// Defaults:
+let client = HttpClient::new_winhttp();
+
+// Configured:
+let client = HttpClient::builder_winhttp(WinHttpDeps {
```
martintmk: For use of `..Default::default()` we cannot apply `#[non_exhaustive]`. Are we sure we won't be adding new fields later?

crates/fetch_winhttp/docs/design.md
```diff
@@ -0,0 +1,318 @@
...
+  3xx responses are surfaced to the caller unchanged rather than followed, with no knob
+  to re-enable automatic redirects.
+- **Cookies and automatic authentication are disabled.** The transport keeps no cookie
+  store and does not answer `WWW-Authenticate`/407 challenges; `Set-Cookie`/`Cookie` and
+  challenge responses pass through as plain data for the caller to manage. The transport
+  is thus stateless between requests.
+
+(The specific OS options behind each behavior are implementation.md §10.3.)
+
+## 6. Timeouts and time
```
martintmk, line 242: from the perspective of transport, it needs to ensure drop safety, i.e. when fetch cancels the future, transport needs to honor it.

Also, fetch attaches timeout information to each request that can be inspected and may be even applied to transport?

See:
https://docs.rs/http_extensions/latest/http_extensions/timeout/index.html

martintmk: saw this mentioned bellow, drop safety still applies

crates/fetch_winhttp/docs/design.md
```diff
@@ -0,0 +1,318 @@
...
+
+- **Connect timeout** (`TransportOptions.connect_timeout`, default 30 s): honored by this
+  transport as a *total* deadline on connection establishment (§6.2). `fetch` models this
+  option but leaves each transport to enforce it.
+- **Response timeout and body idle timeout** (`http_extensions::ResponseTimeout` /
+  `BodyTimeout`): enforced by `fetch` above the transport and honored here the same way
+  `fetch_hyper` does; both surface as `HttpError::timeout`.
+- **Seatbelt request timeout**: enforced above the transport; the transport is not
+  involved.
+- **Resolve and send timeouts**: `fetch` has no concept for these, so they are exposed as
```
martintmk: `SendTimeout` is really just the `ResposeTimout` in disguise. We send request and wait for the response without touching the response body. After that the `BodyTimeout` applies.

martintmk:
Thanks for fine-tuning the design for humans, the split into design + implementation might be something we should leverage more to avoid overly technical design docs.

I think this is good to go, minor comments. The fetch stabilization is a different topic we need address. Custom winhttp transport implementation will give us a lot of useful insights.

crates/fetch_winhttp/docs/design.md
sandersaares: [Copilot speaking] Agreed - we should expect to add fields. Reworked design.md §1.1 so `WinHttpDeps`, `WinHttpOptions` and `WinHttpTlsConfig` are all `#[non_exhaustive]` and built through builders (`WinHttpDeps::builder().tls(...).options(...).sink(...).build()`) instead of a struct literal with `..Default::default()`, so new knobs can land compatibly. Fixed in 2883b8d8.

crates/fetch_winhttp/docs/design.md
sandersaares: [Copilot speaking] Both points addressed in 2883b8d8.

Drop safety: honored. Dropping the request future closes the WinHTTP handle, which cancels the in-flight operation and delivers the terminal HANDLE_CLOSING that reclaims the request context (design.md §7 cancellation model; implementation.md §4). I added an explicit note in design.md §6.1 that cancellation is the transport's backstop for any timeout it does not enforce natively.

Per-request timeout inspection: adopted where the semantics actually match, which turned out to be a split:
- BodyTimeout (idle gap between body chunks, reset on progress) matches WINHTTP_OPTION_RECEIVE_TIMEOUT (per-receive-operation idle timer, reset each read), so the transport now reads it from the request extensions and programs that native timer per-request.
- ResponseTimeout is a *total* over connect + send + receive-headers. WINHTTP_OPTION_RECEIVE_RESPONSE_TIMEOUT covers only the post-send wait for the first response byte (it excludes connect and send), so it is not a faithful remap. ResponseTimeout therefore stays enforced above the transport (as fetch/http_extensions already do) and the native receive-response timer is set only as a looser liveness backstop.

Mapping detail is in implementation.md §10.4.

crates/fetch_winhttp/docs/design.md
sandersaares: [Copilot speaking] Agreed - fixed in 51169ff2. Removed the standalone "send timeout" framing: §6.1 now states send is not a distinct concept (send + await-headers is exactly the span `ResponseTimeout` governs, then `BodyTimeout` takes over), leaving only a `Resolve` transport-specific knob. Updated the implementation.md §10.4 timeout table to match (WinHTTP send field left as a loose backstop, not driven by a dedicated fetch knob).

martintmk (approved):
I think this is good to go, fetch API design is different discussion

crates/fetch_winhttp/docs/design.md
```diff
@@ -0,0 +1,350 @@
...
+
+// Configured. Every WinHTTP config type is built through a builder, never a struct
+// literal, so all of them are `#[non_exhaustive]` and can gain knobs in later versions
+// without a breaking change:
+let deps = WinHttpDeps::builder()
+    .tls(WinHttpTlsConfig::builder()
+        .accept_invalid_certs(true)                 // Schannel knobs, §4
+        .build())
+    .options(WinHttpOptions::builder()
+        .connect_timeout(Duration::from_secs(10))   // protocol/timeout tuning, §3, §5, §6
```
martin-kolinek: 🤖: `connect_timeout` is configured here via `WinHttpOptions`, but §6.1/§6.2 and implementation.md §10.4 say the transport reads `TransportOptions.connect_timeout` (implementation.md's timeout test table also alternates between the two). Please pick one source of truth or specify precedence — otherwise this public example and the implementation contract can yield different connect-timeout values.

crates/fetch_winhttp/docs/design.md
```diff
@@ -0,0 +1,350 @@
...
+
+`WinHttpTlsConfig` (§4) and `WinHttpOptions` (§3, §5, §6) follow the same
+builder + `#[non_exhaustive]` pattern.
+
+### 1.2 TLS is configured on the transport, not through `fetch`'s `TlsOptions`
+
+`fetch`'s generic `TlsOptions`/`TlsBackend` carries rustls/native-tls material
+(crypto providers, verifiers, client-cert resolvers) that is meaningless to
+Schannel. WinHTTP does TLS itself and accepts only a small set of knobs, so
+`fetch_winhttp` therefore ignores `fetch`'s TLS configuration entirely and takes its
```
martin-kolinek, line 93: 🤖: Silently ignoring `fetch`'s `TlsOptions` is the one dropped option with security consequences. `CustomContext.tls` is always populated, so a caller who set non-default TLS material — a private CA root, a client identity (mTLS), or cert-pinning — silently gets a default system-trust connection with no signal, even though the same code 'works' on fetch_hyper. Please make the design explicit that a non-default `TlsOptions` triggers the §11 build-time warning, and consider a hard error (not just a warn) when security-affecting TLS material is present, since this is a silent security weakening rather than a benign fidelity gap.

crates/fetch_winhttp/docs/implementation.md
```diff
@@ -0,0 +1,1127 @@
...
+    deps: impl Into<CustomDeps<Extras>>,
+) -> HttpClientBuilder
+where
+    F: Fn(CustomContext<Extras>) -> R + Send + Sync + 'static,
+    R: RequestHandler + 'static,
+    Extras: ThreadAware + Send + Sync + Clone + 'static;
+```
+
+`CustomContext` hands the factory a `HttpBodyBuilder` (carrying the clock and
+read-buffer pool), a `PoolIndex`, the generic `TransportOptions`/`TlsOptions`, a
```
martin-kolinek, line 889: 🤖: Ignoring `PoolIndex` collapses `fetch`'s configured `multiple_pools` into a single WinHTTP connection pool: `fetch` invokes the custom transport factory once per pool slot specifically so routing to a `PoolIndex` selects an independent transport pool, but every handler here shares the one OS session — so nominally separate pools all use the same OS connection pool. That contradicts the `fetch` contract and this design's §11 warning policy. Either key WinHTTP sessions by `PoolIndex` (one session per pool slot, shared across cores within a slot), or add `multiple_pools` to the explicitly-unhonored options and emit the §11 warning instead of silently collapsing it.

crates/fetch_winhttp/docs/implementation.md
```diff
@@ -0,0 +1,1127 @@
...
+`Meter`, and the caller's `Extras`. `fetch_winhttp` ignores `PoolIndex` (per-core
+placement comes from `Isolation::Isolated`, §3.2) and ignores `CustomContext::tls` (it
+takes its own `WinHttpTlsConfig` instead; see design.md §1.2).
+
```
martin-kolinek: 🤖: Opening and capturing the session in `builder_winhttp` breaks the per-built-client isolation guarantee (design.md §2) when the builder is cloned: `HttpClientBuilder` is `Clone`, so two clones can be built into independent clients while retaining this same captured `Arc<WinHttpSession>` — and therefore the same WinHTTP pool. Since that isolation is a stated TLS-policy security boundary (a strict client must not share a pooled connection with an `accept_invalid_certs` one), please make the session build-scoped (e.g. create per-client shared state at `build`, lazily initialized from the finalized `CustomContext`) rather than builder-scoped, and add an integration test asserting that two independently built clients targeting the same authority establish separate connections.

crates/fetch_winhttp/docs/implementation.md
```diff
@@ -0,0 +1,1127 @@
...
+Gated behind `#[cfg(windows)]` and `#[cfg_attr(miri, ignore)]`, against a
+localhost server (a small `std::net`-based server, or `wiremock` as used
+elsewhere in `fetch`). These validate the real OS path end to end:
+
+- GET/POST with small and large bodies; response body correctness and size.
+- Streaming upload (unknown length -> chunked) and streaming download; assert
+  incremental delivery, not just final bytes.
+- Real gzip/deflate responses are transparently decoded.
+- `https` against a localhost TLS server with a self-signed cert: fails by
```
martin-kolinek: 🤖: The integration plan only exercises `accept_invalid_certs`, but the design exposes `accept_invalid_certs` and `accept_invalid_hostnames` as independent security relaxations. Please add a localhost TLS scenario with both an untrusted cert and a hostname mismatch, covering all four flag combinations: either flag alone must leave the other validation active, and only both together should succeed. That catches accidentally coupling the two Schannel ignore flags.

crates/fetch_winhttp/docs/DESIGN.md
sandersaares: [Copilot speaking] Addressed: the document was split into design.md (user-facing contract) and implementation.md (strategy) and trimmed substantially. Resolving as done.

crates/fetch_winhttp/docs/DESIGN.md
sandersaares: [Copilot speaking] Addressed: constructors are exposed via the `HttpClientWinHttpExt` extension trait (design.md 1.1), giving the `HttpClient::builder_winhttp`/`new_winhttp` experience. The residual asymmetry between bundled and downstream transports is captured as fetch stabilization item 1.

crates/fetch_winhttp/docs/DESIGN.md
sandersaares: [Copilot speaking] Addressed: stabilization.md item 2 states that `TlsOptions` is consumed by `fetch_hyper` (the Hyper transport), while Schannel-based `fetch_winhttp` needs its own knobs - TLS config is a per-transport property, not a `fetch`-level one.

crates/fetch_winhttp/docs/implementation.md
sandersaares: [Copilot speaking] Captured as a future opportunity in implementation.md 3.2 (per-core instancing could become a knob for low-traffic clients), but deliberately not in v1 scope: a knob earns its place only with demonstrated value, and per-core is a sound default.

crates/fetch/docs/stabilization.md
sandersaares: [Copilot speaking] Captured as reviewer input under stabilization.md item 1 (batteries-included default transport kept terse; current bundling a stopgap pending runtime/security stabilization, with future `builder_oxidizer`/`new_oxidizer`). Recorded as collected input for the stabilization phase, not an agreed decision.

crates/fetch/docs/stabilization.md
sandersaares: [Copilot speaking] Captured as reviewer input under stabilization.md item 2 (fetch_tls as transport-agnostic descriptors interpreted per TLS technology; open problem of representing capabilities like rustls validation callbacks that native-tls/Schannel lack). Collected for the stabilization phase, not an agreed decision.

crates/fetch/docs/stabilization.md
sandersaares: [Copilot speaking] Captured as reviewer input under stabilization.md item 2 (require TLS explicitly at construction, e.g. `builder_tokio(deps, tls)`, rather than threading it through the builder). Collected for the stabilization phase, not an agreed decision.

crates/fetch/docs/stabilization.md
sandersaares: [Copilot speaking] Captured as reviewer input under stabilization.md item 4 (keep connect timeout as a common, transport-agnostic knob each transport interprets; drop safety as the last-resort mechanism). Collected for the stabilization phase, not an agreed decision.

crates/fetch/docs/stabilization.md
sandersaares: [Copilot speaking] Captured as reviewer input under stabilization.md item 1 (`HttpClient::builder()` returning `HttpClientBuilder<NeedsTransport>`, exposing transport-agnostic pipeline config). Collected for the stabilization phase, not an agreed decision.

crates/fetch/docs/stabilization.md
sandersaares: [Copilot speaking] Captured as reviewer input under stabilization.md item 1 (batteries-included plays favorites: `builder()` -> `HttpClientBuilder<NeedsTransport>`, `builder_tokio()` -> `HttpClientBuilder<DefaultTransport>`, generic defaulted `T = DefaultTransport` and kept off `HttpClient`; deeper customization via `.hyper(...)`). Collected for the stabilization phase, not an agreed decision.

crates/fetch_winhttp/docs/design.md
sandersaares: [Copilot speaking] Fixed in b2bcb930. `TransportOptions.connect_timeout` is now the single source of the connect deadline; the overloaded `WinHttpOptions.connect_timeout` is removed from the 1.1 example (illustrated with `resolve_timeout` instead), design.md 6.1 states WinHttpOptions exposes no separate connect knob, and implementation.md 7.2/10.4 are aligned.

crates/fetch_winhttp/docs/design.md
sandersaares: [Copilot speaking] Deferred to the fetch TLS-configuration redesign rather than per-transport special-casing. b2bcb930 records the silent-ignore hazard in stabilization.md item 2: because `CustomContext::tls` is always populated, security-affecting `TlsOptions` a transport cannot consume are dropped with no signal. The fix is to configure TLS on the transport that can honor it so unhonorable combinations are unrepresentable; playing whack-a-mole with per-field warnings/rejections is explicitly rejected as papering over a mis-shaped API.

crates/fetch_winhttp/docs/implementation.md
sandersaares: [Copilot speaking] Pool ownership is unresolved and deferred to the v2 sessions/pools discussion. b2bcb930 notes in implementation.md 8 that ignoring `PoolIndex` collapses `multiple_pools` onto WinHTTP own per-authority pool, and stabilization.md item 3 records that pool ownership most likely belongs on the transport and may retire the `PoolIndex` surface. v1 leans on WinHTTP own pooling rather than keying sessions by `PoolIndex`.

crates/fetch_winhttp/docs/implementation.md
sandersaares: [Copilot speaking] Fixed in b2bcb930. The session is no longer opened eagerly in `builder_winhttp` and captured in the clone-shared factory closure; each per-core transport instance opens its own session inside the factory when fetch materializes it (implementation.md 3.2, 8), so two independently built clients - including builds of a cloned builder - never share a session or pool. Added a cross-client pool-isolation integration test (7.3) asserting two independently built clients to the same authority establish separate connections. Consequence: per-core sessions under Isolation::Isolated; a single cross-core session is not expressible with today custom-transport API, noted as fetch feedback.

crates/fetch_winhttp/docs/implementation.md
sandersaares: [Copilot speaking] Done in b2bcb930. implementation.md 7.3 now tests all four `accept_invalid_certs` x `accept_invalid_hostnames` combinations (neither -> fail; certs-only -> still fails on hostname; hostnames-only -> still fails on cert; both -> succeed), and 7.2 asserts each flag sets only its own SECURITY_FLAGS bit so the two are not coupled.

crates/fetch/docs/stabilization.md
```diff
@@ -0,0 +1,156 @@
...
+
+## 1. First-class transport plug-in
+
+There is no first-class way to plug in a downstream transport. `HttpClient` and its
+builder live in `fetch`, and the orphan rule forbids a downstream crate from adding
+inherent constructors, so `fetch_winhttp` ships an extension trait
+(`HttpClientWinHttpExt`) the caller must import to get `HttpClient::builder_winhttp`/
+`new_winhttp`. That works, but it is asymmetric: transports bundled in `fetch` get
+first-class inherent methods (`HttpClient::builder_tokio`/`new_tokio`), while a
+downstream transport is reachable only through an imported trait. The extension trait is a
```
martin-kolinek, line 33: An obvious alternative is to bundle winhttp into fetch itself - right now, the extensibility exists due to internal fetch variant, but once that also moves to this repo, that requirement goes away. The fundamental split between transport and pipeline obviously still makes sense, but the question of whether the transport list needs to be open-ended is not clear to me.

crates/fetch/docs/stabilization.md
```diff
@@ -0,0 +1,156 @@
+# `fetch` API stabilization feedback
```
martin-kolinek, line 1: A general question we need to answer is the configurability that fetch provides to its users. There are two distinct use cases in my head for fetch:
* Application uses it to directly call an API. This is the simpler use case as the application controls everything and can make all the decisions. If the application wants to use hyper, native-tls, configure the cert validation, connection lifetime and resilience policies, everything is fine as it has all the knobs it needs.
* Fetch is used by a library that calls endpoints, and the application uses this library - for example the ECS use case. This is a bit more complicated as the responsibility for what is configured by the application vs library is split. There are different ways of looking at this:
  * Application should get no way to configure the client, it's all the responsibility of the library. This gives the library the freedom to choose the transport layer and configure everything. The obvious disadvantage is that it makes it really hard for the application to work around configuration issues, and it takes away the dependency configuration from the application (if the library chooses to use rustls, the application has to live with that dependency in the tree). 
  * Application configures the whole client, library only gets completed HttpClient. Here, the library doesn't get the ability to configure even the basic things, such as the retry policy, which is arguably a thing that the library should clearly have control over.
  * Some form of a mix, where the application provides transport, library can configure things it cares about, and optionally, application still gets a way to override some of the properties. This potentially solves both of the issues with the other approaches, but it implies that we can figure out a set of configuration that libraries care about.

In my opinion, the third option is the most ergonomic one for the users. I could see a workable alternative with the second option, where we would expect libraries to expose a function that configures the HttpClientBuilder the way they want it, but that sounds like a fairly significant UX wart.

The challenge with the third option is that the configuration that libraries care about doesn't lie cleanly on the pipeline side of the transport/pipelines split. For example, ECS cares deeply about the connection lifetime, which is generally a transport concern. I was looking through the different configuration that fetch provides and early analysis indicated that we could make the configuration that clients are likely to care about work with both hyper and winhttp, but it does require the transport to be a bit richer than just layered/tower service.

martin-kolinek (approved):
The comments are related to the stabilization of fetch itself, not to the winhttp work - that looks good to me. I'd appreciate having the comments worked into the stabilization.md doc, but it's not a blocker.

crates/fetch/docs/stabilization.md
martintmk: Even in C# world, Win HTTP is now delivered as an optional transport handler:
https://www.nuget.org/packages/System.Net.Http.WinHttpHandler/

I would like to keep the same concepts for fetch, `fetch` is an enabler with some reasonable defaults not a monolithic all-in-one + everything crate. The win-http plumbing and infra will be quite heavy anyway.

One alternative is to have `fetch_core` crate that excludes the transport layer and then fetch that just re-exports form it and adds bunch of supported transports. Libraries would only depend on `fetch_core`, apps would use `fetch`.

sandersaares: [Copilot speaking] Folded into stabilization.md item 1 in 24cc85d2 as reviewer input (collected for stabilization, not an agreed decision): whether the transport list needs to be open-ended at all, given the internal fetch variant moving in-tree may remove the downstream-transport requirement, and the option of bundling fetch_winhttp directly while keeping the transport/pipeline split.

crates/fetch/docs/stabilization.md
sandersaares: [Copilot speaking] Folded into stabilization.md as new item 7 "Application-vs-library configuration responsibility" in 24cc85d2 (reviewer input, collected for stabilization, not an agreed decision): the direct-app vs library (ECS) usage split, the three ownership models (library owns all / app owns finished client / mixed), the preference for the mixed model, and the obstacle that library-relevant knobs like connection lifetime do not fall cleanly on the pipeline side and need a transport abstraction richer than a plain layered/Tower service. Cross-referenced item 3.

crates/fetch/docs/stabilization.md
sandersaares: [Copilot speaking] The fetch_core/fetch split idea (keeping fetch an enabler with reasonable defaults rather than a monolithic all-in-one crate, with libraries depending only on fetch_core and applications on fetch) has been integrated into stabilization.md item 1 in 24cc85d2, captured as collected reviewer input for the stabilization phase rather than an agreed decision.

crates/fetch_winhttp/docs/implementation.md
```diff
@@ -0,0 +1,1164 @@
...
+completion callback" - not "blocks on I/O". The setup calls the transport makes
+(`WinHttpConnect`, `WinHttpOpenRequest`, the `WinHttpSetOption`/`WinHttpSetTimeouts`
+family, `WinHttpQueryHeaders`) do no network, DNS, or socket work; they allocate and
+configure handles, deferring all I/O to the async steps. They are therefore safe to
+run inline on an executor thread, and the transport needs **no** `anyspawn::Spawner`
+and no blocking pool.
+
+The sole exception is the very first `WinHttpOpen` in a process, which runs
+WinHTTP's one-time global initialization (a lock plus registry reads) and can
+briefly block. Since the transport opens its single session once at build time
```
martin-kolinek: 🤖: Stale after the session-ownership change made in this revision. §3.2 (around line 246) now opens a session lazily inside the per-core factory when etch materializes the instance — one session per core, explicitly not a single session at build time. This line still says the transport `opens its single session once at build time (§3.2)`, contradicting the very section it cites. (WinHTTP's one-time global init still runs on the first WinHttpOpen, but that is now the first per-core session open, not a build-time single-session open.) Please reword to match the per-core factory model.

crates/fetch_winhttp/docs/implementation.md
```diff
@@ -0,0 +1,1164 @@
...
+keyed by authority, independent of connect-handle identity (§9.1), so opening a
+fresh connect handle per request costs nothing in reuse terms.
+
+**We do set `WINHTTP_OPTION_DISABLE_GLOBAL_POOLING` on the session at creation.**
+By default WinHTTP may share pooled connections process-wide across sessions. That
+would let two independent `fetch_winhttp` clients - for instance a strict one and
+one built with `accept_invalid_certs` (design.md §4) - reuse each other's pooled
+connections, collapsing the security boundary between them. Disabling global
+pooling scopes the connection pool to this session, so each `HttpClient` gets its
+own pool (still one pool per client, shared across that client's cores, §3.2)
```
martin-kolinek: 🤖: Stale after the same session-ownership flip. §3.2 now gives a client one connection pool **per core**, but this parenthetical still describes the superseded model — `still one pool per client, shared across that client's cores, §3.2` — the opposite of what §3.2 now states (and it cites §3.2). Inter-client isolation via DISABLE_GLOBAL_POOLING is still the point here, but the mechanism is now per-core pools; please reword to per-core pools.

crates/fetch_winhttp/docs/implementation.md
```diff
@@ -0,0 +1,1164 @@
...
+### 7.3 Integration tests (real WinHTTP, localhost)
+
+Gated behind `#[cfg(windows)]` and `#[cfg_attr(miri, ignore)]`, against a
+localhost server (a small `std::net`-based server, or `wiremock` as used
+elsewhere in `fetch`). These validate the real OS path end to end:
+
+- GET/POST with small and large bodies; response body correctness and size.
+- Streaming upload (unknown length -> chunked) and streaming download; assert
+  incremental delivery, not just final bytes.
+- Real gzip/deflate responses are transparently decoded.
```
martin-kolinek, line 784: 🤖: This integration plan validates TLS and pooling end-to-end but never proves that REDIRECT_POLICY_NEVER and DISABLE_COOKIES (§5, set around lines 728 / 1038-1044) actually override WinHTTP's OS defaults — the mock plan only checks the options are passed. Please add a localhost scenario where a 302 points at a sentinel endpoint (assert the 3xx is surfaced unchanged and the sentinel is never hit), and one where a response sets a cookie and a later request confirms it is not replayed. This catches setting these options at the wrong handle scope, or WinHTTP ignoring the intended configuration.

martin-kolinek (approved):

crates/fetch/docs/stabilization.md
martin-kolinek: For future reference, here's the output of a copilot session where this was explored: https://github.com/microsoft/oxidizer/blob/u/makolnek/winhttp/crates/fetch_winhttp/docs/design/builder-architecture.md

martin-kolinek (approved):

crates/fetch_winhttp/docs/implementation.md
sandersaares: [Copilot speaking] Fixed in 38b99d19. The §2.1 note now reads that each per-core transport instance opens its own session lazily inside the factory when fetch materializes it (§3.2), so the WinHttpOpen cost is one-time per core off the request path and the process-wide global init runs only on the first such open - no longer claiming a single build-time session.

crates/fetch_winhttp/docs/implementation.md
sandersaares: [Copilot speaking] Fixed in 38b99d19. The §9.3 DISABLE_GLOBAL_POOLING parenthetical now reads "one pool per core under the per-core session model (§3.2)" instead of one pool per client shared across cores. Inter-client isolation via DISABLE_GLOBAL_POOLING is still the point; the mechanism is now correctly per-core pools.

crates/fetch_winhttp/docs/implementation.md
sandersaares: [Copilot speaking] Added in 38b99d19. §7.3 now has two localhost scenarios: a 302 whose Location points at a sentinel endpoint asserts the 3xx and Location are surfaced unchanged and the sentinel is never hit (server-side counter stays zero), proving REDIRECT_POLICY_NEVER overrides the default and is set at the right handle scope; and a Set-Cookie response followed by a second request that asserts no Cookie header is replayed (verified server-side), proving DISABLE_COOKIES. These validate behavior end-to-end, not just that the options are passed.

martintmk (approved):

crates/fetch_winhttp/docs/implementation.md
```diff
@@ -0,0 +1,1203 @@
...
+  revisit whether one shared session per client is the better default. Tracked as `fetch`
+  API feedback (../../fetch/docs/stabilization.md, connection-management item) so the
+  divergence is not forgotten when that API becomes more expressive.
+- **Per-core vs shared instancing as a knob.** Whether per-core instancing is the right
+  default at all could become configurable: a low-traffic client gains nothing from
+  per-core instances and might prefer a single shared instance (§3.2). Left unconfigurable
+  in v1 - a knob earns its place only with demonstrated value - but a candidate if
+  profiling shows it matters.
+- **Session-keyed connection pools (`PoolIndex`).** v1 leans on WinHTTP's own per-authority
+  pooling and ignores `fetch`'s `PoolIndex`, collapsing nominally separate pools onto one
```
martin-kolinek: 🤖: This "collapsing nominally separate pools onto one OS pool" claim (also in §8 around line 883, and mirrored in fetch/docs/stabilization.md item 3) is now inconsistent with the per-invocation session model this revision adopted. `fetch`'s `create_dispatch_handler` invokes the transport factory once **per pool slot** (`0..pool_count` in `client_builder.rs`), and each invocation opens a fresh WinHTTP session (§3.2). Because pooling is per-session (`DISABLE_GLOBAL_POOLING`, §9.3 — "one pool per core"), distinct `PoolIndex` slots already land in **distinct** sessions/pools rather than collapsing onto one. So ignoring `PoolIndex` does not collapse `multiple_pools` onto a single OS pool; the real v1 topology is one session/pool per (core × pool slot). Please reword this future item, §8, and stabilization.md item 3 to describe the `cores × pool slots` resource profile instead of a single-pool collapse.

sandersaares: [Copilot speaking] Verified and fixed in b733bc62. Confirmed against client_builder.rs: create_dispatch_handler invokes create_transport_handler once per pool slot (0..pool_count) with a distinct PoolIndex, and custom.rs shows the factory is a PerCore thread-aware Arc. Each slot therefore opens its own WinHTTP session (§3.2), and with per-session DISABLE_GLOBAL_POOLING (§9.3) distinct slots already hold distinct pools. Reworded implementation.md §8, the chapter 14 future item, and fetch/docs/stabilization.md item 3 to state that ignoring the PoolIndex value does not collapse multiple_pools; the real v1 topology is one session/pool per (core × pool slot). design.md carries no such claim, so no change there.

crates/fetch_winhttp/docs/implementation.md
```diff
@@ -0,0 +1,1210 @@
...
+transport instance, from the finalized `CustomContext`, not eagerly in `builder_winhttp`.
+This is deliberate: `HttpClientBuilder` is `Clone`, so a session opened up front and
+captured in the (clone-shared) factory closure would be shared by every client built from
+that builder or any clone of it, letting two independently built clients reuse each
+other's pooled connections - violating the pool-isolation boundary of design.md §2.
+Opening it inside the factory (exactly as the Tokio transport builds its hyper client in
+its own factory) scopes the session to the built client: two independently built clients -
+including two builds of a cloned builder - never share a session or its pool.
+
+Under `Isolation::Isolated` the factory runs once per core, so each core opens its own
```
martin-kolinek: 🤖: §8/§14 and stabilization.md now describe the v1 topology as one session/pool per (core × pool slot), but this §3.2 lifecycle summary still frames it as once per core — "the factory runs once per core, so each core opens its own session and a client holds one connection pool per core." Under `Isolation::Isolated`, each core builds a dispatch handler for every configured `multiple_pools` slot (`0..pool_count`), so the actual profile is one session/pool per (core × pool slot), one slot being the default. Please reconcile §3.2 (and any architecture-overview / §9.3 mirrors that still say per-core) with the (core × pool slot) profile the corrected §8 now states.

crates/fetch_winhttp/docs/implementation.md
```diff
@@ -0,0 +1,1210 @@
...
+
+`CustomContext` hands the factory a `HttpBodyBuilder` (carrying the clock and
+read-buffer pool), a `PoolIndex`, the generic `TransportOptions`/`TlsOptions`, a
+`Meter`, and the caller's `Extras`. `fetch_winhttp` ignores `PoolIndex` (per-core
+placement comes from `Isolation::Isolated`, §3.2) and ignores `CustomContext::tls` (it
+takes its own `WinHttpTlsConfig` instead; see design.md §1.2). Ignoring the `PoolIndex`
+*value* does not collapse `fetch`'s `multiple_pools`: `fetch` invokes the factory once per
+pool slot (`0..pool_count` in `client_builder.rs`), so each slot opens its own WinHTTP
+session (§3.2), and because pooling is per-session (`DISABLE_GLOBAL_POOLING`, §9.3) those
+sessions already hold distinct pools. Distinct `PoolIndex` slots therefore land in distinct
```
martin-kolinek, line 896: 🤖: The corrected design now leans on a behavioural invariant the test plan does not exercise: each `multiple_pools` slot must get a distinct WinHTTP session/pool even though the transport ignores the `PoolIndex` *value*. The §7.3 pool-isolation test only compares two independently built clients, not two slots of one client. Please add a Windows scenario that configures ≥2 pool slots, routes requests through both, and asserts via server-side connection counting that a connection cannot be reused across slots (optionally paired with a mock assertion that `WinHttpOpen` runs once per slot).

crates/fetch_winhttp/docs/implementation.md
sandersaares: [Copilot speaking] Fixed in 305a4089. Reconciled the §3.2 lifecycle summary, the architecture-at-a-glance bullets, and the §9.3 pooling note (plus the §14 consolidation item) with the corrected §8: the factory runs once per core and once per configured multiple_pools slot (0..pool_count, a single slot by default), so the profile is one session/pool per (core × pool slot), reducing to one pool per core in the default single-slot case.

crates/fetch_winhttp/docs/implementation.md
sandersaares: [Copilot speaking] Added in 305a4089. §7.3 now has a within-client pool-slot scenario: a single HttpClient built with multiple_pools(2) routes requests through both slots to the same authority and asserts via server-side connection counting that a connection opened for one slot is never reused by the other - proving each slot lands in its own session/pool despite the transport ignoring the PoolIndex value - optionally paired with a mock assertion that WinHttpOpen runs once per slot.

crates/fetch_winhttp/docs/implementation.md
```diff
@@ -0,0 +1,1221 @@
...
+
+This is the whole design in one picture; the numbered chapters below elaborate each part.
+
+- **One OS session per (core × pool slot), scoped to the built client.** Each per-core
+  transport instance opens its own session - one per configured `multiple_pools` slot,
+  a single slot by default - when `fetch` materializes it, so two independently
+  built clients (or clones) never share a session or its connection pool (§3.2).
+- **One transport instance per core.** `fetch` clones and relocates the transport
+  per core (`Isolation::Isolated`, §3.2); each instance owns its object and event
+  pools (§5) and one session `Arc` per pool slot.
```
martin-kolinek, line 15: 🤖: The overview still describes the wrong object model. Under `multiple_pools`, `create_dispatch_handler` invokes the factory once per pool slot (`(0..pool_count).map(|index| create_transport_handler(..., PoolIndex::new(index)))` in `client_builder.rs`), and each invocation returns a *distinct* handler. So the runtime shape is one `WinHttpTransport` per `(core × pool slot)`, each owning exactly one session - not a single per-core transport instance that owns "one session `Arc` per pool slot". §3.2 already says the factory runs once per slot, so this bullet contradicts it. Please align these overview bullets (and any §3.2/§5 mirrors that inherit the single-transport framing) to model one transport/session/pool set per `(core × pool slot)`.

sandersaares: Deferring further alignment to next PR.

crates/fetch_winhttp/docs/design.md
```diff
@@ -0,0 +1,352 @@
...
+#[derive(thread_aware::ThreadAware)]
+#[non_exhaustive]
+pub struct WinHttpDeps { /* tls, options, sink - private; set via the builder */ }
+
+impl WinHttpDeps {
+    /// Starts building a `WinHttpDeps`. `tls`/`options`/`sink` default when unset.
+    pub fn builder() -> WinHttpDepsBuilder;
+}
+
+/// Adds WinHTTP-transport constructors to `fetch::HttpClient`.
```
psandana, line 76: [design] `fetch_winhttp` silently ignores a caller's `fetch::TlsOptions` (private CA root, client identity, cert pinning) with no warning — even though this design already implements a "warn on unhonorable option" mechanism (implementation.md §11) for less security-sensitive things like `connection_lifetime`. Since stabilization.md item 2 itself calls this a "silent-ignore hazard," could the existing §11 warn policy be extended now to cover a non-default `TlsOptions` (detect `CustomContext::tls` is populated and emit the same build-time warning), rather than waiting on the broader fetch-level TLS redesign? This seems like a self-contained, low-risk addition that closes a gap the PR's own docs flag.

crates/fetch_winhttp/docs/design.md
```diff
@@ -0,0 +1,352 @@
...
+use fetch_winhttp::{HttpClientWinHttpExt, WinHttpDeps, WinHttpOptions, WinHttpTlsConfig};
+
+// Defaults:
+let client = HttpClient::new_winhttp();
+
+// Configured. Every WinHTTP config type is built through a builder, never a struct
+// literal, so all of them are `#[non_exhaustive]` and can gain knobs in later versions
+// without a breaking change:
+let deps = WinHttpDeps::builder()
+    .tls(WinHttpTlsConfig::builder()
```
psandana, line 46: [design] This ships `HttpClientWinHttpExt::builder_winhttp`/`new_winhttp` even though stabilization.md item 1 calls this exact extension-trait shape "a workaround, not a fix" for the asymmetry with `HttpClient::builder_tokio`/`new_tokio`. Since design.md is what downstream consumers read for construction, could it include a forward-reference/stability note here (e.g. "this constructor shape may change before 1.0 — see stabilization.md item 1") so early adopters aren't surprised if it changes?

crates/fetch/docs/stabilization.md
```diff
@@ -0,0 +1,205 @@
...
+`HttpClient::build(fetch_winhttp::transport().tls(cfg))` - would be more predictable,
+avoid a Hyper-colored default, and treat every transport uniformly. The slightly more
+verbose hello-world is worth the consistency.
+
+**Reviewer input (collected for stabilization, not a decision).** A counter-view holds
+that `fetch` is deliberately *batteries-included* and so should *play favorites*: pick a
+recommended default transport per supported runtime and keep its hello-world maximally
+terse, so the 90% of users who just want the default runtime pay no extra ceremony, while
+custom transports remain possible without degrading the default ergonomics. One shape
+suggested for this: `HttpClient::builder()` returns `HttpClientBuilder<NeedsTransport>`
```
psandana, line 47: [design] The "reviewer input" here proposes splitting a `fetch_core` crate out of `fetch` and questions whether the whole downstream-transport extensibility point should exist at all — a broad, unrelated architectural redesign of the `fetch` crate embedded inside this Windows-transport design PR. Even labeled "not agreed decisions," baking this into a merged doc risks misattributed git history and can pull review of *this* PR toward relitigating fetch's whole architecture. Would it be better to move this kind of open-ended brainstorming into a tracking issue/discussion, keeping stabilization.md limited to the concrete, PR-motivated feedback items?

crates/fetch_winhttp/docs/implementation.md
```diff
@@ -0,0 +1,1221 @@
...
+instead of the canonical `HttpError::timeout`. Second, it is a liveness backstop.
+The `fetch`-level response and body timeouts are futures driven by the caller's
+async executor; if that executor stalls, those timeouts cannot fire. The native
+timers run on WinHTTP's own threads, independent of the caller's executor, so a
+hung network read is always aborted at the OS level and its socket released,
+regardless of executor liveness. In the normal case the `fetch`-level timeout
+still fires first and reports the canonical error; the native timer only bites
+when the upper layer cannot.
+
+## 11. Handling options the transport cannot honor
```
psandana, line 1122: [design] The unhonorable-option signal here is log/metric-only — there's no described way for a caller to assert deterministically at build time that no option was silently downgraded (e.g. in a startup health-check or test). Worth considering also exposing this as inspectable data, e.g. `HttpClientBuilder::unhonored_options() -> &[UnhonoredOption]`, so callers/tests don't have to depend on log/metric plumbing being wired up and monitored.

crates/fetch_winhttp/docs/design.md
```diff
@@ -0,0 +1,352 @@
...
+  already governs, after which `BodyTimeout` takes over; there is no separate send
+  deadline to honor.
+- **Resolve timeout**: `fetch` has no concept for a standalone DNS-resolution deadline
+  (it is otherwise subsumed by `connect_timeout`), so it is exposed as a transport-specific
+  `WinHttpOptions` knob for callers that need to bound resolution finely, as discussed in
+  the fetch API stabilization feedback (../../fetch/docs/stabilization.md).
+
+Cancellation is the transport's backstop for every timeout it does not enforce natively:
+when `fetch` (or any layer above) drops the request future on a timeout, the transport
+honors it by closing the WinHTTP handle and tearing the request down (the drop-safety
```
psandana, line 281: [doc-consistency] This cites "the drop-safety contract in §7 and implementation.md §4", but design.md's own §7 is "Error handling model" (Win32 error mapping / retry classification) — it doesn't define a drop-safety contract. Only the `implementation.md §4` half of this reference is correct. Suggest dropping the stray `§7` reference:

```suggestion
honors it by closing the WinHTTP handle and tearing the request down (the drop-safety contract is in implementation.md §4).
```

crates/fetch_winhttp/docs/implementation.md
```diff
@@ -0,0 +1,1221 @@
...
+  exclusivity does **not** rely on any undocumented per-handle callback
+  serialization; it follows from "one completion per op" plus "one op outstanding".
+- The `events_once` send-then-receive is the release/acquire edge that transfers
+  buffer ownership back to the driver. Only after the receiver resolves does the
+  driver read the returned buffer.
+
+This is exactly the "WinHTTP takes exclusive ownership via a leaked pointer, we
+recover it at the callback" model: the leaked pointer *is* the ownership token,
+and the two sides never hold it at the same time. No lock is needed on the
+sender/buffer fields; the temporal hand-off does the work.
```
psandana, line 505: [language] Inconsistent spelling of "handoff" (used at lines 216, 286, 432) vs "hand-off" (used here and at 507, 516) for the same callback→future concept. Suggest picking one spelling and using it consistently throughout the document.

crates/fetch_winhttp/docs/design.md
```diff
@@ -0,0 +1,352 @@
...
+
+## 5. WinHTTP-managed HTTP behavior
+
+The OS handles several HTTP behaviors internally. The transport configures each so it
+behaves consistently with the rest of `fetch`:
+
+- **Automatic decompression (always on).** The transport advertises
+  `Accept-Encoding: gzip, deflate`; gzip/deflate responses are transparently decoded
+  before the body is streamed up, with `Content-Encoding`/`Content-Length` stripped, so
+  callers always see a decoded body. `fetch` itself has no content decoding, so there is
```
psandana, line 225: [language] "streamed up" reads awkwardly here; "streamed out"/"streamed back" would match the verb used elsewhere in the doc (e.g. §6.2's "the body is streamed afterward").

```suggestion
gzip/deflate responses are transparently decoded before the body is streamed out, with `Content-Encoding`/`Content-Length` stripped, so callers always see a decoded body.
```

psandana (commented):
Design review notes from a walkthrough of this design PR (fetch_winhttp docs). See inline comments.

crates/fetch_winhttp/docs/design.md
sandersaares: Deferring - will follow-up in next PR together with implementation.

crates/fetch_winhttp/docs/design.md
sandersaares: Deferring - will follow-up in next PR together with implementation.

crates/fetch/docs/stabilization.md
sandersaares: Deferring - will follow-up in next PR together with implementation.

crates/fetch_winhttp/docs/implementation.md
sandersaares: Deferring - will follow-up in next PR together with implementation.

crates/fetch_winhttp/docs/design.md
sandersaares: Deferring - will follow-up in next PR together with implementation.

crates/fetch_winhttp/docs/implementation.md
sandersaares: Deferring - will follow-up in next PR together with implementation.

crates/fetch_winhttp/docs/design.md
sandersaares: Deferring - will follow-up in next PR together with implementation.

sandersaares:
Going to merge this now - addressing some comments deferred to next phase (implementation).

martin-kolinek (approved):

PR #596, opened 2026-07-22
ci(anvil): update to cargo-anvil 0.3.0

st-dev-gh:
## Summary

- Regenerate Anvil-managed workflows, actions, and Just recipes with cargo-anvil 0.3.0  
  see microsoft/ox-tools#60


crates/cachet/examples/stampede_protection.rs
```diff
@@ -23,16 +24,16 @@ impl CacheTier<String, String> for SlowBackend {
         Ok(Some(CacheEntry::new(format!("value_for_{key}"))))
     }
 
-    async fn insert(&self, _: String, _: CacheEntry<String>) -> Result<(), Error> {
-        Ok(())
+    fn insert(&self, _: String, _: CacheEntry<String>) -> impl Future<Output = Result<(), Error>> + Send {
```
martin-kolinek: What is the reason for these changes? I'd assume we can use async fn to implement these.

st-dev-gh: Recent clippy introduced a lint about async fn not using awaitL: https://rust-lang.github.io/rust-clippy/master/index.html#unused_async_trait_impl

But first, the version run on server does not know it yet (hence pr-fast failing)
second, I probably rushed too much into changing those functions.

ralfbiedert: Is the expectation that if anvil is updated it also applies all new refactorings that come with new rules?

st-dev-gh: That is a great question. In this case it is not really the question of anvil update, but rather what version of clippy we use. And ideally how to use the same locally and on the server.

st-dev-gh: As far as I can see, clippy is part of the toolchain and anvil deliberately leaves that at
> msrustup owns it on ADO/1ESPT, the runner image owns it on GH, the user owns it locally
https://github.com/microsoft/ox-tools/blob/main/crates/cargo-anvil/docs/design/local.md#35-the-rust-toolchain

.github/actions/anvil-setup/action.yml
```diff
@@ -18,15 +20,61 @@ inputs:
...
 runs:
   using: composite
   steps:
+    # Standard GitHub-hosted runners guarantee only 14 GB of SSD space. Heavy
+    # build and test jobs can exceed that with target artifacts alone, so this
+    # opt-in cleanup removes toolchains unused by anvil's Rust checks.
+    # See https://github.com/microsoft/oxidizer/pull/583.
+    - name: Free disk space (Linux)
+      if: inputs.free-disk-space == 'true' && runner.environment == 'github-hosted' && runner.os == 'Linux'
+      shell: bash
```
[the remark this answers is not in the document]
st-dev-gh: runner.environment is documented in the GitHub Actions runner context and reports either github-hosted or self-hosted: https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#runner-context

The condition therefore correctly limits cleanup to GitHub-hosted runners.

The same was already asked and answered https://github.com/microsoft/ox-tools/pull/60#discussion_r3624486320

psandana (approved):
Reviewed: mechanical regeneration via cargo-anvil 0.3.0. Verified all pinned Action SHAs resolve to their claimed tags (no supply-chain tampering). No logic concerns found.

justfiles/anvil/checks/semver-check.just
```diff
@@ -31,9 +37,13 @@
...
-# transitions ("no library targets found").
+# set, verify the PR branch baseline is present, skip newly-added
+# manifest paths, then invoke per-crate against that baseline with
+# selective error tolerance for renamed crates ("package `<name>` not
+# found in ...") and bin->bin+lib transitions ("no library targets
+# found").
+
+# Check affected crates for semantic versioning violations.
 [script("pwsh")]
 anvil-semver-check: anvil-semver-check-validate-prereqs
```
Vaiz, line 48: is it possible to move the script inside of cargo-anvil and invoke cargo-anvil instead?

Vaiz (approved):

justfiles/anvil/checks/semver-check.just
st-dev-gh: `cargo-anvil` is a one-off action to copy/update the templates from the crate to the target repository.

The tests are then invoked as `just anvil-...`, .e.g `just anvil-semver-check`. `anvil-semver-check` is also part of `anvil-pr-fast`, which in turn is part of `anvil-pr`.
See https://github.com/microsoft/ox-tools/blob/main/crates/cargo-anvil/docs/design/checks.md for more details.

crates/cachet/examples/stampede_protection.rs
st-dev-gh: > Is the expectation that if anvil is updated it also applies all new refactorings that come with new rules?

I reckon it is up to a repository owner to decide how to deal with new recipes/checks. It may make sense to address them right away or it may be more reasonable to just mark them to be addressed later.

But again, the problem here was not the cargo-anvil itself, but rather that different toolchain versions possibly bring new clippy lints.

PR #600, opened 2026-07-23
feat(rest_over_grpc): add REST transcoding for gRPC services

geeknoid:
## Summary

- generate typed REST/JSON routers and service dispatch from `google.api.http` descriptors
- support unary and server-streaming responses, query/body/path binding, status mapping, metadata, OpenAPI, and Tower, Layered, and Axum adapters
- include end-to-end tests, benchmarks, worked examples, and comprehensive API documentation

## Validation

- `cargo test -p rest_over_grpc -p rest_over_grpc_examples --all-features --all-targets --locked`
- `cargo clippy -p rest_over_grpc -p rest_over_grpc_examples --all-features --all-targets --locked -- -D warnings`
- `cargo hack -p rest_over_grpc check --feature-powerset --locked`
- strict rustdoc, doctests, formatting, spelling, README, and manifest checks


crates/rest_over_grpc/src/http_response.rs
```diff
@@ -0,0 +1,368 @@
...
+            status,
+            content_type,
+            mut headers,
+            body,
+        } = self;
+        let Ok(content_type) = HeaderValue::from_maybe_shared(Bytes::from_static(content_type.as_bytes())) else {
+            let mut fallback = Response::new(Vec::new());
+            *fallback.status_mut() = status;
+            return fallback;
+        };
```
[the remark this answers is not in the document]
geeknoid: Addressed in the amended commit: response content types are now stored as validated `HeaderValue`s, so `into_http` no longer has an invalid-content-type fallback and always preserves the status, body, and custom headers.

crates/rest_over_grpc/src/build/response_body.rs
```diff
@@ -0,0 +1,49 @@
...
+}
+
+impl fmt::Display for ResponseBody {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Whole => f.write_str("*"),
+            Self::Field(field) => f.write_str(field),
+        }
+    }
+}
```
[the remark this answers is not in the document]
geeknoid: Fixed in c02b2076: `ResponseBody::Whole` now displays as the empty string, matching an unset `HttpRule.response_body`, with the docs and regression test updated.

crates/rest_over_grpc/src/build/request_body.rs
```diff
@@ -0,0 +1,55 @@
...
+/// ```
+/// use rest_over_grpc::build::RequestBody;
+///
+/// // Defaults to consuming no body.
+/// assert_eq!(RequestBody::default(), RequestBody::None);
+///
+/// // `Display` renders the `HttpRule.body` value.
+/// assert_eq!(RequestBody::None.to_string(), "none");
+/// assert_eq!(RequestBody::Whole.to_string(), "*");
+/// assert_eq!(RequestBody::Field("book".to_owned()).to_string(), "book");
```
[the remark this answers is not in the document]
geeknoid: Fixed in c02b2076: `RequestBody::None` now displays as the empty string, matching an unset `HttpRule.body`, with the docs and regression test updated.

crates/rest_over_grpc/src/transcode/error.rs
```diff
@@ -0,0 +1,229 @@
...
+        Status::new(self.code(), format!("{}: {}", self.summary(), self.detail))
+    }
+
+    /// The human-readable summary for this error's kind.
+    const fn summary(&self) -> &'static str {
+        match self.kind {
+            TranscodeErrorKind::Body => "invalid request body JSON",
+            TranscodeErrorKind::Structure => "invalid request structure",
+            TranscodeErrorKind::Deserialize => "request did not match the expected message",
+            TranscodeErrorKind::Serialize => "failed to serialize the response message",
```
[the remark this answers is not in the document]
geeknoid: Fixed in c02b2076: serialize-kind failures now use the broader summary `failed to encode the response`, covering both serialization failures and response-shape mismatches. Regression coverage includes an absent selected field.

crates/rest_over_grpc/src/status.rs
```diff
@@ -0,0 +1,222 @@
...
+    pub fn details(&self) -> &[Value] {
+        &self.details
+    }
+}
+
+impl fmt::Display for Status {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:?}: {}", self.code, self.message)
+    }
+}
```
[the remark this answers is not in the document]
geeknoid: Fixed in c02b2076: `Status` now formats `Code` through its `Display` implementation, producing canonical names such as `NOT_FOUND`. Added an exact formatting regression test.

crates/rest_over_grpc/src/transcode_api.rs
```diff
@@ -0,0 +1,141 @@
...
+            self.try_transcode(method, target, headers, body)
+                .await
+                .unwrap_or_else(|| TranscodeResponse::Unary(HttpResponse::not_found()))
+        }
+    }
+}
+
+/// Forwards to the wrapped transcoder, so an `Arc`-shared `Transcoder` (cloned
+/// cheaply into an adapter) still implements [`Transcode`].
+impl<D: Transcode + Send> Transcode for std::sync::Arc<D> {
```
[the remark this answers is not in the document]
geeknoid: The bound is required by the trait hierarchy, so I retained it. `Transcode` has a `Sync` supertrait; for `Arc<D>` itself to satisfy `Sync`, Rust requires `D: Send + Sync`. `D: Transcode` supplies `Sync` but not `Send`, and removing `Send` fails to compile at the impl with E0277. The existing bound is therefore the minimum valid bound.

crates/rest_over_grpc/src/status.rs
```diff
@@ -0,0 +1,227 @@
...
+    #[must_use]
+    pub fn new(code: Code, message: impl Into<String>) -> Self {
+        Self {
+            code,
+            message: message.into(),
+            details: Vec::new(),
+        }
+    }
+
+    /// Creates an [`Code::InvalidArgument`] status.
```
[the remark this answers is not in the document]
geeknoid: Fixed in ffc1498f: the documentation now reads “Creates a `Code::InvalidArgument` status.”

crates/rest_over_grpc/CHANGELOG.md
```diff
@@ -0,0 +1,29 @@
...
+  generated dispatchers.
+- `transcode` module: serde-based JSON⇄message request/response transcoding
+  (`decode_request`, `encode_response`, `BodyKind`, `ResponseBodyKind`,
+  `status_response`, `not_found_response`), composing with `pbjson`-generated
+  proto3-canonical serde.
+- Query-string helpers `split_query` and `parse_query`.
+- Web-stack adapters: `HttpResponse::into_http`, `adapter::transcode_http`, and
+  `adapter::RestService`, bridging the neutral dispatcher to the `http`/`http-body`
+  ecosystem. `RestService` implements both `tower::Service` (feature `tower`) and
+  `layered::Service` (feature `layered`) over the same dispatcher.
```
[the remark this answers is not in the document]
geeknoid: Fixed in ffc1498f: the initial release notes now name the actual code/status methods, build types, transcoding primitives, response types, serving adapters, and runnable example categories.

crates/rest_over_grpc_examples/examples/handling/client_streaming_upload.rs
```diff
@@ -0,0 +1,155 @@
...
+    // and buffer the reply.
+    let method = request.method().clone();
+    let target = request.uri().path_and_query().map_or("/", |pq| pq.as_str()).to_owned();
+    let headers = request.headers().clone();
+    let bytes = request
+        .into_body()
+        .collect()
+        .await
+        .map(http_body_util::Collected::to_bytes)
+        .unwrap_or_default();
```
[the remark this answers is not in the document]
geeknoid: Fixed in 1ca92f3b: unary request-body collection now handles stream errors explicitly and returns `400 Bad Request` instead of transcoding an empty payload. Added a regression test using a body that yields a read error.

crates/rest_over_grpc_examples/examples/serving/custom_body_handling.rs
```diff
@@ -0,0 +1,142 @@
...
+
+/// Collects a body, failing if it exceeds `max`.
+///
+/// A real server reading a streaming body would track the running total and bail
+/// as soon as it crosses `max`; a [`Full`] body is already in memory, so here we
+/// simply check the collected length.
+async fn read_capped(body: Full<Bytes>, max: usize) -> Result<Bytes, TooLarge> {
+    let bytes = body.collect().await.map(http_body_util::Collected::to_bytes).unwrap_or_default();
+    if bytes.len() > max { Err(TooLarge) } else { Ok(bytes) }
+}
```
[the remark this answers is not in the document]
geeknoid: Fixed in 1ca92f3b: `read_capped` now uses an explanatory `expect`, documenting that `Full<Bytes>` cannot fail because its body error type is `Infallible`, rather than silently substituting empty bytes.

crates/rest_over_grpc/README.md
```diff
@@ -0,0 +1,261 @@
+<div align="center">
+ <img src="./logo.png" alt="Rest Over Grpc Logo" width="96">
```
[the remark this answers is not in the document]
geeknoid: Fixed in f8694c7a at the generator source: the shared README template now uses the absolute repository logo URL, and all generated crate READMEs were refreshed. This avoids relying on unshipped Git LFS assets.

crates/rest_over_grpc_examples/build.rs
```diff
@@ -0,0 +1,124 @@
...
+fn main() {
+    let manifest = PathBuf::from(env::var("CARGO_MANIFEST_DIR").expect("CARGO_MANIFEST_DIR is set for build scripts"));
+    let out_dir = PathBuf::from(env::var("OUT_DIR").expect("OUT_DIR is set for build scripts"));
+    let proto_dir = manifest.join("proto");
+
+    build_tonic_bridge(&proto_dir, &out_dir.join("tonic_bridge"));
+    build_custom(&proto_dir, &out_dir.join("custom"));
+
+    println!("cargo:rerun-if-changed=proto/library.proto");
+    println!("cargo:rerun-if-changed=build.rs");
```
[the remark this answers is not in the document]
geeknoid: Fixed in f8694c7a: the examples build script now emits explicit `rerun-if-changed` directives for both imported Google API proto files.

crates/rest_over_grpc_tests/build.rs
```diff
@@ -0,0 +1,154 @@
...
+    // they must not share a directory). These mirror the `rest_over_grpc_examples`
+    // crate but are generated here so this crate stays independent of it.
+    let proto_dir = manifest.join("proto");
+    build_tonic_bridge(&proto_dir, &out_dir.join("tonic_bridge"));
+    build_custom(&proto_dir, &out_dir.join("custom"));
+
+    println!("cargo:rerun-if-changed=bench_routes.rs");
+    println!("cargo:rerun-if-changed=proto/greeter.proto");
+    println!("cargo:rerun-if-changed=proto/library.proto");
+    println!("cargo:rerun-if-changed=build.rs");
```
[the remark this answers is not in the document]
geeknoid: Fixed in f8694c7a: the test fixture build script now emits explicit `rerun-if-changed` directives for both `proto/google/api/annotations.proto` and `proto/google/api/http.proto`.

crates/rest_over_grpc/README.md
```diff
@@ -0,0 +1,261 @@
...
+
+# Rest Over Grpc
+
+[![crate.io](https://img.shields.io/crates/v/rest_over_grpc.svg)](https://crates.io/crates/rest_over_grpc)
+[![docs.rs](https://docs.rs/rest_over_grpc/badge.svg)](https://docs.rs/rest_over_grpc)
+[![MSRV](https://img.shields.io/crates/msrv/rest_over_grpc)](https://crates.io/crates/rest_over_grpc)
+[![CI](https://github.com/microsoft/oxidizer/actions/workflows/main.yml/badge.svg?event=push)](https://github.com/microsoft/oxidizer/actions/workflows/main.yml)
+[![Coverage](https://codecov.io/gh/microsoft/oxidizer/graph/badge.svg?token=FCUG0EL5TI)](https://codecov.io/gh/microsoft/oxidizer)
+[![License](https://img.shields.io/badge/license-MIT-blue.svg)](../../LICENSE)
+<a href="../.."><img src="../../logo.svg" alt="This crate was developed as part of the Oxidizer project" width="20"></a>
```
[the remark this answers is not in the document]
geeknoid: Fixed in f8694c7a at the shared README template: project/logo/footer links are absolute, and the license badge now targets the repository LICENSE directly. (`cargo package --list` confirms these crates currently use SPDX metadata and do not ship a package-local LICENSE file.) All generated crate READMEs were refreshed.

crates/rest_over_grpc/src/stream.rs
```diff
@@ -0,0 +1,446 @@
...
+        StreamEncoding::JsonArray => frame.push(if first { b'[' } else { b',' }),
+        StreamEncoding::NdJson => {}
+        StreamEncoding::Sse => frame.extend_from_slice(b"data: "),
+    }
+    match response_body {
+        ResponseBodyKind::Whole => {
+            to_writer(&mut frame, message).map_err(|error| Status::internal(format!("failed to serialize a streamed message: {error}")))?;
+        }
+        ResponseBodyKind::Field(_) => {
+            let body = encode_response(message, response_body).map_err(TranscodeError::into_status)?;
```
psandana: [perf] `ResponseBodyKind::Field` allocates a fresh `Vec<u8>` via `encode_response` and then copies it into `frame` with `extend_from_slice`, unlike the `Whole` arm which serializes straight into `frame` via `to_writer`. On a long-lived server-streaming RPC using field-selected `response_body`, this pays an extra heap allocation + memcpy per frame. Consider a `FieldSerializer`-based helper that writes directly into the caller's `frame` buffer to avoid the per-frame allocation on this hot path.

crates/rest_over_grpc/src/transcode/overlay.rs
```diff
@@ -0,0 +1,812 @@
...
+
+    let mut body_entries = match body {
+        Some(bytes) if !bytes.is_empty() => match from_slice::<BodyTop<'_>>(bytes) {
+            Ok(top) => top.entries,
+            Err(_) => return Err(classify_body_error(bytes)),
+        },
+        _ => Vec::new(),
+    };
+
+    body_entries.retain(|(key, _)| !overlay.iter().any(|(overlay_key, _)| *overlay_key == key.as_str()));
```
psandana: [perf] `decode_flat`'s `body_entries.retain(|(key, _)| !overlay.iter().any(|(overlay_key, _)| *overlay_key == key.as_str()))` is O(body_entries.len() * overlay.len()), and `upsert` (used to build `overlay`) itself does a linear `iter_mut().find(...)` per query pair, making `overlay` construction O(query.len()^2). For the typical handful of query parameters this is negligible, but both `query` and the JSON body are attacker-controlled per-request inputs on every REST call, so a request with many repeated/distinct query keys and a body with many top-level keys degrades quadratically. If large query strings/bodies are otherwise within trust boundaries this is fine as-is; otherwise consider a `HashMap`/sorted-key lookup for the overlay/body-key membership check.

crates/rest_over_grpc/src/build/binding.rs
```diff
@@ -0,0 +1,88 @@
...
+        &self.method
+    }
+
+    /// The parsed path template this binding matches.
+    #[must_use]
+    #[expect(clippy::missing_panics_doc, reason = "only unreachable panics")]
+    pub fn template(&self) -> PathTemplate<'_> {
+        // The pattern was parsed and validated when this `Binding` was created,
+        // so re-parsing it with the affix-enabled grammar preserves its AST.
+        PathTemplate::parse(&self.pattern, Grammar::default().with_segment_affixes())
```
psandana, line 81: [guidelines] `Binding::template()` re-parses the stored pattern with `Grammar::default().with_segment_affixes()`, while the doc example on `Binding::new()` (and the public constructor contract) uses the plain `Grammar::default()`. This is likely intentional -- the affix grammar is a superset needed to recover `Segment::Affix` info that a plain re-parse would lose, and other consumers (`HttpRule::path_variable_field_paths`, `Route::template()`) do the same re-parse. Worth a one-line comment cross-referencing why the two grammars differ, so a future maintainer does not assume they must match.

crates/rest_over_grpc/src/serving.rs
```diff
@@ -0,0 +1,516 @@
...
+impl<T> RestService<T> {
+    /// Wraps a [`Transcode`] implementation
+    /// (typically a generated `Transcoder`) as a service.
+    ///
+    /// The request body is buffered without a size cap; call
+    /// [`with_max_body_bytes`](Self::with_max_body_bytes) to bound it.
+    pub const fn new(transcoder: T) -> Self {
+        Self {
+            transcoder,
+            max_body_bytes: None,
```
psandana, line 411: [security] `RestService::new()` defaults to `max_body_bytes: None` (unbounded request-body buffering) unless callers opt in via `with_max_body_bytes`. This appears to be a deliberate, documented trade-off (matching the neutral, policy-free contract of the free `serve_http` helper) rather than an oversight, so flagging as a low-severity hardening note rather than a bug: an integrator who copies one of the serving examples verbatim inherits unbounded body buffering, which is a memory-exhaustion/DoS vector on any endpoint exposed to untrusted clients without an external limit (e.g. reverse proxy). Consider either a conservative default cap (still overridable) or making sure every serving example demonstrates `with_max_body_bytes` so integrators do not copy an unbounded configuration into production.

crates/rest_over_grpc/src/build/descriptor.rs
```diff
@@ -0,0 +1,1631 @@
+// Copyright (c) Microsoft Corporation.
```
psandana, line 1: [design] Every other file under `crates/rest_over_grpc/src/build/` (binding.rs, route.rs, request_body.rs, response_body.rs, descriptor_options.rs, generator.rs, generator_builder.rs, http_rule.rs, service_method.rs, emit.rs) is well under ~330 lines with a single clear responsibility. `descriptor.rs` (1631 lines) is 5-6x larger and bundles descriptor decoding, type-path resolution, and enum-field classification. This makes it a maintenance/review hotspot inconsistent with the granularity used elsewhere in the same crate. Consider splitting into decode / type-resolution / annotation-parsing submodules (see also the equivalent-sized `service_definition.rs`).

crates/rest_over_grpc/src/build/service_definition.rs
```diff
@@ -0,0 +1,1628 @@
+// Copyright (c) Microsoft Corporation.
```
psandana, line 1: [design] Similar to `descriptor.rs`, this file (1628 lines) is a 5-6x outlier versus every sibling module in `build/`, bundling service-definition storage, trait codegen, tonic-bridge codegen, and OpenAPI glue in one file. Consider moving the codegen rendering (trait_code / tonic_bridge / openapi_spec) into their own module(s), leaving `ServiceDefinition` as a thinner data/builder type.

crates/rest_over_grpc/src/build/mod.rs
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) Microsoft Corporation.
```
psandana, line 1: [design] This PR adds a complete build-time codegen pipeline inside `rest_over_grpc::build` (descriptor decoding via `prost-reflect`, `ServiceDefinition`/`HttpRule`/`Binding`/`Route` builders, a `Generator` emitting `proc_macro2::TokenStream` via `quote`, and OpenAPI emission) that runs as its own `build.rs`-time step, separate from `routerama_build` (already a workspace dependency here). It is not clear from the diff alone whether this is the best-fit approach vs. extending `routerama_build` (which already owns build-time codegen conventions for routerama-based routing) to host the descriptor-to-route lowering, or whether a proc-macro invoked at compile time would have been simpler for consumers than wiring a second `build.rs` step alongside `tonic_prost_build`/`pbjson_build`. If the split is intentional -- e.g. because reading the `google.api.http` extension requires build-time reflection over descriptors that a proc-macro over source text could not do -- it would help to state that rationale explicitly in the README/module docs.

crates/rest_over_grpc/src/build/generator_builder.rs
```diff
@@ -0,0 +1,71 @@
...
+/// ```
+/// use rest_over_grpc::build::Generator;
+///
+/// let generator = Generator::builder().emit_tonic_bridge(true).build();
+/// ```
+#[derive(Debug, Default, Clone)]
+pub struct GeneratorBuilder {
+    // Stored inverted so the derived `Default` yields output with the `tonic`
+    // bridge emitted.
+    no_tonic: bool,
```
psandana: [design] `GeneratorBuilder`/`Generator` store `no_tonic: bool` (negated) instead of `emit_tonic: bool`, purely so `#[derive(Default)]` yields "tonic bridge on" by default. Negated-boolean fields are a common source of double-negative bugs (`!self.no_tonic`, `self.no_tonic = !emit`) and are easy to get backwards during future edits; the public setter (`emit_tonic_bridge`) already reads positively, so the internal representation fighting the derive is avoidable. Consider keeping the field positive (`emit_tonic: bool`) with a small hand-written `impl Default` (or an `Option<bool>`), trading one derive for a clearer invariant.

crates/rest_over_grpc/src/build/binding.rs
```diff
@@ -0,0 +1,88 @@
...
+///     HttpMethod::GET,
+///     PathTemplate::parse("/v1/shelves/{shelf}/info", Grammar::default()).expect("valid"),
+/// )
+/// .request_body(RequestBody::Whole);
+/// assert_eq!(binding.method().as_str(), "GET");
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct Binding {
+    method: HttpMethod,
+    pattern: String,
```
psandana, line 30: [design] `Binding` (and `HttpRule`/`Route`) store the rendered path pattern as an owned `String` and re-parse it via `PathTemplate::parse` (with an `.expect("pattern was validated when ... was created")`) on every `template()`/`path_variable_field_paths()` call, rather than caching a parsed representation. This is build-time-only code so the cost is likely negligible in absolute terms, but parse failures that "can't happen" are enforced only by convention/tests rather than the type system. If a self-owned parsed representation is wanted, consider an owned/`'static` `PathTemplate` variant, or caching the parse behind `Arc`/`OnceCell` -- worth a quick sanity check this pattern does not leak into any hot path outside `build/`.

crates/rest_over_grpc/src/transcode/overlay.rs
geeknoid: Fixed in bc096b84. Flat overlays now retain their ordered vectors but use hash indexes for query grouping and query-over-body membership, and top-level body duplicate handling is indexed as well. Regression tests preserve repeated-query order, last-write body semantics, and query precedence.

crates/rest_over_grpc/src/build/service_definition.rs
geeknoid: Agreed this is a reasonable follow-up refactor, but I am deferring it from this PR. The rendering paths share private route/capture structures and `ServiceDefinition` state extensively, so splitting them cleanly is a substantial mechanical change with no runtime/API behavior benefit. Keeping that structural work separate avoids adding review risk to this already broad feature PR.

crates/rest_over_grpc/src/build/descriptor.rs
geeknoid: I checked this split, but am deferring it from this PR. Production code ends at line 475; the remaining ~1,150 lines are colocated exhaustive descriptor tests, so the raw file size substantially overstates the implementation responsibility. Moving those tests mechanically would reduce the file count without improving the production design and would add churn to an already broad change.

crates/rest_over_grpc/src/build/binding.rs
geeknoid: Confirmed this is confined to the feature-gated build-time path and does not enter request routing. `PathTemplate` borrows its source, so caching it beside the owned `String` would require a self-referential representation; `Arc`/`OnceCell` alone cannot provide that lifetime. bc096b84 documents the ownership/reparse reason, and the validated-pattern invariant remains explicit in the `expect` message.

crates/rest_over_grpc/src/build/generator_builder.rs
geeknoid: Fixed in bc096b84. Both `GeneratorBuilder` and `Generator` now store the positive `emit_tonic` invariant and have explicit `Default` implementations that preserve bridge emission by default. Existing default/toggle tests cover the behavior.

crates/rest_over_grpc/src/build/binding.rs
geeknoid: Clarified in bc096b84. The comment now explains that `PathTemplate` borrows its input and that the affix grammar is intentionally the constructor grammar’s superset, recovering rendered affixes as `Segment::Affix`.

crates/rest_over_grpc/src/serving.rs
geeknoid: Hardened in bc096b84 without changing the policy-neutral default. Every copyable `RestService` example now demonstrates a 1 MiB `with_max_body_bytes` cap, including the tower example; the axum example was already capped. The free `serve_http` helper remains neutral because it receives an already-buffered `Request<Vec<u8>>`, so collection limits belong to its caller.

crates/rest_over_grpc/src/stream.rs
geeknoid: Fixed in bc096b84. Selected-field streaming now uses an internal `encode_response_into` helper that serializes directly into the frame buffer, including the `serde_json::Value` fallback, so there is no temporary `Vec` or frame copy per item. A regression test covers appending selected-field output to an existing buffer.

crates/rest_over_grpc/src/build/mod.rs
geeknoid: Documented the rationale in the `build` module docs in bc096b84. The HTTP annotations and protobuf type information become available from descriptor sets produced during `build.rs`; a source-level procedural macro cannot naturally consume that build output. `routerama_build` remains transport/protobuf agnostic, while this module lowers those descriptor semantics into routerama routes.

psandana (approved):

ralfbiedert (approved):
Approved with review comments. Please consider follow-ups for: (1) OpenAPI body/query exclusion compares JSON and proto names inconsistently for multi-word fields; (2) generated identifier sanitization should match prost, including self/extern/gen; (3) validate path-capture fields during descriptor processing; (4) invalid Accept q-values should not become 1.0; (5) StreamingResponse::encode exposes hidden StreamEncoding. Performance opportunities: avoid HashMap/per-value Vec allocation for small query sets and make TranscodeError backtrace capture lazy.

PR #605, opened 2026-07-25
feat(multitude): add fallible streaming and generation stats

geeknoid:
## Summary

- add fallible top-level JSON array callbacks that stop immediately and preserve callback errors separately from JSON and limit failures
- add `ArenaStats` generation counters for backing allocations, allocated bytes, cache reuse, and relocations since the last successful reset
- replace the optimistic index-only refresh fixture with an escaped-payload global-selection workload and paired Criterion/Gungraun coverage
- correct the shared Anvil Bolero recipe for cargo-bolero 0.13.4
- bump Multitude to 0.7.1

## Selective-array decision

The realistic workload did not justify a specialized public selected-array API. Direct `Vec<&RawValue>` indexing measured 3.66 ms / 41.93M instructions versus 3.47 ms / 41.89M instructions for the existing raw callback, with higher allocation counts. This PR keeps the benchmark evidence but adds no selected-array API.

## Validation

- `cargo test -p multitude --all-features --lib --tests --locked`
- `cargo clippy -p multitude --all-targets --all-features --locked -- -D warnings`
- `cargo check -p multitude --no-default-features --features serde_json,stats --locked`
- short cargo-bolero 0.13.4 libfuzzer campaign using the corrected invocation
- Multitude doctests, formatting, generated README check, spelling, Criterion, and Gungraun


crates/multitude/benches/multitude_record_batch/shared.rs
```diff
@@ -415,56 +467,70 @@ pub(crate) fn sparse_arena_hot_path(arena: &Arena, input: &[u8]) {
...
+        .filter(|(index, _)| selected.binary_search(index).is_ok())
+        .map(|(_, record)| (record.id, (record.name, record.payload)))
         .collect()
 }
 
 #[inline(never)]
 pub(crate) fn refresh_arena_each_hot_path(arena: &mut Arena, input: &[u8]) -> ArenaRetained {
     arena.reset();
-    let mut retained = BTreeMap::new();
+    let mut values = Vec::with_capacity(REFRESH_RECORDS);
```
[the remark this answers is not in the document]
geeknoid: Fixed in 420c6ac8. The full-callback benchmark now collects into an arena-backed vector, and its warm-up preallocates the same vector capacity. I also refreshed the Criterion and Gungraun figures in PERF.md.

crates/multitude/tests/arena.rs
```diff
@@ -173,23 +200,60 @@ fn arena_stats_report_cache_reuse_and_resets() {
...
+}
+
+#[cfg(feature = "stats")]
+#[test]
+fn arena_stats_report_oversized_bytes_and_relocations_since_reset() {
+    let mut arena = Arena::new();
+    let source = vec![0_u8; OVERSIZED_BYTES];
+    let oversized = arena.alloc_slice_copy_box(&source);
+    let mut values = arena.alloc_vec_with_capacity::<u64>(1);
+    values.extend(0..1_000);
```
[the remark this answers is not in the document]
geeknoid: Fixed in 420c6ac8. The test now advances the bump cursor with a separate allocation before growing the vector, so the relocation assertion is deterministic.

ralfbiedert:
Reviewed at `a869e5809f85c4925e6d6d08faa3c3947e4d7f08`.

No correctness or unsafe defect found. Two follow-ups:

1. The benchmark/report path does not measure either new hot-path surface. `crates/multitude/src/de/json.rs:366-428` adds fallible streaming, but the Criterion and Callgrind suites exercise only infallible streaming. The stats benchmark is also omitted from the generated report. Add success and early-rejection cases for `try_deserialize_json_each` and a separately reported stats-enabled reset/allocation workload.
2. `crates/multitude/src/de/json_each_error.rs:13-20` introduces a handwritten public error enum, which conflicts with the repository's canonical-error guidance and M365-OHNO.

The added atomics are on chunk slow paths and did not expose a credible implementation regression. Tests, Clippy, rustfmt, benchmarks, CI, and `multitude` docs succeeded.

ralfbiedert (approved):
Approved after the detailed review comment above.

crates/multitude/src/de/json.rs
```diff
@@ -288,6 +349,83 @@ impl<A: Allocator + Clone> Arena<A> {
...
+    /// [`JsonEachError::Callback`]. Because parsing stops at that element, the
+    /// remaining input is not validated. If every callback succeeds, the
+    /// complete array and trailing input are validated normally.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`JsonEachError::Json`] for malformed JSON, a shape mismatch,
+    /// trailing input, or allocation failure. Returns
+    /// [`JsonEachError::Callback`] when `callback` fails.
+    pub fn try_deserialize_json_each<'de, T, I, F, E>(&self, input: &'de I, mut callback: F) -> Result<(), JsonEachError<E>>
```
martin-kolinek, line 366: 🤖: Could we add a public-API success case for this method (and the limited variant) using a complete valid array, asserting `Ok(())` and that every element reaches the callback in order? The new tests currently exercise callback, JSON, trailing-input, wrong-root, and limit failures, but every invocation ends in `unwrap_err()`, so successful completion of the new APIs is not regression-tested.

crates/multitude/src/arena/mod.rs
```diff
@@ -412,7 +426,11 @@ impl<A: Allocator + Clone> Arena<A> {
         unsafe { displaced.release_with_refund(refund) };
         self.current_has_reference.set(false);
         #[cfg(feature = "stats")]
-        self.resets.set(self.resets.get() + 1);
+        {
+            self.provider.reset_generation_stats();
```
martin-kolinek, line 430: 🤖: Could the stats integration test perform fresh oversized allocation/forced relocation activity after this reset and assert the `*_since_reset` counters now describe only the second generation while the lifetime counters include both? The current coverage checks that these fields clear (and that cache reuse increments afterward), but never verifies that oversized/backing-byte/relocation counters repopulate correctly in a later generation.

crates/multitude/src/de/json.rs
geeknoid: Added a public-API success regression covering both `try_deserialize_json_each` variants with complete arrays. It asserts `Ok(())` and verifies every element reaches the callback in order.

crates/multitude/src/arena/mod.rs
geeknoid: Extended the stats integration test with a second post-reset generation that performs a fresh oversized allocation and forced relocation. It verifies the generation counters repopulate from zero while lifetime oversized-allocation and relocation counters include both generations.

crates/multitude/tests/serialization.rs
```diff
@@ -8,7 +8,9 @@
 
 use std::sync::atomic::{AtomicUsize, Ordering};
 
-use multitude::de::{DeserializationLimits, DeserializationResource, DeserializeIn, DeserializeInSeed, Entry, Number, Value};
+use multitude::de::{
+    DeserializationLimits, DeserializationResource, DeserializeIn, DeserializeInSeed, Entry, JsonEachError, JsonError, Number, Value,
+};
```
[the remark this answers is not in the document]
geeknoid: Fixed by gating the integration test module on `serde_json`, which transitively enables `serde`. Confirmed that `cargo check -p multitude --test serialization --features serde --locked` now succeeds and the JSON-enabled integration tests still pass.

martin-kolinek (approved):

PR #610, opened 2026-07-27
fix: eliminate missed mutation cases

st-dev-gh:
Catch mutations previously missed by anvil-mutants-full.

- handle nonzero cargo exits explicitly in run_cargo
- test cargo failure propagation and PID-based process termination
- make the zero-length Drain drop path explicit


crates/automation/src/process.rs
```diff
@@ -107,3 +107,48 @@ fn kill_by_pid(pid: u32) {
         let _ = Command::new("taskkill").args(["/F", "/PID", &pid.to_string()]).status();
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::env;
+    use std::time::Instant;
+
+    use super::*;
```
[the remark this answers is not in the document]
st-dev-gh: "use super::*;" is a common practice in tests, see https://github.com/search?q=repo%3Amicrosoft%2Foxidizer%20%22use%20super%3A%3A*%3B%22&type=code

ralfbiedert:
Reviewed at `bfe5490e0540a1477aa8b85b9332604e91eecf49`.

No production correctness, performance, or unsafe defect found. Explicit handling of nonzero Cargo exits is correct, PID-based process termination behaves correctly, and the `Drain::drop` branch inversion preserves guard ordering and exactly-once destruction.

One non-blocking guideline concern: `crates/automation/src/cargo.rs:37-47` tests only the public `run_cargo` contract and would fit better as an integration test under `crates/automation/tests/`.

Focused tests, strict-provenance Miri, Clippy, rustfmt, CI, and `multitude` docs succeeded.

ralfbiedert (approved):
Approved after the detailed review comment above.

crates/automation/src/process.rs
```diff
@@ -107,3 +107,48 @@ fn kill_by_pid(pid: u32) {
...
+
+    #[test]
+    #[ignore = "helper process spawned by kill_by_pid_terminates_process"]
+    fn sleeping_child() {
+        thread::sleep(Duration::from_secs(30));
```
Vaiz: I suggest to exclude `automation` crate from mutants testing

Vaiz (changes requested):

crates/automation/src/process.rs
st-dev-gh: Yes, `automation` is excluded in `scripts\mutants.rs` along with 3 other crates.
Let's update `.cargo/mutants.toml` exclusions to reflect that.

.cargo/mutants.toml
```diff
@@ -3,7 +3,9 @@
 
 examine_globs = ["crates/**"]
 exclude_globs = [
+    "crates/automation/**",
     "crates/rest_over_grpc_examples/**",
+    "crates/rest_over_grpc_tests/**",
     "crates/testing_aids/**",
```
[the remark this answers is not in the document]
st-dev-gh: It is already documented here: https://github.com/microsoft/oxidizer/blob/f0ef8998b57e0647c43aabe1756d67b45d4928a6/crates/automation/src/cargo.rs#L8-L11

Vaiz (approved):

.cargo/mutants.toml
```diff
@@ -3,7 +3,9 @@
 
 examine_globs = ["crates/**"]
 exclude_globs = [
+    "crates/automation/**",
```
martin-kolinek, line 6: 🤖: This correctly brings `exclude_globs` into line with `automation::INTERNAL_CRATES` (crates/automation/src/cargo.rs:11), which scripts/mutants.rs:64 already enforces for the orchestrated run. The concern is that the internal-crate exclusion policy now lives in three hand-synced places (this toml as path globs, `INTERNAL_CRATES` as names, and transitively scripts/mutants.rs) with nothing failing when they disagree — this PR fixes today's drift but the next crate added to `INTERNAL_CRATES` will silently re-open the gap here. Consider a small unit test in the `automation` crate that reads .cargo/mutants.toml and asserts every `INTERNAL_CRATES` entry has a matching `crates/<name>/**` exclude line, turning this manual invariant into a checked one.

martin-kolinek (approved):

.cargo/mutants.toml
st-dev-gh: Yes. this is meant as a quick fix only.  Copilot suggested a few longer-term options to consider but they exceed the intended scope of this PR.

> Longer-term options:
> 
> 1. Add a drift check: verify that  INTERNAL_CRATES  and  exclude_globs  contain matching crates.
> 2. Use package metadata as the source of truth: for example,  [package.metadata.oxidizer-ci] mutation = false . A wrapper script could translate this into repeated  cargo mutants --exclude ...  arguments.
> 3. Use  publish = false : all four current internal crates are private, so CI scripts could automatically skip private crates. Explicit metadata is safer if a private crate might still deserve mutation testing.
> 4. Route Anvil through  scripts/mutants.rs : it already filters  INTERNAL_CRATES , but the Anvil justfiles are managed and this change should be made upstream or through a repository-specific wrapper.

PR #612, opened 2026-07-28
feat(tick): add EcmaScript timestamp format

sandersaares:
[Copilot speaking]

## Summary

Adds `EcmaScript` as a first-class `tick::fmt` timestamp format, alongside `Iso8601`, `Rfc2822`, and `UnixSeconds`. It formats the ECMAScript Date Time String Format produced by `Date.prototype.toISOString`, with exactly three fractional digits and the UTC designator `Z`.

For years `0000` through `9999`, output has the fixed 24-character shape `YYYY-MM-DDTHH:MM:SS.sssZ`. On platforms whose `SystemTime` supports earlier years, negative years use the ECMAScript expanded-year form, such as `-009999-01-02T01:59:59.000Z`.

## API and behavior

- Adds `tick::fmt::EcmaScript` with `MAX`, `UNIX_EPOCH`, `FromStr`, `Display`, `TryFrom<SystemTime>`, infallible conversion to `SystemTime`, serde support, and thread-awareness conformance.
- Parses RFC 3339 timestamps and the expanded-year syntax supported by `jiff::Timestamp`, normalizing UTC offsets to `Z`.
- Canonicalizes values to millisecond precision at construction. Finer precision is floored rather than rounded, including before the Unix epoch, so equality, ordering, hashing, display, and serialization agree.
- Validates constructed values against the platform's `SystemTime` range. Any `EcmaScript` value can therefore be converted to `SystemTime` without a range panic.
- Does not provide direct conversions between timestamp format types. Conversions use `SystemTime` as the explicit fallible bridge, consistent with the rest of `tick::fmt`.
- Adds `SystemTimeExt::display_ecmascript()`, an infallible formatter that saturates values outside jiff's supported timestamp range.
- Extends the `tick::fmt` module documentation, examples, serde coverage, range tests, and shared format tests for `EcmaScript`.

## Unix seconds consistency

`UnixSeconds` now stores only canonical whole-second values. Construction floors sub-second `Duration` values to whole seconds, keeping equality, ordering, hashing, display, and serialization consistent. `UnixSeconds::MAX` is the whole-second floor of the common upper timestamp boundary: `253402207200` (`9999-12-30T22:00:00Z`).

Pre-epoch `SystemTime` values remain rejected because `UnixSeconds` represents an unsigned offset from the Unix epoch.

## Validation

- Windows: 187 `tick` unit tests and 61 doctests pass.
- Linux/WSL: all ECMAScript-specific tests, including expanded negative years, and the shared platform-range test pass.
- `clippy`, `format-check`, `spellcheck`, and README generation pass.


crates/tick/src/system_time_ext.rs
```diff
@@ -22,6 +22,27 @@ pub trait SystemTimeExt: sealed::Sealed {
...
     #[cfg(any(feature = "fmt", test))]
     fn display_iso_8601(&self) -> impl std::fmt::Display;
+
+    /// Returns a value that formats the [`SystemTime`] in the ECMAScript Date Time String Format.
+    ///
+    /// The output has the fixed 24-character shape `YYYY-MM-DDTHH:MM:SS.sssZ`, truncated to
+    /// millisecond precision. See [`EcmaScript`][crate::fmt::EcmaScript] for details.
+    ///
+    /// Times outside the valid range (before year -9999 or after year 9999) are saturated
+    /// to the nearest boundary.
```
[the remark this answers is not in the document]
sandersaares: [Copilot speaking] Fixed in 91690c13. The display_ecmascript doc now scopes the fixed 24-character shape to non-negative years and explicitly states that saturating below the minimum yields a negative year, which renders in the wider ECMAScript expanded-year form rather than the fixed 24-character shape.

crates/tick/src/fmt/ecma_script.rs
```diff
@@ -0,0 +1,339 @@
...
+
+use crate::Error;
+use crate::fmt::{Iso8601, Rfc2822, UnixSeconds};
+
+/// Fixed-width parser and formatter for the ECMAScript Date Time String Format.
+///
+/// For every wall-clock instant the output has the fixed 24-character shape
+/// `YYYY-MM-DDTHH:MM:SS.sssZ`: a four-digit year, two-digit calendar and clock
+/// fields, exactly three fractional digits (milliseconds, truncated rather than
+/// rounded), and the UTC designator `Z`. For example: `2024-08-06T21:30:00.123Z`.
```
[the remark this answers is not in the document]
sandersaares: [Copilot speaking] Fixed in 91690c13. The opening description now scopes the fixed 24-character shape to non-negative years and notes that pre-1 CE years (reachable only by saturation) are the sole exception, rendering wider in the expanded-year form described in the Range section below.

Vaiz (approved):

st-dev-gh (approved):

crates/tick/src/fmt/ecma_script.rs
```diff
@@ -0,0 +1,341 @@
...
+use crate::Error;
+use crate::fmt::{Iso8601, Rfc2822, UnixSeconds};
+
+/// Fixed-width parser and formatter for the ECMAScript Date Time String Format.
+///
+/// For any non-negative year the output has the fixed 24-character shape
+/// `YYYY-MM-DDTHH:MM:SS.sssZ`: a four-digit year, two-digit calendar and clock
+/// fields, exactly three fractional digits (milliseconds, truncated rather than
+/// rounded), and the UTC designator `Z`. For example: `2024-08-06T21:30:00.123Z`.
+/// Years before 1 CE - reachable only by saturation - are the sole exception and
```
martin-kolinek: 🤖: The fixed-width and `lexicographic == chronological` guarantees stated here aren't enforced by the type, and `reachable only by saturation` is inaccurate. Negative expanded years are also reachable through the public `FromStr` and `TryFrom<SystemTime>` paths — the round-trip test at lines 275-284 parses `-009999-01-02T01:59:59.000Z` straight back into an `EcmaScript`. Those paths therefore also produce the wider expanded-year form, and lexicographic order is not chronological across negative expanded years (and, since milliseconds are truncated, ordering is only meaningful at ms precision). Either reject pre-1-CE / non-four-digit years consistently in every constructor and conversion if fixed width is the intended invariant, or document the expanded-year and ordering limitations as general properties instead of scoping them to saturation.

crates/tick/src/fmt/ecmascript.rs
```diff
@@ -0,0 +1,341 @@
...
+    /// Wraps a [`Timestamp`] already known to be within the representable range.
+    ///
+    /// Used by [`SystemTimeExt::display_ecmascript`][crate::SystemTimeExt::display_ecmascript]
+    /// to build a value from an already-saturated timestamp.
+    pub(crate) const fn from_timestamp(timestamp: Timestamp) -> Self {
+        Self(timestamp)
+    }
+}
+
+impl FromStr for EcmaScript {
```
martin-kolinek, line 140: 🤖: Please add a behavioural parsing test for an offset-bearing RFC 3339 input — e.g. parse `2024-08-06T14:30:00-07:00` and assert it formats as `2024-08-06T21:30:00.000Z`. The public contract says this ISO/RFC 3339 superset is accepted and normalized to UTC, but every current parse test uses `Z`, so a regression that rejected or mishandled offsets would pass unnoticed.

crates/tick/src/system_time_ext.rs
```diff
@@ -30,6 +56,11 @@ impl SystemTimeExt for SystemTime {
         // jiff's Timestamp implements Display that outputs ISO 8601 format
         to_timestamp(*self)
     }
+
+    #[cfg(any(feature = "fmt", test))]
+    fn display_ecmascript(&self) -> impl std::fmt::Display {
```
martin-kolinek, line 61: 🤖: Please exercise the lower saturation boundary through this public method: construct a `SystemTime` earlier than `Timestamp::MIN` and assert `display_ecmascript()` returns `-009999-01-02T01:59:59.000Z`. The existing out-of-range test covers only `MAX`, while the negative-year test wraps `Timestamp::MIN` directly and so cannot catch a regression in the public lower-saturation path.

crates/tick/src/fmt/ecma_script.rs
```diff
@@ -0,0 +1,341 @@
...
+/// render wider, in the expanded-year form described in the Range section below.
+///
+/// Unlike the variable-precision output of [`Iso8601`] - which trims trailing
+/// fractional zeros and so varies in width - this fixed width keeps tabular
+/// columns aligned and guarantees that lexicographic ordering matches
+/// chronological ordering.
```
martintmk: The fixed width is the selling point of this type, but nothing enforces it and "reachable only by saturation" isn't true. Both ordinary public paths reach a negative year with no saturation involved — verified on this branch:

```
"-000500-01-01T00:00:00Z".parse::<EcmaScript>()
  -> Ok -> "-000500-01-01T00:00:00.000Z"   (27 chars)

EcmaScript::try_from(SystemTime::UNIX_EPOCH - Duration::from_secs(140_000_000_000))
  -> Ok -> "-002467-07-30T15:06:40.000Z"
```

So a consumer who sizes a `[u8; 24]`, a fixed-width log column, or a lexicographic index on the word "guarantees" here gets broken silently by parsed input.

Please scope the claim to years `0..=9999` and state that `FromStr` / `TryFrom<SystemTime>` can yield wider output. The same wording needs fixing at L56-59 and in the test comment at L276.

The alternative — actually enforcing it, by rejecting negative years in `FromStr`/`TryFrom` and saturating low to `0000-01-01T00:00:00.000Z` — would make 24 characters a real invariant. But it forces `From<Iso8601 | Rfc2822 | UnixSeconds>` to become `TryFrom` and breaks the family symmetry you deliberately built here. I don't think that trade is worth it, so: fix the docs.

crates/tick/src/fmt/ecma_script.rs
```diff
@@ -0,0 +1,341 @@
...
+/// # Range
+///
+/// The wrapped instant is a [`jiff::Timestamp`], whose year is in the range
+/// `-9999..=9999`. Construct one fallibly with [`TryFrom`], or infallibly (with
+/// out-of-range instants saturated to the nearest boundary) via
+/// [`SystemTimeExt::display_ecmascript`][crate::SystemTimeExt::display_ecmascript].
+/// The 24-character width above assumes a non-negative year; years before 1 CE -
+/// reachable only by saturating an instant before the minimum - render using the
+/// ECMAScript expanded-year form (a leading `-` and six year digits) and are
+/// correspondingly wider.
```
martintmk: Same inaccuracy as flagged on the type doc above — years before 1 CE are reachable by plain `FromStr` and `TryFrom<SystemTime>`, not only by saturating below the minimum. Since this is the Range section, it's the place to state the real contract: the wrapped `Timestamp` spans `-9999..=9999`, and only the `0..=9999` sub-range renders at the fixed 24-character width.

crates/tick/src/fmt/ecma_script.rs
```diff
@@ -0,0 +1,341 @@
...
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct EcmaScript(pub(super) Timestamp);
+
+crate::thread_aware_move!(EcmaScript);
+
+/// Prints the fixed-width ECMAScript representation.
+///
+/// `precision(Some(3))` pins the fractional component to exactly three digits
+/// and truncates (does not round) any finer precision. Years outside `1..=9999`
+/// use the ECMAScript expanded-year form, keeping the output round-trippable.
```
martintmk: nit: round-trip is broken for anything with sub-millisecond precision, `MAX` included — `EcmaScript::MAX.to_string().parse() != EcmaScript::MAX` (verified). The claim is only about the year *syntax*; worth saying so explicitly, otherwise it reads as a general round-trip guarantee that the truncating printer can't offer.

crates/tick/src/fmt/ecmascript.rs
```diff
@@ -0,0 +1,341 @@
...
+/// `precision(Some(3))` pins the fractional component to exactly three digits
+/// and truncates (does not round) any finer precision. Years outside `1..=9999`
+/// use the ECMAScript expanded-year form, keeping the output round-trippable.
+static ECMASCRIPT_PRINTER: temporal::DateTimePrinter = temporal::DateTimePrinter::new().precision(Some(3));
+
+impl EcmaScript {
+    /// The largest value that can be represented by `EcmaScript`.
+    ///
+    /// This represents a Unix system time of `31 December 9999 23:59:59 UTC`.
+    pub const MAX: Self = Self(Timestamp::MAX);
```
martintmk: nit: the value is `9999-12-30T22:00:00.999999999Z`, not `31 December 9999 23:59:59 UTC` — your own `max_constant_renders_truncated` asserts the former. Copied verbatim from the siblings so it's pre-existing and family-wide; only worth correcting because this is a fresh copy.

crates/tick/src/fmt/ecmascript.rs
```diff
@@ -0,0 +1,341 @@
...
+    fn negative_year_renders_in_expanded_form() {
+        // Years before 1 CE are reachable only by saturating an instant before
+        // the minimum. They render in the ECMAScript expanded-year form (a
+        // leading `-` and six year digits), which - unlike a bare `-9999`
+        // prefix - round-trips back to the same value.
+        let ecma = EcmaScript::from_timestamp(Timestamp::MIN);
+        let rendered = ecma.to_string();
+
+        assert_eq!(rendered, "-009999-01-02T01:59:59.000Z");
+        assert_eq!(rendered.parse::<EcmaScript>().unwrap(), ecma);
```
martintmk, line 327: This reaches the negative-year case through the `pub(crate)` `from_timestamp`, which is exactly what makes the "reachable only by saturation" framing look true. Worth pinning the public path too:

```suggestion
        let ecma = EcmaScript::from_timestamp(Timestamp::MIN);
        let rendered = ecma.to_string();

        assert_eq!(rendered, "-009999-01-02T01:59:59.000Z");
        assert_eq!(rendered.parse::<EcmaScript>().unwrap(), ecma);

        // Negative years are also reachable straight from `FromStr`, with no
        // saturation involved, so the fixed 24-character width is not an
        // invariant of the type.
        let parsed: EcmaScript = "-000500-01-01T00:00:00Z".parse().unwrap();
        assert_eq!(parsed.to_string(), "-000500-01-01T00:00:00.000Z");
```

crates/tick/src/system_time_ext.rs
```diff
@@ -22,6 +22,32 @@ pub trait SystemTimeExt: sealed::Sealed {
...
+    /// use tick::SystemTimeExt;
+    ///
+    /// let time = SystemTime::UNIX_EPOCH + Duration::from_hours(1);
+    /// assert_eq!(
+    ///     time.display_ecmascript().to_string(),
+    ///     "1970-01-01T01:00:00.000Z"
+    /// );
+    /// ```
+    #[cfg(any(feature = "fmt", test))]
+    fn display_ecmascript(&self) -> impl std::fmt::Display;
```
martintmk, line 50: Naming breaks the family convention: `Iso8601` (module `iso_8601`) gets `display_iso_8601`, so `EcmaScript` (module `ecma_script`) should get `display_ecma_script`.

```suggestion
    #[cfg(any(feature = "fmt", test))]
    fn display_ecma_script(&self) -> impl std::fmt::Display;
```

Public surface — cheap now, breaking once `tick` ships it. The impl at L61 and the doc links in `ecma_script.rs` (L55, L104) need the same rename.

crates/tick/src/fmt/unix_seconds.rs
```diff
@@ -164,6 +164,12 @@ impl From<Iso8601> for UnixSeconds {
     }
 }
 
+impl From<EcmaScript> for UnixSeconds {
+    fn from(value: EcmaScript) -> Self {
+        Self(value.to_unix_epoch_duration())
+    }
+}
```
martintmk: non-blocking: this new conversion inherits `to_unix_epoch_duration`'s `unsigned_abs`, so pre-epoch instants come back mirrored. Verified:

`"1969-12-31T23:00:00.250Z"` → `UnixSeconds(3599)` → `1970-01-01T00:59:59Z` — an hour *after* the epoch instead of an hour before.

You flag this as pre-existing (ADO 7661495) and consistent with the siblings, which is fair. But this PR ships a *new* public conversion carrying it, so at minimum worth a doc note on the limitation rather than leaving it to be discovered in production.

crates/tick/src/fmt/mod.rs
```diff
@@ -59,10 +67,12 @@
 //! // Output: Time: 1970-01-01T01:00:00Z
 //! ```
```
martintmk, line 128: nit: the format list above now has four entries, but this `SystemTimeExt` example still demonstrates only `display_iso_8601`. One more line for the ECMAScript formatter would make both extension methods discoverable.

martintmk (changes requested):
[Copilot speaking]

_Automated review by an AI agent (GitHub Copilot CLI), requested by @martintmk._

---

Clean, convention-following addition — family symmetry is complete, the tests are behaviour-shaped rather than coverage-shaped, and it validates locally (168 tests, 57 doctests, `clippy --all-features --all-targets`, default-feature build — all pass).

Two things to fix: the fixed-width guarantee doesn't actually hold and the docs assert that it does, and the `SystemTimeExt` method name breaks the family convention. The rest are non-blocking nits.

**Design note, no change requested:** `EcmaScript` is state-identical to `Iso8601` with a byte-identical `FromStr`, differing only in `Display` precision, and the conversion matrix is now 12 `From` impls growing O(N²). I think the separate type is the right call — serde field typing is the payoff a display adapter on `Iso8601` can't give you. Noting it only because a fifth format costs 8 more impls.

martintmk:
tried a new review agent, address only relevant parts

crates/tick/src/fmt/ecma_script.rs
martintmk: [Human] I think we should reject negative values to keep the invariants (unchanged format and size),

crates/tick/src/fmt/ecma_script.rs
sandersaares: [Copilot speaking] Addressed in 0b72ccdf. The type doc no longer states the fixed 24-character width or `lexicographic == chronological` ordering as guarantees. It now scopes the fixed width to years `0000..=9999` and states that years outside that range render in the ECMAScript expanded-year form, reachable through any constructor (including `FromStr`), matching `Date.prototype.toISOString`.

crates/tick/src/fmt/ecmascript.rs
sandersaares: [Copilot speaking] Added in 0b72ccdf as `parse_with_offset_normalizes_to_utc`: parsing `"2024-08-06T14:30:00-07:00"` renders `2024-08-06T21:30:00.000Z`, confirming the numeric offset is normalized to UTC.

crates/tick/src/system_time_ext.rs
sandersaares: [Copilot speaking] Added in 0b72ccdf as `display_ecmascript_saturates_below_min`: a `SystemTime` far below `Timestamp::MIN` renders `-009999-01-02T01:59:59.000Z` through the public `display_ecmascript` method. It is gated `#[cfg(not(windows))]` because a `SystemTime` before the FILETIME epoch (1601-01-01) is unconstructable on Windows.

crates/tick/src/fmt/ecma_script.rs
sandersaares: [Copilot speaking] Docs fixed in 0b72ccdf: the opening description, the Range section, and the test comment now scope the fixed width to `0000..=9999` and note that `FromStr` / `TryFrom<SystemTime>` can yield wider expanded-year output.

On the enforcement alternative, and @martintmk's follow-up to reject negative values: this stays as document-don't-reject, for three reasons. (1) Conformance: real ECMAScript `Date.prototype.toISOString()` renders expanded years rather than rejecting them (verified in Node: `-009999-...`, `+275760-...`); rejecting would make this type non-conformant to the spec it implements. (2) Rejecting negatives would not create a real `SystemTime` invariant: on Windows `SystemTime::from(EcmaScript)` panics for any instant before 1601-01-01 (the FILETIME epoch), so years `1..=1600` are already unsafe - the boundary is 1601, not 0. That pre-existing family-wide issue is filed as ADO 7663342. (3) It preserves the infallible `From<Iso8601 | Rfc2822 | UnixSeconds>` family symmetry; enforcing would force those to `TryFrom`.

This is a design call for a human to confirm - I can't dismiss the CHANGES_REQUESTED review, so leaving this thread open for @martintmk.

crates/tick/src/fmt/ecma_script.rs
sandersaares: [Copilot speaking] Addressed in 0b72ccdf. The Range section now states that years outside `0000..=9999` render in the expanded-year form and are reachable through any constructor, not only by saturation.

crates/tick/src/fmt/ecma_script.rs
sandersaares: [Copilot speaking] Addressed in 0b72ccdf by dropping the round-trip claim entirely from the type and printer docs rather than qualifying it.

crates/tick/src/fmt/ecmascript.rs
sandersaares: [Copilot speaking] Fixed in 0b72ccdf. The MAX doc now reads `9999-12-30T22:00:00.999999999Z`, corrected here and in the sibling formats (`Iso8601`, `Rfc2822`, `UnixSeconds`) plus the shared alignment test comment.

crates/tick/src/fmt/ecmascript.rs
sandersaares: [Copilot speaking] Applied in 0b72ccdf. `negative_year_renders_in_expanded_form` now also parses `"-000500-01-01T00:00:00Z"` directly and asserts it renders `-000500-01-01T00:00:00.000Z`, exercising the public `FromStr` path with no saturation.

crates/tick/src/system_time_ext.rs
sandersaares: [Copilot speaking] Resolved the other direction in 0b72ccdf. "ECMAScript" is one word, so instead of splitting the method into `display_ecma_script` we renamed the module `ecma_script` -> `ecmascript` to match. The family stays consistent: type `EcmaScript`, module `ecmascript`, method `display_ecmascript`, doc prose "ECMAScript".

crates/tick/src/fmt/unix_seconds.rs
sandersaares: [Copilot speaking] This is the pre-existing `unsigned_abs` mirroring behavior, identical across the whole format family, filed as ADO 7661495. Keeping the new conversion consistent with its siblings in this PR; the fix belongs in that bug so every format is corrected together rather than one diverging here.

crates/tick/src/fmt/mod.rs
sandersaares: [Copilot speaking] Fixed in 0b72ccdf. The `SystemTimeExt` example in `fmt/mod.rs` now also calls `display_ecmascript()` and shows its `1970-01-01T01:00:00.000Z` output.

crates/tick/src/fmt/ecma_script.rs
martintmk: ok, this is clarified. ECMA allows going negative, so let's allow it too

martintmk (approved):

ralfbiedert (commented):
Review findings at current head `0b72ccdf`:

1. **Pre-epoch conversion changes the instant** (`crates/tick/src/fmt/ecmascript.rs:100-102`, `unix_seconds.rs:167-170`). `unsigned_abs()` maps `1969-12-31T23:59:59Z` to Unix second `+1`, which round-trips as `1970-01-01T00:00:01Z`. This conversion should be fallible for pre-epoch values, or Unix seconds should be signed.
2. **Negative years can panic after the infallible RFC conversion** (`rfc_2822.rs:119-124,154-157`). `From<EcmaScript> for Rfc2822` accepts values that the RFC printer cannot represent; ordinary `to_string()` then panics on `fmt::Error`. This should be a checked `TryFrom` conversion.
3. **The millisecond wire representation is not canonical** (`ecmascript.rs:77-87,113-119,168-186`). Constructors and equality retain nanoseconds while Display/Serde truncate to milliseconds, so unequal values serialize identically and round-trip unequally. Please normalize all construction paths to milliseconds or preserve precision and name/document the abstraction accordingly.
4. **The advertised ECMAScript range is unsupported** (`ecmascript.rs:21-33,54-61`, `system_time_ext.rs:67-80`). ECMAScript Date supports approximately years -271821 through +275760, while Jiff is limited to roughly -9999 through +9999. Valid values such as `+010000` are rejected or silently saturated despite the `Date.prototype.toISOString` compatibility claim.
5. **The fixed-width wording remains contradictory** (`fmt/mod.rs:12-14`, `ecmascript.rs:42-50`): expanded-year output is 27 characters, so parsing/formatting/serialization are not unconditionally fixed-width.
6. **Deserialization allocates unnecessarily** (`ecmascript.rs:178-186`). A Serde string visitor can parse borrowed `&str` input without allocating a `String` for ordinary inputs.

The latest commit removes the incorrect ordering/round-trip promises and corrects the Jiff `MAX` documentation, but the substantive conversion, range, precision, panic, and allocation issues remain.

crates/tick/src/fmt/mod.rs
```diff
@@ -9,6 +9,10 @@
 //! - [`Iso8601`]: Parsing and formatting of system time in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) format.
 //!   For example, `2024-08-06T21:30:00Z`.
 //!
+//! - [`EcmaScript`]: Fixed-width parsing and formatting of system time in the
```
martin-kolinek: 🤖: This module summary still advertises `EcmaScript` as "Fixed-width parsing and formatting" unconditionally, but the type docs in `ecmascript.rs` now scope the fixed 24-character width to years `0000..=9999` and explicitly disclaim it as a type-wide invariant. The Parsing section (`ecmascript.rs:44`, "formatting always emits the fixed-width profile above") and the Serialization section (`ecmascript.rs:50`, serialized using the "fixed-width ECMAScript Date Time String Format") repeat the same unqualified claim. Please carry the `0000..=9999` qualification through these three remaining public descriptions so rustdoc doesn't reintroduce the type-wide 24-byte bound the type docs now disclaim.

crates/tick/src/system_time_ext.rs
```diff
@@ -30,6 +56,11 @@ impl SystemTimeExt for SystemTime {
         // jiff's Timestamp implements Display that outputs ISO 8601 format
         to_timestamp(*self)
     }
+
```
psandana, line 59: [style, minor] The newly added `display_ecmascript` method leaves a stray blank line right before the closing `}` of the `impl SystemTimeExt for SystemTime` block. Purely cosmetic - `cargo fmt` should collapse it.

crates/tick/src/system_time_ext.rs
```diff
@@ -22,6 +22,32 @@ pub trait SystemTimeExt: sealed::Sealed {
     /// ```
     #[cfg(any(feature = "fmt", test))]
     fn display_iso_8601(&self) -> impl std::fmt::Display;
+
+    /// Returns a value that formats the [`SystemTime`] in the ECMAScript Date Time String Format.
+    ///
+    /// For any non-negative year the output has the fixed 24-character shape
```
psandana: [doc-consistency] "For any non-negative year the output has the fixed 24-character shape" doesn't quite match the more precise `0000..=9999` framing used in `ecmascript.rs` (which says "For years 0000 through 9999..." and "Only the `0000..=9999` sub-range renders at the fixed 24-character width..."). A reader of this doc alone could think any large positive year gets the fixed width, when the real bound is `EcmaScript`'s own `-9999..=9999` range. Suggest aligning the wording, e.g. "For years `0000` through `9999` the output has the fixed 24-character shape...", to match `EcmaScript`'s doc.

crates/tick/src/fmt/unix_seconds.rs
```diff
@@ -164,6 +164,12 @@ impl From<Iso8601> for UnixSeconds {
     }
 }
 
+impl From<EcmaScript> for UnixSeconds {
```
psandana: [correctness] `to_unix_epoch_duration` uses `self.0.duration_since(Timestamp::UNIX_EPOCH).unsigned_abs()`, which discards the sign of the signed duration. This is the sole implementation behind `impl From<EcmaScript> for UnixSeconds`, so a pre-epoch `EcmaScript` instant (e.g. `1969-12-31T23:59:59Z`) converts to the *same* `UnixSeconds` value as the instant one second *after* epoch - two distinct instants silently collapse to the same (wrong-signed) value, and there's no test covering this conversion path.

I understand from the PR description that the `unsigned_abs` pattern mirroring pre-epoch instants is a known family-wide latent bug already filed separately (ADO 7661495/7661499) for the `Iso8601`/`Rfc2822` siblings, and that it's intentionally being kept consistent here. Could you confirm those filed bugs explicitly cover this new `EcmaScript -> UnixSeconds` call site as well, and consider adding a regression test that documents the current (known-bad) behavior so it isn't silently changed or reintroduced differently later?

crates/tick/src/fmt/ecmascript.rs
```diff
@@ -0,0 +1,353 @@
...
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde_core::Serializer,
+    {
+        serializer.collect_str(self)
+    }
+}
+
+#[cfg(any(feature = "serde", test))]
+impl<'de> serde_core::Deserialize<'de> for EcmaScript {
```
psandana, line 189: [perf, minor] `Deserialize` allocates an owned `String` before parsing (`String::deserialize(deserializer)?.parse::<Self>()`), even though `EcmaScript::from_str` only needs a `&str`. A borrowed-friendly deserialize path (e.g. via a visitor or `deserializer.deserialize_str`) could avoid this allocation on the hot path. This likely mirrors the same pattern in the sibling `Iso8601`/`Rfc2822`/`UnixSeconds` types for consistency, so may be an accepted trade-off, but flagging per M-MEM-REUSE in case it's worth addressing crate-wide.

crates/tick/src/fmt/ecmascript.rs
```diff
@@ -0,0 +1,353 @@
...
+
+use crate::Error;
+use crate::fmt::{Iso8601, Rfc2822, UnixSeconds};
+
+/// Parser and formatter for the ECMAScript Date Time String Format.
+///
+/// For years 0000 through 9999 the output has the fixed 24-character shape
+/// `YYYY-MM-DDTHH:MM:SS.sssZ`: a four-digit year, two-digit calendar and clock
+/// fields, exactly three fractional digits (milliseconds, truncated rather than
+/// rounded), and the UTC designator `Z`. For example: `2024-08-06T21:30:00.123Z`.
```
psandana, line 19: [language, minor] A few small wording/style nits in this doc comment block:
- Line 19 says "the UTC designator `Z`" while the "# UTC and time zones" section below calls the same thing "the `Z` designator" - consider using one consistent term.
- Line 43: "Parsing accepts any RFC 3339 or ISO 8601 timestamp, of which the ECMAScript profile is a subset." reads a bit indirectly; something like "...timestamp - the ECMAScript profile is a subset of both." may be clearer.
- Lines 21-22 and 28-29 use a spaced hyphen (` - `) as a parenthetical dash (e.g. "expanded-year form - a sign..."). Consider an em dash (-) for consistency with typical Rust doc style and to avoid visual ambiguity with the numeric/range content nearby.

None of these are blocking, just polish.

crates/tick/src/system_time_ext.rs
sandersaares: [Copilot speaking] This blank line (system_time_ext.rs:59) is the standard separator between the two trait methods `display_iso_8601` and `display_ecmascript`, not a trailing line before the closing brace: the `}` at line 64 directly follows the last method body with no intervening blank line. `cargo fmt` keeps this separator (format-check is green), so there is nothing to collapse. No change.

crates/tick/src/fmt/ecmascript.rs
sandersaares: [Copilot speaking] Agreed it's worth addressing, and you're right that it mirrors the sibling `Iso8601`/`Rfc2822`/`UnixSeconds` types, so it's a crate-wide concern rather than something to diverge here. Filed ADO 7667382 to remove the intermediate `String` allocation across the family, with a request for Criterion + Callgrind benchmarks as evidence of the improvement. Keeping the current consistent pattern in this PR.

crates/tick/src/fmt/mod.rs
sandersaares: [Copilot speaking] Done in e6cd8381. Dropped the unqualified "Fixed-width" from the `fmt` module summary, and removed the two remaining unqualified claims in `ecmascript.rs`: the Parsing section now says formatting "emits the ECMAScript profile described above", and the Serialization section no longer says "fixed-width". The `0000..=9999` qualification in the type docs now stands without being contradicted elsewhere.

crates/tick/src/fmt/ecmascript.rs
sandersaares: [Copilot speaking] Applied the first two in e6cd8381:
- Line 19 and the "# UTC and time zones" section now consistently use "the UTC designator `Z`".
- The Parsing sentence now reads "...or ISO 8601 timestamp; the ECMAScript profile is a subset of both."

On the third: keeping the spaced hyphens rather than switching to em dashes, as a matter of house style here.

crates/tick/src/fmt/unix_seconds.rs
sandersaares: [Copilot speaking] Confirmed. ADO 7661495 now explicitly covers both new EcmaScript occurrences, which I annotated on the bug: `EcmaScript::to_unix_epoch_duration` (same `unsigned_abs` pattern) and `From<EcmaScript> for UnixSeconds` (built on it). We're deliberately keeping the family-wide behavior consistent for now and fixing it holistically under that bug rather than diverging one format. We chose not to add a characterization test that locks in the known-bad mirroring, to avoid cementing it; the fix under 7661495 will add correct-behavior tests instead.

crates/tick/src/system_time_ext.rs
sandersaares: [Copilot speaking] Fixed in e6cd8381. `display_ecmascript`'s doc now reads "For years `0000` through `9999` the output has the fixed 24-character shape...", matching the `EcmaScript` type docs.

sandersaares:
[Copilot speaking] @ralfbiedert thanks for the thorough review. Disposition of the six findings (fixes are in e6cd8381):

**3. Non-canonical millisecond representation — fixed.** All construction paths (`FromStr`, `TryFrom<SystemTime>`, `From<Iso8601>`, `From<Rfc2822>`, `From<UnixSeconds>`, and the internal `from_timestamp`) now truncate the wrapped `Timestamp` to the millisecond, so equality, ordering, hashing, `Display`, and serde all agree and serde round-trips are identity. Truncation floors via `div_euclid` (toward negative infinity) to match the printer's civil-fraction flooring, so pre-epoch fractions stay correct (`1969-12-31T23:59:59.999999999Z` -> `.999Z`). `MAX` is now the millisecond-aligned boundary so it is itself canonical. Added regression tests for canonical equality, pre-epoch civil flooring, and sub-millisecond serde round-trip identity.

**5. Contradictory fixed-width wording — fixed.** Dropped the unqualified "fixed-width" from the module summary and the Parsing/Serialization sections; the 24-character width is now consistently scoped to years `0000..=9999`, with expanded years called out as wider.

**4. Advertised ECMAScript range unsupported — documented.** The type docs now state the representable year range is `-9999..=9999`, narrower than ECMAScript's own `-271821..=275760`, and that values beyond it (such as `+010000`) are rejected rather than represented, so the `toISOString` claim is scoped to the supported range. This matches the team decision to represent what jiff supports and document the bound rather than widen the backing type.

**1. Pre-epoch conversion (`unsigned_abs`) — tracked, deferred.** This is a pre-existing family-wide latent bug shared with the `Iso8601`/`Rfc2822`/`UnixSeconds` siblings, tracked in ADO 7661495. I annotated that bug with the two new EcmaScript call sites (`EcmaScript::to_unix_epoch_duration` and `From<EcmaScript> for UnixSeconds`). We're keeping the behavior consistent across the family and fixing it holistically there rather than diverging one format in this PR.

**2. Negative-year panic after infallible RFC conversion — tracked, deferred.** Same underlying `Rfc2822` negative-year `Display` panic, tracked in ADO 7661499, which I annotated with the new `From<EcmaScript> for Rfc2822` reachability. Making the RFC conversion a checked `TryFrom` is the family-wide fix under that bug.

**6. Deserialization allocation — tracked.** Filed ADO 7667382 to remove the intermediate `String` allocation across the whole format family (this pattern is shared by all four types), with a request for Criterion + Callgrind benchmarks as evidence.

In short: the precision, range, and wording findings (3, 4, 5) are addressed in this PR; the conversion and allocation findings (1, 2, 6) are pre-existing family-wide issues kept consistent here and tracked in ADO 7661495 / 7661499 / 7667382 for a holistic fix.

wukchung (approved):

crates/tick/src/fmt/unix_seconds.rs
```diff
@@ -58,11 +58,15 @@ pub struct UnixSeconds(pub(super) Duration);
...
-    /// This represents a Unix system time of `31 December 9999 23:59:59 UTC`.
-    // NOTE: This value is aligned with the max jiff timestamp for easier interoperability.
-    pub const MAX: Self = Self(Duration::new(253_402_207_200, 999_999_999));
+    /// This represents a Unix system time of `9999-12-30T22:00:00Z`, the
+    /// whole-second floor of the maximum jiff timestamp.
+    #[expect(
+        clippy::duration_suboptimal_units,
+        reason = "UnixSeconds represents whole seconds since the Unix epoch, so seconds is the natural unit, matching to_secs() and Display"
+    )]
+    pub const MAX: Self = Self(Duration::from_secs(253_402_207_200));
```
martin-kolinek, line 83: 🤖: This PR silently changes the semantics of the already-shipped `UnixSeconds` type, not just adds `EcmaScript`: `MAX` moves from `...200.999999999` to the whole-second floor `...200`, and the new `canonicalize` now floors sub-second `Duration` state in every constructor. The fix itself is right — a whole-second type retaining sub-second state was a latent `Eq`/`Hash`/serde-identity inconsistency — but because the rendered string is unchanged the release tooling won't surface it, while any downstream that compares/hashes/serializes a `UnixSeconds` carrying sub-second precision, or references the old `MAX` value, will observe the change. Please make this a conscious, documented decision (release/changelog note + a sign-off that no consumer relied on the old semantics); ideally split the `UnixSeconds` canonicalization into its own small change so it isn't an invisible side effect of adding EcmaScript.

crates/tick/src/fmt/unix_seconds.rs
```diff
@@ -134,13 +147,14 @@ impl TryFrom<Duration> for UnixSeconds {
     type Error = Error;
 
     fn try_from(value: Duration) -> Result<Self, Self::Error> {
-        if value > Self::MAX.0 {
+        let canonical = Self::canonicalize(value);
```
martin-kolinek, line 167: 🤖: Please add a boundary test for the new canonicalize-before-range-check: `UnixSeconds::try_from(Duration::new(253_402_207_200, 999_999_999))` should now succeed and equal `UnixSeconds::MAX`, since the sub-second part is floored away before the bound check. The existing tests cover exact `MAX`, ordinary fractional durations, and `Duration::MAX`, but none exercises the final fractional-second band just above `MAX.0`; they would still pass if the canonicalize/compare order were swapped and values in that band were wrongly rejected.

crates/tick/src/fmt/ecmascript.rs
```diff
@@ -0,0 +1,408 @@
...
+    ///
+    /// This represents a Unix system time of `9999-12-30T22:00:00.999Z`.
+    pub const MAX: Self = Self(Timestamp::constant(253_402_207_200, 999_000_000));
+
+    /// The Unix epoch represented as `EcmaScript`.
+    ///
+    /// This represents a Unix system time of `1 January 1970 00:00:00 UTC`.
+    pub const UNIX_EPOCH: Self = Self(Timestamp::UNIX_EPOCH);
+
+    pub(super) fn to_unix_epoch_duration(self) -> Duration {
```
psandana: `to_unix_epoch_duration` uses `self.0.duration_since(Timestamp::UNIX_EPOCH).unsigned_abs()`, which discards the sign for pre-epoch instants -- the same known bug already tracked for `Iso8601`/`Rfc2822` as ADO 7661495/7661499. This PR adds a third call site with the identical defect (which `UnixSeconds::from(EcmaScript)` routes through), but there's no code-level pointer tying it to those bugs, and no test covers the pre-epoch case here (only `to_unix_epoch_duration_ok`, which uses a post-epoch instant). Could we extend ADO 7661495/7661499 to explicitly cover `EcmaScript`, or file a new bug, and add a short `// TODO(ADO-xxxx)` comment at this call site so it doesn't get rediscovered independently later?

crates/tick/src/fmt/mod.rs
```diff
@@ -57,12 +65,17 @@
 //! let time = SystemTime::UNIX_EPOCH + Duration::from_hours(1);
 //! println!("Time: {}", time.display_iso_8601());
 //! // Output: Time: 1970-01-01T01:00:00Z
+//!
+//! println!("Time: {}", time.display_ecmascript());
+//! // Output: Time: 1970-01-01T01:00:00.000Z
 //! ```
 
+mod ecmascript;
```
psandana, line 134: Minor naming nit: the module is `ecmascript` (one word) while every sibling format keeps the module name in exact snake_case sync with its type (`iso_8601.rs` maps to `Iso8601`, `rfc_2822.rs` maps to `Rfc2822`, `unix_seconds.rs` maps to `UnixSeconds`). `EcmaScript` would map to `ecma_script.rs` under that convention. Not blocking, but worth a quick check that this is a deliberate convention choice ('ECMAScript' is one word) rather than an oversight, given how strictly this crate otherwise mirrors type/module names.

crates/tick/src/fmt/ecmascript.rs
```diff
@@ -0,0 +1,408 @@
...
+///
+/// `precision(Some(3))` pins the fractional component to exactly three digits and
+/// truncates (does not round) any finer precision. Years outside `0000..=9999` use
+/// the ECMAScript expanded-year form.
+static ECMASCRIPT_PRINTER: temporal::DateTimePrinter = temporal::DateTimePrinter::new().precision(Some(3));
+
+impl EcmaScript {
+    /// The largest value that can be represented by `EcmaScript`.
+    ///
+    /// This represents a Unix system time of `9999-12-30T22:00:00.999Z`.
```
psandana, line 100: Every constructor path (`FromStr`, `TryFrom<SystemTime>`, the cross-format `From` impls, `from_timestamp`) silently floors to millisecond precision via `canonicalize`. That's documented at the struct level, but since rustdoc renders each trait impl's docs separately, a reader landing directly on `impl From<Iso8601> for EcmaScript` (for example) won't see that caveat. Worth adding a one-line 'truncates to millisecond precision' note to each `From` impl doc, or at least the ones most likely to carry sub-millisecond input (`Iso8601`)?

crates/tick/src/fmt/ecmascript.rs
```diff
@@ -0,0 +1,408 @@
...
+/// Years outside `0000..=9999` render in the ECMAScript expanded-year form - a
+/// sign and six year digits, e.g. `-009999-01-02T01:59:59.000Z` - matching the
+/// shape the ECMAScript `Date.prototype.toISOString` method produces. Such years
+/// are reachable through any constructor (for example [`FromStr`]), not only by
+/// saturation, so the 24-character width is a property of the `0000..=9999` range,
+/// not an invariant of the type.
+///
+/// The representable year range is `-9999..=9999`, narrower than the
+/// `-271821..=275760` range ECMAScript itself supports; years ECMAScript can
+/// express beyond it (such as `+010000`) are rejected rather than represented.
```
psandana: Question rather than a defect: `EcmaScript` only represents years `-9999..=9999`, a small fraction of the full ECMAScript range (`-271821..=275760`) that `Date.prototype.toISOString` actually supports. That's clearly documented and mirrors the existing `Iso8601` fallible/saturating split, so it's internally consistent -- but given the type is named directly after ECMAScript conformance, is the narrower range an intentional, permanent characteristic we're comfortable with, or something we'd want to flag as a known limitation more prominently (e.g., a short 'this is permanent, not a TODO' note) so it isn't mistaken for an oversight later?

crates/tick/src/fmt/ecmascript.rs
```diff
@@ -0,0 +1,408 @@
...
+
+/// Parser and formatter for the ECMAScript Date Time String Format.
+///
+/// For years 0000 through 9999 the output has the fixed 24-character shape
+/// `YYYY-MM-DDTHH:MM:SS.sssZ`: a four-digit year, two-digit calendar and clock
+/// fields, exactly three fractional digits (milliseconds, truncated rather than
+/// rounded), and the UTC designator `Z`. For example: `2024-08-06T21:30:00.123Z`.
+///
+/// Years outside `0000..=9999` render in the ECMAScript expanded-year form - a
+/// sign and six year digits, e.g. `-009999-01-02T01:59:59.000Z` - matching the
```
psandana: Nit: 'a sign and six year digits' reads a little awkwardly -- 'a sign and a six-digit year' is more natural and matches how ISO 8601/ECMAScript spec prose usually phrases the extended year form. Same phrase is repeated verbatim in the test comment at line 330, so worth updating both together.

crates/tick/src/fmt/unix_seconds.rs
sandersaares: [Copilot speaking] This is no longer applicable after merging `main`: direct format-to-format conversions were removed, along with `EcmaScript::to_unix_epoch_duration` and `From<EcmaScript> for UnixSeconds`. Conversions now go through the explicit fallible `SystemTime` bridge, and `UnixSeconds::try_from` rejects pre-epoch values.

crates/tick/src/fmt/mod.rs
```diff
@@ -57,12 +65,17 @@
 //! let time = SystemTime::UNIX_EPOCH + Duration::from_hours(1);
```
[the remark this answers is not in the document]
sandersaares: [Copilot speaking] No change is needed. The workspace MSRV is Rust 1.93 (`Cargo.toml`), where `std::time::Duration::from_hours` is stable. I compiled a direct probe with the installed 1.93 toolchain, and the complete `tick` unit and documentation test suites pass under Rust 1.93.

crates/tick/src/system_time_ext.rs
```diff
@@ -22,6 +22,32 @@ pub trait SystemTimeExt: sealed::Sealed {
...
+    /// renders in the wider ECMAScript expanded-year form rather than the fixed 24-character shape.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use std::time::{Duration, SystemTime};
+    ///
+    /// use tick::SystemTimeExt;
+    ///
+    /// let time = SystemTime::UNIX_EPOCH + Duration::from_hours(1);
```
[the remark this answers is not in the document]
sandersaares: [Copilot speaking] No change is needed. The workspace MSRV is Rust 1.93 (`Cargo.toml`), where `std::time::Duration::from_hours` is stable. I compiled a direct probe with the installed 1.93 toolchain, and the complete `tick` unit and documentation test suites pass under Rust 1.93.

crates/tick/src/fmt/ecmascript.rs
```diff
@@ -0,0 +1,408 @@
...
+        assert_eq!(whole.to_string().len(), fractional.to_string().len());
+        assert_eq!(fractional.to_string(), "1970-01-01T00:00:00.000Z");
+    }
+
+    #[test]
+    fn parse_then_display() {
+        let stamp: EcmaScript = "1970-01-01T01:00:00Z".parse().unwrap();
+
+        assert_eq!(stamp.to_string(), "1970-01-01T01:00:00.000Z");
+        assert_eq!(SystemTime::from(stamp), SystemTime::UNIX_EPOCH + Duration::from_hours(1));
```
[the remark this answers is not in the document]
sandersaares: [Copilot speaking] No change is needed. The workspace MSRV is Rust 1.93 (`Cargo.toml`), where `std::time::Duration::from_hours` is stable. I compiled a direct probe with the installed 1.93 toolchain, and the complete `tick` unit and documentation test suites pass under Rust 1.93.

crates/tick/src/fmt/unix_seconds.rs
sandersaares: [Copilot speaking] This is now a conscious and visible change. The canonicalization is isolated in commit `e0fcfae4` (`fix(tick): normalize UnixSeconds to whole-second precision`), so the release tooling can classify it independently from the ECMAScript feature. The PR description now has a dedicated "Unix seconds consistency" section describing the equality/hash/serialization and `MAX` changes. A repository-wide search found no consumer outside `tick` relying on sub-second `UnixSeconds` state or the previous fractional `MAX`.

crates/tick/src/fmt/unix_seconds.rs
sandersaares: [Copilot speaking] Added in `88860829` as `try_from_fractional_jiff_max_floors_to_max`. It passes `Duration::new(253_402_207_200, 999_999_999)` through `TryFrom<Duration>` and asserts that canonicalization succeeds with `UnixSeconds::MAX`.

crates/tick/src/fmt/ecmascript.rs
sandersaares: [Copilot speaking] This is no longer applicable after merging `main`: direct format-to-format conversions were removed, along with `EcmaScript::to_unix_epoch_duration` and `From<EcmaScript> for UnixSeconds`. Conversions now go through the explicit fallible `SystemTime` bridge, and `UnixSeconds::try_from` rejects pre-epoch values.

crates/tick/src/fmt/mod.rs
sandersaares: [Copilot speaking] Deliberate choice: "ECMAScript" is a single word in the specification and public documentation, so its conventional Rust module spelling is `ecmascript`, matching the terminology rather than splitting it into `ecma_script`.

crates/tick/src/fmt/ecmascript.rs
sandersaares: [Copilot speaking] This is no longer applicable after merging `main`. The direct `From<Iso8601>`, `From<Rfc2822>`, and `From<UnixSeconds>` implementations were removed; format conversions now go through `SystemTime`. The remaining public construction paths document the canonical millisecond representation at the type level.

crates/tick/src/fmt/ecmascript.rs
sandersaares: [Copilot speaking] The narrower range is intentional and is treated as a permanent API characteristic, not a pending TODO. The public docs now state prominently that ECMAScript itself supports `-271821..=275760`, while this type is bounded by jiff's `-9999..=9999` range and may be narrowed further by the platform's `SystemTime` range.

crates/tick/src/fmt/ecmascript.rs
sandersaares: [Copilot speaking] Fixed in `88860829`: the documentation now says "a sign and a six-digit year."

crates/tick/src/system_time_ext.rs
```diff
@@ -22,6 +22,32 @@ pub trait SystemTimeExt: sealed::Sealed {
...
+    /// assert_eq!(
+    ///     time.display_ecmascript().to_string(),
+    ///     "1970-01-01T01:00:00.000Z"
+    /// );
+    /// ```
+    #[cfg(any(feature = "fmt", test))]
+    fn display_ecmascript(&self) -> impl std::fmt::Display;
```
martintmk, line 50: this must not fail for any valid system time, this is a once place where saturating behavior can be observed. Let's verify this in tests, system time representing year 10001 is displayed as saturated ECMA

martintmk (approved):

psandana (approved):

crates/tick/src/system_time_ext.rs
sandersaares: [Copilot speaking] Added explicit coverage in `fd9295f4` and retained it after the Clippy follow-up in `2854c5e2`. `display_ecmascript_saturates_year_10001` constructs the `SystemTime` for `10001-01-01T00:00:00Z` and verifies that `display_ecmascript()` saturates it to `9999-12-30T22:00:00.999Z`. The test passes on Windows and Linux.

PR #613, opened 2026-07-28
feat(heapwatch): add crate scaffolding and initial design

martintmk:
Introduces `heapwatch`, a `#[global_allocator]` wrapper that continuously accounts heap usage -- live bytes, cumulative bytes allocated and freed, and allocation / free / reallocation counts.

This PR lands **the design ahead of the code**, so the architecture can be reviewed on its own terms rather than through an implementation diff. The crate compiles and is registered in the workspace, but exposes no public API yet. The implementation follows in a separate change.

## What is heapwatch

It answers *how much*, never *who*. No backtraces, no per-allocation metadata, no pointer-to-size shadow map -- which is what keeps the per-allocation cost to a few non-atomic adds, cheap enough to leave enabled in production.

Each thread accumulates its own totals without synchronization and publishes them in batches into atomic totals owned by the `HeapWatch` instance, once churn since its last publication crosses a compile-time threshold. That removes the atomic read-modify-write per allocation that an exact accounting wrapper pays, in exchange for a bounded, stated inaccuracy of at most `threshold x live threads` -- a bound that does not grow with uptime. Reading goes through a `Watch` handle obtained from the instance, and is O(1) in the thread count.

## Contents

| File | What it is |
|---|---|
| `crates/heapwatch/docs/DESIGN.md` | The shipped architecture: measurement boundary, the per-thread / batched-commit mechanism and rejected alternatives, ordering and signedness, the instance-owned public surface with registration and reading examples, the accuracy contract, re-entrancy, operation counting, edge cases, invariants, and the verification strategy. |
| `crates/heapwatch/docs/TODO.md` | Ideas deliberately excluded from the design (peak tracking, reported error bar, telemetry integration, usable-size accounting, scoped accounting, sampled attribution, size histogram), each with its reasoning. |
| `crates/heapwatch/src/lib.rs` | Crate-level rustdoc summarising the above; no items. |
| `crates/heapwatch/Cargo.toml` | Package metadata at `0.1.0`, no dependencies. |
| `crates/heapwatch/README.md` | Generated by `cargo doc2readme` against `crates/README.j2`. |
| `crates/heapwatch/CHANGELOG.md` | Seeded, as for other brand-new crates. |
| `Cargo.toml` / `Cargo.lock` | Registered in `[workspace.dependencies]`. |
| `.spelling` | Adds `heapwatch` / `Heapwatch`. |

## Notes for reviewers

- **No tests.** There is no behavior yet; the compile and lint gates are the verification. Coverage is unaffected -- a crate with no executable lines contributes nothing to the 100% project target.
- **No `logo.png` / `favicon.ico`.** Each crate has bespoke artwork, so rather than copy another crate's, the `html_logo_url` / `html_favicon_url` doc attributes are omitted for now (matching `data_privacy_core` and `multitude_macros`).
- The design notes that heapwatch requires `std` -- a deliberate exception to the workspace `no_std` preference, since the mechanism needs a thread-local with a thread-exit destructor.
- **Peak is deferred, not dropped forever.** It was cut from the design in review: a consumer that samples live bytes can keep its own maximum, and folding it in-crate drags along a reset operation and a subtler contract, since a peak candidate mixes a local maximum and a total read at different instants and so acquires a *persistent* bias where the other counters carry only a transient one. The rationale is recorded as item 1 in `TODO.md`.

## Validation

Run locally against the workspace:

- `cargo check --workspace --all-features --all-targets --locked`
- `cargo clippy -p heapwatch --all-features --all-targets --locked`
- `cargo doc -p heapwatch --no-deps --all-features` with `RUSTDOCFLAGS=-D warnings`
- `cargo fmt -p heapwatch -- --config-path ./unstable-rustfmt.toml --check`
- `cargo sort --check --grouped --workspace`
- `cargo heather` (license headers)
- `cargo doc2readme --check --lib --template ../README.j2`

`cargo spellcheck` could not run locally (its `libclang` build dependency is unavailable on this machine); CI covers it.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -78,6 +78,7 @@ Fundle
 Fundle's
 HRESULT
 Hardcoded
+Heapwatch
 How-tos
 Hyper
 IOCP
@@ -345,6 +346,7 @@ growable
 hashbrown
 hasher
 hashers
+heapwatch
 hedges
 honour
 hostname
--- a/crates/heapwatch/docs/DESIGN.md
+++ b/crates/heapwatch/docs/DESIGN.md
@@ -69,6 +69,16 @@ classes. This is inherent to the trait, not a shortcut, and it says nothing
 about how the total compares to RSS: heap pages requested but never touched, or
 swapped out, are not resident at all.
 
+**Demand, not footprint.** Because the wrapper counts what callers asked for,
+its numbers are invariant across inner allocators: swapping the system allocator
+for mimalloc leaves every counter identical, since the call pattern did not
+change. That invariance is deliberate — it is what makes heapwatch the control
+variable when something else changes — but it also means heapwatch alone cannot
+say whether an allocator swap reduced the memory a process actually holds.
+Retention, arena caching, and fragmentation are supply-side quantities, visible
+only to the allocator or the OS. Pairing the two is sketched under *Allocator
+footprint reporting* in [`TODO.md`](./TODO.md).
+
 ## The mechanism
 
 One idea: **accumulate per thread without synchronization, publish in batches**.
--- a/crates/heapwatch/docs/DESIGN.md
+++ b/crates/heapwatch/docs/DESIGN.md
@@ -93,8 +93,8 @@ One idea: **accumulate per thread without synchronization, publish in batches**.
         │  live · bytes · op counts         │
         └───────────────┬───────────────────┘
                         │  churn ≥ threshold ─┐
-                        │  thread exit ───────┼─► one batched commit
-                        │  explicit flush ────┘
+                        │                     ├─► one batched commit
+                        │  thread exit ───────┘
                         ▼
         ┌───────────────────────────────────┐
         │  atomic totals owned by the       │
@@ -235,20 +235,25 @@ The rest of the surface:
   before/after comparison across an allocator change a dashboard filter rather
   than a reconstruction from deployment timestamps.
 - **`watch()`** — a `Copy + Send + Sync` handle valid for the life of the
-  process. `stats()` and `flush_thread()` are callable on the instance and on
-  the handle alike.
+  process. `stats()` is callable on the instance and on the handle alike.
 - **`stats()`** — a plain `Copy` snapshot: current bytes, cumulative bytes
   allocated and freed, allocation, free, and reallocation counts, plus the
   derived net allocation count. It is `#[non_exhaustive]`, so adding a counter
   later stays additive.
-- **`flush_thread()`** — fold the caller's pending values in before reading.
+
+**There is deliberately no flush operation.** Committing is entirely internal —
+the churn threshold and thread exit — so a caller has one verb, *read*, and
+nothing to sequence before it. Exposing a flush would offer callers a way to
+tighten a bound that is already documented and small, in exchange for an API
+whose correct use is subtle: it only affects the calling thread, so tightening
+a process-wide reading would mean calling it on every thread, which is not
+something a diagnostic endpoint can do.
 
 **Concurrency contract.** Every entry point is callable from any thread at any
-time; none blocks, none allocates, and none can fail. `flush_thread()` touches
-only the calling thread's block, and is the one operation no other thread can
-perform on its behalf. Concurrent readers and writers otherwise race in the
-ordinary way; since the counters are a gauge rather than a protocol, that is not
-a correctness problem, only the imprecision the accuracy section bounds.
+time; none blocks, none allocates, and none can fail. Concurrent readers and
+writers race in the ordinary way; since the counters are a gauge rather than a
+protocol, that is not a correctness problem, only the imprecision the accuracy
+section bounds.
 
 Heapwatch requires `std`, a deliberate exception to the workspace's `no_std`
 preference: the mechanism needs a thread-local with a thread-exit destructor,
@@ -278,10 +283,10 @@ gigabytes.
 
 Two caveats. The bound does not *converge*: a thread that parks below the
 threshold hides its residue for as long as it sleeps, so a large idle pool can
-hold the full bound indefinitely — which is why a worker about to block should
-flush first. And it is stated in live threads, a quantity `stats()` does not
-report, so a caller wanting a numeric error bar must supply the thread-count
-ceiling from its own deployment knowledge.
+hold the full bound indefinitely, and with no flush operation there is nothing a
+reader can do about it. And it is stated in live threads, a quantity `stats()`
+does not report, so a caller wanting a numeric error bar must supply the
+thread-count ceiling from its own deployment knowledge.
 
 Lowering the threshold tightens the bound and raises the per-allocation share of
 the atomic cost; raising it does the reverse. Zero flushes every operation:
--- a/crates/heapwatch/docs/DESIGN.md
+++ b/crates/heapwatch/docs/DESIGN.md
@@ -1,0 +1,440 @@
+# Heapwatch — Architecture
+
+How the heap accountant works, and the accuracy contract its mechanism buys.
+Implementation-agnostic — for the API see the crate-level rustdoc, for
+forward-looking ideas see [`TODO.md`](./TODO.md).
+
+## What heapwatch is
+
+Heapwatch **wraps a global allocator and accounts the volume passing through
+it**. Installed as a binary's `#[global_allocator]`, it reports how many bytes
+that binary's Rust heap holds, how high it has been, and how much has moved
+through it — continuously, in production, at a cost low enough that leaving it
+on is not a decision anyone has to weigh.
+
+It sits among four neighbours:
+
+- Unlike a **heap profiler** (`dhat`, `heaptrack`, jemalloc's profiler), it does
+  not attribute bytes to call sites — no backtraces, no per-allocation metadata,
+  no pointer-to-size shadow map. It answers *how much*, never *who*, which is
+  what lets its per-allocation cost be a few adds instead of a stack walk.
+- Unlike an **exact accounting wrapper** (`stats_alloc`), it does not pay an
+  atomic read-modify-write on a shared cache line per allocation, trading a
+  small, bounded, stated inaccuracy for removing the one cost that scales badly
+  with core count.
+- Unlike **allocator-native statistics** (jemalloc's `mallctl`, mimalloc's
+  stats), it is independent of what is underneath — including the system
+  allocator, where no such statistics exist — and counts *requested* bytes,
+  the figure a code change moves. Native counters also report retention and
+  fragmentation, which heapwatch cannot see; the two are complementary.
+- Unlike **OS-level metrics** (RSS, working set, cgroup counters), it isolates
+  the Rust heap from stacks, static data, code, and non-Rust allocations.
+
+The design target is a long-running server: a heap in gigabytes, hundreds of
+threads, millions of allocations per second, and an operator who wants a gauge
+to alert on and a regression signal to bisect.
+
+Four properties hold regardless of workload:
+
+- **Transparency** — the wrapper never changes what the inner allocator does,
+  which pointer it returns, or which layout it sees.
+- **The recording path is allocation-free and cannot fail** — a fixed number of
+  arithmetic operations, never a call back into the allocator, never a panic.
+- **The error does not grow with uptime** — it is bounded by the thread count
+  and the flush threshold, neither of which depends on how long the process has
+  been running.
+- **Reading is O(1) in the thread count** — a fixed number of relaxed loads.
+
+Everything else in this document is the price of those four.
+
+## The measurement boundary
+
+The boundary is **successful calls through this `GlobalAlloc`**.
+
+**Inside**: every allocation Rust's heap routes through the registered global
+allocator — `Box`, `Vec`, `String`, `Arc`, collection growth, and the same
+inside any dependency that uses the standard allocator.
+
+**Outside**: allocations made by native or FFI dependencies that manage their
+own memory, even when Rust triggered the work; direct OS calls (`mmap`,
+`VirtualAlloc`); anything deliberately routed to `std::alloc::System`, which
+bypasses the registered allocator entirely — including Rust's own thread-local
+bookkeeping; thread stacks, static data, and the executable image; and the inner
+allocator's own metadata, free-list retention, and fragmentation.
+
+**Counted as requested, not as reserved.** `GlobalAlloc` hands the wrapper a
+`Layout` — what the caller *asked for*. Every real allocator rounds that up to a
+size class, and the trait has no hook to ask what the rounded figure was. Each
+individual allocation is therefore recorded below what the allocator actually
+reserved for it, by a margin set by the inner allocator's size classes. This is
+inherent to the trait, not a shortcut, and it says nothing about how the total
+compares to RSS: heap pages that are requested but never touched, or swapped
+out, are not resident at all.
+
+## The mechanism
+
+One idea: **accumulate per thread without synchronization, publish in batches**.
+
+```text
+        alloc / dealloc / realloc
+                    │
+                    │  plain, non-atomic adds — no atomics, no shared state
+                    ▼
+        ┌───────────────────────────────────┐
+        │  Pending (this thread only)       │ ◄── reset to zero after a flush
+        │  live · peak · bytes · op counts  │
+        └───────────────┬───────────────────┘
+                        │  churn ≥ threshold ─┐
+                        │  thread exit ───────┼─► one batched commit
+                        │  explicit flush ────┘
+                        ▼
+        ┌───────────────────────────────────┐
+        │  process-wide atomic totals       │
+        │  (one fixed set: no shard         │
+        │   registry, no aggregation walk)  │
+        └───────────────┬───────────────────┘
+                        │  a handful of relaxed loads
+                        ▼
+                     stats()
+```
+
+**Churn, not live bytes, triggers the flush.** A thread publishes once *bytes
+allocated plus bytes freed* since its last flush reaches the threshold. Churn,
+because a thread that allocates and frees in a loop barely moves `live` while
+still running the cumulative counters up — and because `|live| ≤ churn`, so
+bounding churn bounds both. A thread also flushes when it exits, or live bytes
+would drift upward for the life of the process as threads came and went.
+
+### One window, end to end
+
+```text
+ alloc(4 KiB) ........ thread block: live +4096, allocated +4096, alloc_ops +1,
+                       local peak raised if live is a new local high.
+                       Nothing global moves — stats() cannot see this yet.
+ ~1000 more ops ...... churn accumulates; between operations the residue is
+                       always below the threshold. This residue IS the error.
+ op crossing 64 KiB .. one batched commit:
+                         · one fetch_add per cumulative counter
+                         · base = CURRENT.fetch_add(live)   ← base read HERE
+                         · PEAK.fetch_max(base + local peak)
+                       thread block resets to zero; error returns to zero
+ dealloc(4 KiB) ...... live −4096, freed +4096, free_ops +1; no peak work
+ thread exits ........ the destructor commits whatever remains
+ stats() ............. relaxed loads of the committed totals only
+```
+
+### Cost
+
+| Path | How often | Work |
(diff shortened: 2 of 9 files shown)


.spelling
```diff
@@ -78,6 +78,7 @@ Fundle
 Fundle's
 HRESULT
 Hardcoded
+Heapwatch
 How-tos
 Hyper
```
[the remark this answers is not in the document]
martintmk: &#x1F916; I don't think this change is needed. `cargo spellcheck` never sees `crates/*/docs/*.md`, and there are three independent reasons `HeapWatch` cannot fail it:

1. **Empirically disproven by this PR's own CI.** The `spell-check` job passed on the first push (commit `7a31d994`), which already contained `HeapWatch<HeapWatch<_>>` in `DESIGN.md`. That same file also contains `mallctl`, `nallocx`, `heaptrack`, `dhat`, `cgroup` and `mi_usable_size` -- none of which are in `.spelling`. If `DESIGN.md` were in scope, all six would have failed first.

2. **`cargo spellcheck` operates on cargo targets**, i.e. Rust doc comments plus the README (`skip_readme = false` in `spellcheck.toml`). Loose Markdown under a crate's `docs/` directory is not a cargo target and is not passed to the checker. The `Heapwatch` / `heapwatch` entries were added for the crate-level rustdoc in `src/lib.rs` and the README generated from it, which are in scope.

3. **Even in a file that were checked, it would still pass twice over.** `HeapWatch<HeapWatch<_>>` is inside a Markdown code span, which cargo-spellcheck skips; and `spellcheck.toml` sets `allow_concatenation = true`, so the CamelCase identifier decomposes into `Heap` + `Watch`, both ordinary dictionary words.

Adding `HeapWatch` to `.spelling` would therefore be a dead entry, and renaming the identifier in the design doc would make it disagree with the type name the implementation will actually use.

---

crates/heapwatch/docs/DESIGN.md
```diff
@@ -0,0 +1,449 @@
+# Heapwatch — Architecture
```
martintmk, line 1: Feedback: 

- lets simplify and drop peak counting for now, consumer of the metrics can keep track of that
- rather than having a bunch of statics lets introduce some instance based handle that can be used to retrieve the data
- the design should include snippet of public surface for registration of tracker and also reading part
- i feel the design is still overly verbose and too "wordly", shorten it

martintmk: &#x1F916; Agreed - working on this now.

Plan:
- Drop peak tracking entirely (consumers can derive it from the metrics).
- Replace the static-based API with an instance handle that owns registration and reading.
- Add concrete public-surface snippets for both registration and reading.
- Tighten the prose throughout.

---

martintmk, line 1: genuine concern with the design.

This crate tracks allocated bytes, however when i switch from standard allocator so something like mimalloc the caller still call the same APis so numbers are identical. However, mimalloc can retain it's own internal caches that won't be tracked. 

How is it possible to calculate the impact of mimalloc on memory used by the particular dll?

Scenario I want to cover is:

- Process that uses FFI to rust library compiled with standard allocator. We can track the memory allocated by the allocator.

- Switch rust library to mimalloc. We want to measure memory usage of the dll, including memory retained by mimalloc. 

Can you analyze and suggest a solution?

martintmk: &#x1F916; Fixed in 79e8ca9.

- **Peak dropped.** Gone from the mechanism, the accuracy table, the counters, the public surface, and the invariant list. Moved to `TODO.md` as deferred work, with the reasoning for deferring: a consumer that samples live bytes can keep its own maximum, and doing it in-crate drags in a reset operation plus a subtler contract, since a candidate mixes a local peak and a total read at different instants and so acquires a *persistent* bias where the other counters carry only a transient one.
- **Instance-owned counters.** A `HeapWatch` now owns its totals instead of writing to process-wide statics, and reading goes through a `Watch` handle obtained from the instance. The handle has no type parameters, so a diagnostic endpoint takes a `Watch` rather than a generic `&HeapWatch<A, N>`. Added a short section on which totals a thread's pending block commits to, since the block is one thread-local rather than one per instance.
- **Public surface snippets.** Two examples: registration (`#[global_allocator]` `static` with the `const` constructor, plus the explicit-threshold form) and reading (a `report(Watch)` function calling `stats()`).
- **Shortened.** 3872 to 3121 words, despite adding the public-surface section and both examples.

`README.md` was regenerated with `cargo doc2readme` after the rustdoc edit.

---

martintmk: &#x1F916; Genuine, and the design was only implying it. Analysis and a proposal below; docs updated in c457298.

## Why the numbers do not move

Heapwatch measures **demand** -- the `Layout` the caller asked for. `GlobalAlloc` has no hook to ask what the allocator rounded that up to, let alone what it kept. So a mimalloc swap leaves every counter bit-identical: same call sites, same requested sizes. What changes is **supply** -- size-class rounding, free-list retention, arena caching, purge policy, fragmentation, metadata -- and all of it is invisible from inside the wrapper.

That invariance is worth keeping rather than fixing: it is what makes heapwatch the *control variable*. When you A/B two allocators, identical heapwatch numbers are the proof that the workload really was identical, so any RSS delta is attributable to the allocator rather than to drift in the benchmark.

## Where the missing number comes from

The allocators already report it:

- **mimalloc** -- `mi_process_info(...)` gives `current_commit` / `peak_commit` (precise on Windows, estimated elsewhere) plus `current_rss` / `peak_rss`. Call `mi_stats_merge()` first so thread-local stats are folded in.
- **jemalloc** -- `mallctl` gives `stats.allocated` (equivalent to heapwatch's live bytes), `stats.active`, `stats.resident`, `stats.mapped`, `stats.retained`. Needs an `epoch` advance to refresh.
- **system allocator** -- nothing comparable, which is the asymmetry that makes your baseline awkward. See below.

`committed / live_requested` is the allocator's overhead factor, and it is exactly the figure that moves on a swap.

## The DLL attribution question

The key point for your scenario: **the allocator instance is the attribution boundary, and it is a better one than anything the OS offers.** There is no per-DLL memory accounting in Windows -- once a page is committed, it belongs to the process, not to the module that asked for it. But if mimalloc is statically linked into the Rust library and the host process does not share it, then mimalloc's committed figure *is* that library's heap footprint, retention included. You get per-component attribution precisely because the allocator is per-component.

So for the second half of your A/B, the answer is direct: link mimalloc into the Rust DLL only, and read `mi_process_info`.

## The baseline half is the harder half

With the system allocator the Rust code shares the host process heap, so there is no symmetric supply-side number to compare against. Three options, in increasing order of effort:

1. **Process-level A/B.** Run identical load in both builds, measure process private commit / working set, and use heapwatch to prove the workload matched. Crude, but it answers the question you actually asked -- did the swap cost or save memory -- and needs no new code.
2. **Give the baseline an isolated heap too.** Wrap a private Windows heap (`HeapCreate`) as a `GlobalAlloc` and read its committed size. Now both arms are per-component and symmetric. This is a test-only allocator, not something to ship.
3. **Skip the system allocator as a baseline** and A/B two reporting allocators (mimalloc vs snmalloc vs jemalloc). If the real decision is "which allocator", the system allocator only needs to appear in the process-level measurement from option 1.

## Measurement protocol

Retention is partly *policy*, not waste, so a single sample conflates the two. Reach steady state, quiesce, sample; then force a collect (`mi_collect(true)`, or jemalloc's purge) and sample again. The gap between the two readings is reclaimable cache; what survives the collect is genuine fragmentation. Reporting only the first number makes every caching allocator look worse than it is.

## What changed in the docs

- `DESIGN.md`, measurement boundary: a new **Demand, not footprint** paragraph saying outright that the counters are invariant across inner allocators, that the invariance is deliberate, and that heapwatch alone therefore cannot say whether a swap reduced memory.
- `TODO.md` item 8, **Allocator footprint reporting**: sketches the pairing -- an optional trait an inner allocator implements to report committed bytes, queried on demand rather than per allocation, so demand and footprint are reported side by side. Notes the per-component attribution property, and why it has to stay optional (the figures' cost and freshness vary by allocator, and the system allocator has no equivalent).

I kept it in `TODO.md` rather than the shipped design since it is a real feature with its own trait and cross-allocator concerns, not a tweak. Happy to promote it into this PR if you would rather the trait land with the design.

---

crates/heapwatch/docs/DESIGN.md
```diff
@@ -0,0 +1,410 @@
...
+
+The rest of the surface:
+
+- **`HeapWatch::new(inner, name)`** — `const`, generic over the inner allocator
+  and the threshold. `name()` returns the stored name: heapwatch emits no
+  telemetry itself, so whatever does emit can tag the series with it, making a
+  before/after comparison across an allocator change a dashboard filter rather
+  than a reconstruction from deployment timestamps.
+- **`watch()`** — a `Copy + Send + Sync` handle valid for the life of the
+  process. `stats()` and `flush_thread()` are callable on the instance and on
```
martintmk: I don't want any `flush_thread` like APIs that should be callable by the application. The only thing we care is retrieve status, even thought these might be not completely accurate. No explicit flushing, keep it simple. Potential measurement error is documented.

martintmk: &#x1F916; Agreed - dropping it. `stats()` and the handle stay; no caller-facing flush. Flushing becomes purely internal (churn threshold plus thread exit), and the bound is documented. Working on it now.

---

martintmk: &#x1F916; Fixed in 228dfbb.

Surface is now `HeapWatch::new` / `name()` / `watch()` / `stats()` -- one verb, read. Committing is entirely internal: churn threshold plus thread exit.

Also removed the explicit-flush arrow from the mechanism diagram, the flush clause from the concurrency contract, and `or on request` from the rustdoc (README regenerated). The accuracy section previously advised flushing before a worker blocks; that caveat now just states the consequence -- an idle pool can hold the full bound for as long as it sleeps, and nothing can be done about it.

Added a short note recording *why* there is no flush, so it does not get re-proposed: it only affects the calling thread, so tightening a process-wide reading would mean calling it on every thread, which a diagnostic endpoint cannot do -- all to tighten a bound that is already documented and small.

---

crates/heapwatch/docs/DESIGN.md
```diff
@@ -0,0 +1,415 @@
...
+forward-looking ideas see [`TODO.md`](./TODO.md).
+
+## What heapwatch is
+
+Heapwatch **wraps a global allocator and accounts the volume passing through
+it**. Installed as a binary's `#[global_allocator]`, it reports how many bytes
+that binary's Rust heap holds and how much has moved through it — continuously,
+in production, at a cost low enough to leave on.
+
+It sits among four neighbours:
```
sandersaares, line 14: How does this compare to `alloc_tracker`? Are we reinventing some wheels here? It is not clear why we need yet another memory tracker. If there is something novel this design adds, we should make it clear what that novelty is and why it cannot be solved with existing tools. This is the 2nd seeming duplication of an existing OSS crate I see today - our mission is not to reinvent the universe, so I find this general approach of just adding yet another tool to our already large list of crates dubious (but perhaps a clear motivation here will help).

crates/heapwatch/docs/DESIGN.md
```diff
@@ -0,0 +1,415 @@
+# Heapwatch — Architecture
+
+How the heap accountant works, and the accuracy contract its mechanism buys.
+Implementation-agnostic — for the API see the crate-level rustdoc, for
+forward-looking ideas see [`TODO.md`](./TODO.md).
+
+## What heapwatch is
```
sandersaares, line 7: There is a lot of text here about what it is but I do not see text about what the actual use cases or problems to solve are. What is the specific motivation? It is not possible today to judge whether/how the design addresses the problem space.

crates/heapwatch/docs/DESIGN.md
```diff
@@ -0,0 +1,415 @@
...
+                       included. The thread block resets to zero and the error
+                       returns to zero.
+ dealloc(4 KiB) ...... live −4096, freed +4096, free_ops +1
+ thread exits ........ the destructor commits whatever remains
+ stats() ............. relaxed loads of the committed totals only
+```
+
+### Which totals a thread commits to
+
+Totals live on the instance, but the pending block is one thread-local rather
```
martintmk, line 133: **Pending thread-local state can outlive its HeapWatch instance**

The design stores the target instance in thread-local pending state and commits to that target later on an instance switch or thread exit, but `HeapWatch::new` is public and does not require static storage. A locally owned allocator can therefore be dropped while pending state still targets its counters, making the later commit dereference dead storage. The same lifetime gap affects the non-generic `Watch` handle described below. Make the invariant enforceable—for example, require a `&'static self` receiver for handles and ensure pending targets can only refer to permanently static storage, or carry the instance lifetime explicitly.

martintmk (commented):
Found two findings.

**The new crate is missing from both root crate indexes**

`heapwatch` is registered in the workspace but omitted from the root README crate list and root CHANGELOG index. The repository's `scripts/add-crate.ps1` onboarding path updates both lists, so the crate will otherwise be absent from the published repository indexes. Add the sorted `heapwatch` entries to both files.

The implemented surface remains only an empty module scaffold (pub mod heapwatch). No vote was submitted because this is my own PR.

geeknoid (commented):
Early feedback on the draft: I found four design issues worth resolving before the implementation fixes the contracts in place. The highest-impact one is that `net_alloc_ops()` derives a leak signal from two independently loaded atomics even though the snapshot contract explicitly forbids equations across fields. I also flagged an API-surface contradiction, the missing verification matrix for the stated invariants, and the absence of a benchmark/regression plan for the per-allocation hot path.

I did not duplicate existing feedback about the instance/`Watch` lifetime hole, missing root README/CHANGELOG index entries, motivation, or overlap with `alloc_tracker`; those discussions remain the authoritative threads. The measurement boundary, requested-versus-reserved distinction, threshold error bound at quiescent boundaries, relaxed ordering for individual gauges, and workspace scaffold/metadata read clean.

All five diff-scoped taxonomies were assessed: 150 security, 162 correctness, 182 testing, 183 performance, and 143 conformance classes. The unsafe/UB protocol was applied to the proposed allocator boundary; because this head contains no implementation, no `unsafe`, FFI, or confirmed UB claim was filed. Nothing was executed, posted, built, tested, benchmarked, profiled, fuzzed, or measured.

crates/heapwatch/docs/DESIGN.md
```diff
@@ -0,0 +1,415 @@
...
+thread-count ceiling from its own deployment knowledge.
+
+Lowering the threshold tightens the bound and raises the per-allocation share of
+the atomic cost; raising it does the reverse. Zero flushes every operation:
+exact accounting with the cost profile of a conventional atomic-per-allocation
+tracker — useful in tests, not in production.
+
+**Snapshot consistency is not claimed.** Reading is a sequence of independent
+relaxed loads, not a stop-the-world snapshot, so fields of one reading may be
+mutually inconsistent by a few operations. Treat each as a gauge; do not solve
```
geeknoid: **The derived net count violates the snapshot contract** — *Correctness · Medium · High confidence*

A commit updates each counter with a separate `fetch_add`, while `stats()` reads each field separately. A reader can therefore load `alloc_ops` on one side of a commit and `free_ops` on the other, then `net_alloc_ops()` subtracts values that never formed one state. This is more than the documented batching error: one commit may represent roughly a threshold's worth of operations, so the derived value can emit a false leak spike or dip even though this paragraph says not to solve equations across fields. The advertised derivation is at [`DESIGN.md:352-358`](https://github.com/microsoft/oxidizer/blob/228dfbb27bfe57f610ce2c3e8bfd0783523b4bc3/crates/heapwatch/docs/DESIGN.md#L352-L358).

**Direction:** make net allocations a directly maintained signed counter updated once per batch, or introduce a coherent snapshot protocol; otherwise remove the derived leak signal from the API.

**Done when:** every advertised snapshot value is either one independently valid gauge or is produced by a documented consistency mechanism.

crates/heapwatch/docs/DESIGN.md
```diff
@@ -0,0 +1,415 @@
...
+
+## Transparency of the wrapper
+
+Pointers and layouts pass through unchanged — no layout adjustment for metadata
+and no per-allocation header, which is also why `dealloc` can rely on the
+caller's `Layout` rather than a shadow map. `realloc` and `alloc_zeroed` are
+forwarded rather than left to their `GlobalAlloc` defaults, which would
+respectively discard the inner allocator's in-place growth and forfeit pages the
+OS already zeroed — regressions far costlier than the tracking itself. The inner
+allocator needs nothing beyond `GlobalAlloc` and is reachable by reference, so
```
geeknoid: **The public surface does not preserve access to allocator controls** — *Correctness · Low · High confidence*

`HeapWatch::new` takes ownership of the inner allocator, and the exhaustive surface immediately above lists only `name()`, `watch()`, and `stats()`. Nothing in that surface makes the inner allocator reachable by reference, so allocator-specific controls such as purge/collection or native statistics become inaccessible through the wrapper despite this guarantee.

**Direction:** add an explicit `inner(&self) -> &A`-style accessor to the settled surface, or document the retained external-handle construction pattern that makes this statement true.

**Done when:** the public-surface section contains a concrete, type-correct route to every inner allocator control the design promises remains usable.

crates/heapwatch/docs/DESIGN.md
```diff
@@ -0,0 +1,415 @@
...
+  "shrink to zero" is not a valid reallocation. Heapwatch neither installs nor
+  intercepts an allocation-error hook; an aborting OOM leaves pending values
+  unpublished.
+- **Counter widths.** Cumulative byte totals are 64-bit and would need centuries
+  at gigabytes per second to wrap; live bytes are signed 64-bit, beyond the
+  reach of any real heap.
+
+## Design invariants at a glance
+
+1. **The accounting path never allocates**, and never panics. Every step is
```
geeknoid: **Most settled invariants have no verification plan** — *Testing · Medium · High confidence*

The PR description says this document includes the verification strategy, but only the TLS first-touch re-entrancy assumption is mapped to a test. The remaining invariants and edge cases have no stated oracle: successful-only accounting, threshold boundaries, cross-thread allocation/free ordering, failed `realloc`, multi-instance retargeting, thread-exit/post-TLS teardown, wrapping behavior, and forwarding transparency. This is not a request to add implementation tests to the scaffold PR; it is a request to settle how the implementation will prove the architecture it inherits.

**Direction:** add a verification matrix mapping each invariant and edge case to a unit/integration/property/model-checking test and the observable assertion it must make.

**Done when:** every numbered invariant and supported edge case names evidence that would fail if that contract were implemented incorrectly.

crates/heapwatch/docs/DESIGN.md
```diff
@@ -0,0 +1,415 @@
...
+mismatch commits the pending values to the previous target and re-targets the
+block. A binary has one `#[global_allocator]`, so the compare is a predictable
+never-taken branch in practice, and the fallback exists only so that a second
+instance cannot silently mix its numbers into the first's.
+
+### Cost
+
+| Path | How often | Work |
+|---|---|---|
+| Record | every operation | one thread-local address, one target compare, a few non-atomic adds, one compare-and-branch on churn |
```
geeknoid: **The core overhead claims have no benchmark or regression plan** — *Performance · Medium · Structural evidence*

This record path runs on every allocation, and avoiding shared atomic traffic is the crate's central production value proposition, yet the design specifies no benchmark matrix, baseline, or regression gate. Without one, claims such as “a few adds”, production suitability, and favorable core-count scaling cannot be accepted or protected. This also leaves [M-HOTPATH](https://microsoft.github.io/rust-guidelines/guidelines/performance/#M-HOTPATH)'s requirement to create benchmarks around identified hot paths only partially met.

**Direction:** define paired wall-clock and instruction/allocation benchmarks against the unwrapped allocator and an exact accounting wrapper, varying thread count, allocation size, and threshold; state which result is gated versus recorded for investigation.

**Done when:** the design names representative benchmark scenarios and an acceptance/regression policy, or explicitly tracks that measurement work as a prerequisite to claiming production-low overhead.

PR #615, opened 2026-07-29
feat!: support no_std in bytesbuf and thread_aware

geeknoid:
## Summary

- add default-on `std` features to `thread_aware` and `bytesbuf`, moving portable code to `core` and `alloc`
- keep per-affinity storage, `GlobalPool`, standard I/O adapters, hosted integrations, and metrics behind `std`
- let `no_std` applications that already own specialized memory providers integrate them through the existing `Memory` APIs
- preserve hosted behavior through explicit feature propagation and gate hosted-only tests, doctests, and benchmarks
- document the `alloc` and pointer-width atomic requirements and the `thread_aware` derive feature recipe

`bytesbuf` intentionally does not provide a general-purpose allocator-backed memory provider without `std`; its `no_std` surface is intended for applications that already own a specialized provider.

## Breaking change

Consumers that disable default features must now explicitly enable `std` to retain `bytesbuf`'s `GlobalPool` and standard I/O APIs or `thread_aware`'s per-affinity `Arc` and hosted implementations.

## Validation

- hosted package tests, doctests, clippy, formatting, generated READMEs, and spelling
- no-default unit-test and doctest compilation for both crates
- `thumbv7em-none-eabihf` checks for portable feature combinations
- no-default benchmark compilation and the hosted `multitude` integration

### DIFF
--- a/crates/bytesbuf/Cargo.toml
+++ b/crates/bytesbuf/Cargo.toml
@@ -32,24 +32,27 @@ allowed_external_types = [
 all-features = true
 
 [features]
-default = []
+default = ["std"]
+# Enables the global memory pool, standard I/O adapters, and metrics. Disable for `#![no_std]`
+# environments; the crate still requires `alloc`.
+std = ["dep:nm", "dep:plurality", "thread_aware/std", "bytes?/std"]
 # Interoperability with the `bytes` crate.
 bytes-compat = ["dep:bytes"]
-test-util = []
+test-util = ["std"]
 
 [dependencies]
-bytes = { workspace = true, features = ["std"], optional = true }
-new_zealand = { workspace = true }
-nm = { workspace = true }
-plurality = { workspace = true }
+bytes = { workspace = true, optional = true }
+nm = { workspace = true, optional = true }
+plurality = { workspace = true, optional = true }
 smallvec = { workspace = true, features = ["const_new", "union"] }
-thread_aware = { workspace = true, features = ["derive"] }
+thread_aware = { workspace = true, default-features = false, features = ["derive"] }
 
 [dev-dependencies]
 alloc_tracker = { workspace = true }
 bytes = { workspace = true, features = ["std"] }
 criterion = { workspace = true }
 mutants = { workspace = true }
+new_zealand = { workspace = true }
 static_assertions = { workspace = true }
 testing_aids = { path = "../testing_aids" }
 
@@ -75,6 +78,7 @@ required-features = ["test-util"]
 [[bench]]
 name = "global_pool"
 harness = false
+required-features = ["std"]
 
 [[bench]]
 name = "view"
@@ -94,6 +98,7 @@ required-features = ["test-util"]
 [[bench]]
 name = "global_pool_cg"
 harness = false
+required-features = ["std"]
 
 [[example]]
 name = "bb_has_memory_forwarding"
--- a/crates/bytesbuf/src/lib.rs
+++ b/crates/bytesbuf/src/lib.rs
@@ -3,8 +3,18 @@
 
 #![cfg_attr(coverage_nightly, feature(coverage_attribute))]
 #![cfg_attr(docsrs, feature(doc_cfg))]
+#![cfg_attr(not(test), no_std)]
 
 //! Types for creating and manipulating byte sequences.
+//!
+//! # Crate features
+//!
+//! * The **`std` Cargo feature** *(enabled by default)* enables the global memory pool, standard I/O
+//!   adapters, and metrics. Disable it for `#![no_std]` environments; the crate still requires
+//!   `alloc` and pointer-width atomics.
+//! * **`bytes-compat`** enables interoperability with the `bytes` crate and is available with or
+//!   without `std`.
+//! * **`test-util`** enables memory providers intended for tests and implies `std`.
 //!
 //! <img src="https://media.githubusercontent.com/media/microsoft/oxidizer/refs/heads/main/crates/bytesbuf/docs/diagrams/Introduction.png" alt="Diagram showing byte sequences inside BytesView and BytesBuf" />
 //!
@@ -35,10 +45,13 @@
 //! * [`get_byte()`] reads a single byte.
 //! * [`copy_to_slice()`] copies bytes into a provided slice.
 //! * [`copy_to_uninit_slice()`] copies bytes into a provided uninitialized slice.
-//! * [`BytesView`] implements [`std::io::Read`] and [`std::io::BufRead`] directly, since
+//! * With the `std` feature, [`BytesView`] implements [`std::io::Read`] and [`std::io::BufRead`]
+//!   directly, since
 //!   it is already a buffered byte sequence.
 //!
 //! ```
+//! # fn main() {
+//! # #[cfg(feature = "std")] {
 //! # let memory = bytesbuf::mem::GlobalPool::new();
 //! # let message = BytesView::copied_from_slice(b"1234123412341234", &memory);
 //! use bytesbuf::BytesView;
@@ -55,6 +68,8 @@
 //!     println!("Message received. The sum of all words in the message is {sum}.");
 //! }
 //! # consume_message(message);
+//! # }
+//! # }
 //! ```
 //!
 //! If the helper methods are not sufficient, you can access the byte sequence behind the [`BytesView`]
@@ -69,6 +84,8 @@
 //!   will return a new slice of bytes starting from the new front position of the view.
 //!
 //! ```
+//! # fn main() {
+//! # #[cfg(feature = "std")] {
 //! # let memory = bytesbuf::mem::GlobalPool::new();
 //! # let mut bytes = BytesView::copied_from_slice(b"1234123412341234", &memory);
 //! use bytesbuf::BytesView;
@@ -86,12 +103,16 @@
 //! }
 //!
 //! println!("Inspected a view over {len} bytes with slice lengths: {slice_lengths:?}");
+//! # }
+//! # }
 //! ```
 //!
 //! To reuse a [`BytesView`], clone it before consuming the contents. This is a cheap
 //! zero-copy operation.
 //!
 //! ```
+//! # fn main() {
+//! # #[cfg(feature = "std")] {
 //! # let memory = bytesbuf::mem::GlobalPool::new();
 //! # let mut bytes = BytesView::copied_from_slice(b"1234123412341234", &memory);
 //! use bytesbuf::BytesView;
@@ -107,6 +128,8 @@
 //!
 //! // Operations on the clone have no effect on the original view.
 //! assert_eq!(bytes.len(), 16);
+//! # }
+//! # }
 //! ```
 //!
 //! # Producing Byte Sequences
@@ -127,12 +150,17 @@
 //!    instance of a shared [`GlobalPool`]. In a typical web application, the global memory pool
 //!    is a service exposed by the application framework. In a different context (e.g. example
 //!    or test code with no framework), you can create your own instance via [`GlobalPool::new()`].
+//!    In a `no_std` environment, applications that already own a specialized memory provider can
+//!    integrate it by implementing [`Memory`]. The crate does not provide a general-purpose
+//!    allocator-backed provider without `std`.
 //!
 //! Once you have a memory provider, you can reserve memory from it by calling
 //! [`Memory::reserve()`] on it. This returns a [`BytesBuf`] with at least the requested
 //! number of bytes of memory capacity.
 //!
 //! ```
+//! # fn main() {
+//! # #[cfg(feature = "std")] {
 //! # struct Connection {}
 //! # impl Connection { fn memory(&self) -> impl Memory { bytesbuf::mem::GlobalPool::new() } }
 //! # let connection = Connection {};
@@ -141,6 +169,8 @@
 //! let memory = connection.memory();
 //!
 //! let mut buf = memory.reserve(100);
+//! # }
+//! # }
 //! ```
 //!
 //! Now that you have the memory capacity in a [`BytesBuf`], you can fill the memory
@@ -158,6 +188,8 @@
 //! * [`put_bytes()`] appends an existing [`BytesView`].
 //!
 //! ```
+//! # fn main() {
+//! # #[cfg(feature = "std")] {
 //! # struct Connection {}
 //! # impl Connection { fn memory(&self) -> impl Memory { bytesbuf::mem::GlobalPool::new() } }
 //! # let connection = Connection {};
@@ -170,6 +202,8 @@
 //! buf.put_u64_be(1234);
 //! buf.put_u64_be(5678);
 //! buf.put_slice(*b"Hello, world!");
+//! # }
+//! # }
 //! ```
 //!
 //! If the helper methods are not sufficient, you can write contents directly into mutable byte slices
@@ -190,6 +224,8 @@
 //! to identify how much unused memory capacity is available.
 //!
 //! ```
+//! # fn main() {
+//! # #[cfg(feature = "std")] {
 //! # struct Connection {}
 //! # impl Connection { fn memory(&self) -> impl Memory { bytesbuf::mem::GlobalPool::new() } }
 //! # let connection = Connection {};
@@ -207,12 +243,16 @@
 //! // Remember that a memory provider can always provide more memory than requested.
 //! assert!(buf.capacity() >= 100 + 80);
 //! assert!(buf.remaining_capacity() >= 80);
+//! # }
+//! # }
 //! ```
 //!
 //! When you have written your byte sequence into the memory capacity of the [`BytesBuf`], you can consume
 //! the data in the buffer as a [`BytesView`].
 //!
 //! ```
+//! # fn main() {
+//! # #[cfg(feature = "std")] {
 //! # struct Connection {}
 //! # impl Connection { fn memory(&self) -> impl Memory { bytesbuf::mem::GlobalPool::new() } }
 //! # let connection = Connection {};
@@ -227,12 +267,16 @@
 //! buf.put_slice(*b"Hello, world!");
 //!
 //! let message = buf.consume_all();
+//! # }
+//! # }
 //! ```
 //!
 //! This can be done piece by piece, and you can continue writing to the [`BytesBuf`]
 //! after consuming some already written bytes.
 //!
 //! ```
+//! # fn main() {
+//! # #[cfg(feature = "std")] {
 //! # struct Connection {}
 //! # impl Connection { fn memory(&self) -> impl Memory { bytesbuf::mem::GlobalPool::new() } }
 //! # let connection = Connection {};
@@ -251,6 +295,8 @@
 //! buf.put_slice(*b"Hello, world!");
 //!
 //! let final_contents = buf.consume_all();
+//! # }
+//! # }
 //! ```
 //!
 //! If you already have a [`BytesView`] that you want to write into a [`BytesBuf`], call
@@ -258,6 +304,8 @@
 //! that reuses the memory capacity of the view you are appending.
 //!
 //! ```
+//! # fn main() {
+//! # #[cfg(feature = "std")] {
 //! # struct Connection {}
 //! # impl Connection { fn memory(&self) -> impl Memory { bytesbuf::mem::GlobalPool::new() } }
 //! # let connection = Connection {};
@@ -272,6 +320,8 @@
 //! let mut buf = memory.reserve(128);
 //! buf.put_bytes(header);
 //! buf.put_slice(*b"Hello, world!");
+//! # }
+//! # }
 //! ```
 //!
 //! Note that there is no requirement that the memory capacity of the buffer and the
@@ -307,6 +357,8 @@
 //! for full code):
 //!
 //! ```
+//! # fn main() {
+//! # #[cfg(feature = "std")] {
 //! # #[derive(Debug)]
 //! # struct ConnectionZeroCounter {
 //! #     connection: Connection,
@@ -322,12 +374,16 @@
 //! # #[derive(Debug)] struct Connection;
 //! # impl Connection { fn write(&mut self, mut _message: BytesView) {} }
 //! # impl HasMemory for Connection { fn memory(&self) -> impl MemoryShared { bytesbuf::mem::GlobalPool::new() } }
+//! # }
+//! # }
 //! ```
 //!
 //! Example of returning a memory provider that performs configuration for optimal memory (see
 //! `examples/bb_has_memory_optimizing.rs` for full code):
 //!
 //! ```
+//! # fn main() {
+//! # #[cfg(feature = "std")] {
 //! # #[derive(Debug)]
 //! # struct UdpConnection {
 //! #     io_context: IoContext,
@@ -373,12 +429,16 @@
 //! # }
 //! # #[derive(Clone, Copy)]
 //! # struct MemoryConfiguration { requires_page_alignment: bool, zero_memory_on_release: bool, requires_registered_memory: bool }
+//! # }
+//! # }
 //! ```
 //!
 //! Example of returning a global memory pool when the type is agnostic toward memory configuration
 //! (see `examples/bb_has_memory_global.rs` for full code):
 //!
 //! ```
+//! # fn main() {
+//! # #[cfg(feature = "std")] {
 //! # #[derive(Debug)]
 //! # struct ChecksumCalculator {
 //! #     memory: GlobalPool,
@@ -392,6 +452,8 @@
 //!         self.memory.clone()
 //!     }
 //! }
+//! # }
+//! # }
 //! ```
 //!
 //! It is generally expected that all types work with byte sequences using memory from any provider.
@@ -484,6 +546,8 @@
 //! intended usage.
 //!
 //! ```
+//! # fn main() {
+//! # #[cfg(feature = "std")] {
 //! use std::sync::OnceLock;
 //!
 //! use bytesbuf::BytesView;
@@ -518,6 +582,8 @@
 //! #     fn accept() -> Self { Connection }
 //! #     fn memory(&self) -> impl bytesbuf::mem::Memory { bytesbuf::mem::GlobalPool::new() }
 //! #     fn write(&self, _data: BytesView) {}
+//! # }
+//! # }
 //! # }
 //! ```
 //!
@@ -559,6 +625,10 @@
 #![doc(html_logo_url = "https://media.githubusercontent.com/media/microsoft/oxidizer/refs/heads/main/crates/bytesbuf/logo.png")]
 #![doc(html_favicon_url = "https://media.githubusercontent.com/media/microsoft/oxidizer/refs/heads/main/crates/bytesbuf/favicon.ico")]
 
+extern crate alloc;
+#[cfg(all(feature = "std", not(test)))]
+extern crate std;
+
 // The root level contains "byte sequence" types, whereas the closely related
 // "memory management" types are shoved away into the `mem` module. This is largely
 // for organizational purposes, to help navigate the API documentation better. Both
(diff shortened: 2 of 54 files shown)


crates/bytesbuf/Cargo.toml
```diff
@@ -32,16 +32,18 @@ allowed_external_types = [
...
 [dependencies]
-bytes = { workspace = true, features = ["std"], optional = true }
-new_zealand = { workspace = true }
-nm = { workspace = true }
-plurality = { workspace = true }
+bytes = { workspace = true, optional = true }
+nm = { workspace = true, optional = true }
+plurality = { workspace = true, optional = true }
 smallvec = { workspace = true, features = ["const_new", "union"] }
 thread_aware = { workspace = true, features = ["derive"] }
```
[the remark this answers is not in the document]
geeknoid: Added `default-features = false` explicitly to the `thread_aware` dependency. The workspace dependency already disabled defaults, but the package declaration now makes the no-std requirement local and unambiguous. The thumb target check passes.

crates/uniflight/Cargo.toml
```diff
@@ -29,7 +29,7 @@ ahash = { workspace = true, default-features = false, features = ["std"] }
 async-once-cell = { workspace = true }
 dashmap = { workspace = true }
 futures-util = { workspace = true, default-features = false, features = ["std", "alloc"] }
-thread_aware = { workspace = true }
+thread_aware = { workspace = true, features = ["std"] }
```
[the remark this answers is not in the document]
geeknoid: Added `default-features = false`, so this dependency now explicitly enables only `std` and does not pull in the derive proc macro.

crates/seatbelt/Cargo.toml
```diff
@@ -66,7 +66,7 @@ layered = { workspace = true }
 opentelemetry = { workspace = true, optional = true }
 recoverable = { workspace = true }
 serde = { workspace = true, features = ["std", "derive"], optional = true }
-thread_aware = { workspace = true }
+thread_aware = { workspace = true, features = ["std"] }
```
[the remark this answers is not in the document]
geeknoid: Added `default-features = false`, so this dependency now explicitly enables only `std` and does not pull in the derive proc macro.

crates/thread_aware/src/lib.rs
```diff
@@ -3,9 +3,20 @@
...
 
 //! Essential building blocks for thread-per-core libraries.
 //!
+//! # Crate features
+//!
+//! * The **`std` Cargo feature** *(enabled by default)* enables the per-affinity `Arc` and
+//!   hosted-only type implementations.
+//! * **`derive`** *(default)* re-exports the `#[derive(ThreadAware)]` macro.
+//! * **`threads`** enables the `registry` module and implies `std`.
+//! * Disable default features for `#![no_std]` environments. The [`ThreadAware`] trait,
```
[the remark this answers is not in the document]
geeknoid: Disambiguated `ThreadAware` links across the API docs with `trait@ThreadAware`. The crate-level docs use an explicit trait-path reference so cargo-doc2readme 0.6.4 generates links to `thread_aware`, while the two derive-macro references explicitly target `thread_aware_macros::ThreadAware`. The regenerated README now separates trait and derive-macro links correctly.

crates/tick/Cargo.toml
```diff
@@ -48,7 +48,7 @@ futures-core = { workspace = true }
 jiff = { workspace = true, features = ["std"], optional = true }
 pin-project-lite = { workspace = true }
 serde_core = { workspace = true, features = ["std"], optional = true }
-thread_aware = { workspace = true }
+thread_aware = { workspace = true, features = ["std"] }
```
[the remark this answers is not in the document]
geeknoid: Added `default-features = false`, so this dependency now explicitly enables only `std` and does not pull in the derive proc macro.

martin-kolinek (approved):

Vaiz:
what's the motivation for this change? we never targeted no_std environments before

crates/thread_aware/src/lib.rs
```diff
@@ -118,10 +134,11 @@
 //!
 //! # Examples
 //!
-//! ## Deriving [`ThreadAware`]
+//! ## Deriving the [`ThreadAware` derive macro](thread_aware_macros::ThreadAware)
```
sandersaares: "Deriving the derive macro" does not English for me - I think something is a bit messed up in the wording.

crates/bytesbuf/src/lib.rs
```diff
@@ -127,6 +138,7 @@
 //!    instance of a shared [`GlobalPool`]. In a typical web application, the global memory pool
 //!    is a service exposed by the application framework. In a different context (e.g. example
 //!    or test code with no framework), you can create your own instance via [`GlobalPool::new()`].
+//!    In a `no_std` environment, the application must supply a custom [`Memory`] implementation.
```
sandersaares: [Copilot speaking]

## [P1] Add the promised alloc-backed provider or narrow the `bytesbuf` `no_std` claim

**Impacted locations**:
- Pull request description
- `crates/bytesbuf/src/lib.rs:10-14,132-145`
- `crates/bytesbuf/docs/snippets/choosing_memory_provider.md:1-9`
- `crates/bytesbuf/src/mem/mod.rs:4-34`
- `crates/bytesbuf/src/mem/block.rs:25-48`
- `crates/bytesbuf/src/mem/block_ref.rs:52-83,151-199`

The pull request description says this change adds an alloc-backed `bytesbuf` memory provider, but the reviewed head contains no `AllocMemory`, `alloc_block`, or equivalent production provider. `GlobalPool` is now available only with `std`, the `multitude::Arena` integration opts into hosted mode, and the remaining providers are test-only. The new public guidance therefore leaves a custom `Memory` implementation as the only way to create usable buffers without `std`.

That is not a small adapter: a downstream provider must use the unsafe `Block::new` and `BlockRef::new` constructors and correctly implement allocation ownership, dynamic vtable dispatch, and reference counting. The public `BlockRefVTable` documentation also points to `examples/bb_basic.rs` as a complete implementation example, but that example only uses the hosted `GlobalPool` and implements none of this lifecycle.

**Reproducible reasoning**:
1. Search the reviewed tree for `AllocMemory` or an allocation-backed production provider; no implementation exists.
2. Disable `bytesbuf` default features; the `global` module and `GlobalPool` re-export disappear.
3. Follow the new memory-provider guidance; it directs the application to implement `Memory`.
4. Follow the construction APIs needed by such an implementation; they require unsafe block construction and a correct reference-counted vtable.

**Consequence**: Ordinary `no_std + alloc` consumers can compile the portable container types but cannot reserve memory without introducing duplicated unsafe memory-management code. The implementation therefore does not deliver the provider promised by the pull request and makes the advertised support impractical for consumers that do not already own a specialized provider.

**Recommended action**: Add the promised non-pooling provider backed by the global `alloc` allocator, or expose a safe reusable helper that turns allocator-owned memory into a `Block`. Point the custom-provider documentation to a real portable implementation. If the intended scope is only to make the core types portable for applications that already have a specialized provider, narrow the pull request description and public feature documentation to state that explicitly.

**References**:
- `crates/multitude/src/bytesbuf.rs:35-143`
- Existing motivation question: https://github.com/microsoft/oxidizer/pull/615#issuecomment-5127025820

crates/bytesbuf/src/mem/mod.rs
```diff
@@ -48,6 +49,8 @@ mod opaque_memory;
 pub use block::{Block, BlockSize};
 pub use block_ref::{BlockMeta, BlockRef, BlockRefDynamic, BlockRefDynamicWithMeta, BlockRefVTable};
 pub use callback_memory::CallbackMemory;
+#[cfg(feature = "std")]
```
sandersaares, line 53: [Copilot speaking]

## [P1] Gate hosted-only `bytesbuf` unit tests and doctests

**Impacted locations**:
- `crates/bytesbuf/src/mem/mod.rs:42-54`
- `crates/bytesbuf/src/mem/opaque_memory.rs:73-153`
- `crates/bytesbuf/src/buf.rs:60-78,231-247,403-421,453-468,502-521,534-557,578-598,700-713,731-753,896-923,943-955,970-997,1062-1091,1427-1438,2632-2635`
- `crates/bytesbuf/src/buf_put.rs:54-64,124-139,152-162,175-191`
- `crates/bytesbuf/src/bytes_compat/view.rs:4-24,45-101`
- `crates/bytesbuf/src/bytes_compat/to_bytes.rs:15-28`
- `crates/bytesbuf/src/mem/callback_memory.rs:27-57`
- `crates/bytesbuf/src/view.rs:37-50,119-131,165-178,209-221,254-264,501-516,537-556,571-587,601-610,629-641,661-680,727-737,757-770`
- `crates/bytesbuf/src/view_get.rs:68-78,117-128,160-174`

The production APIs are gated behind the new `std` feature, but their tests and public examples are not gated with the APIs they exercise.

Targeted validation confirms both package surfaces fail:
- `cargo test -p bytesbuf --no-default-features --lib --no-run` exits 101 because unit tests import the absent `GlobalPool` (`mem/opaque_memory.rs:82` is the first diagnostic), with additional unconditional uses of `BytesBuf::into_writer`, `IoSlice`, and vectored methods.
- `cargo test -p bytesbuf --no-default-features --doc` exits 101 because portable API examples construct the absent `GlobalPool` (`buf.rs:231` is the first diagnostic). The same pattern occurs throughout `buf.rs`, `buf_put.rs`, `view.rs`, `view_get.rs`, `callback_memory.rs`, and the `bytes-compat` conversion docs.

**Consequence**: The new no-default configuration cannot compile its own unit tests or API examples. Workspace validation can remain green because it uses default/all features or feature unification from hosted dependents, so regressions in the portable surface remain hidden.

**Recommended action**: Gate hosted-only tests and examples with `feature = "std"`, or use a portable test provider where the behavior under test is not inherently hosted. The repository-root `AGENTS-feature-gated-doctests.md` documents the expected hidden wrapper patterns for feature-dependent doctests. Add package-scoped no-default unit-test and doctest commands to required CI so downstream feature unification cannot mask this configuration.

**References**:
- `AGENTS-feature-gated-doctests.md`
- [Cargo feature unification](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification)
- Existing summaries identify the same pattern only in the crate-root doctests: https://github.com/microsoft/oxidizer/pull/615#pullrequestreview-4809882977 and https://github.com/microsoft/oxidizer/pull/615#pullrequestreview-4813621357

crates/bytesbuf/Cargo.toml
```diff
@@ -32,24 +32,27 @@ allowed_external_types = [
 all-features = true
 
 [features]
-default = []
+default = ["std"]
```
sandersaares, line 35: [Copilot speaking]

## [P2] Classify the next releases of both affected packages as breaking

**Impacted locations**:
- `crates/bytesbuf/Cargo.toml:34-38`
- `crates/bytesbuf/src/mem/mod.rs:42-54`
- `crates/bytesbuf/src/buf.rs:1171-1175`
- `crates/thread_aware/Cargo.toml:56-62`
- `crates/thread_aware/src/lib.rs:189-191,257-259`
- `crates/thread_aware/src/impls.rs:44-48,144-151`

Before this change, `bytesbuf` had an empty default feature set, so `default-features = false` still exposed `GlobalPool`, `BytesBufWriter`, and `BytesBuf::into_writer`. `thread_aware` defaulted only to `derive`, so disabling defaults still exposed the per-affinity `Arc` API and hosted type implementations. This change creates a new default-on `std` feature in both packages and moves those existing public APIs or implementations behind it.

Existing consumers that already specify `default-features = false` therefore lose APIs after an otherwise compatible update. The repository release classifier invokes `cargo semver-checks` with `--all-features`, which keeps the moved APIs visible and cannot detect this feature-surface break.

**Consequence**: Publishing either crate as a compatible `0.7.x` or `0.8.x` update can make existing no-default consumers fail to compile through an automatic Cargo update.

**Recommended action**: Classify the next `bytesbuf` release as breaking (`0.8.0`) and the next `thread_aware` release as breaking (`0.9.0`). Add a release-process guard or manual-review rule for changes that move existing APIs behind a new feature, because an all-features semver check cannot classify this pattern.

**References**:
- `docs/releasing.md`, "How the change type is determined"
- `scripts/lib/releasing.ps1:613-659`
- [Cargo default features](https://doc.rust-lang.org/cargo/reference/features.html#the-default-feature)
- [ADO bug 7675347: Semver validation misses breakage in lesser-featured API surfaces](https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7675347)

crates/bytesbuf/src/lib.rs
```diff
@@ -3,9 +3,19 @@
...
 #![cfg_attr(docsrs, feature(doc_cfg))]
+#![cfg_attr(not(test), no_std)]
 
 //! Types for creating and manipulating byte sequences.
 //!
+//! # Crate features
+//!
+//! * The **`std` Cargo feature** *(enabled by default)* enables the global memory pool, standard I/O
+//!   adapters, and metrics. Disable it for `#![no_std]` environments; the crate still requires
+//!   `alloc`.
```
sandersaares: [Copilot speaking]

## [P2] Document the pointer-atomic prerequisite for `no_std`

**Impacted locations**:
- `crates/bytesbuf/src/vec.rs:4-10,65-71`
- `crates/thread_aware/src/wrappers.rs:4,61-66`
- `crates/bytesbuf/src/lib.rs:10-14`
- `crates/thread_aware/src/lib.rs:10-18`

The new public guidance says the crates require only `alloc` after disabling `std`, but the portable configurations also require pointer-width atomics. `bytesbuf` compiles `core::sync::atomic::AtomicUsize` unconditionally, while `thread_aware` exposes `Unaware::into_arc` through `alloc::sync::Arc`; neither API exists on targets such as `thumbv6m-none-eabi`.

**Reproducible reasoning**:
1. `bytesbuf::vec` stores an `AtomicUsize` reference count for the portable `From<Vec<u8>> for BytesView` implementation.
2. `thread_aware::wrappers` imports `alloc::sync::Arc` for the portable `Unaware::into_arc` API.
3. Neither module or API is guarded by `cfg(target_has_atomic = "ptr")`.
4. Rust does not provide these types on targets without pointer atomics.

**Consequence**: Consumers can reasonably interpret "`no_std` with `alloc`" as covering alloc-capable Cortex-M0/M0+ and similar targets, then discover only at compilation that these targets are outside the intended support boundary. A representative `thumbv7em` check would not communicate this limitation because that target has pointer atomics.

**Recommended action**: State in the crate-level feature documentation that `no_std` support requires `alloc` and pointer-width atomics. Also document the repository policy in `AGENTS.md`: `no_std` is a best-effort, low-complexity capability, so contributors should not complicate the implementation or add extensive conditional-compilation branches to support atomics-free targets. A tailored compile-time diagnostic may be added if the author considers it useful, but is not required.

**References**:
- [`alloc::sync::Arc` platform availability](https://doc.rust-lang.org/alloc/sync/struct.Arc.html)
- [`AtomicUsize` platform availability](https://doc.rust-lang.org/core/sync/atomic/type.AtomicUsize.html)
- [`target_has_atomic`](https://doc.rust-lang.org/reference/conditional-compilation.html#target_has_atomic)

crates/bytesbuf/Cargo.toml
```diff
@@ -32,24 +32,27 @@ allowed_external_types = [
...
+# Enables the global memory pool, standard I/O adapters, and metrics. Disable for `#![no_std]`
```
sandersaares, line 36: [Copilot speaking]

## [P2] Add the permanent bare-metal check promised by the PR

**Impacted locations**:
- `crates/bytesbuf/Cargo.toml:35-38`
- `crates/thread_aware/Cargo.toml:56-61`
- `.github/workflows/main.yml:103-188,453-454,589-607`
- Pull request description

The pull request description promises permanent bare-metal CI coverage, but the reviewed workflow has no `thumbv7em`, `none-eabi`, or other bare-metal target job. The current `cargo hack build --each-feature` runs on hosted Linux and Windows targets, while `cargo ensure-no-default-features` validates dependency feature policy rather than target compatibility. The resolved thread establishes that a manual thumb check passed, but it does not provide permanent regression coverage.

**Consequence**: Branch protection can report success even when a hosted-only dependency or API leaks into the newly documented embedded configuration.

**Recommended action**: Add a required package-scoped `cargo check --locked --target thumbv7em-none-eabihf --no-default-features` job for `bytesbuf` and `thread_aware`, plus their advertised non-`std` feature combinations. Add any new delta-gated job to `required-checks.needs`.

**References**:
- `AGENTS.md`, "Required CI Checks"
- [Rust target support: `thumbv7em-none-eabi*`](https://doc.rust-lang.org/rustc/platform-support/thumbv7em-none-eabi.html)
- Existing review summary claiming the embedded check: https://github.com/microsoft/oxidizer/pull/615#pullrequestreview-4809597716
- Related resolved thread mentioning the thumb check: https://github.com/microsoft/oxidizer/pull/615#discussion_r3675649230

crates/thread_aware/src/lib.rs
sandersaares: [Copilot speaking]

## [P2] Make the documented `thread_aware` `no_std` recipe match the feature graph

**Impacted locations**:
- `crates/thread_aware/src/lib.rs:10-18,116-151`
- `crates/thread_aware/Cargo.toml:56-84`
- `crates/thread_aware/src/lib.rs:255-256`

The new guidance says to disable default features for `no_std` and then states that derive support remains available. In the manifest, however, `default = ["std", "derive"]`, so `default-features = false` disables the derive macro as well. The nearby `Point` doctest still invokes `#[derive(ThreadAware)]` unconditionally, and `cargo test -p thread_aware --no-default-features --doc` exits 101 at `src/lib.rs:147` because only the trait remains available.

The feature list also presents `http` alongside the portable third-party integrations without saying that `http = ["std", "dep:http"]` opts the crate back into hosted mode.

**Consequence**: Consumers following the documented no-default recipe cannot use the derive example, while bare-metal consumers can select the documented `http` integration and unexpectedly reactivate `std`.

**Recommended action**: Document `default-features = false, features = ["derive"]` as the `no_std` recipe when derive support is wanted, state that no-feature builds omit the macro, mark `http` as requiring/enabling `std`, and gate the derive doctest with `feature = "derive"`.

**References**:
- [Cargo default features](https://doc.rust-lang.org/cargo/reference/features.html#the-default-feature)
- [Cargo feature activation](https://doc.rust-lang.org/cargo/reference/features.html)

crates/bytesbuf/Cargo.toml
```diff
@@ -32,24 +32,27 @@ allowed_external_types = [
...
+# environments; the crate still requires `alloc`.
+std = ["dep:nm", "dep:plurality", "thread_aware/std", "bytes?/std"]
```
sandersaares, line 38: [Copilot speaking]

## [P2] Propagate `bytesbuf/std` through the hosted `multitude` integration

**Impacted locations**:
- `crates/bytesbuf/Cargo.toml:36-38,45`
- `crates/multitude/Cargo.toml:50,66`
- `crates/bytesbuf/src/view.rs:89-91,1037-1049`

The new `bytesbuf/std` feature now owns the `nm` dependency and the `bytesbuf_view_created_spans` observation, but `multitude`'s hosted `bytesbuf` integration enables only `dep:bytesbuf` and its own local `std` feature. The workspace dependency disables `bytesbuf` default features, and a local `"std"` feature entry does not activate `bytesbuf/std`.

**Reproducible reasoning**:
1. The workspace declaration sets `bytesbuf` `default-features = false`.
2. This change makes `nm` optional and gates the span-count event on `bytesbuf/std`.
3. `multitude = { features = ["bytesbuf"] }` activates `dep:bytesbuf` and `multitude/std`, but not `bytesbuf/std`.
4. Arena-backed `BytesView` creation therefore stops emitting the metric even though this is a hosted integration.

**Consequence**: Real workloads using the arena-backed provider silently stop contributing span-count telemetry used to tune `MAX_INLINE_SPANS`, biasing the repository's metrics-driven optimization data.

**Recommended action**: Propagate the dependency feature, for example with `bytesbuf = ["dep:bytesbuf", "std", "bytesbuf/std"]`, or enable `features = ["std"]` on the optional dependency.

**References**:
- `crates/bytesbuf/AGENTS.md`, "Metrics-driven fine-tuning"
- [Cargo dependency features](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features)

crates/bytesbuf/src/mem/mod.rs
```diff
@@ -3,9 +3,9 @@
 
 //! Types for using and implementing memory providers.
 //!
-//! The only production-grade memory provider published by this crate is [`GlobalPool`], which uses
-//! memory from the Rust global memory allocator and adds a layer of pooling to reduce the overhead
-//! from memory allocation churn.
+//! With the `std` feature, this crate provides `GlobalPool`, which uses memory from the Rust
+//! global allocator and adds pooling to reduce allocation churn. In all configurations, custom
+//! memory providers can be implemented with [`Memory`].
```
sandersaares: [Copilot speaking]

## [P2] Align `Memory::reserve` with the newly documented portable providers

**Impacted locations**:
- `crates/bytesbuf/src/lib.rs:132-145`
- `crates/bytesbuf/src/mem/mod.rs:4-8`
- `crates/bytesbuf/src/mem/memory.rs:18-35`
- `crates/bytesbuf/src/mem/callback_memory.rs:97-98`
- `crates/bytesbuf/src/mem/opaque_memory.rs:38-39`

The new `no_std` guidance tells applications to implement `Memory`, but its existing contract still promises that every reservation is fulfilled by obtaining more memory from "the operating system." The same promise is repeated by generic forwarding adapters that do not control the wrapped provider's allocation strategy.

A bare-metal implementation may use a fixed arena, device memory, or the global allocator without any operating system. Because the API returns no `Result`, exhaustion may panic or abort according to the provider, but the current contract promises an OS-specific source and documents only an OS out-of-memory panic.

**Consequence**: The providers newly invited by this change cannot truthfully satisfy the public trait contract, and callers can infer allocation and failure behavior that generic providers do not guarantee.

**Recommended action**: Define the portable contract in terms of successful return: if `reserve` returns, the buffer has at least the requested capacity. Document that implementations may panic or abort when capacity cannot be obtained, and keep stronger OS/global-pool guarantees only on concrete providers that actually provide them.

**References**:
- [The Embedded Rust Book: a `no_std` environment](https://docs.rust-embedded.org/book/intro/no-std.html)

crates/bytesbuf/Cargo.toml
```diff
@@ -32,24 +32,27 @@ allowed_external_types = [
...
 # Interoperability with the `bytes` crate.
 bytes-compat = ["dep:bytes"]
-test-util = []
+test-util = ["std"]
```
sandersaares, line 41: [Copilot speaking]

## [P2] Require `std` for both `GlobalPool` benchmark targets

**Impacted locations**:
- `crates/bytesbuf/Cargo.toml:35-38,78-80,97-99`
- `crates/bytesbuf/benches/global_pool.rs:10`
- `crates/bytesbuf/benches/global_pool_cg.rs:40`

The new `std` feature makes `GlobalPool` available only when `std` is enabled, but the `global_pool` and `global_pool_cg` benchmark targets do not declare `required-features = ["std"]`. Cargo therefore selects these targets during `cargo bench -p bytesbuf --no-default-features`, even though their imports are unavailable.

`cargo bench -p bytesbuf --no-default-features --no-run` exits 101 at `benches/global_pool.rs:10`; the Linux Callgrind target has the same registration problem.

**Consequence**: The newly supported no-default configuration cannot use the package's normal benchmark command, and future all-target benchmark validation fails before running applicable benchmarks.

**Recommended action**: Add `required-features = ["std"]` to both benchmark target declarations.

**References**:
- `docs/benchmarks.md:33`
- [Cargo target `required-features`](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-required-features-field)

geeknoid:
> what's the motivation for this change? we never targeted no_std environments before

I have several no_std crate which want to use bytesbuf and are currently blocked on this.

crates/bytesbuf/src/lib.rs
geeknoid: Documented that both portable configurations require `alloc` and pointer-width atomics, and added the requested best-effort, low-complexity `no_std` policy to `AGENTS.md`.

crates/bytesbuf/src/mem/mod.rs
geeknoid: Made the `Memory::reserve` contract provider-neutral: successful return guarantees the requested capacity, while implementations may panic or abort on exhaustion. Applied the same wording to the generic forwarding providers.

crates/bytesbuf/Cargo.toml
geeknoid: Removed the stale permanent-CI promise from the PR description. The portable combinations were checked directly on `thumbv7em-none-eabihf`, but this PR intentionally does not change the global workflow.

crates/bytesbuf/Cargo.toml
geeknoid: Updated the hosted `multitude` integration to propagate `bytesbuf/std`, preserving the span-count metrics.

crates/thread_aware/src/lib.rs
geeknoid: Fixed the heading to “Using the `ThreadAware` derive macro.”

crates/bytesbuf/Cargo.toml
geeknoid: Added `required-features = ["std"]` to both `global_pool` benchmark targets. The no-default benchmark compilation now succeeds and skips those hosted targets.

crates/bytesbuf/src/lib.rs
sandersaares: [Copilot speaking] The prerequisites are now documented. The policy wording was subsequently clarified: adopting or expanding `no_std` is optional, but once a crate documents a configuration as supported it is a real compatibility promise, not best-effort support. PR #618 contains that clarified wording. Please remove this new `AGENTS.md` block from this PR and let #618 establish the policy, or update it to match #618; otherwise the two PRs can leave contradictory guidance.

crates/bytesbuf/Cargo.toml
sandersaares: [Copilot speaking] This remains open after revalidation. The PR description no longer promises permanent coverage, but `bytesbuf` and `thread_aware` now publicly document supported `no_std` configurations, and the clarified policy in #618 requires documented configurations to be covered by CI. Manual thumb validation cannot prevent regressions, and the workflow still lacks both package-scoped no-default unit/doctest commands and a representative bare-metal check. Please add required CI coverage for those surfaces; ADO bug 7675218 tracks the broader validation gap.

geeknoid: Not in scope, not important enough right now. The ADO item is sufficient,

crates/bytesbuf/src/lib.rs
geeknoid: Removed the `AGENTS.md` policy block from this PR so #618 can establish the clarified repository-wide policy without conflicting guidance. The crate-level pointer-atomic prerequisites remain documented here.

crates/bytesbuf/Cargo.toml
sandersaares: [Copilot speaking] The latest head removes the conflicting `AGENTS.md` policy block, which addresses the other reopened thread, but it does not change `.github/workflows/main.yml` or otherwise add the missing CI coverage. The supported `no_std` configurations therefore remain protected only by manual validation. Reopening this thread pending permanent package-scoped no-default tests/doctests and representative bare-metal coverage, or an explicit maintainer decision to defer that coverage to the tracked validation work.

geeknoid: I'm a maintainer, I'm deferring. STOP MENTIONING THIS.

sandersaares (approved):

PR #616, opened 2026-07-29
fix(fetch)!: stop the `tokio` feature from selecting a rustls crypto provider

Vaiz:
`fetch`'s `tokio` feature enabled `rustls?/aws-lc-rs`. `tokio` is a runtime/transport selector and must not choose a TLS crypto provider.

Because `rustls?/...` is a weak dependency feature it looks inert in isolation, but as soon as anything else in the same workspace enables `dep:rustls` (for example via `fetch/rustls`), Cargo's workspace-wide feature unification activates it and forces rustls onto aws-lc-rs, dragging in `aws-lc-rs` + `aws-lc-sys` and its native NASM/C build.

This bit a real consumer: crates selecting only
`fetch = { features = ["tokio", "native-tls"] }` still ended up with `aws-lc-rs`/`aws-lc-sys` in `Cargo.lock` in a workspace that deliberately drives rustls with `default-features = false` plus a different provider. Dropping `"tokio"` from those manifests made aws-lc-rs disappear, confirming `fetch/tokio` was the trigger.

Changes:

- `tokio` no longer implies any crypto provider.
- New opt-in `rustls-aws-lc-rs` feature carries `rustls/aws-lc-rs`.
- `tls` is now `rustls` + `rustls-aws-lc-rs`, so it keeps its documented "rustls with aws-lc-rs" meaning and stays the recommended way to turn on HTTPS.
- `rustls` alone is provider-neutral: the Tokio transport resolves the process-wide default installed via `CryptoProvider::install_default`, and panics with an actionable message when none is installed.
- Examples/tests that actually negotiate TLS now require `tls` instead of `rustls`, as does `fetch_azure`'s dev-dependency on `fetch`.
- Crate docs/README describe the provider choice and the new feature.

Verified with `cargo tree -e normal -i aws-lc-rs`: `fetch` with `tokio,rustls` no longer pulls aws-lc-rs, while `tokio,rustls,rustls-aws-lc-rs` (and `tokio,tls`) still does.

BREAKING CHANGE: `fetch/tokio` no longer implies the `aws-lc-rs` rustls crypto provider. Users who relied on `features = ["tokio", "rustls"]` giving them a working provider must switch to `features = ["tokio", "tls"]` (or add `rustls-aws-lc-rs`), or install a provider themselves with `rustls::crypto::CryptoProvider::install_default`.

### DIFF
--- a/crates/fetch/Cargo.toml
+++ b/crates/fetch/Cargo.toml
@@ -53,13 +53,22 @@ ignored = ["ctor"]
 
 [features]
 default = []
-tls = ["rustls"]
+# Batteries-included TLS: the `rustls` backend together with the `aws-lc-rs`
+# crypto provider.
+tls = ["rustls", "rustls-aws-lc-rs"]
+# The `rustls` backend without a crypto provider. Callers must either enable
+# `rustls-aws-lc-rs` (or `tls`), or install a process-wide provider through
+# `rustls::crypto::CryptoProvider::install_default`.
 rustls = [
     "dep:rustls",
     "dep:rustls-platform-verifier",
     "fetch_hyper?/rustls",
     "fetch_tls/rustls",
 ]
+# Opt-in crypto provider for the `rustls` backend. Kept separate from the
+# runtime/transport features so that selecting a runtime never selects a crypto
+# provider (and never drags in `aws-lc-sys`' native build).
+rustls-aws-lc-rs = ["rustls", "rustls/aws-lc-rs"]
 native-tls = ["dep:native-tls", "fetch_hyper?/native-tls", "fetch_tls/native-tls"]
 json = ["http_extensions/json"]
 test-util = ["tick/test-util", "http_extensions/test-util"]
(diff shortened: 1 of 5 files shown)


martintmk:
One downgraded scenario of this change would be customer that explicitly constructs TlsOptions, provides his own override for crypto provider and then when constructing fetch it fails with static provider not being installed.

https://docs.rs/fetch_tls/latest/fetch_tls/struct.TlsBackendBuilder.html#method.configure_rustls.

I would change `fetch_tls::TlsBackendBuilder::configure_rustls` to accept in `Option<Arc<CryptoProvider>>` rather than `Arc<CryptoProvider>`. This way when consumer overrides [crypto provider ](https://docs.rs/fetch_tls/latest/fetch_tls/struct.TlsOptionsBuilder.html#method.crypto_provider), `TlsBackendBuilder` doesn't need to access installed provider to crate `TlsBackend` and will just use one provided by the consumer.

PR #617, opened 2026-07-29
feat(multitude): support trait object coercion

geeknoid:
Add compiler-checked coercion APIs for Box, Rc, and Arc using hybrid metadata. Sized values, strings, and slices retain thin handles, while trait objects carry their vtable metadata in each erased handle.

Define unwind-safety contracts for Multitude and Plurality, strengthen DST and refcount coverage, and correct Serde ignored-value dispatch.

Audit API documentation and make Criterion and Callgrind benchmark setup, teardown, naming, and reporting consistent.


crates/multitude/src/coerce.rs
```diff
@@ -0,0 +1,104 @@
...
+
+/// Applies `coercion` and returns the resulting pointer metadata.
+#[inline]
+pub(crate) fn unsize_metadata<T, U: ?Sized + Pointee, F: FnOnce(*const T) -> *const U>(
+    ptr: NonNull<u8>,
+    coercion: Coercion<T, U, F>,
+) -> U::Metadata {
+    let source = ptr.cast::<T>().as_ptr().cast_const();
+    let target = (coercion.coerce)(source);
+    debug_assert!(core::ptr::addr_eq(source, target));
```
psandana: [correctness] The coercion pointer-identity contract is only checked via `debug_assert!(core::ptr::addr_eq(source, target))` (line 57), which compiles away in release builds. Since `Coercion::new` is a public `unsafe fn` escape hatch accepting an arbitrary `FnOnce(*const T) -> *const U`, a caller that violates the safety contract (e.g., returns a pointer to a different object or with an offset) will silently produce a `Box`/`Arc`/`Rc` whose vtable is paired with the wrong address — every virtual call and `Drop` through that handle is then UB, with zero diagnostics in release. Since `unsize`/`unsize_pin` are called at most once per coercion (never hot-path), the cost of promoting this to an always-on `assert!` is negligible and would close an easy-to-hit misuse channel for anyone bypassing the `coerce!` macro.

```rust
assert!(
    core::ptr::addr_eq(source, target),
    "Coercion::new contract violated: coercion function changed the pointer address"
);
```

crates/multitude/src/box.rs
```diff
@@ -54,15 +55,18 @@ use crate::vec::Vec;
...
 /// ```
-pub struct Box<T: ?Sized + Pointee, A: Allocator + Clone = Global> {
-    /// **Thin** pointer to the first byte of the contained value, which
-    /// lives in a 64K-aligned [`Chunk`]'s payload. The chunk
-    /// header is recovered by masking, and `T`'s pointer metadata (if
-    /// any) is stored in the `size_of::<T::Metadata>()` bytes
-    /// immediately preceding the payload (read with
-    /// [`core::ptr::read_unaligned`]). This makes `Box<T>` 8 bytes
-    /// uniformly, even for DST `T`.
+pub struct Box<
```
psandana, line 58: [design] `Box<T, A>`/`Arc<T, A>`/`Rc<T, A>` gain a third, `#[doc(hidden)]`-bounded generic parameter (`M: HandleMetadata<T> = <T as SmartPointerPointee>::StoredMetadata`) to carry either `()` or a `DynMetadata` per handle. `M` is defaulted everywhere internal, so ordinary callers never write it explicitly, but it's a real third type parameter on a previously two-parameter public type — it will show up in rustdoc's struct definition, in compiler error messages for misused bounds, and in any downstream code trying to be fully generic over `Box<T, A, M>`. It also permanently commits the public API to this exact metadata-storage strategy.

Would it be worth hiding the metadata field behind an internal enum (`enum StoredMeta { Thin, Vtable(DynMetadata) }`) instead, accepting one extra branch on the (non-hot) trait-object path, in exchange for a stable, simpler public type signature? If the current approach is intentional, it may be worth documenting in DESIGN.md why the extra type parameter was chosen over an internal enum.

crates/plurality/src/alloced.rs
```diff
@@ -32,6 +32,12 @@ pub struct Alloc<'pool, T, A: Allocator = Global> {
     _pool: PhantomData<&'pool Pool<T, A>>,
 }
 
+// `Alloc` semantically owns its slot. Moving it across an unwind boundary is
+// safe under the same bounds as an owning box; borrowing it mutably still
+// produces `&mut Alloc`, which remains `!UnwindSafe`.
+impl<T: core::panic::RefUnwindSafe, A: Allocator> core::panic::RefUnwindSafe for Alloc<'_, T, A> {}
```
psandana, line 40: [correctness] `Box<T,A>`, `Arc<T,A>`, `Rc<T,A>`, and `Pool<T,A>` all gate their `RefUnwindSafe`/`UnwindSafe` impls on `A: Allocator + RefUnwindSafe`, reflecting that a shared/interior-mutable allocator observed through another owner after an unwind is unsound to treat as unwind-safe. `Alloc<'pool, T, A>` here drops the `A` bound entirely (`impl<T: RefUnwindSafe, A: Allocator> RefUnwindSafe for Alloc<'_, T, A> {}` / the analogous `UnwindSafe` impl on the next line), even though `Alloc` holds `PhantomData<&'pool Pool<T, A>>` — a shared borrow of a `Pool<T, A>` which itself is only unwind-safe when `A: RefUnwindSafe`. `tests/unwind_safe.rs` doesn't appear to cover `Alloc<'_, T, InteriorMutableAllocator>`, so this asymmetry (possibly intentional per the surrounding comment, but unverified by a test) isn't exercised. Could this be confirmed as intentional, or should the `A: RefUnwindSafe` bound be added here to match `Box`/`Arc`/`Rc`/`Pool`?

crates/multitude/src/internal/thin_dst.rs
```diff
@@ -1,52 +1,266 @@
...
+    T::Metadata: MetadataPolicy<T, Stored = ()>,
+{
+    #[inline]
+    unsafe fn resolve(self, value_ptr: NonNull<u8>) -> T::Metadata {
+        // SAFETY: forwarded from the caller.
+        unsafe { <T::Metadata as MetadataPolicy<T>>::resolve(value_ptr, self) }
+    }
+}
+
+impl<T, D> HandleMetadata<T> for DynMetadata<D>
```
psandana, line 247: [correctness] `HandleMetadata<T>: Copy + Send + Sync + ...` is implemented for `DynMetadata<D>` with no `D: Send + Sync`/`'static` bound. This relies entirely on `ptr_meta::DynMetadata<D>` itself being unconditionally `Send + Sync` regardless of `D`. That happens to be true for the current `ptr_meta` implementation, but since `Box`/`Arc`/`Rc`'s own `unsafe impl Send/Sync` blanket on `M: HandleMetadata<T>` transitively depends on this external crate's auto-trait behavior for soundness, it would be worth an explicit compile-time assertion (e.g. a `const _: () = { fn assert_send_sync<T: Send + Sync>() {} ... };`) or at least a doc comment citing the `ptr_meta` guarantee, rather than relying implicitly on an upstream invariant that isn't otherwise checked in this repository.

crates/multitude/src/coerce.rs
```diff
@@ -0,0 +1,104 @@
+// Copyright (c) Microsoft Corporation.
```
psandana, line 1: [design] The new public unsizing API is a hand-rolled `Coercion<T, U, F>` token plus a `coerce!` macro that generates a monomorphic helper performing the built-in unsizing coercion, rather than implementing `core::ops::CoerceUnsized` (which would let `Box<Square> as Box<dyn Shape>`-style implicit coercions work, at the cost of nightly-only trait bounds). This is a fairly unusual API shape for a smart pointer library, and a natural first question from any reviewer or contributor is "why not `CoerceUnsized`?" The PR touches DESIGN.md extensively but never states this rationale (likely: targeting stable Rust and emulating `CoerceUnsized`/`Unsize` with a compiler-checked function pointer instead).

Could a short "Trait-object coercion" rationale paragraph be added to DESIGN.md explicitly ruling out `CoerceUnsized` (stable-toolchain constraint) so the API shape is self-justifying for future maintainers? Was nightly `CoerceUnsized` considered and rejected purely for stable-toolchain support, or are there other reasons (e.g., supporting coercions `CoerceUnsized` can't express cleanly)?

psandana:
[design] Box/Arc/Rc/Pool/Alloc in both `multitude` and `plurality` each get hand-written `RefUnwindSafe`/`UnwindSafe` impls with very similar reasoning comments, and each crate has its own near-identical `tests/unwind_safe.rs` with the same `AmbiguousIfUnwindSafe`/`AmbiguousIfRefUnwindSafe` negative-assertion harness copy-pasted. This is exactly the kind of cross-cutting contract that's easy to let drift — if a future refactor changes ownership semantics in one crate's smart pointer but the parallel crate is forgotten, the contracts could silently diverge with no shared enforcement point.

Would it be worth factoring the `assert_(not_)unwind_safe`/`Ambiguous*` test harness (and ideally the impl-generating pattern, via a shared internal macro if feasible given the different field sets) into a small shared internal testing/dev-dependency crate so both `multitude` and `plurality` consume one canonical implementation?

geeknoid:
Thanks for the suggestion. I considered sharing the unwind assertion harness, but I prefer keeping these small test-only helpers local for now. A shared dev crate or macro would add dependency and macro indirection while centralizing only test syntax, not the intentionally distinct ownership and allocator contracts each crate must assert. Keeping each contract self-contained also makes drift visible as an explicit test change in that crate; extraction would make more sense if the harness grows materially.

crates/multitude/src/coerce.rs
geeknoid: Implemented in 53f5d8bc. `unsize_metadata` now uses an always-on address-preservation assertion with the requested diagnostic. I also added a regression that supplies a wrong-address unsafe coercion, verifies the panic, and confirms the original owner is dropped exactly once.

crates/multitude/src/box.rs
geeknoid: This tradeoff is intentional. `M = ()` is zero-sized, so sized values, strings, and slices retain one-word handles; a uniform enum or raw metadata field would add storage to every handle. 53f5d8bc documents that representation benefit and explicitly acknowledges that the defaulted parameter is a real public-signature and semver commitment.

crates/plurality/src/alloced.rs
geeknoid: Confirmed intentional in 53f5d8bc. `Alloc` exposes no path to the pool or allocator; its phantom pool borrow carries lifetime and thread-affinity constraints, while drop destroys `T` and returns the slot without accessing `A`. The comment now states that rationale, and the unwind test explicitly locks `Alloc<u64, InteriorMutableAllocator>` as unwind-safe while `Pool` and `Box` remain allocator-sensitive.

crates/multitude/src/coerce.rs
geeknoid: Added the rationale in 53f5d8bc. The reason is the stable-toolchain constraint: downstream `CoerceUnsized`/`Unsize` implementations require unstable Rust. `coerce!` instead uses the compiler built-in raw-pointer unsizing conversion inside a generated local function, preserving compiler validation on stable Rust; `Coercion::new` remains the explicit unsafe escape hatch.

crates/multitude/src/internal/thin_dst.rs
geeknoid: Added both safeguards in 53f5d8bc: the implementation now documents the `ptr_meta::DynMetadata<D>: Send + Sync` dependency contract, and `tests/send_sync.rs` asserts that `DynMetadata<dyn ThreadAffineMetadataTarget>` is Send + Sync even though the target object is neither. The `HandleMetadata` supertraits also make an upstream contract change fail compilation.

crates/multitude/src/de/value.rs
```diff
@@ -707,7 +707,7 @@ impl<'de, A: Allocator + Clone> de::Deserializer<'de> for &'de Value<A> {
     }
 
     fn deserialize_ignored_any<V: Visitor<'de>>(self, visitor: V) -> Result<V::Value, Self::Error> {
-        visitor.visit_unit()
+        self.deserialize_any(visitor)
```
sandersaares: [Copilot speaking]

# F1 [P2] Restore shape-independent, constant-time ignored-value replay

**Impacted current locations**:
- `crates/multitude/src/de/value.rs:709-710`
- `crates/multitude/src/de/value.rs:825-843`
- `crates/multitude/src/de/value/tests.rs:532-552`

`Value` is already materialized, so `deserialize_ignored_any` can discard it immediately. Forwarding to `deserialize_any` instead redispatches the stored shape into Serde's `IgnoredAny` visitor, changing every sequence and map from an O(1) discard to recursive traversal and also rejecting some explicitly represented enum payloads.

**Reproducible reasoning**: Serde's pinned `IgnoredAny::visit_enum` consumes the variant and unconditionally calls `VariantAccess::newtype_variant()`. `EnumReplay::newtype_variant_seed` accepts `EnumPayload::Value` and explicit `EnumValue::Newtype`, but explicit `Value::Enum` unit, tuple, and struct payloads return `invalid enum payload, expected a newtype variant`. This enum failure is limited to the public model's explicitly represented `Value::Enum` form; ordinary externally tagged string/map replay follows different paths. The recursive-work regression applies universally to stored sequences and maps.

**Concrete consequence**: Derived replay that skips an unknown field can fail when that field contains an explicitly represented unit, tuple, or struct enum, while large ignored collections incur a needless second traversal and recursion depth.

**Recommended action**: Restore `visitor.visit_unit()`. Replace the custom boolean visitor test with unknown-field replay regressions for explicit unit, newtype, tuple, and struct enum payloads, plus a nested collection regression if the O(1) discard property should be locked down.

**References**: [`Deserializer::deserialize_ignored_any`](https://docs.rs/serde/latest/serde/de/trait.Deserializer.html#tymethod.deserialize_ignored_any); [Serde `IgnoredAny` at the pinned revision](https://github.com/serde-rs/serde/blob/747814f7d5fbab872df3b02f070c165b91bde062/serde_core/src/de/ignored_any.rs); `Cargo.lock:4078-4081`.

crates/multitude/src/de/reuse.rs
```diff
@@ -42,7 +42,8 @@ impl<A: Allocator + Clone> String<'_, A> {
     /// capacity.
     ///
     /// The string is cleared before reading input. If deserialization fails,
-    /// it remains valid but may contain a partially deserialized value.
+    /// it remains empty because the replacement string is appended only after
```
sandersaares: [Copilot speaking]

# F3 [P2] Do not promise transactional String reuse without staging the replacement

**Impacted current locations**:
- `crates/multitude/src/de/reuse.rs:41-47,152-166`
- `crates/multitude/src/de/mod.rs:156-161`

The new documentation promises that `String::deserialize_reusing` remains empty on every error and calls replacement “all-or-nothing,” but the implementation clears the destination and writes directly into it before `Deserializer::deserialize_str` returns.

**Reproducible reasoning**: `StringVisitor::visit_str` calls `try_push_str` on the destination. A conforming custom deserializer can call `visitor.visit_str("replacement")`, receive `Ok(())`, and then fail during framing, checksum, or other post-token validation. The method returns `Err` while the destination contains `replacement`; no staging buffer, commit step, or rollback restores the documented empty state.

**Concrete consequence**: Error-recovery code may treat `Err` as proof that the destination is empty and instead observe data from a failed operation.

**Recommended action**: Either restore the conservative contract that the string remains valid but its post-error contents are unspecified, or deserialize into staged storage and commit only after the deserializer returns success. Add a custom deserializer that fails after a successful visitor callback to lock the chosen contract.

**References**: [`serde::Deserializer::deserialize_str`](https://docs.rs/serde/latest/serde/de/trait.Deserializer.html#tymethod.deserialize_str).

crates/multitude/src/internal/thin_dst.rs
```diff
@@ -1,52 +1,270 @@
...
+    const ALLOCATION_BYTES: usize;
+
+    /// Converts pointer metadata into its per-handle representation.
+    fn store(metadata: Self) -> Self::Stored;
+
+    /// Recovers pointer metadata from the allocation and handle.
+    ///
+    /// # Safety
+    ///
+    /// `value_ptr` must point to a live value allocated using this policy.
```
sandersaares, line 49: [Copilot speaking]

# F7 [P2] Complete the public metadata resolution safety and panic contracts

**Impacted current locations**:
- `crates/multitude/src/internal/thin_dst.rs:38-69`
- `crates/multitude/src/internal/thin_dst.rs:142-161`
- `crates/multitude/src/internal/thin_dst.rs:200-216`
- `crates/multitude/src/lib.rs:533-534`

The doc-hidden metadata traits are publicly re-exported and their methods are callable downstream. The safety contracts on `MetadataPolicy::resolve` and `SmartPointerPointee::resolve_metadata` require only a live allocation, without stating that handle-resident metadata must describe the allocation's actual concrete pointee/type. The allocation-metadata constructors also lack canonical `# Panics` documentation even though the `DynMetadata` policy always panics.

**Reproducible reasoning**: A trait-object data pointer paired with a vtable for a different concrete implementation can create an invalid fat pointer and cause undefined behavior. Matching metadata need not be extracted from the identical pointer value—for example, a vtable from another allocation of the same concrete type can be valid—but it must match the allocation's concrete pointee/type. `MetadataPolicy::from_allocation` reliably panics for `DynMetadata`, and `SmartPointerPointee::metadata_from_allocation` forwards that behavior without documenting it.

**Concrete consequence**: Advanced downstream code can satisfy the written safety text while violating the actual pointer-metadata invariant, or encounter an undocumented panic through a public helper.

**Recommended action**: Give every public `resolve` method a canonical `# Safety` section requiring the data pointer and supplied/stored metadata to describe the same concrete pointee/type and allocation layout. Add canonical `# Panics` sections to `from_allocation` and `metadata_from_allocation` for metadata policies that require handle-supplied metadata.

**References**: Microsoft Pragmatic Rust Guidelines `M-CANONICAL-DOCS` and `M-UNSAFE`; [`ptr::from_raw_parts`](https://doc.rust-lang.org/std/ptr/fn.from_raw_parts.html); Rust Reference invalid values.

crates/multitude/src/coerce.rs
```diff
@@ -0,0 +1,107 @@
...
+        core::ptr::addr_eq(source, target),
+        "Coercion::new contract violated: coercion function changed the pointer address"
+    );
+    ptr_meta::metadata(target)
+}
+
+/// Builds a [`Coercion`] that unsizes to a trait object.
+///
+/// The syntax is `coerce!(dyn Trait)`, including bounds such as
+/// `coerce!(dyn Trait + Send)`. If the trait object refers to an enclosing
```
sandersaares: [Copilot speaking]

# F2 [P2] Document and test the actual trait-object coercion support boundary

**Impacted current locations**:
- `crates/multitude/src/coerce.rs:13-36,64-72`
- `crates/multitude/src/box.rs:198-209`
- `crates/multitude/src/arc.rs:252-263`
- `crates/multitude/src/rc.rs:241-252`
- `crates/multitude/src/lib.rs:450-453`
- `crates/multitude/tests/unsize.rs:20-43`

The API is presented as compiler-checked trait-object coercion, including `coerce!(dyn Trait + Send)`, but every consuming method requires `U: ptr_meta::Pointee<Metadata = DynMetadata<U>>`. With pinned `ptr_meta` 0.3.1, that excludes most ordinary foreign trait objects and some explicitly advertised custom-trait spellings.

**Reproducible reasoning**: `ptr_meta` supplies foreign trait-object implementations only for `dyn Any` and `dyn Error`, including their auto-trait combinations. Common targets such as `dyn Debug`, `dyn Display`, `dyn Future`, and `dyn Iterator` have no implementation, and downstream code cannot add one because both the trait and target type are foreign. The `#[pointee]` attribute generates an implementation for `dyn LocalTrait`, not separate `dyn LocalTrait + Send` or `+ Sync` types; the tests avoid that spelling by making `Send + Sync` supertraits and coercing to plain `dyn Shape`. Multitude's consumers also require `DynMetadata`, so the general raw-pointer `Coercion<T, U, F>` token does not imply general unsizing support.

**Concrete consequence**: Natural code such as `Box::unsize(arena.alloc_box(7_u32), coerce!(dyn Debug))` fails at a hidden metadata bound, and the documented `coerce!(dyn Trait + Send)` form fails for an ordinarily annotated local trait. Users discover the usable target set only from compiler errors.

**Recommended action**: First define and document the exact support boundary: which foreign trait objects work, that custom traits require the `dst` feature and `#[pointee]`, how explicit auto-trait spellings differ from supertraits, and that the smart-pointer consumers accept only `DynMetadata` targets. Add compile examples/tests for representative supported cases, `dyn Debug`, and the documented `dyn LocalTrait + Send` spelling so the selected boundary is explicit and enforced. Consider representation changes only if the intended contract is broader than that verified surface.

**References**: [`ptr_meta::Pointee` implementors](https://docs.rs/ptr_meta/0.3.1/ptr_meta/trait.Pointee.html); [`ptr_meta` 0.3.1 trait-object implementations](https://docs.rs/ptr_meta/0.3.1/src/ptr_meta/impls.rs.html); [`ptr_meta_derive` generated implementation](https://docs.rs/ptr_meta_derive/0.3.1/src/ptr_meta_derive/lib.rs.html#115-124); Rust Reference orphan rules; `Cargo.lock:3461-3467`. This is distinct from the representation discussion at [r3678715156](https://github.com/microsoft/oxidizer/pull/617#discussion_r3678715156): it concerns the undisclosed accepted target set, not whether the metadata parameter should exist.

crates/multitude/benches/criterion_arc_array_cg/linux.rs
```diff
@@ -1,10 +1,9 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-//! Instruction-precise `Arc<[Arc<[u8]>]>` build benchmarks for multitude.
+//! Linux Callgrind `Arc<[Arc<[u8]>]>` build benchmarks for multitude.
 //!
```
sandersaares, line 5: [Copilot speaking]

# F4 [P2] Make the paired Criterion and Callgrind array workloads equivalent

**Impacted current locations**:
- `crates/multitude/benches/criterion_arc_array.rs:78-110`
- `crates/multitude/benches/criterion_rc_array.rs:78-110`
- `crates/multitude/benches/criterion_arc_array_cg/linux.rs:42-45,101-158`
- `crates/multitude/benches/criterion_rc_array_cg/linux.rs:42-45,101-158`

Each Criterion routine performs one build, with each arena build receiving a fresh warmed 128-KiB arena. Each paired Callgrind routine instead performs 1,000 builds, pushes every result into a retained `Vec`, and reuses one arena for the entire run.

**Reproducible reasoning**: Criterion's `iter_batched` setup calls `warm_arena` for one routine invocation. Callgrind setup creates one arena and an output `Vec::with_capacity(N)`, then the timed body loops over `0..1_000`. For the full arena build, inner payloads alone consume `1_000 * 8 * 16 = 128,000` bytes; outer arrays, prefixes, and alignment necessarily exceed the 128-KiB warm allocation. Later iterations therefore measure refill/chunk growth and an increasing live set absent from Criterion. Global variants likewise add 1,000 loop branches and `Vec::push` operations and retain 1,000 live results.

**Concrete consequence**: Wall-clock and instruction-count results with paired names do not describe the same operation or allocator state, so they cannot cross-validate one another. Refill and retained-live-set effects can be misattributed to one array build.

**Recommended action**: Make each Callgrind scenario perform one build from the same setup state as Criterion and return the result/state for destruction outside the toggle. If amplification is necessary, add an explicitly matching Criterion batch scenario and report a normalized per-build value.

**References**: `docs/callgrind-benchmarks.md:42-46,267-281`; `docs/benchmarks.md:13-20`; Criterion 0.8.2 `Bencher::iter_batched`. This is distinct from the teardown-wording concern in [review 4813982517](https://github.com/microsoft/oxidizer/pull/617#pullrequestreview-4813982517): the issue is the one-build versus 1,000-retained-build workload and allocator-state mismatch.

crates/multitude/tests/loom.rs
```diff
@@ -73,6 +82,32 @@ mod loom_arc {
         });
     }
 
+    #[cfg(feature = "dst")]
```
sandersaares, line 85: [Copilot speaking]

# F6 [P3] Update the pre-existing local Loom runner to cover the new DST race

**Impacted current locations**:
- `crates/multitude/tests/loom.rs:84-109`
- `crates/multitude/scripts/correctness.rs:180-205`
- `crates/multitude/Cargo.toml:188-193`
- `justfiles/anvil/checks/loom.just:19-30,105-109`

The crate-local correctness runner already invoked the `loom` test target without enabling its required `loom` feature before this PR. The PR now adds a mixed sized/erased-handle regression additionally gated on `dst`, making that pre-existing runner break tightly coupled to the new coverage.

**Reproducible reasoning**: The target has `required-features = ["loom"]`, so the unchanged `cargo test --test loom` command cannot select it. A narrow fix that adds only `--features loom` would compile out `sized_and_erased_arc_clone_drop_race` because that test is also `#[cfg(feature = "dst")]`. The workspace Loom recipe already uses `--all-features` and documents that the feature gate and `--cfg loom` are independent requirements.

**Concrete consequence**: The documented crate-local correctness workflow does not run the Loom target, and an incomplete repair can report success without exercising the PR's new mixed-representation race.

**Recommended action**: Add `--all-features` to `run_loom`, matching the authoritative workspace recipe.

**References**: `justfiles/anvil/checks/loom.just:8-34,78-109`.

crates/multitude/docs/BUMPALO.md
```diff
@@ -14,15 +14,15 @@ spirit; here's how multitude differs.
 | Growable collections                              | ✅ `bumpalo::collections`   | ✅ `Vec`, `String`, `Utf16String`                                                                                                    |
 | `format!`-style macro                             | ✅                          | ✅                                                                                                                                   |
 | `#![no_std]`                                      | ✅                          | ✅                                                                                                                                   |
-| Smart-pointer width                               | 16 bytes                   | 8 bytes                                                                                                                             |
+| Smart-pointer width                               | 16 bytes                   | 8 bytes for sized, string, and slice pointees; 16 bytes for trait objects                                                           |
```
sandersaares: [Copilot speaking]

# F8 [P3] Correct the changed `bumpalo` pointer-width and DST rows

**Impacted current locations**:
- `crates/multitude/docs/BUMPALO.md:17,25`

The changed comparison rows say `bumpalo` smart pointers are uniformly 16 bytes and that it does not support dynamically sized types. Both statements are too broad for the pinned `bumpalo` 3.20.3 dependency.

**Reproducible reasoning**: `bumpalo::boxed::Box<'a, T: ?Sized>` is `#[repr(transparent)]` over `NonNull<T>` plus zero-sized `PhantomData`, so a sized `Box<T>` is one pointer wide and only wide pointees carry two words. The crate directly supports owned `Box<[T]>` and `Box<str>` forms, so the blanket DST ❌ is false even though stable automatic/custom trait-object coercion remains more limited.

**Concrete consequence**: The newly edited rows overstate Multitude's width and DST-support advantages and can mislead library selection.

**Recommended action**: Split `bumpalo` width by sized versus wide pointee and describe its supported slice/string DST ownership separately from its narrower trait-object coercion/construction surface.

**References**: [bumpalo 3.20.3 `Box` representation](https://github.com/fitzgen/bumpalo/blob/v3.20.3/src/boxed.rs#L147); `Cargo.lock:613-616`.

crates/multitude/src/de/json_error.rs
```diff
@@ -18,9 +18,10 @@ enum JsonErrorKind {
 /// rejection from malformed or incompatible JSON. The underlying
 /// [`serde_json::Error`] remains available through
 /// [`as_json_error`](Self::as_json_error) and [`Error::source`].
-/// [`Display`](fmt::Display) reports only the high-level classification so
-/// error reporters can render the source chain and backtrace without
-/// duplicating them.
+/// For parse and shape errors, [`Display`](fmt::Display) reports only a
```
sandersaares: [Copilot speaking]

# F9 [P3] Do not codify the pre-existing incomplete `JsonError` formatter

**Impacted current locations**:
- `crates/multitude/src/de/json_error.rs:15-24,96-111`

The changed type documentation explicitly codifies a pre-existing `Display` implementation that renders parse and shape errors only as `JSON deserialization failed`, despite the repository-selected canonical-error guideline requiring the formatter to include a summary, captured backtrace, and upstream cause information.

**Reproducible reasoning**: `JsonError` stores both an actionable `serde_json::Error` and, with `std`, a captured backtrace. Its unchanged formatter omits the parse message, line/column, and backtrace; the edited documentation now makes that omission the stated public behavior rather than merely leaving the formatter underdocumented.

**Concrete consequence**: Ordinary `{error}` logging cannot distinguish malformed, truncated, or shape-incompatible input and discards context already stored by the error.

**Recommended action**: Bring `Display` into the selected guideline by rendering the classification, underlying `serde_json::Error`, and captured backtrace, including limit details for limit violations, then align the documentation with that complete output.

**References**: Microsoft Pragmatic Rust Guidelines `M-ERRORS-CANONICAL-STRUCTS` and `M-PUBLIC-DISPLAY`.

crates/multitude/src/lib.rs
```diff
@@ -31,10 +31,11 @@
...
 //! 4. **Drop Support.** Owning handles run value destructors eagerly.
 //!
-//! 5. **Uniformly Thin Smart Pointers.** `multitude`'s escape-capable smart
-//!    pointers — [`Arc<T>`](Arc), [`Rc<T>`](Rc), and [`Box<T>`](Box) — are
-//!    **8 bytes** on 64-bit for *every* `T`, even DSTs like `str` and `[T]`
-//!    (the metadata lives in a chunk prefix). The arena-lifetime
+//! 5. **Compact Smart Pointers.** `multitude`'s escape-capable smart pointers —
+//!    [`Arc<T>`](Arc), [`Rc<T>`](Rc), and [`Box<T>`](Box) — are **8 bytes** on
+//!    64-bit for sized values, `str`, and `[T]`. Slice lengths live in a chunk
+//!    prefix; trait-object handles additionally carry one vtable word. The
```
sandersaares: [Copilot speaking]

# F12 [P3] Describe the extra handle word by metadata kind, not only trait-object syntax

**Impacted current locations**:
- `crates/multitude/src/lib.rs:34-37,102-104,450-453`
- `crates/multitude/tests/unsize.rs:344-357,433-449`

The crate-level documentation says only trait-object handles carry the additional vtable word. The actual rule is that every `DynMetadata`-backed pointee carries it, including a custom struct with a trailing trait-object DST.

**Reproducible reasoning**: `HeaderAndShape` is a struct, not itself a trait object, but its trailing `dyn Shape` gives the struct `DynMetadata`; the new test asserts `size_of::<Box<HeaderAndShape>>() == 2 * size_of::<usize>()`. `docs/DESIGN.md` already states the precise `DynMetadata` rule.

**Concrete consequence**: Users planning handle density can incorrectly assume custom trailing-DST owners remain one word unless the outer pointee type is literally `dyn Trait`.

**Recommended action**: Say that vtable-metadata pointees—including trait objects and structs with trait-object tails—use two-word handles. Rename `only_trait_object_handles_pay_for_metadata` to match and retain both direct and trailing-DST assertions.

**References**: [`ptr_meta::Pointee` pointer-metadata classes](https://docs.rs/ptr_meta/0.3.1/ptr_meta/trait.Pointee.html); `crates/multitude/docs/DESIGN.md:302-323`.

crates/multitude/src/box.rs
```diff
@@ -160,7 +194,29 @@ impl<T: ?Sized + Pointee, A: Allocator + Clone> Box<T, A> {
 
 impl_thin_smart_ptr_common!(Box);
 
-impl<T: ?Sized + Pointee, A: Allocator + Clone> From<Box<T, A>> for Pin<Box<T, A>> {
+impl<T, A: Allocator + Clone> Box<T, A> {
+    /// Converts this owner into a trait-object owner without moving or
```
sandersaares: [Copilot speaking]

# F13 [P3] Keep the new public API summaries within one rendered line

**Impacted current locations**:
- `crates/multitude/src/box.rs:198-199`
- `crates/multitude/src/arc.rs:252-253,272-273`
- `crates/multitude/src/rc.rs:241-242,261-262`
- `crates/multitude/src/arena/alloc_unsized.rs:725-726`

The new `unsize`, `unsize_pin`, and `alloc_dst_box_pin` first paragraphs render as 87-112 characters because summary and implementation detail remain in one contiguous paragraph. The applicable Rust API-documentation instruction requires each public item's first rendered paragraph to fit on one 80-character line.

**Reproducible reasoning**: Markdown joins each listed pair of contiguous `///` lines into one paragraph; no markup hides enough text to meet the limit. This rule applies to these public API items.

**Concrete consequence**: Rustdoc wraps the summary paragraphs and violates the repository's scannable-summary convention on every new coercion method.

**Recommended action**: Use a concise first sentence such as “Converts this owner into a trait-object owner in place.” or “`Pin` variant of `alloc_dst_box`.” Move no-move, no-reallocation, and vtable-storage details into a separate paragraph.

**References**: `C:\Users\sasaares\dev-env\agents\.github\instructions\rust.instructions.md` (“API documentation”).

crates/multitude/src/internal/thin_dst.rs
```diff
@@ -1,52 +1,270 @@
...
+    /// `value_ptr` must point to a live value allocated for `Self`.
+    unsafe fn resolve_metadata(value_ptr: NonNull<u8>, stored: Self::StoredMetadata) -> Self::Metadata;
+}
+
+mod sealed {
+    use ptr_meta::{DynMetadata, Pointee};
+
+    use super::MetadataPolicy;
+
+    pub trait MetadataSealed<T: ?Sized + Pointee> {}
```
sandersaares: [Copilot speaking]

# F15 [P3] Use crate visibility for the private sealed helper traits

**Impacted current locations**:
- `crates/multitude/src/internal/thin_dst.rs:169,180,184`

`MetadataSealed`, `Sealed`, and `HandleMetadataSealed` are declared `pub` even though every use is crate-internal and their enclosing `sealed` module is private. Their effective privacy therefore relies on the parent module rather than expressing the intended visibility on the items.

**Reproducible reasoning**: The traits exist only to seal `MetadataPolicy`, `SmartPointerPointee`, and `HandleMetadata`; no public signature needs the sealed traits themselves to be exported. `pub(crate)` preserves all current bounds and implementations. The applicable Rust visibility instruction explicitly says not to rely on private parents to restrict a `pub` item.

**Concrete consequence**: The new implementation-only traits violate the repository's explicit visibility convention and obscure which hidden traits are unavoidable public-signature artifacts versus purely internal sealing machinery.

**Recommended action**: Change the three declarations to `pub(crate) trait`, retaining narrowly justified `private_bounds` allowances only on the public hidden traits that actually appear in public smart-pointer signatures.

**References**: `C:\Users\sasaares\dev-env\agents\.github\instructions\rust.instructions.md` (“pub is pub”).

crates/multitude/src/coerce.rs
geeknoid: Note to robot: This comment is also applicable to plurality's equivalent functionality, so the fix we do here should also be applied to plurality. We should do this as part of this PR.

crates/multitude/src/de/value.rs
geeknoid: Addressed in cf7ff79c. `deserialize_ignored_any` again calls `visit_unit()`, so discarding a materialized value is shape-independent and constant-time. The old boolean-specific test is replaced with unknown-field replay coverage for explicit unit, newtype, tuple, and struct enum payloads plus a nested collection.

crates/multitude/src/de/reuse.rs
geeknoid: Addressed in cf7ff79c. The implementation now preserves the stronger empty-on-error contract without staging: it clears before deserialization and clears again whenever `deserialize_str` returns `Err`, retaining the existing allocation. A late-failing custom deserializer regression visits `"replacement"`, then errors, and verifies the string is empty with the same allocation.

crates/multitude/src/internal/thin_dst.rs
geeknoid: Addressed in cf7ff79c. The public hidden resolution helpers now require the stored or handle metadata to describe the live allocation actual concrete pointee type and layout. Canonical `# Panics` sections now document allocation-metadata construction for policies that require handle-supplied metadata.

crates/multitude/src/coerce.rs
geeknoid: Addressed in cf7ff79c for both crates. Multitude now documents and tests its exact `ptr_meta` boundary: `Any`/`Error` combinations work without `dst`, annotated custom traits require `dst`, explicit local `+ Send` spellings are not generated, and unsupported foreign targets such as `dyn Debug` are compile-fail examples. Plurality now explicitly documents its broader native-fat-pointer boundary and tests `dyn LocalTrait + Send`. Both APIs also distinguish compiler-checked safe constructors from unsafe `Coercion::new`.

crates/multitude/benches/criterion_arc_array_cg/linux.rs
geeknoid: Addressed in cf7ff79c. Each paired Callgrind scenario now performs exactly one build from the corresponding Criterion setup state and returns the result and setup state for teardown outside the toggle. The 1,000-build loop, retained output vector, and arena-refill/live-set mismatch are removed.

crates/multitude/tests/loom.rs
geeknoid: Addressed in cf7ff79c. The crate-local Loom runner now passes `--all-features`, matching the authoritative workspace recipe and enabling both the `loom` target and the `dst` mixed sized/erased race. The serialized release invocation exercises all 21 models.

crates/multitude/docs/BUMPALO.md
geeknoid: Addressed in cf7ff79c. The comparison now reports `bumpalo::boxed::Box` as one word for sized pointees and wide for wide pointees, and recognizes owned slice/string DST support while accurately describing its more limited trait-object construction surface.

crates/multitude/src/de/json_error.rs
geeknoid: Addressed in cf7ff79c. `JsonError::Display` now renders the classification (including limit details), the captured backtrace when `std` is enabled, and the underlying `serde_json::Error`. The type documentation and display assertions now reflect the complete canonical output.

geeknoid: Addressed in cf7ff79c. `JsonError::Display` now renders the classification (including limit details), the captured backtrace when `std` is enabled, and the underlying `serde_json::Error`. The type documentation and display assertions now reflect the complete canonical output.

crates/multitude/src/lib.rs
geeknoid: Addressed in cf7ff79c. Crate and table documentation now describe the two-word case by vtable metadata rather than outer syntax, explicitly including custom DSTs with trait-object tails. The layout test was renamed accordingly and asserts `Box`, `Rc`, and `Arc` widths for both direct trait objects and `HeaderAndShape`.

crates/multitude/src/box.rs
geeknoid: Addressed in cf7ff79c. The affected public items now have concise one-line summary paragraphs, with no-move, no-reallocation, exposure, target-boundary, and vtable details moved into separate paragraphs.

crates/multitude/src/internal/thin_dst.rs
geeknoid: Addressed in cf7ff79c. All three traits inside the private `sealed` module now use `pub(crate)`. Narrow `private_bounds` allowances remain only on the public hidden implementation-detail traits whose signatures require those sealing bounds.

geeknoid:
Addressed the two suppressed alignment-documentation findings from the latest Copilot review in a501b81a. The scalar `BytemuckView::{alloc,try_alloc}` and `ZerocopyView::{alloc,try_alloc}` contracts now state their actual 32 KiB ceiling inherited from `Arena::try_alloc_with`; the separate slice contracts retain their correct 64 KiB arena-reference ceiling.

geeknoid:
Addressed the latest suppressed no-`std` serialization-test findings in 830f674e. Backtrace method calls and the dedicated capture test are now gated on `std`; display tests assert the backtrace section is present with `std` and absent without it while still checking the underlying JSON cause in both modes. The full `--no-default-features --features serde_json` serialization target now passes.

psandana (approved):

sandersaares (approved):

PR #618, opened 2026-07-30
docs: clarify no_std support policy

sandersaares:
[Copilot speaking]

## Summary
- treats adoption or expansion of `no_std` support as optional when it would add disproportionate implementation complexity
- defines every documented `no_std` configuration as a real compatibility promise rather than best-effort support
- requires supported configurations to work correctly, be covered by CI, and document their prerequisites and target boundary

## Validation
- `just spellcheck`


Vaiz:
this clarification is only for agents. what about developers? do they also have to follow it?

wukchung (approved):

sandersaares:
> this clarification is only for agents. what about developers? do they also have to follow it?

The way I see it, we are in a world where the agents are the developers.

sgalkin (approved):

PR #619, opened 2026-07-31
feat: Clarify that .unwrap() is fine in tests

sandersaares:
Addressing some feedback on PR feedback items.


Vaiz (approved):

ralfbiedert (commented):
🤖 Review findings at `f9eae21c`:

1. **The new `.unwrap()` guidance conflicts with the enforced lint policy** (`AGENTS.md:87-89`, `Cargo.toml:325`, `.github/workflows/main.yml:168-169`). Tests and examples are told to use `.unwrap()`, while `clippy.unwrap_used = "warn"` is promoted to an error by CI. Following the guidance therefore blocks the PR unless every use gets a scoped `#[expect(clippy::unwrap_used, reason = "...")]`. Please either require that override explicitly or retain `.expect()` as the default. The term “test code” should also exclude public test utilities under `src/testing`, which are library code.
2. **The local Clippy recipe does not reproduce CI** (`justfiles/basic.just:59-60`). `just clippy` does not pass `-D warnings`, so the checklist can succeed locally while CI rejects the same warnings. Please add `-- -D warnings` or explicitly require a warning-free result.

sandersaares:
> The new .unwrap() guidance conflicts with the enforced lint policy

Lints are not laws - it is implicit that when needed, they can be suppressed. This does not need to be said.

> The local Clippy recipe does not reproduce CI

Out of scope of this PR. Presumably fixed by Anvil migration.

sgalkin (approved):

PR #622, opened 2026-07-31
refactor(fetch_tls): unify client identity configuration

kate-shine:
## Summary

- represent key-material identities and rustls client-certificate resolvers with one internal `ClientAuth` state
- preserve the portable `client_identity` and rustls-specific `client_identity_resolver` APIs
- make repeated identity configuration predictable: the last builder call wins instead of the resolver silently taking precedence
- share configured key material through `Arc` so cloning TLS options does not duplicate private-key bytes

## Validation

- `fetch_tls` formatting
- 71 crate tests
- Clippy across all targets and features with warnings denied
- API docs with no features and all features
- generated README and spelling checks

### DIFF
--- a/crates/fetch_tls/src/options.rs
+++ b/crates/fetch_tls/src/options.rs
@@ -3,7 +3,11 @@
 
 //! [`TlsOptions`] and its type-state builder [`TlsOptionsBuilder`].
 
+use std::sync::Arc;
+
 use http::Version;
+#[cfg(any(feature = "rustls", test))]
+use rustls::client::ResolvesClientCert;
 
 use crate::client_identity::ClientIdentity;
 use crate::{TlsBackend, TlsBackendBuilder};
@@ -103,7 +107,14 @@ pub struct TlsOptions {
 #[derive(Debug, Clone, Default)]
 pub(crate) struct SharedOptions {
     pub(crate) supported_http_versions: Option<Vec<Version>>,
-    pub(crate) client_identity: Option<ClientIdentity>,
+    pub(crate) client_auth: Option<ClientAuth>,
+}
+
+#[derive(Debug, Clone)]
+pub(crate) enum ClientAuth {
+    Identity(Arc<ClientIdentity>),
+    #[cfg(any(feature = "rustls", test))]
+    Resolver(Arc<dyn ResolvesClientCert>),
 }
 
 impl SharedOptions {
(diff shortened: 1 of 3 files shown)


crates/fetch_tls/src/options.rs
```diff
@@ -103,7 +107,14 @@ pub struct TlsOptions {
 #[derive(Debug, Clone, Default)]
 pub(crate) struct SharedOptions {
     pub(crate) supported_http_versions: Option<Vec<Version>>,
-    pub(crate) client_identity: Option<ClientIdentity>,
+    pub(crate) client_auth: Option<ClientAuth>,
+}
+
+#[derive(Debug, Clone)]
+pub(crate) enum ClientAuth {
+    Identity(Arc<ClientIdentity>),
```
geeknoid: **The featureless configuration now fails denied-warning checks** — *Correctness · Medium · High confidence*

`ClientAuth::Identity` is always compiled, but its tuple payload is only read by backend-gated code. In the no-feature configuration, the existing CI annotations report `field '0' is never read` here; because warnings are denied, the crate's testing and external-type checks fail before exercising the refactor. This also violates the repository's `M-STATIC-VERIFICATION` gate, so I am aggregating the correctness and conformance passes into this one root cause.

**Direction:** keep the featureless API available while making the conditional-use intent lint-clean, for example with a narrowly justified dead-code expectation/allowance on the payload or an equivalent representation that is read in every supported configuration.

**Done when:** the no-feature and backend-feature static-verification configurations accept `fetch_tls` without suppressing unrelated warnings.

geeknoid (commented):
Early feedback on the draft: one introduced configuration regression is blocking the current head. The new `ClientAuth::Identity` tuple payload is unread when neither TLS backend feature is enabled, and the repository's denied-warning checks reject that configuration. I left one inline comment with the verified anchor and acceptance criterion.

The security/TLS read was otherwise clean: server-certificate verification and TLS-version selection are unchanged, resolver and key-material variants remain mutually exclusive, and `ClientIdentity` debug output still redacts private-key bytes. The intended last-setter-wins compatibility change is documented on both public setters and covered in both call orders.

All five diff-scoped taxonomies were assessed (150 security, 162 correctness, 182 testing, 183 performance, 143 conformance), including feature combinations, secret lifetime/debugging, `Arc` sharing, API compatibility, and the unsafe/UB protocols. Nothing was executed; conclusions use source, repository configuration, existing PR comments, and existing CI annotations only. No review was posted.

PR #632, opened 2026-08-03
refactor(tick): align API with Rust guidelines

martintmk:
## Overview

Aligns the `tick` public API with the Rust API Guidelines and Microsoft Rust Guidelines. The change makes time conversion failures explicit, replaces mutable test-clock configuration with a builder, strengthens error and future contracts, and fixes timer waker handling so callbacks run outside internal locks.

Workspace consumers are migrated to the revised API to keep behavior consistent. This is justified by clearer failure semantics, a smaller and more predictable public surface, feature-safe documentation, and correct async task wake behavior.

## Validation

- Every `tick` feature combination compiles
- 192 unit tests and 62 all-feature doctests pass
- Strict Clippy, rustdoc warnings, formatting, generated README, and spelling checks pass
- Nightly line coverage reaches 100% across all-feature and no-default-feature configurations
- Mutation testing reports no surviving mutants or timeouts
- Independent Opus and Sonnet reviews found no significant issues
- `fetch`, `fetch_hyper`, `http_extensions`, `seatbelt`, and `seatbelt_http` compile with all features and targets

### DIFF
--- a/crates/tick/src/clock.rs
+++ b/crates/tick/src/clock.rs
@@ -43,10 +43,11 @@ use crate::timers::TimerKey;
 /// # Clock construction
 ///
 /// The clock requires a runtime to drive the registered timers. This crate provides built-in support
-/// for Tokio via [`Clock::new_tokio()`] (available with the `tokio` feature). For other async runtimes,
+/// for Tokio via [`Clock::new_tokio`] (available with the `tokio` feature). For other async runtimes,
 /// you can use types in the [`runtime`][crate::runtime] module to drive the clock.
 ///
-/// In tests, the clock can be constructed directly using [`ClockControl`][crate::ClockControl] or via [`Clock::new_frozen`][crate::Clock::new_frozen]
+/// In tests, the clock can be constructed directly using [`ClockControl`] or via
+/// [`Clock::new_frozen`](https://docs.rs/tick/latest/tick/struct.Clock.html#method.new_frozen)
 /// (available with the `test-util` feature) because the passage of time is controlled manually.
 ///
 /// See the [Testing](#testing) section for more information.
@@ -59,7 +60,7 @@ use crate::timers::TimerKey;
 ///
 /// The ability to jump forward in time makes tests faster, more reliable and gives you complete control over the passage of time.
 /// By default, the clock does not allow you to control the passage of time. However, when the `test-util` feature is enabled,
-/// this crate provides a [`ClockControl`][crate::ClockControl] type that can be used to control time.
+/// this crate provides a [`ClockControl`] type that can be used to control time.
 ///
 /// # Cloning and shared state
 ///
@@ -93,10 +94,10 @@ use crate::timers::TimerKey;
 ///
 /// - **`ClockControl` clocks** (`test-util`): Relocation is a no-op. All clones share the same
 ///   controlled time state regardless of which thread they are on. This is intentional, a single
-///   [`ClockControl`][crate::ClockControl] controls time for all clocks derived from it, even
+///   [`ClockControl`] controls time for all clocks derived from it, even
 ///   across threads.
 ///
-/// - **Tokio clocks** (created via [`Clock::new_tokio()`]): Relocation is a no-op. The Tokio clock
+/// - **Tokio clocks** (created via [`Clock::new_tokio`]): Relocation is a no-op. The Tokio clock
 ///   is driven by a single background task that advances a shared set of timers, so all clones
 ///   share the same timer storage regardless of which thread they are on. Per-core relocation
 ///   would create independent timer storage on the destination thread that the background driver
@@ -172,6 +173,9 @@ use crate::timers::TimerKey;
 /// }
 /// # }
 /// ```
+///
+/// [`ClockControl`]: https://docs.rs/tick/latest/tick/struct.ClockControl.html
+/// [`Clock::new_tokio`]: https://docs.rs/tick/latest/tick/struct.Clock.html#method.new_tokio
 #[derive(Clone)]
 pub struct Clock {
     state: ClockState,
@@ -210,6 +214,7 @@ impl Clock {
     ///
     /// Panics if called outside of a Tokio runtime context.
     #[cfg(any(feature = "tokio", test))]
+    #[cfg_attr(docsrs, doc(cfg(feature = "tokio")))]
     #[must_use]
     #[cfg_attr(test, mutants::skip)] // Causes test timeout.
     pub fn new_tokio() -> Self {
@@ -283,6 +288,7 @@ impl Clock {
     /// assert_eq!(instance, clock.instant());
     /// ```
     #[cfg(any(feature = "test-util", test))]
+    #[cfg_attr(docsrs, doc(cfg(feature = "test-util")))]
     #[must_use]
     pub fn new_frozen() -> Self {
         crate::ClockControl::new().to_clock()
@@ -311,6 +317,7 @@ impl Clock {
     /// assert_eq!(system_time, clock.system_time());
     /// ```
     #[cfg(any(feature = "test-util", test))]
+    #[cfg_attr(docsrs, doc(cfg(feature = "test-util")))]
     #[must_use]
     pub fn new_frozen_at(time: impl Into<SystemTime>) -> Self {
         crate::ClockControl::new_at(time).to_clock()
@@ -339,7 +346,7 @@ impl Clock {
         self.simple_clock().system_time()
     }
 
-    /// Retrieves the current system time converted to a target type.
+    /// Retrieves the current system time and converts it to `T`.
     ///
     /// This is a convenience method that retrieves the current [`SystemTime`] via
     /// [`system_time()`][Self::system_time] and converts it to the specified target type.
@@ -349,26 +356,26 @@ impl Clock {
     /// * `T` - The target type that implements [`TryFrom<SystemTime>`]. Common examples include
     ///   timestamp types from external crates that can be constructed from a [`SystemTime`].
     ///
-    /// # Panics
+    /// # Errors
     ///
-    /// Panics if the current system time cannot be represented by the target type `T`.
+    /// Returns the target type's conversion error when the current [`SystemTime`] is outside
+    /// the target type's representable range.
     ///
-    /// Callers must choose a target type whose representable range covers the
-    /// [`SystemTime`] values they expect. The conversion can fail in two cases:
+    /// # Examples
     ///
-    /// - **In production**, if `T` has a narrower representable range than [`SystemTime`] and the
-    ///   current system time falls outside it.
-    /// - **In tests** using manual time control (via the `test-util` feature), if controlled time
-    ///   is moved outside the target type's supported range.
+    /// ```
+    /// use std::time::SystemTime;
     ///
-    /// If `T` can represent every [`SystemTime`] value, this conversion never fails and this method
-    /// never panics — the panic above only occurs for target types whose representable range does
-    /// not cover the current time.
+    /// use tick::Clock;
     ///
-    /// Kept consistent with [`SimpleClock::system_time_as`][crate::SimpleClock::system_time_as].
-    #[must_use]
-    pub fn system_time_as<T: TryFrom<SystemTime>>(&self) -> T {
-        self.simple_clock().system_time_as()
+    /// # fn example(clock: &Clock) -> Result<(), std::convert::Infallible> {
+    /// let time = clock.try_system_time_as::<SystemTime>()?;
+    /// assert_eq!(time, clock.system_time());
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn try_system_time_as<T: TryFrom<SystemTime>>(&self) -> Result<T, T::Error> {
+        self.simple_clock().try_system_time_as()
     }
 
     /// Retrieves the current [`Instant`] time.
@@ -425,7 +432,7 @@ impl Clock {
     /// assert!(stopwatch.elapsed() >= Duration::from_millis(10));
     /// # }
     /// ```
-    #[must_use]
+    #[must_use = "futures do nothing unless awaited or polled"]
     pub fn delay(&self, duration: Duration) -> crate::Delay {
         crate::Delay::new(self, duration)
     }
@@ -468,6 +475,14 @@ impl Clock {
             #[cfg(any(feature = "test-util", test))]
             ClockState::ClockControl(control) => control.unregister_timer(key),
             ClockState::System(timers) => timers.with_timers(|t| t.unregister(key)),
+        }
+    }
+
+    pub(super) fn update_timer_waker(&self, key: TimerKey, waker: &Waker) {
+        match self.clock_state() {
+            #[cfg(any(feature = "test-util", test))]
+            ClockState::ClockControl(control) => control.update_timer_waker(key, waker),
+            ClockState::System(timers) => timers.with_timers(|timers| timers.update_waker(key, waker)),
         }
     }
 
@@ -656,8 +671,7 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "target type cannot represent the current SystemTime")]
-    fn system_time_as_panics_on_conversion_failure() {
+    fn try_system_time_as_returns_conversion_failure() {
         /// A newtype that always fails conversion from `SystemTime`.
         struct AlwaysFailsConversion;
 
@@ -670,7 +684,10 @@ mod tests {
         }
 
         let clock = Clock::new_frozen();
-        let _: AlwaysFailsConversion = clock.system_time_as();
+        assert_eq!(
+            clock.try_system_time_as::<AlwaysFailsConversion>().err(),
+            Some("conversion always fails")
+        );
     }
 
     #[test]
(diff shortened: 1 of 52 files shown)


crates/tick/src/timeout.rs
```diff
@@ -1,35 +1,37 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-use std::io::ErrorKind;
 use std::pin::Pin;
 use std::task::{Context, Poll};
```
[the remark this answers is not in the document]
martintmk: &#x1F916; This module compiles as written because the workspace uses Rust edition 2024, whose prelude includes `Future` and `IntoFuture`. The explicit import would be redundant, and the branch passes both `cargo check --workspace --all-features` and the `tick` test suite.

---

crates/tick/src/runtime/clock_gone.rs
```diff
@@ -11,7 +11,7 @@ use std::fmt::{Display, Formatter, Result};
 ///
 /// Runtime integrations should use this error to determine when to stop the timer
 /// advancement loop.
-#[derive(Debug)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
```
martintmk: Why do we want to add all these derives that are not necessary for this type?

Adding derive to type is not a breaking change, we can do it later if required. Review unnecessarily added derives and remove them

crates/tick/src/runtime/mod.rs
```diff
@@ -69,6 +69,9 @@ mod clock_driver;
 mod clock_gone;
 mod inactive_clock;
 
+#[doc(inline)]
```
martintmk: why inlining docs?

crates/tick/src/clock.rs
```diff
@@ -210,6 +214,7 @@ impl Clock {
     ///
     /// Panics if called outside of a Tokio runtime context.
     #[cfg(any(feature = "tokio", test))]
+    #[cfg_attr(docsrs, doc(cfg(feature = "tokio")))]
```
martintmk: not needed, this code `#![cfg_attr(docsrs, feature(doc_cfg))]` in lib.rs takes care of this across all crate.

crates/tick/src/periodic_timer.rs
```diff
@@ -78,6 +78,7 @@ use crate::timers::TIMER_RESOLUTION;
 /// # }
 /// ```
 #[derive(Debug)]
+#[must_use = "streams do nothing unless polled"]
```
martintmk, line 81: is this really necessary?

crates/tick/src/periodic_timer.rs
```diff
@@ -142,7 +143,10 @@ impl Stream for PeriodicTimer {
                 Poll::Ready(Some(()))
             }
             // Timer is registered and will fire later in the future.
-            Some(_) => Poll::Pending,
+            Some(key) => {
+                this.clock.update_timer_waker(key, cx.waker());
```
martintmk, line 147: why was this change added?

crates/tick/src/stopwatch.rs
```diff
@@ -64,17 +64,22 @@ impl Stopwatch {
...
+    /// ```
+    /// use tick::SimpleClock;
+    ///
+    /// let clock = SimpleClock::new_system();
+    /// let stopwatch = clock.stopwatch();
+    ///
+    /// assert!(stopwatch.start() <= clock.instant());
+    /// ```
+    #[must_use]
+    pub const fn start(&self) -> Instant {
```
martintmk: we could add a start method later for Stowpatch, suggest alternative names

crates/tick/src/runtime/clock_gone.rs
martintmk: &#x1F916; Agreed - I am removing the unnecessary ClockGone derives and retaining only the traits its error role currently needs.

---

crates/tick/src/runtime/mod.rs
martintmk: &#x1F916; Agreed - the private-module re-exports are already inlined by rustdoc, so I am removing the redundant attributes.

---

crates/tick/src/clock.rs
martintmk: &#x1F916; Agreed - I am removing the per-method doc(cfg) additions and leaving the crate-level docs.rs configuration as the existing source of truth.

---

crates/tick/src/stopwatch.rs
martintmk: &#x1F916; Agreed - there is no current need to expose the start instant, so I am removing this method. If added later, started_at is the clearest name; start_instant is the more type-explicit alternative.

---

crates/tick/src/runtime/clock_gone.rs
martintmk: &#x1F916; Fixed in `bd24b1e5`: `ClockGone` now derives only `Debug`, which is required for its `Error` implementation.

---

crates/tick/src/runtime/mod.rs
martintmk: &#x1F916; Fixed in d24b1e5: removed the redundant doc(inline) attributes from the private-module re-exports.

---

crates/tick/src/clock.rs
martintmk: &#x1F916; Fixed in d24b1e5: removed the three newly added per-method doc(cfg) attributes.

---

crates/tick/src/periodic_timer.rs
martintmk: &#x1F916; Yes. PeriodicTimer is a lazy Stream, so dropping it without polling performs no work. The type-level must_use also catches values returned through helper functions, whereas the constructor-level attribute only catches a directly ignored constructor call. I kept this contract.

---

crates/tick/src/periodic_timer.rs
martintmk: &#x1F916; This fixes a correctness bug when a pending timer is polled again with a different task waker. Without replacing the stored waker, timer expiry wakes the old task and the current task may never be polled again. The latest_waker_is_notified regression test verifies that only the latest waker fires, so I kept this change.

---

crates/tick/src/stopwatch.rs
martintmk: &#x1F916; Fixed in d24b1e5: removed the unused accessor and the associated Clone addition. If this API is needed later, I recommend started_at; start_instant is a more explicit alternative.

---

crates/tick/src/fmt/mod.rs
```diff
@@ -136,9 +136,13 @@ mod iso_8601;
 mod rfc_2822;
 mod unix_seconds;
 
+#[doc(inline)]
```
martintmk: this is not needed

crates/tick/src/periodic_timer.rs
martintmk: could it be that multiple wakers are awating the poll?

crates/tick/src/simple_clock.rs
```diff
@@ -46,6 +49,7 @@ enum TimeKind {
     System,
     /// Reads time controlled by a [`ClockControl`][crate::ClockControl].
     #[cfg(any(feature = "test-util", test))]
+    #[cfg_attr(docsrs, doc(cfg(feature = "test-util")))]
```
martintmk: not needed (and others too)

crates/tick/src/state.rs
```diff
@@ -123,7 +123,14 @@ impl SynchronizedTimers {
 
     #[cfg_attr(test, mutants::skip)] // Causes test timeout.
     pub(crate) fn try_advance_timers(&self, now: Instant) -> Option<Instant> {
-        self.with_timers(|timers| timers.advance_timers(now))
+        let mut ready = Vec::new();
```
martintmk: we need to avoid allocating this vector, why we need to capture the wakerss , why we cannot advance directly inside advance_timers

crates/tick/src/periodic_timer.rs
martintmk: &#x1F916; A stream has one current task waker at a time. Each `Poll::Pending` must arrange for the waker from that poll to be notified; if the stream is moved to another task, the previous task is no longer the current waiter. Concurrent polling of the same stream requires external synchronization, which serializes access and still presents one waker per poll. Replacing the stale waker is therefore the correct behavior.

---

crates/tick/src/fmt/mod.rs
martintmk: &#x1F916; Agreed - I will remove the redundant `doc(inline)` attributes.

---

crates/tick/src/simple_clock.rs
martintmk: &#x1F916; Agreed - I will remove the newly added per-item `doc(cfg)` attributes throughout the crate.

---

crates/tick/src/state.rs
martintmk: &#x1F916; Agreed on avoiding the extra vector allocation. Wakers still need to be detached from the locked timer state before invocation because `wake()` may synchronously re-enter timer code and deadlock. I will accumulate the already-allocated expired map nodes instead of copying them into a `Vec`.

---

crates/tick/src/clock_control.rs
```diff
@@ -92,6 +93,43 @@ impl std::fmt::Debug for ClockControl {
...
+/// ```
+#[derive(Debug, Clone)]
+pub struct ClockControlBuilder {
+    time: SystemTime,
+    auto_advance: Duration,
+    auto_advance_total_max: Option<Duration>,
+    auto_advance_timers: bool,
+}
+
+impl Default for ClockControlBuilder {
```
martintmk: not needed

crates/tick/src/fmt/mod.rs
martintmk: &#x1F916; Fixed: removed all redundant `doc(inline)` attributes added by this PR.

---

crates/tick/src/simple_clock.rs
martintmk: &#x1F916; Fixed: removed all per-item `doc(cfg)` attributes added by this PR.

---

crates/tick/src/state.rs
martintmk: &#x1F916; Fixed: expired wakers now remain in their existing `BTreeMap` nodes and are accumulated with `BTreeMap::append`, avoiding the extra `Vec` allocation. They are still invoked after the timer or clock-control lock is released to prevent synchronous re-entry from deadlocking.

---

crates/tick/src/clock_control.rs
martintmk: &#x1F916; Agreed - all builder fields already have the intended standard defaults, so I will derive `Default` and remove the manual implementation.

---

martintmk: &#x1F916; Fixed: removed the `Default` implementation and `Default` trait assertion from `ClockControlBuilder`. Builder defaults remain defined only by `ClockControl::builder()`.

---

crates/tick/src/runtime/inactive_clock.rs
```diff
@@ -15,17 +15,18 @@ use crate::state::ClockState;
 /// This is the default mode. Clones can be relocated to different threads via
 /// [`ThreadAware::relocate`], producing independent timer storage per core. Suitable for
 /// thread-per-core runtimes.
-#[derive(Debug, Default, Clone, Copy)]
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Hash)]
```
martintmk:
```suggestion
#[derive(Debug)]
```

Why not just this?

martintmk: &#x1F916; Agreed that the newly added derives are unnecessary. I will restore the pre-PR surface: `Isolated` keeps its existing `Debug`, `Default`, `Clone`, and `Copy` implementations to avoid an unrelated breaking change, while `Shared` returns to `Debug` only.

---

martintmk: &#x1F916; Fixed: removed the newly added `PartialEq`, `Eq`, and `Hash` derives from `Isolated`, and restored `Shared` to `Debug` only. The pre-existing `Default`, `Clone`, and `Copy` implementations on `Isolated` remain unchanged.

---

crates/tick/src/timers.rs
```diff
@@ -86,6 +88,17 @@ impl Timers {
         self.wakers.remove(&id);
     }
 
+    /// Replaces the waker for an existing timer when the awaiting task changes.
+    pub(crate) fn update_waker(&mut self, id: TimerKey, waker: &Waker) {
```
martintmk, line 92: **Posted by an AI agent**

**Treat a missing timer during waker replacement as already fired**

**Problem**
`update_waker` silently returns when the timer entry is gone, while `Delay::poll` and `PeriodicTimer::poll_next` still return `Pending`. The driver can remove the timer after the caller checks the deadline but before this lookup; it then wakes the previously registered waker, not a newly supplied waker.

**Why this matters**
If the future moved to a different task, that task may never be woken.

**Suggested fix**
Return whether the timer was still registered and treat a miss as fired (clear the key and return `Ready`) instead of returning `Pending`.

crates/tick/src/delay.rs
```diff
@@ -112,7 +113,10 @@ impl Future for Delay {
 
                 Poll::Ready(())
             }
-            Some(_) => Poll::Pending,
+            Some(key) => {
+                this.clock.update_timer_waker(key, cx.waker());
```
martintmk, line 117: **Posted by an AI agent · Non-blocking**

**Avoid taking the shared timer mutex on every pending poll**

**Problem**
The pending arm now calls `update_timer_waker` on every poll. For `Clock::new_tokio`, this acquires the single shared `Mutex<Timers>`, even when `will_wake` later proves the waker unchanged.

**Why this matters**
Timeouts around frequently-polled I/O therefore add global lock contention to a path the timer-state documentation says is not hot.

**Suggested fix**
Cache the last registered waker in the future and only enter the timer map when it changes, or use a per-timer waker slot.

martintmk (commented):
**Posted by an AI agent**

Review completed across all changed files and the paired all-features `tick` public API. Three findings: a missing timer during waker replacement can leave a moved future pending, unchanged wakers still take the shared timer mutex on every poll, and the breaking public-API migration is incomplete. No vote was cast because this is my own PR.

**Document or preserve the breaking public-API migration**

**Problem**
The paired all-features API diff removes `system_time_as`, four direct `ClockControl` configuration methods, both `From<Stopwatch>` conversions, and changes `Timeout<F, D>` to `Timeout<F>`. Rustdoc explains the new builder and fallible conversion, but does not cover every migration (notably `advance_millis`, `Timeout` arity, and `Stopwatch` conversions).

**Why this matters**
Consumers can encounter undocumented source-breaking changes across several public paths.

**Suggested fix**
Ensure release/versioning and migration documentation cover all removed paths, or retain deprecated compatibility shims where practical.

crates/tick/src/clock_control.rs
```diff
@@ -439,14 +343,87 @@ impl ClockControl {
...
+    }
+
+    /// Builds the configured [`ClockControl`].
+    ///
+    /// # Panics
+    ///
+    /// Panics if the initial time is outside the range supported by the platform's
+    /// [`SystemTime`] or [`Instant`].
+    #[must_use]
+    pub fn build(self) -> ClockControl {
```
geeknoid: **Return builder validation failures instead of panicking** — *Conformance · Low · High confidence*

`ClockControlBuilder::time` accepts any `SystemTime`, but `build()` returns `ClockControl` even though `State::set_time` can fail through the checked `Instant` conversion and panic. That makes this newly introduced builder fallible through an unwind rather than through its type. [M-BUILD-RESULT](https://microsoft.github.io/rust-guidelines/guidelines/libs/resilience/#m-build-result) says that where builders are fallible, they should offer a `Result`-carrying `.build()` so final and cross-field validation is handled explicitly.

**Direction:** return a typed error from `build()`, or use a validated input type that makes the conversion infallible; keep any panicking constructor as an explicitly documented convenience wrapper.

**Done when:** an initial time that cannot be represented by the paired `Instant` is reported as a typed build error, with a boundary test, or the builder is made structurally infallible.

geeknoid (commented):
Early static feedback on the draft: I found one additional API-guideline issue in the new builder. `ClockControlBuilder::build()` is documented as panicking on invalid initial-time conversion even though M-BUILD-RESULT calls for fallible builders to return `Result`; the inline comment has the specific direction and acceptance criterion.

The highest-impact correctness and performance concerns are already covered by existing comments, so I did not duplicate them: the timer-removal/waker-replacement race at `timers.rs:92`, and the shared timer-mutex acquisition on every pending re-poll at `delay.rs:117`. The previously raised incomplete public-API migration documentation is likewise not repeated.

I traced the before/after timer state machine, wake-outside-lock change, builder migration, error/timeout shape, feature-gated exports, all changed workspace consumers, tests/doctests, and repository instructions. The explicit conversion errors, wake-after-unlock design, exact-deadline handling, and consumer migrations otherwise hold under static inspection.

All 820 classes were assessed (150 security, 162 correctness, 182 testing, 183 performance, 143 conformance). Nothing was executed, so the PR’s compilation, test, coverage, and mutation claims were not independently revalidated.

PR #637, opened 2026-08-04
feat(fetch): add socket tuning and HTTP/2 stream window options

kate-shine:
Closes a C#-to-Rust HTTP client parity gap tracked by AB#7690295.

## Parity

| Setting | Rust API |
|---|---|
| `SocketNoDelay` | `SocketOptions::no_delay` |
| `SocketReceiveBufferSizeInBytes` | `SocketOptions::receive_buffer_size` |
| `SocketSendBufferSizeInBytes` | `SocketOptions::send_buffer_size` |
| `InitialHttp2StreamWindowSizeInBytes` | `Http2Options::initial_stream_window_size` |
| `InitialCongestionWindowLevel` | Deferred |

## Changes

- Adds reusable `SocketOptions` and re-exports it from `fetch::options`.
- Adds `TokioTransportOptions` and `HttpClient::builder_tokio_with_options`; existing `builder_tokio` callers remain source-compatible.
- Applies HTTP/2 stream-window configuration in `fetch_hyper`.
- Delegates Tokio TCP dialing and address-family fallback to hyper-util's `HttpConnector`.

## Behavior

- `None` requests no override; transports and lower layers choose effective defaults.
- Socket settings are transport-scoped and applied best-effort. Rejected settings are logged without failing the connection.
- Socket buffer values are forwarded verbatim; the operating system may accept, adjust, or reject them.
- The bundled Tokio transport uses hyper-util's RFC 6555 fallback when both address families are available.
- Unsupported schemes are rejected before dialing without exposing host, port, or path data.
- DNS resolution and connection attempts share the existing `connect_timeout` budget.
- `fetch_hyper` clamps the HTTP/2 stream window to the RFC 9113 maximum.

## Deferred

`InitialCongestionWindowLevel` requires the Windows-only `SIO_TCP_SET_ICW` API and unsafe FFI, which conflicts with the repository's `unsafe_code = "deny"` policy.

## Validation

- Tests and doctests pass with all features for `fetch`, `fetch_hyper`, and `fetch_options`.
- Clippy passes with all features, all targets, and warnings denied.
- Formatting, spellcheck, generated READMEs, and mutation testing are clean.


sandersaares:
This also feeds into the `docs/stabilization.md` in `fetch` - how and whether these options are supported by different transports may vary, so these should likely be seen as transport-level options, not fetch-level options.

crates/fetch/src/tokio.rs
```diff
@@ -102,24 +189,168 @@ impl HttpClient {
...
+    /// a `flavor = "multi_thread"` runtime, or a migrated task will read `None` and report a
+    /// missing option that was in fact delivered.
+    static LAST_SOCKET_OPTIONS: std::cell::Cell<Option<SocketOptions>> = const { std::cell::Cell::new(None) };
+}
+
+impl TokioConnector {
+    /// Builds a connector that applies `options` to every connection it opens.
+    fn new(options: SocketOptions) -> Self {
+        #[cfg(test)]
+        LAST_SOCKET_OPTIONS.set(Some(options));
```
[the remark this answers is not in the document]
kate-shine: GHCP: `LocalKey<Cell<T>>` does provide specialized `set()` and `get()` methods directly; `with(|cell| ...)` is not required for this case. This code compiles and the propagation test runs across the PR CI matrix, so no change is needed here.

kate-shine:
GHCP: Addressed this by moving the new socket tuning off the shared `TransportOptions`/`HttpClientBuilder` surface. It now lives in `TokioTransportOptions` and is supplied through `HttpClient::builder_tokio_with_options`, so transports that cannot own/configure sockets cannot accept and silently ignore it. I also added section 8 to `fetch/docs/stabilization.md` documenting the remaining pre-existing transport-specific HTTP/2/pool settings on the shared surface and why moving those is a separate compatibility effort.

crates/fetch_options/src/http2.rs
```diff
@@ -33,11 +55,78 @@ impl Http2Options {
...
+    /// options.initial_stream_window_size = Some(u32::MAX);
+    ///
+    /// assert_eq!(
+    ///     options.effective_initial_stream_window_size(),
+    ///     Some(MAX_HTTP2_WINDOW_SIZE)
+    /// );
+    /// ```
+    #[must_use]
+    #[inline]
+    pub fn effective_initial_stream_window_size(&self) -> Option<u32> {
```
martintmk: I don't think this is needed, we have fields and fetch_hyper can do the clamp if necessary

crates/fetch_options/src/socket.rs
```diff
@@ -0,0 +1,323 @@
...
+    /// // A directly assigned value bypasses the builder, so it is clamped on read instead.
+    /// options.receive_buffer_size = Some(0);
+    /// assert_eq!(
+    ///     options.effective_receive_buffer_size(),
+    ///     Some(MIN_SOCKET_BUFFER_SIZE)
+    /// );
+    /// ```
+    #[must_use]
+    #[inline]
+    pub fn effective_receive_buffer_size(&self) -> Option<u32> {
```
martintmk: same as other comment, we can just drop both effective_ methods and also constants, those can be kept internal in fetch_hyper or not?

crates/fetch_options/src/socket.rs
```diff
@@ -0,0 +1,323 @@
...
+/// `fetch::tokio::TokioTransportOptions`.
+// The reference above is intentionally not an intra-doc link: `fetch` depends on this crate,
+// so linking the other way round would invert the dependency direction.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+#[non_exhaustive]
+pub struct SocketOptions {
+    /// Whether the Nagle algorithm is disabled (`TCP_NODELAY`).
+    ///
+    /// `Some(true)` sends small writes immediately at the cost of extra packets.
+    /// `None` keeps the operating system default, which enables Nagle on every platform
```
sandersaares: We should not make defaults be "operating system" defaults. If overrides are not specified, default values may come from fetch itself or anything on the lower levels. Promising operating system defaults is a big promise that we should not make.

crates/fetch_options/src/socket.rs
```diff
@@ -0,0 +1,323 @@
...
+    /// `fetch` supports and is therefore equivalent to `Some(false)` in practice.
+    pub no_delay: Option<bool>,
+    /// Size of the kernel receive buffer (`SO_RCVBUF`), in bytes.
+    ///
+    /// `None` keeps the operating system default. Values are clamped into
+    /// <code>[MIN_SOCKET_BUFFER_SIZE]..=[MAX_SOCKET_BUFFER_SIZE]</code> when set. The kernel may
```
sandersaares: Why do we need to document this clamping and expose these constants? It feels like some implementation detail. Actually, why do we need to clamp at all? This feels like some implementation detail that we do not need to expose in the API but perhaps I am missing some context.

crates/fetch_options/src/socket.rs
```diff
@@ -0,0 +1,323 @@
...
+    /// [`effective_receive_buffer_size`][Self::effective_receive_buffer_size] to get a value that
+    /// is safe to hand to the kernel.
+    pub receive_buffer_size: Option<u32>,
+    /// Size of the kernel send buffer (`SO_SNDBUF`), in bytes.
+    ///
+    /// `None` keeps the operating system default. Values are clamped into
+    /// <code>[MIN_SOCKET_BUFFER_SIZE]..=[MAX_SOCKET_BUFFER_SIZE]</code> when set. The kernel may
+    /// clamp or scale the request further, so the effective size can still differ.
+    ///
+    /// Assigning to this field directly bypasses the clamping; read it back through
```
sandersaares: Why would we implement some clamping and then enable a way to bypass it? This is very strange. Why are two ways (clamp/no-clamp) of working with the configuration required?

crates/fetch_options/src/socket.rs
```diff
@@ -0,0 +1,323 @@
...
+    /// );
+    /// ```
+    #[must_use]
+    #[inline]
+    pub fn effective_send_buffer_size(&self) -> Option<u32> {
+        self.send_buffer_size.map(clamp_buffer_size)
+    }
+}
+
+/// Clamps a requested buffer size into the accepted range.
```
sandersaares: Such unusual operations need justification to explain why such a thing happens. This raises more questions than it answers in its current form.

crates/fetch/src/options/mod.rs
```diff
@@ -7,11 +7,20 @@
 //! and re-exported here. `ClientOptions` bundles those transport options
 //! together with the response-body, routing, redaction, and TLS configuration
 //! owned by the `fetch` request pipeline.
+//!
+//! [`SocketOptions`] is the exception: it is re-exported here for convenience, but it is not
+//! part of [`TransportOptions`] and has no [`HttpClientBuilder`](crate::HttpClientBuilder)
+//! setter, because only a transport that dials its own sockets can honor it. Supply it through
+//! `fetch::tokio::TokioTransportOptions` instead.
+//!
+//! That reference is intentionally not an intra-doc link: the `tokio` module is feature-gated,
```
sandersaares: This sort of meta-documentation feels rather inappropriate to publish to our customers - they do not seem like the intended audience for it.

Also, we can easily solve the actual problem it talks about by making the link conditional:

```rust
#![cfg_attr(feature = "tokio", doc = ...)]
#![cfg_attr(not(feature = "tokio"), doc = ...)]
```

(Not that I am saying it is strictly required, just an option)

crates/fetch/src/custom.rs
```diff
@@ -116,6 +116,13 @@ pub struct CustomContext<Extras = ()> {
 /// thread-aware dependencies through to `factory` without resorting to globals.
 /// Leave it defaulted to `()` when no extras are needed.
 ///
+/// Transport-specific tuning is the transport's own concern: options that only some
+/// transports can honor (socket buffer sizes, protocol-specific windows) do not belong on
+/// [`TransportOptions`], because a transport that cannot apply them would silently ignore
+/// them. Accept them in the transport's own options type instead, the way the bundled Tokio
+/// transport does with `fetch::tokio::TokioTransportOptions` (not linked here, because that
+/// module is feature-gated).
```
sandersaares: Another instance of meta-documentation that seems inappropriate. If you want to link, make a conditional link. Or accept defeat. Both are fine.

crates/fetch_options/src/socket.rs
```diff
@@ -0,0 +1,323 @@
...
+pub const MIN_SOCKET_BUFFER_SIZE: u32 = 4 * 1024;
+
+/// Largest accepted socket buffer size, in bytes.
+///
+/// Requested sizes are clamped down to this value. 64 MiB is far above any realistic
+/// bandwidth-delay product for a service-to-service hop, so a larger request indicates a
+/// misconfiguration rather than an intent to reserve that much kernel memory per connection.
+pub const MAX_SOCKET_BUFFER_SIZE: u32 = 64 * 1024 * 1024;
+
+/// Socket-level settings applied to outbound `TCP` connections.
```
sandersaares: [Copilot speaking]

## M-01: Separate the API contract from delegated transport behavior

**Finding:** The new surface is primarily a set of requests forwarded to Hyper and hyper-util, but its public documentation repeatedly describes lower-layer behavior as guarantees owned by `fetch_options` or `fetch`: socket settings are “applied” to every connection, unset values select operating-system or protocol defaults, and delegated dialing implements a specific Happy Eyeballs RFC.

Those claims do not form a stable contract. The lower layer may reject a socket setting while continuing, may choose its own value when no override is supplied, and only races address families when the resolution result permits it. The precise behavior also depends on the pinned dependency version and platform. As a result, dependency upgrades or another conforming transport can violate the published wording even though the option was forwarded correctly.

**Recommendation:** Define these options uniformly as requested overrides: the bundled transport attempts to forward supported values, and `None` requests no override from this API. Promise an effective value or dialing algorithm only where this crate enforces and observes it. Keep current Hyper/hyper-util behavior in non-contractual implementation notes with version-specific references, and align the pull request description with that boundary.

crates/fetch_options/src/socket.rs
```diff
@@ -0,0 +1,323 @@
...
+    /// `None` keeps the operating system default, which enables Nagle on every platform
+    /// `fetch` supports and is therefore equivalent to `Some(false)` in practice.
+    pub no_delay: Option<bool>,
+    /// Size of the kernel receive buffer (`SO_RCVBUF`), in bytes.
+    ///
+    /// `None` keeps the operating system default. Values are clamped into
+    /// <code>[MIN_SOCKET_BUFFER_SIZE]..=[MAX_SOCKET_BUFFER_SIZE]</code> when set. The kernel may
+    /// clamp or scale the request further, so the effective size can still differ.
+    ///
+    /// Assigning to this field directly bypasses the clamping; read it back through
```
sandersaares: [Copilot speaking]

## M-02: Use one representation for each option value

**Finding:** Both new option families have path-dependent semantics. Their fluent setters store clamped values, while direct assignment to the public fields stores the caller's raw values. The API then adds `effective_*` accessors, requires every transport to avoid reading the fields directly, and adds another wrapper in `fetch_hyper`.

Consequently, the same requested configuration has different `Debug`, equality, and snapshot representations depending on how it was constructed, while every future consumer must remember an out-of-band rule to recover the actual value. This split representation is the common source of the duplicate clamping, public limit constants, effective-value APIs, and transport wrapper covered by the related findings.

**Recommendation:** Choose one canonical representation. With public fields, have setters store the requested value verbatim and perform any necessary validation or transformation exactly once in the transport that owns the lower-layer API. Alternatively, make the fields private and expose only a validated representation. Do not retain both raw public fields and separately normalized “effective” values.

crates/fetch/src/tokio.rs
```diff
@@ -102,24 +189,168 @@ impl HttpClient {
 /// Named pipes / Unix-domain sockets are intentionally not supported; the
 /// connector opens a TCP stream to the request authority and hands the wrapped
 /// stream to hyper. TLS, when required, is layered on top by the transport.
-#[derive(Clone)]
-struct TokioConnector;
+///
+/// The actual dialing is delegated to hyper's own [`HttpConnector`], which already
```
sandersaares: [Copilot speaking]

## F-01: Correct the Happy Eyeballs specification reference and scope it to the transport

**Finding:** `TokioConnector` delegates dialing unchanged to the pinned hyper-util 0.1.20 `HttpConnector`, whose default `set_happy_eyeballs_timeout` behavior is explicitly documented as RFC 6555, not RFC 8305. Its implementation splits a resolved address list into preferred and fallback address families and starts the fallback family after the configured delay. If the target is an IP literal or DNS returns only one family, the fallback list is empty and no address-family race occurs.

The connector documentation therefore overstates the upstream implementation as RFC 8305. RFC 8305 obsoletes RFC 6555 and specifies additional resolution, sorting, and attempt-sequencing behavior that this wrapper does not add.

Separately, the pull request description presents address racing as a property of the change rather than of one transport, stating that “every target gets address racing”. Connection establishment is transport-owned: the bundled WinHTTP transport and any custom transport dial by their own means, and no `fetch`-level contract requires them to race address families at all, let alone with the same algorithm or timing. The claim is also inaccurate even for the Tokio transport, because single-family resolutions and IP literals never race.

References: [hyper-util 0.1.20 `HttpConnector`](https://docs.rs/hyper-util/0.1.20/hyper_util/client/legacy/connect/struct.HttpConnector.html), [RFC 6555](https://www.rfc-editor.org/rfc/rfc6555.html), and [RFC 8305](https://www.rfc-editor.org/rfc/rfc8305.html).

**Recommendation:** In `TokioConnector`, describe the behavior as hyper-util's RFC 6555 IPv4/IPv6 fallback, or simply as address-family racing, and state that the alternate family is attempted only when both families are available and the preferred-family attempt has not completed before the configured delay. In the pull request description, attribute the behavior explicitly to the bundled Tokio transport rather than to `fetch`, so that no cross-transport dialing guarantee is implied.

crates/fetch/src/tokio.rs
```diff
@@ -53,15 +59,96 @@ impl TokioDeps {
...
+///         .send_buffer_size(256 * 1024),
+/// );
+/// ```
+// `Copy` is deliberately not derived: this type exists to grow, and the first knob that is
+// not a plain scalar would force its removal. `Http2Options` and `ConnectionPoolOptions` are
+// `Clone`-only for the same reason.
+#[derive(Debug, Clone, Default)]
+#[non_exhaustive]
+pub struct TokioTransportOptions {
+    /// Socket-level tuning applied to every outbound `TCP` connection.
```
sandersaares: [Copilot speaking]

## F-02: Qualify socket tuning as best-effort

**Finding:** The public field and setter promise socket tuning “applied to every outbound TCP connection,” but hyper-util 0.1.20 treats all three settings as best-effort. It logs and continues when `SO_SNDBUF` or `SO_RCVBUF` is rejected before connect, and likewise logs and continues when applying `TCP_NODELAY` to the connected stream fails. A successful request therefore does not establish that the requested tuning took effect.

Reference: [hyper-util 0.1.20 `HttpConnector` source](https://raw.githubusercontent.com/hyperium/hyper-util/v0.1.20/src/client/legacy/connect/http.rs).

**Recommendation:** Say that the transport attempts to apply the settings to every outbound connection and that lower-layer rejection is logged without failing the connection. This change only weakens the documentation guarantee; the runtime behavior remains unchanged.

crates/fetch/src/tokio.rs
```diff
@@ -102,24 +189,168 @@ impl HttpClient {
...
+///
+/// Keepalive, local-address binding, `SO_REUSEADDR` and `TCP_USER_TIMEOUT` are also exposed by
+/// [`HttpConnector`] but are deliberately left at its defaults, because [`SocketOptions`] does
+/// not model them yet.
+#[derive(Clone, Debug)]
+struct TokioConnector {
+    connector: HttpConnector,
+}
+
+#[cfg(test)]
```
sandersaares: [Copilot speaking]

## F-08: Remove test-only state from the production connector

**Finding:** `LAST_SOCKET_OPTIONS` exists only in test builds, and `TokioConnector::new` conditionally writes to it. The constructor under unit test therefore executes instructions that are absent from the constructor shipped to consumers. The associated test also depends on reset, construction, and inspection staying on one thread; the extensive thread-affinity warning is complexity created solely by this recorder.

The recorder verifies a short value-capture path rather than observable connector behavior. The connector tests already exercise the real `HttpConnector` dependency for `TCP_NODELAY`, buffer configuration, HTTPS acceptance, and connection failures.

**Recommendation:** Remove `LAST_SOCKET_OPTIONS`, the `#[cfg(test)]` write in `TokioConnector::new`, and the recorder-based propagation test. Retain the tests that observe the real connector. Do not add a production observer or factory abstraction solely to preserve this implementation-level assertion.

crates/fetch_hyper/src/builder.rs
```diff
@@ -247,11 +247,21 @@ fn apply_pool_options(hyper_builder: &mut legacy::Builder, pool: &ConnectionPool
 
 #[cfg_attr(test, mutants::skip)] // cannot be verified with hyper APIs
 fn apply_http2_options(hyper_builder: &mut legacy::Builder, http_2: &Http2Options) {
+    // Order matters: hyper's `http2_adaptive_window(true)` resets the stream window to the
+    // spec default, so the explicit size has to be applied first for adaptive tuning to win
+    // when both are configured. That precedence is documented on `initial_stream_window_size`.
     hyper_builder
         .http2_initial_max_send_streams(http_2.initial_max_send_streams)
+        .http2_initial_stream_window_size(initial_stream_window_size(http_2))
```
sandersaares, line 255: [Copilot speaking]

## F-15: Remove the duplicate Hyper stream-window wrapper

**Finding:** `initial_stream_window_size` is currently a one-line pass-through to `Http2Options::effective_initial_stream_window_size`; it does not perform transport-specific validation or transformation. Its test only retests the public accessor and remains green if `apply_http2_options` stops calling Hyper's setter. The same-crate `#[inline]` hint is also unsupported by benchmark or disassembly evidence, contrary to the workspace performance guidance.

**Recommendation:** Choose one clamping owner. With the current public effective accessor, call it directly from `apply_http2_options` and remove the private wrapper, its duplicate test, and the `#[inline]` hint. If the public accessor is removed and clamping moves into `fetch_hyper`, keep a private helper only when it performs that clamp itself.

sandersaares (commented):
[Copilot speaking]

Automated review of the socket tuning and HTTP/2 stream window changes.

| Category | Findings |
|---|---|
| API contract accuracy | 4 |
| Design and layering | 1 |
| Simplification | 2 |
| Test design | 1 |
| **Total** | **8** |

Three findings continue existing review threads; two of those were posted as replies in the corresponding threads rather than as new comments.

| Item | Value |
|---|---|
| Cache | Miss |

</details>

crates/fetch_options/src/socket.rs
sandersaares: [Copilot speaking]

## F-03: Describe unset values as lower-layer defaults

**Finding:** The new documentation promises defaults owned by layers that this API does not control.

For `Http2Options::initial_stream_window_size`, `None` causes `apply_http2_options` to pass `None` to Hyper, whose setter then does nothing. Hyper 1.10.1 currently chooses a 2 MiB stream window when adaptive flow control is disabled, not RFC 9113's 65,535-byte initial value, and Hyper documents its option defaults as unstable. Calling this the “protocol default” is therefore incorrect.

For socket settings, unset buffer sizes are left unspecified for hyper-util, but `no_delay: None` is converted to `false` and hyper-util applies `TCP_NODELAY = false` to the connected socket. The bundled connector's current behavior happens to match common operating-system defaults, but `SocketOptions` cannot promise that every transport or future lower layer will preserve those defaults.

References: [Hyper 1.10.1 HTTP/2 configuration defaults](https://raw.githubusercontent.com/hyperium/hyper/v1.10.1/src/proto/h2/client.rs) and [hyper-util 0.1.20 `HttpConnector` source](https://raw.githubusercontent.com/hyperium/hyper-util/v0.1.20/src/client/legacy/connect/http.rs).

**Recommendation:** Define `None` as “no override requested by this option.” For the HTTP/2 field, say that the transport chooses its default. For socket fields, say that the lower layer chooses the effective behavior; if useful, document separately that the bundled connector currently uses `TCP_NODELAY = false` and leaves buffer sizes unset. Apply the same wording to the tests and pull request description.

crates/fetch_options/src/socket.rs
sandersaares: [Copilot speaking]

## F-12: Remove unsupported socket-buffer clamping policy

**Finding:** The new 4 KiB lower bound and 64 MiB upper bound silently rewrite caller requests, but neither boundary follows from the cited platform behavior. Windows' `SO_SNDBUF = 0` semantics explain what zero does; they do not justify replacing every smaller positive send or receive request with 4 KiB. Likewise, bandwidth-delay product has no universal 64 MiB ceiling, so the upper bound can reduce a deliberate request on high-bandwidth or high-latency links.

Hyper-util already hands requested values to the socket API and logs while continuing if the operating system rejects them. Adding cross-platform policy here therefore creates public constants, effective-value accessors, duplicate clamping at setter and consumption time, and tests for limits that the operating system remains free to adjust again.

References: [Windows `SOL_SOCKET` options](https://learn.microsoft.com/en-us/windows/win32/winsock/sol-socket-socket-options) and [Linux `socket(7)`](https://man7.org/linux/man-pages/man7/socket.7.html).

**Recommendation:** Store buffer sizes verbatim and pass the public fields directly to hyper-util after the `u32`-to-`usize` conversion. Remove the public min/max constants, `clamp_buffer_size`, the `effective_*` accessors, and their clamping examples and tests. `None` and values inside the former range behave as before; values outside it reach the platform, which may accept, adjust, or reject them. If zero must be forbidden by contract, reject it explicitly rather than silently substituting a different request.

crates/fetch_options/src/http2.rs
kate-shine: GHCP: Agreed. Addressed in `e909ffdb`: `Http2Options` now preserves the raw field/setter value, the public `effective_initial_stream_window_size` API and public maximum constant are removed, and `fetch_hyper` privately clamps to the RFC 9113 limit immediately before configuring hyper.

crates/fetch_options/src/socket.rs
kate-shine: GHCP: Agreed. Addressed in `e909ffdb`: both socket `effective_*` methods and all public clamp constants are gone. `SocketOptions` is now only a raw reusable configuration bag; the bundled Tokio connector owns its private 4 KiB..64 MiB clamp, while `fetch_hyper` owns the private HTTP/2 protocol clamp. Custom transports can apply their own appropriate limits.

crates/fetch_hyper/src/builder.rs
```diff
@@ -247,11 +247,24 @@ fn apply_pool_options(hyper_builder: &mut legacy::Builder, pool: &ConnectionPool
...
         .http2_adaptive_window(http_2.adaptive_window);
 }
 
+/// Largest legal HTTP/2 stream flow-control window, in bytes.
+const MAX_HTTP2_STREAM_WINDOW_SIZE: u32 = i32::MAX as u32;
+
+/// Returns the protocol-valid stream window forwarded to hyper.
+#[inline]
+fn initial_stream_window_size(http_2: &Http2Options) -> Option<u32> {
+    http_2.initial_stream_window_size.map(|size| size.min(MAX_HTTP2_STREAM_WINDOW_SIZE))
```
Vaiz: 🤖 **Clawpilot here!** Automated review comment from an AI agent -- not a human reviewer. Please verify before acting.
AI reviewers: Claude Opus 5, GPT-5.6 Sol, Gemini 3.1 Pro Preview

Two comments in this block spend their line restating the code, and both drop something a reader genuinely cannot recover.

`MAX_HTTP2_STREAM_WINDOW_SIZE` (line 260) is documented as "Largest legal HTTP/2 stream flow-control window, in bytes," which restates the name. The rationale that used to live on `fetch_options::MAX_HTTP2_WINDOW_SIZE` -- that RFC 9113 caps `SETTINGS_INITIAL_WINDOW_SIZE` at `2^31 - 1`, and that exceeding it is a protocol error surfacing as a connection failure far from the misconfiguration -- was lost when the constant moved here in `e909ffdb`. The equivalent rationale was kept on the socket constants in `fetch/src/tokio.rs`, so this looks like an oversight in the move rather than a decision.

`initial_stream_window_size` (line 262) has a similar problem, and it now matters more than it did before that refactor. The doc says only "Returns the protocol-valid stream window forwarded to hyper," while the thing a maintainer cannot see is why the function exists separately at all: `apply_http2_options` is `#[cfg_attr(test, mutants::skip)]`, so the clamp on line 265 would never be mutation-tested if it were written inline there. Hoisting it here is what lets `http2_stream_window_is_clamped_at_the_transport_boundary` kill that mutant. Since this helper now carries the clamp itself rather than delegating to an `effective_*` accessor, inlining it would remove the only mutation-tested copy of the HTTP/2 clamp -- and all tests would still pass.

This is also the branch that #discussion_r3733659857 (F-15) explicitly left open: it asked for the wrapper to be removed *while* it was a pass-through, but added "if the public accessor is removed and clamping moves into `fetch_hyper`, keep a private helper only when it performs that clamp itself." That is exactly what happened, so the helper now earns its place -- it just does not say why.

Suggestion: replace both restatements with the missing rationale, for example:

```rust
/// Largest legal HTTP/2 stream flow-control window, in bytes.
///
/// RFC 9113 caps `SETTINGS_INITIAL_WINDOW_SIZE` at `2^31 - 1`; a larger value is a
/// protocol error that would surface as a connection failure far from the
/// misconfiguration, so requests are clamped to this value instead.
const MAX_HTTP2_STREAM_WINDOW_SIZE: u32 = i32::MAX as u32;

// Kept out of `apply_http2_options` on purpose: that function is `mutants::skip`ped
// because its effect cannot be observed through hyper's APIs, so this clamp would
// never be mutation-tested if it were written inline there. Keeping it here lets
// `http2_stream_window_is_clamped_at_the_transport_boundary` cover it. Do not inline.
#[inline]
fn initial_stream_window_size(http_2: &Http2Options) -> Option<u32> {
    http_2.initial_stream_window_size.map(|size| size.min(MAX_HTTP2_STREAM_WINDOW_SIZE))
}
```

Non-blocking.

crates/fetch_options/src/socket.rs
```diff
@@ -0,0 +1,185 @@
...
+    /// ```
+    #[must_use]
+    pub fn send_buffer_size(mut self, size: impl Into<Option<u32>>) -> Self {
+        self.send_buffer_size = size.into();
+        self
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
Vaiz, line 114: 🤖 **Clawpilot here!** Automated review comment from an AI agent -- not a human reviewer. Please verify before acting.
cargo-evaluate guideline: [M-INTEGRATION-TESTS](https://microsoft.github.io/rust-guidelines/guidelines/libs/resilience/#M-INTEGRATION-TESTS)

Reason: "Rust source should not carry an inline `#[cfg(test)] mod tests {}` module for public-API tests."

Recommendation: "Move tests that exercise only the public API into a file under `tests/`; keep `src/` focused on implementation. See M-INTEGRATION-TESTS."

Context from verification, in fairness to this PR: every one of the 8 modules in `fetch_options` uses an inline `#[cfg(test)] mod tests`, and the crate has no `tests/` directory at all, so this new file follows the only convention the crate has. Flagging it purely so the decision is deliberate -- if the convention is worth changing, that seems like a separate crate-wide change rather than something this PR should carry.

crates/fetch_options/src/http2.rs
kate-shine: GHCP: Addressed. The public effective accessor and limit constant were removed; `fetch_hyper` now owns the protocol clamp privately.

crates/fetch_options/src/socket.rs
kate-shine: GHCP: Addressed. The socket effective accessors and public clamping constants were removed. `SocketOptions` now preserves the raw requested values.

crates/fetch_options/src/socket.rs
kate-shine: GHCP: Addressed in 9ea7eb64. `None` is now documented uniformly as requesting no override; the transport and lower layers choose the effective behavior.

crates/fetch_options/src/socket.rs
kate-shine: GHCP: Agreed. The public constants and clamping documentation were removed, and the Tokio connector no longer imposes an arbitrary socket-buffer range.

crates/fetch_options/src/socket.rs
kate-shine: GHCP: Addressed by using one representation. Public fields and fluent setters both preserve the caller's raw value, with protocol validation performed once by the consuming transport.

crates/fetch_options/src/socket.rs
kate-shine: GHCP: Agreed; the unusual effective-value path and its associated public policy surface were removed.

crates/fetch/src/options/mod.rs
kate-shine: GHCP: Addressed in 9ea7eb64. The customer-facing meta-documentation was removed; the plain code reference remains without commentary about rustdoc mechanics.

crates/fetch/src/custom.rs
kate-shine: GHCP: Addressed in 9ea7eb64. The feature-gating meta note was removed and the documentation keeps only the relevant API guidance.

crates/fetch_options/src/socket.rs
kate-shine: GHCP: Addressed. The public contract now defines these values as requested overrides, says `None` requests no override, and qualifies Tokio socket application as best-effort. The PR description was updated to the same boundary.

crates/fetch_options/src/socket.rs
kate-shine: GHCP: Addressed. Setters and public fields now store identical raw values. Socket values are forwarded by Tokio without policy clamping, while `fetch_hyper` privately enforces only the HTTP/2 protocol maximum.

crates/fetch/src/tokio.rs
kate-shine: GHCP: Addressed in 9ea7eb64. The connector documentation and PR now describe hyper-util's RFC 6555 address-family fallback, scoped to the bundled Tokio transport, including delayed fallback, immediate fallback after preferred-family failure, and the dual-family condition.

crates/fetch/src/tokio.rs
kate-shine: GHCP: Addressed. The public field and setter now describe requested tuning, and the connector documentation states that lower-layer rejection is logged without failing the connection.

crates/fetch/src/tokio.rs
kate-shine: GHCP: Addressed in 9ea7eb64. `LAST_SOCKET_OPTIONS`, the conditional constructor write, the recorder-based test, and the factory extraction used solely by that test were removed. Observable connector tests remain.

crates/fetch_hyper/src/builder.rs
kate-shine: GHCP: Addressed under the alternative described here. The public effective accessor is gone; the private helper now performs the RFC 9113 clamp itself. The unsupported `#[inline]` hint was removed, and the test now covers `None`, pass-through below the maximum, the boundary, and clamping above it.

crates/fetch_hyper/src/builder.rs
kate-shine: GHCP: Addressed in aaf2a293. The constant now documents the RFC 9113 limit and distant failure mode, and the helper documents why it remains separate for mutation coverage. I intentionally did not restore `#[inline]`, since there is no benchmark or disassembly evidence for it.

crates/fetch_options/src/socket.rs
kate-shine: GHCP: Agreed with the context here. The tests follow the existing `fetch_options` crate convention; moving all public-API tests to integration tests should be a separate crate-wide change rather than an isolated refactor in this PR.

Vaiz (approved):

martintmk (approved):

Vaiz (approved):

PR #639, opened 2026-08-05
feat(msvc_spectre_libs): import crate

psandana:
## Summary

- copy the five `msvc_spectre_libs` 0.1.3 crate files byte-for-byte from `microsoft/rust_msvc_spectre_libs`
- explicitly exclude the verbatim legacy import from Oxidizer workspace policy checks
- make Cargo manifest sorting operate on actual workspace members so excluded imports are respected

The imported crate files are unchanged from version 0.1.3.


crates/msvc_spectre_libs/build.rs
```diff
@@ -0,0 +1,48 @@
...
+#[cfg(all(target_os = "windows", target_env = "msvc"))]
+fn add_spectre_link_search() {
+    use std::env;
+
+    use cc::windows_registry;
+
+    let target = env::var("TARGET").expect("Cargo must set TARGET when executing a package build script");
+    let arch = env::var("CARGO_CFG_TARGET_ARCH").expect("Cargo must set CARGO_CFG_TARGET_ARCH when executing a package build script");
+    let arch = architecture::spectre_directory(&arch)
+        .expect("Windows MSVC targets must use an architecture supported by Visual Studio Spectre libraries");
```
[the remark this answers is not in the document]
psandana: We are importing as is today. We are working on top of this later

crates/msvc_spectre_libs/build.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    let tool =
+        windows_registry::find_tool(&target, "cl.exe").expect("the Windows MSVC target requires cl.exe from a Visual Studio installation");
+    let spectre_libs = tool.path().join(format!(r"..\..\..\..\lib\spectre\{arch}"));
+
```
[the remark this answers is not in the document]
psandana: We are importing as is today. We are working on top of this later

wukchung (commented):
&#x1F916; Automated deep review of this import. Six independent reviewer models were fanned out over the diff; every finding below was then re-verified by hand against the code in a local worktree, and anything that could not be proven was dropped. 5 of 6 reviewer slots returned (the GPT-5.6 slot timed out and is excluded). 20 raw findings in, 5 verified findings out.

All five are correctness, security, or licensing issues; style and formatting were out of scope. Three of them were reproduced end-to-end on a Windows box with Visual Studio 18 installed but **without** the Spectre-mitigated libraries, which is exactly the configuration the crate is meant to handle gracefully.

The common thread: the crate documents "warn and continue by default, fail only under the `error` feature", but in practice the default path can hard-fail, the `error` path can silently not fail, and a successful build can silently omit the mitigation entirely. For a crate whose only purpose is a security mitigation, the silent-no-op cases are the ones worth fixing before release.

Checked and deliberately **not** raised: the `assert!` polarity is correct (not inverted); `#[cfg(feature = "error")]` does work in a build script; the four `.parent()` calls correctly resolve the MSVC version directory; `cargo package` ships both `build.rs` and `src/architecture.rs`; and the `arm` -> `arm32` mapping could not be disproven, so it is left alone (it matches upstream). Generated READMEs and changelogs were treated as out of scope per `AGENTS.md`.

No vote has been cast - that is left to a human reviewer.

crates/msvc_spectre_libs/build.rs
```diff
@@ -0,0 +1,55 @@
...
+        return;
+    }
+
+    let target = env::var("TARGET").expect("Cargo must set TARGET when executing a package build script");
+    let arch = env::var("CARGO_CFG_TARGET_ARCH").expect("Cargo must set CARGO_CFG_TARGET_ARCH when executing a package build script");
+    let arch = architecture::spectre_directory(&arch)
+        .expect("Windows MSVC targets must use an architecture supported by Visual Studio Spectre libraries");
+
+    let tool =
+        windows_registry::find_tool(&target, "cl.exe").expect("the Windows MSVC target requires cl.exe from a Visual Studio installation");
```
wukchung: &#x1F916; **Missing `cl.exe` hard-fails the build, even with default features**

`find_tool(...).expect(...)` panics, and a panic in a build script is a hard build failure. This contradicts the crate's own documented default in `src/lib.rs`:

> By default, the build script emits a warning and continues when it cannot find the libraries.

Neither the warning branch nor the `MSVC_SPECTRE_LIBS_ALLOW_MISSING` escape hatch is reachable, because both live below this line.

Reproduced on a machine with VS 18 (x64/x86 toolsets present, ARM64 cross tools absent), using **default features** with `error` NOT enabled:

```
cargo build --target aarch64-pc-windows-msvc
...
thread 'main' panicked at build.rs:36:56:
the Windows MSVC target requires cl.exe from a Visual Studio installation
error: failed to run custom build command for `msvc_spectre_libs v0.1.3`
```

`cc::windows_registry::find_tool("aarch64-pc-windows-msvc", "cl.exe")` returns `None` on that box, while `x86_64` and `i686` both resolve. So taking this crate as a dependency breaks any consumer building for a windows-msvc target whose matching toolset component is not installed - a hard failure where the docs promise a warning.

The expect message is also inaccurate: Rust's `*-pc-windows-msvc` targets require `link.exe`/`lld-link`, not `cl.exe`; `cl.exe` is used here only to locate the toolchain directory. `AGENTS.md` requires `.expect()` to cover a programming error or a situation that can never happen, and a missing VS component is neither.

Suggested fix: treat `None` as the same "not found" case as a missing directory - warn, and honour `error` - instead of panicking.

crates/msvc_spectre_libs/build.rs
```diff
@@ -0,0 +1,55 @@
...
+    let spectre_libs = architecture::spectre_libs_path(tool.path(), arch)
+        .expect("the cl.exe path must be inside a Visual Studio MSVC toolchain directory");
+
+    if spectre_libs.exists() {
```
wukchung, line 28: &#x1F916; **No `rerun-if-*` directives, so a stale cached result can leave the mitigation silently off**

The script emits no `cargo:rerun-if-env-changed` and no `cargo:rerun-if-changed`, so Cargo falls back to its default fingerprint, which only watches files inside the package - and those never change. Two consequences:

1. A developer who sees the warning, installs the Spectre component through the VS Installer, and rebuilds gets the **cached** "warning only" output. No `rustc-link-search` is emitted, so binaries keep linking the non-mitigated CRT while the developer believes the mitigation is now active. Only `cargo clean` clears it.
2. `MSVC_SPECTRE_LIBS_ALLOW_MISSING` is untracked, so toggling it has no effect until a clean rebuild - including toggling it *off* to re-enable enforcement.

Reproduced with a consumer using `features = ["error"]` on a box with no Spectre libs installed. With the variable set the build succeeds; removing it from the environment and rebuilding into the same target directory:

```
cargo build
warning: msvc_spectre_libs@0.1.3: No Spectre-mitigated libraries were found...
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.05s
ExitCode=0
```

The build script never re-ran. With a clean cache the identical command panics at `build.rs:49` as intended, so the difference is purely stale caching rather than any feature/env difference.

Suggested fix - emit these unconditionally, before the `exists()` check:

```rust
println!("cargo:rerun-if-env-changed=MSVC_SPECTRE_LIBS_ALLOW_MISSING");
println!("cargo:rerun-if-changed={}", spectre_libs.display());
```

Cargo re-runs the script when a path named by `rerun-if-changed` is later created, which covers the install-after-warning case.

crates/msvc_spectre_libs/build.rs
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Locates the Spectre-mitigated MSVC libraries and adds them to the linker search path.
+
+#[cfg(all(target_os = "windows", target_env = "msvc"))]
```
wukchung: &#x1F916; **This `cfg` gates on the HOST, so cross-compiling to windows-msvc silently skips the mitigation**

Build scripts are compiled for the host, so `target_os` / `target_env` here describe the machine running the build, not the target being built. Verified directly with a probe build script compiled on this windows-msvc host while targeting Linux:

```
cfg!(windows+msvc) in build.rs = true
CARGO_CFG_TARGET_OS  = Ok("linux")
CARGO_CFG_TARGET_ENV = Ok("gnu")
TARGET               = Ok("x86_64-unknown-linux-gnu")
```

The gate is therefore inverted relative to intent. When cross-compiling **to** `*-pc-windows-msvc` from a Linux/macOS host (cargo-xwin, containerised CI) or from a `windows-gnu` host, `main()` is an empty function: no link-search path is emitted, **no `cargo:warning` is printed**, and the `error` feature cannot fire either, because the `assert!` also lives inside this host-gated function. The build succeeds while producing un-mitigated CRT linkage, with nothing in the output to indicate it.

The runtime `is_windows_msvc_target` check on lines 24-28 only covers the opposite direction (windows-msvc host building for a non-windows target).

Worth noting alongside this: `src/lib.rs` states the crate "does nothing for non-Windows or non-MSVC **targets**", but it also does nothing for windows-msvc *targets* built from a non-MSVC *host*, which a reader will not infer from that sentence.

Suggested fix: drive the gating entirely off `CARGO_CFG_TARGET_*` (making `cc` an unconditional build-dependency), letting the runtime check that already exists do the work.

crates/msvc_spectre_libs/src/lib.rs
```diff
@@ -0,0 +1,21 @@
...
+//! [Spectre-mitigated libraries] installed by Visual Studio. The crate has no
+//! runtime API and does nothing for non-Windows or non-MSVC targets.
+//!
+//! Visual Studio Installer offers the libraries as optional components such as
+//! `MSVC v143 - VS 2022 C++ x64/x86 Spectre-mitigated libs (Latest)`.
+//!
+//! # Missing libraries
+//!
+//! By default, the build script emits a warning and continues when it cannot
+//! find the libraries. Enable the `error` feature to fail the build instead.
```
wukchung: &#x1F916; **The `error` feature has an undocumented environment kill-switch**

`build.rs` gates the `error` failure on an environment variable:

```rust
#[cfg(feature = "error")]
assert!(env::var_os("MSVC_SPECTRE_LIBS_ALLOW_MISSING").is_some(), ...);
```

Upstream `0.1.3` panics unconditionally here, so this is a behavioural change rather than the preserved behaviour the PR description describes. Verified both directions on a box without the Spectre libs: with the variable unset the build fails as documented; with `=1` it succeeds with only a warning.

The variable is documented nowhere - not in these crate docs, not in the generated README, and not in the panic message itself. So someone who hits the failure has no way to discover the knob, and someone whose environment already has it set has no way to discover that their `error` feature is inert. Since `.cargo/config.toml` sets it for every build in this workspace, `--features error` never actually fails here.

Two related details on the same check:
- `is_some()` makes this presence-based, so `MSVC_SPECTRE_LIBS_ALLOW_MISSING=0` (or an empty value) also disables enforcement, which is likely the opposite of what someone setting `0` intends.
- Cargo hides build-script warnings for registry dependencies, so for published consumers the default path is invisible as well as non-failing.

Suggested fix: document the variable and its accepted values here (the README is generated from these docs), and name it in the panic message so the failure is self-explanatory.

crates/msvc_spectre_libs/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+version = "0.1.3"
+description = "Links Spectre-mitigated libraries for Windows MSVC targets"
+readme = "README.md"
+keywords = ["spectre", "msvc", "mitigated", "crt", "windows"]
+categories = ["development-tools::build-utils"]
+
+edition.workspace = true
+rust-version.workspace = true
+authors.workspace = true
+license.workspace = true
```
wukchung: &#x1F916; **This narrows the published license from `MIT OR Apache-2.0` to `MIT`**

`msvc_spectre_libs` already exists on crates.io under the name this crate would publish to, and every published release so far is dual-licensed. From the crates.io API:

| version | license | published |
| --- | --- | --- |
| 0.1.3 | MIT OR Apache-2.0 | 2025-03-03 |
| 0.1.2 | MIT OR Apache-2.0 | 2024-08-30 |
| 0.1.1 | MIT OR Apache-2.0 | 2023-07-20 |
| 0.1.0 | MIT OR Apache-2.0 | 2022-05-23 |

`license.workspace = true` resolves to `MIT`, so the next release published under this name would drop the Apache-2.0 option that existing downstream consumers currently rely on. That is a licensing change to an established crates.io identity rather than a metadata tidy-up, so it seems worth an explicit decision - along with crate-ownership coordination with `microsoft/rust_msvc_spectre_libs`, whose repository is still the one crates.io points at.

Separately, `version = "0.1.3"` on line 6 equals the already-published version. The release tooling bumps from the on-disk version so this should not block publishing, but it is worth a sanity check that the intended first Oxidizer release lands above the upstream line.

crates/msvc_spectre_libs/build.rs
psandana: This PR intentionally imports the published 0.1.3 crate byte-for-byte, without behavioral changes. This is a valid upstream issue, but changing build.rs belongs in a follow-up after ownership transfer rather than in the verbatim import.

Vaiz (approved):

sandersaares (changes requested):
I do not see any justification from excluding an arbitrary crate from the Cargo workspace and related automation/logic. If we want it in this workspace, it needs to fully participate in this workspace. Otherwise, it belongs in a different repo/workspace. If you feel there is some justification for this, I am happy to hear but without very strong justification this does not seem like an appropriate change to merge in its current form.

psandana:
> I do not see any justification from excluding an arbitrary crate from the Cargo workspace and related automation/logic. If we want it in this workspace, it needs to fully participate in this workspace. Otherwise, it belongs in a different repo/workspace. If you feel there is some justification for this, I am happy to hear but without very strong justification this does not seem like an appropriate change to merge in its current form.

I do agree. I wanted to port current released version, so we have a delta on this repo, but seems that is not feasible due our opinionated stack. I marked this PR as draft though, so not sure what happened that is not in that state anymore :D

PR #640, opened 2026-08-06
Test PR for automation testing

sandersaares:


sandersaares (commented):
[Copilot speaking] probe summary: verifying that the review body carries the summary.

sandersaares (commented):
[Copilot speaking] probe summary 2

sandersaares (commented):
[Copilot speaking]

Submitted 4 findings. One finding follows up on an existing discussion thread.

| Diagnostic | Value |
|------------|-------|
| Cache | Miss |
| Instruction expansion | Degraded |

</details>

sandersaares (commented):
[Copilot speaking]

Submitted 4 findings. One finding follows up on an existing discussion thread.

| Diagnostic | Value |
|------------|-------|
| Cache | Miss |
| Instruction expansion | Degraded |

</details>

crates/opentelemetry_rescaled/src/rescale.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+impl Rescale for f64 {
+    fn rescale(self, factor: f64) -> Self {
+        self * factor
+    }
+}
+
+impl Rescale for u64 {
+    #[expect(
```
sandersaares, line 24: [Copilot speaking]

# Validate the rescaling factor before applying it

The rescaling factor is applied to every recorded measurement without being checked against the range the exporter accepts.

**Reproducible reasoning**: The factor is read from configuration, which is outside the crate, while the accepted range is a property of the exporter, so nothing connects the two.

**Consequence**: A factor of zero silently reports every measurement as zero, which reads as a healthy service rather than as a misconfiguration.

**Recommended action**: Reject a factor outside the accepted range when the configuration is read, so the failure names the setting rather than appearing later as implausible telemetry.

**References**:

- [OpenTelemetry metrics specification](https://opentelemetry.io/docs/specs/otel/metrics/)

**Impacted locations**:

- `crates/opentelemetry_rescaled/src/rescale.rs:20-24`

crates/opentelemetry_rescaled/src/config.rs
sandersaares, line 1: [Copilot speaking]

# Document the configuration surface this file defines

The file defines the configuration a caller supplies but does not state which settings are required or what happens when one is omitted. Non-ASCII fidelity is exercised here: ✓ é 日本語.

**Reproducible reasoning**: A caller reading only this file cannot tell a required setting from an optional one, and the defaults live in a different file.

**Consequence**: Callers discover a missing required setting at run time rather than when they write the configuration.

**Recommended action**: State, for each setting, whether it is required and what an omitted value falls back to.

**Impacted locations**:

- `crates/opentelemetry_rescaled/src/config.rs`

crates/opentelemetry_rescaled/src/lib.rs
sandersaares, line 1: [Copilot speaking]

# Name the boundary this file is responsible for

The file changed by this pull request does not state which concern it owns, so a reader cannot tell what belongs in it.

**Reproducible reasoning**: The anchor names a line far beyond any hunk the pull request touches, which exercises the downgrade to a file-level comment.

**Consequence**: Without a stated boundary, unrelated concerns accumulate in the file over time.

**Recommended action**: State the concern the file owns at the top of the file.

**Impacted locations**:

- `crates/opentelemetry_rescaled/src/lib.rs`

sandersaares (commented):
[Copilot speaking]

Published 3 findings. No finding follows up on an existing discussion thread.

| Diagnostic | Value |
|------------|-------|
| Cache | Miss |
| Instruction expansion | Degraded |

</details>

crates/opentelemetry_rescaled/src/rescale.rs
sandersaares, line 24: [Copilot speaking]

# Validate the rescaling factor before applying it

The rescaling factor is applied to every recorded measurement without being checked against the range the exporter accepts.

**Reproducible reasoning**: The factor is read from configuration, which is outside the crate, while the accepted range is a property of the exporter, so nothing connects the two.

**Consequence**: A factor of zero silently reports every measurement as zero, which reads as a healthy service rather than as a misconfiguration.

**Recommended action**: Reject a factor outside the accepted range when the configuration is read, so the failure names the setting rather than appearing later as implausible telemetry.

**References**:

- [OpenTelemetry metrics specification](https://opentelemetry.io/docs/specs/otel/metrics/)

**Impacted locations**:

- `crates/opentelemetry_rescaled/src/rescale.rs:20-24`

crates/opentelemetry_rescaled/src/config.rs
sandersaares, line 1: [Copilot speaking]

# Document the configuration surface this file defines

The file defines the configuration a caller supplies but does not state which settings are required or what happens when one is omitted. Non-ASCII fidelity is exercised here: ✓ é 日本語.

**Reproducible reasoning**: A caller reading only this file cannot tell a required setting from an optional one, and the defaults live in a different file.

**Consequence**: Callers discover a missing required setting at run time rather than when they write the configuration.

**Recommended action**: State, for each setting, whether it is required and what an omitted value falls back to.

**Impacted locations**:

- `crates/opentelemetry_rescaled/src/config.rs`

crates/opentelemetry_rescaled/src/lib.rs
sandersaares, line 1: [Copilot speaking]

# Name the boundary this file is responsible for

The file changed by this pull request does not state which concern it owns, so a reader cannot tell what belongs in it.

**Reproducible reasoning**: The anchor names a line far beyond any hunk the pull request touches, which exercises the downgrade to a file-level comment.

**Consequence**: Without a stated boundary, unrelated concerns accumulate in the file over time.

**Recommended action**: State the concern the file owns at the top of the file.

**Impacted locations**:

- `crates/opentelemetry_rescaled/src/lib.rs`

crates/opentelemetry_rescaled/src/rescale.rs
sandersaares: [Copilot speaking]

# Follow up on the earlier observation about this code

The earlier discussion on this thread raised a concern that the current revision does not yet address.

**Reproducible reasoning**: The thread names a specific behaviour, and the reviewed revision still exhibits it.

**Consequence**: The concern raised in the earlier round remains unresolved while the thread reads as answered.

**Recommended action**: Address the behaviour the thread names, or state why the current form is preferred.

**Impacted locations**:

- `crates/opentelemetry_rescaled/src/config.rs`

sandersaares (commented):
[Copilot speaking]

Published 4 findings. One finding follows up on an existing discussion thread.

| Diagnostic | Value |
|------------|-------|
| Cache | Miss |
| Instruction expansion | Degraded |

</details>

crates/opentelemetry_rescaled/src/instruments.rs
```diff
@@ -0,0 +1,304 @@
...
+                for rule in rules {
+                    let callbacks = Arc::clone(&callbacks);
+                    let factor = rule.factor;
+                    let mut inner = self.inner.$method(rule.target_name.clone());
+                    if let Some(description) = description.clone() {
+                        inner = inner.with_description(description);
+                    }
+                    inner = inner.with_unit(rule.target_unit.clone());
+                    let _instrument = inner
+                        .with_callback(move |observer| {
```
geeknoid: **Observable sidecars can sample a different value from the source** — *Correctness · Medium · High confidence*

Each registration invokes the user's callback independently: once for the original and again for every sidecar. The callback contract permits captured state and live reads, so this fails deterministically for a callback that increments or consumes a value on each invocation: the source observes `v`, while the sidecar observes the later value and exports that value times the factor. The current test at `tests/integration.rs:455-476` confirms two invocations but uses a constant observation, so it cannot detect the broken source-to-sidecar relationship. This contradicts the crate's core promise that sidecars carry the same measurements rescaled.

**Direction:** arrange for one callback evaluation per collection and fan each resulting observation out to the original plus all sidecar observers; if the public OpenTelemetry API cannot support that invariant, narrow the advertised contract rather than assuming callbacks are idempotent.

**Done when:** a stateful-callback test either proves one sample is shared by the source and all sidecars, or the API/design explicitly drops that guarantee and documents the repeated-callback semantics and constraints.

crates/opentelemetry_rescaled/src/instruments.rs
```diff
@@ -0,0 +1,304 @@
...
+            let Some(rules) = self.rules.rules_for(&name) else {
+                return original;
+            };
+
+            let sidecars = rules
+                .iter()
+                .map(|rule| {
+                    let scaled_boundaries = boundaries
+                        .as_ref()
+                        .map(|bounds| bounds.iter().map(|bound| bound * rule.factor).collect());
```
geeknoid: **Scaling can turn valid histogram boundaries into a silently disabled sidecar** — *Correctness · Medium · High confidence*

Valid finite, increasing source boundaries are multiplied without validating the result. A finite positive factor can still overflow a large boundary to `+∞`, or underflow adjacent small boundaries to the same value. OpenTelemetry rejects non-finite or non-increasing explicit boundaries and builds a no-op histogram, so the original continues exporting while the configured sidecar silently drops every measurement. Factor validation alone cannot prevent this because validity depends on both the factor and the instrument's boundaries.

**Direction:** validate the complete scaled boundary vector before constructing the sidecar and surface an explicit creation/configuration failure consistent with the documented fail-fast policy.

**Done when:** overflow and underflow/collapse cases fail explicitly instead of producing a no-op sidecar, while valid scaled boundaries remain finite and strictly increasing.

geeknoid (commented):
Early feedback on this draft found two correctness defects in the new rescaling layer. Observable callbacks are re-run independently for the source and each sidecar, so a stateful or destructive callback can produce sidecars that are not rescaled copies of the source sample. Explicit histogram boundaries can also become non-finite or duplicate after scaling; the SDK then substitutes a no-op sidecar, silently dropping that metric.

The security read found no exploitable path: the change is an in-process metrics library with no authentication, authorization, secret, filesystem, network, web, FFI, or unsafe boundary. Configuration rejects non-finite/non-positive factors, equal source/target names, and duplicate targets. The synchronous fan-out, metadata propagation, scope pass-through, and broad instrument-kind integration coverage otherwise hold under static inspection.

All five static passes were completed: security (150 classes), correctness (162), testing (182), performance (183), and conformance (143). The generated README/logo content was excluded as instructed, and the existing factor/documentation comments were not repeated. The configured hot path has no crate-specific benchmark or regression gate; I kept that lower-priority measurement gap out of the inline budget.

Nothing was executed: no tests, builds, benchmarks, profilers, Miri, coverage, mutation testing, fuzzers, binaries, or live probes.

PR #642, opened 2026-08-06
fix(scripts): link semver-checks baselines with rust-lld

kate-shine:
## Problem

On Windows, `scripts/release-packages.ps1` and `scripts/ci/semver-report.ps1` fail during the
`cargo semver-checks` baseline build:

```
LINK : fatal error LNK1104: cannot open file '...\deps\<crate>-<hash>.exe'
```

`cargo-semver-checks` nests its baseline registry checkout under the workspace target
directory, so intermediate artifact paths grow well past 260 characters. The MSVC `link.exe`
is **not long-path aware** — it fails regardless of the `LongPathsEnabled` registry setting
or a long-path-aware manifest on the calling process.

## Fix

Link the baseline build with `rust-lld`, which prepends the `\\?\` prefix internally and
handles those depths.

Measured on a dev machine with `CARGO_TARGET_DIR` at 239 characters:

| Linker | Result |
| --- | --- |
| `link.exe` (default) | `LNK1104: cannot open file ...\debug\deps\<crate>.exe` |
| `rust-lld` | linked successfully |

### How the override is applied

`Invoke-SemverChecksCli` sets `CARGO_TARGET_<HOST_TRIPLE>_LINKER` (e.g.
`CARGO_TARGET_X86_64_PC_WINDOWS_MSVC_LINKER`) to `rust-lld.exe` for the duration of the call,
restoring or removing the variable in a `finally`.

That variable is just the environment spelling of `[target.<triple>] linker = "..."` in
`.cargo/config.toml` — the same knob, applied only where the problem is rather than to every
build in the repo. CI runs in short paths and does not have this problem, so switching its
linker would be blast radius for no gain.

An environment variable is used rather than a CLI flag because it is the only channel that
reaches the build that actually overflows `MAX_PATH`. `cargo-semver-checks` (0.47.0) exposes
no flag to forward cargo configuration, and as an *external* subcommand it receives
`--config` itself rather than letting cargo interpret it — while the failing build is the one
the tool spawns internally.

Two things keep the implementation small:

- **The bare name is enough.** `rustc` puts its own linker directory on `PATH` when it
  invokes the linker, so `rust-lld` resolves without locating the sysroot. No flavor flag is
  needed either — `rustc` infers the `link` flavor from the `rust-lld` stem.
- **Only the linker is overridden.** Deliberately *not* via `RUSTFLAGS`, which *replaces*
  `target.<triple>.rustflags` from `.cargo/config.toml` and would silently drop the
  repository's `-C target-cpu=x86-64-v3`. Verified with a verbose build that the linker
  override and the repo's `target-cpu` flag coexist.

Non-Windows platforms are unchanged; their default linkers already handle long paths.

## Known limitation

This fixes the **link step only**. `cl.exe` / `lib.exe` invoked by `cc`-based build scripts
remain MAX_PATH-bound. In practice the semver-checks baseline builds have not hit that, but
it is a narrower guarantee than relocating the whole target directory would give.

## Also in this PR

`scripts/ci/semver-report.ps1` now routes its `cargo semver-checks` invocation through the
shared `Invoke-SemverChecksCli` instead of calling cargo directly. Besides picking up the
linker override, this fixes a latent bug where `$LASTEXITCODE` was read after an intervening
command and could be stale.

## Tests

- `scripts/tests/Pester/unit/releasing/SemverChecksLinker.Tests.ps1`: env-var naming for the
  host triple, that `rust-lld` is actually in the environment *while* cargo runs, and
  save/restore including the case where cargo throws.
- `PureFunctions.Tests.ps1`: updated the `LNK1104` hint assertion.
- Full Pester suite green (431 passed, 0 failed, 1 platform skip).


scripts/lib/releasing.ps1
```diff
@@ -610,6 +610,103 @@ function Get-WorkspacePackages {
...
+
+    $targetDir = Get-SemverChecksTargetDir
+    $hadPrevious = Test-Path Env:\CARGO_TARGET_DIR
+    $previous = if ($hadPrevious) { $env:CARGO_TARGET_DIR } else { $null }
+    $applied = $false
+
+    if (-not [string]::IsNullOrWhiteSpace($targetDir)) {
+        try {
+            if (-not (Test-Path -LiteralPath $targetDir)) {
+                New-Item -ItemType Directory -Path $targetDir -Force -ErrorAction Stop | Out-Null
```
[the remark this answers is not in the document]
kate-shine: GHCP: Good catch on the inconsistency, though the suggested fix isn't available: `New-Item` has no `-LiteralPath` parameter (verified on PS 7.6 -- it only exposes `-Path`). Switched to `[System.IO.Directory]::CreateDirectory` instead, which is literal by construction, creates intermediate directories, and is a no-op when the directory already exists -- so the mismatched `Test-Path -LiteralPath` guard is removed entirely rather than just aligned. Also documented that the override must be an absolute path: a relative one would be resolved against this process's working directory but against a possibly different one by the cargo child process.

scripts/release-packages.ps1
```diff
@@ -83,6 +83,19 @@
...
+    build hash) whose generated path segments alone exceed 100 characters. MSVC's
+    `link.exe` still resolves paths against the legacy 260-character MAX_PATH
+    limit and does not honor the `LongPathsEnabled` policy, so a repository
+    cloned any distance from the drive root would fail the baseline build with
+    `LNK1104: cannot open file`. To keep well clear of that limit these runs
+    (and only these runs) are redirected to a short scratch root — `%SystemDrive%\ox-semver`
+    by default. Set the OXIDIZER_SEMVER_TARGET_DIR environment variable to
+    relocate it (for example onto a different drive), or point it at the
+    repository's own `target` directory to restore the previous behavior. On
+    non-Windows platforms Cargo's default target directory is used unchanged.
```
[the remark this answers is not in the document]
kate-shine: GHCP: Correct, that was inaccurate. The override is honored on every platform; only the *default* is Windows-only. Reworded to say so explicitly.

docs/releasing.md
```diff
@@ -256,6 +256,38 @@ up the new dependency version even when their own public API is
...
+therefore overflows the limit, and the baseline build fails with
+`LNK1104: cannot open file` on an intermediate build-script executable.
+
+So on Windows the release scripts point **only these runs** at a short
+scratch root, `%SystemDrive%\ox-semver` by default, leaving every other
+cargo invocation on the repository-local `target/` directory. Set
+`OXIDIZER_SEMVER_TARGET_DIR` to relocate it (for example onto a drive
+with more space), or point it at the repository's own `target` directory
+to restore the previous behaviour. On non-Windows platforms the default
+target directory is used unchanged. If the scratch directory cannot be
```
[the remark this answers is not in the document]
kate-shine: GHCP: Same fix as the equivalent note in release-packages.ps1 -- reworded to make clear the override applies on every platform and that the Windows-only part is the default when it is unset.

kate-shine:
GHCP: Addressed the three suppressed findings from the last reviews. They had no inline threads to reply to, so recording the responses here.

**1. Hardcoded separator in `Clear-LegacySemverChecksScratch` containment check** — correct, and worse than described. The concern was raised for forward slashes on Windows; the same defect broke Linux outright, where `Join-Path` *always* produces `/`, so `StartsWith("$normalizedRepoTarget\")` could never match and the guard was dead code. `script-tests (ubuntu-latest)` caught it. Fixed in 320d8f95 with `[System.IO.Path]::DirectorySeparatorChar` plus a platform-appropriate string comparison (`OrdinalIgnoreCase` on Windows, `Ordinal` elsewhere). Added cases for the target dir being the scratch dir itself, and for a sibling like `target-scratch` that shares a string prefix without being contained.

**2. `OXIDIZER_SEMVER_TARGET_DIR` documented as absolute but not validated** — agreed, fixed in 1a426e03. Now rejected via `IsPathFullyQualified`, which catches both relative and drive-relative (`C:foo`) forms. Worth noting the two failure modes compound: a relative override does not merely create the scratch directory in an unexpected place, it also defeats the containment check in finding 1, which compares against an absolute repository path — so it would conclude that a same-directory override lies elsewhere and delete the live scratch directory.

**3. Test used `'D:\short'` on non-Windows** — correct, and it became load-bearing once the validation above landed: `D:\short` is not fully qualified on Linux, so the test would have started failing there. Switched to a `GetTempPath()`-based absolute path, and added explicit coverage for the relative and drive-relative rejections.

Full unit suite: 343 passed, 0 failed. Both `script-tests` legs were green on 320d8f95.

kate-shine:
GHCP: Both suppressed findings addressed in f5d8cd89.

**1. `semver-report` not passing `-RepoRoot` through** — the change is right, but the reported failure is not reachable today. The call site at `semver-report.ps1:217` sits inside `Push-Location $RepoRoot` (line 121) … `Pop-Location` (line 282), so `$PWD.ProviderPath` already equals `$RepoRoot` and the cleanup targets the correct directory. The real problem is the coupling: correctness depends on a `Push-Location` about a hundred lines earlier, so moving or removing it would silently break the cleanup with no local signal. Passing it explicitly costs nothing and makes the call site self-contained. Applied at both call sites — `Invoke-CrateSemverCheck` (`releasing.ps1:835`) had exactly the same latent coupling, which the review didn't flag. Also softened the parameter comment, which asserted the default "is correct for them" and would have discouraged precisely this fix.

**2. Windows-only example in the error message** — agreed, straightforwardly wrong for a knob documented as working on every platform. The example is now chosen per platform (`/var/tmp/ox-semver` off Windows).

Full unit suite: 343 passed, 0 failed.

ralfbiedert (commented):
🤖 `Clear-LegacySemverChecksScratch` compares lexical paths without canonicalizing dot segments. For example, `OXIDIZER_SEMVER_TARGET_DIR=C:\repo\target\..\semver-out` is fully qualified but is classified as being inside `target`, even though it resolves to a sibling. Cleanup can then be skipped, leaving stale semver-check clones that cause duplicate/ambiguous-package failures, or the wrong cache can be rebuilt. Please normalize both paths with `[IO.Path]::GetFullPath()` before the platform-specific equality/child comparison, and add Pester coverage for `.` and `..` segments.

kate-shine:
GHCP: Fixed in 5cd67099. `Clear-LegacySemverChecksScratch` now canonicalizes both the configured target directory and the repository target directory with `[System.IO.Path]::GetFullPath()` before the platform-specific equality/child comparison. Added Pester coverage proving `target/./custom` is retained while `target/../semver-out` is treated as a sibling and triggers legacy cleanup.

kate-shine:
GHCP: Follow-up simplification in 8b02915c: replaced the platform-specific canonical-prefix comparison with Path.GetRelativePath. The guard now protects only the repository target itself or a configured target inside the legacy scratch being deleted; another target nested under repo/target no longer incorrectly suppresses stale-scratch cleanup. Added coverage for both cases.

kate-shine:
GHCP: Simplified further in 78e9228a: the shared helper now owns RepoRoot/Push-Location, RepoRoot is mandatory, path comments are concise, and user-facing docs only describe the Windows default plus the absolute-path override. This removes 143 lines and adds 25 relative to the previous revision without changing behavior.

kate-shine:
GHCP: Agreed that writing to the drive root is an antipattern. Updated the Windows default to the per-user %TEMP%\ox-semver in 2be4fa83; OXIDIZER_SEMVER_TARGET_DIR remains available for environments that need an even shorter path.

kate-shine:
GHCP: Updated the Windows default to ~\.cache\ox-sv in 8533a7d4. On the reported machine this is 31 characters, projecting the failing path to about 240 characters (247 for the longer crate example), without writing to C:\.

ralfbiedert (dismissed):
🤖 Approved.

scripts/release-packages.ps1
```diff
@@ -83,6 +83,11 @@
     cargo-semver-checks remains a hard dependency for ordinary library packages
     (install the version pinned in constants.env); there is no heuristic fallback.
 
+    On Windows, cargo-semver-checks uses `~\.cache\ox-sv` to avoid path length
```
sgalkin, line 86: are you sure it's short enough? afaik cargo passes fully resolved path to tools, so it will look more like `c:\Users\<user_name>\.cache\ox-sv`

sgalkin (approved):

scripts/lib/releasing.ps1
```diff
@@ -610,6 +610,123 @@ function Get-WorkspacePackages {
...
+            # CreateDirectory treats user-provided paths literally.
+            [void][System.IO.Directory]::CreateDirectory($targetDir)
+            $env:CARGO_TARGET_DIR = $targetDir
+            $applied = $true
+        } catch {
+            Write-Warning "Could not use '$targetDir' as the cargo-semver-checks target directory ($($_.Exception.Message)). Falling back to the default; on Windows a long repository path may overflow MAX_PATH and fail the baseline build. Set OXIDIZER_SEMVER_TARGET_DIR to a short writable path to override."
+        }
+    }
+
+    if ($applied) {
```
sgalkin: Could we move `Clear-LegacySemverChecksScratch` inside the `try` guarded by the `CARGO_TARGET_DIR` restoration? A terminating cleanup error currently occurs before the `finally` is active and leaks the override into later Cargo commands; a forced-cleanup failure test would pin this down.

kate-shine: GHCP: Fixed in 20c9546a. Legacy cleanup now runs inside the existing restoration try/finally, and a regression test forces cleanup to terminate and verifies CARGO_TARGET_DIR is restored.

scripts/tests/Pester/unit/releasing/SemverChecksTargetDir.Tests.ps1
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    It 'removes the in-repo scratch directory when the target dir is redirected elsewhere' {
+        $elsewhere = Join-Path ([System.IO.Path]::GetTempPath()) 'ox-semver-elsewhere'
+        Clear-LegacySemverChecksScratch -RepoRoot $script:repo -TargetDir $elsewhere 6>$null
+
+        Test-Path -LiteralPath $script:legacy | Should -BeFalse
+    }
+    It 'leaves the directory alone when the target dir IS the repository target dir' {
+    It 'leaves the directory alone when the target dir IS the repository target dir' {
```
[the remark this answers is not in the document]
kate-shine: GHCP: Fixed—the duplicate declaration was removed and the focused Pester file passes.

scripts/release-packages.ps1
kate-shine: At least currently it seems fine, the former implementation was putting it in root folder, which I wasn't that comfortable with :/ But if people consider it fine, it's certainly more future proof. It might just be my Linux background, where creating a folder in `/` is basically a crime

kate-shine:
GHCP: Simplified in b56d0353: no linker-output parsing. On Windows the helper now rejects target roots over 32 characters before invoking Cargo and directs users to OXIDIZER_SEMVER_TARGET_DIR. Added focused preflight coverage.

kate-shine:
GHCP: Removed the 32-character preflight in 2d99f2dd. The final path suffix varies by crate, target, and build dependency, so target-root length is not a reliable proof. Windows build failures now include a concise OXIDIZER_SEMVER_TARGET_DIR hint without parsing linker output.

scripts/lib/releasing.ps1
```diff
@@ -610,6 +610,127 @@ function Get-WorkspacePackages {
...
+
+    $targetDir = Get-SemverChecksTargetDir
+    $hadPrevious = Test-Path Env:\CARGO_TARGET_DIR
+    $previous = if ($hadPrevious) { $env:CARGO_TARGET_DIR } else { $null }
+    $applied = $false
+
+    if (-not [string]::IsNullOrWhiteSpace($targetDir)) {
+        if ($IsWindows -and $targetDir.Length -gt 32) {
+            throw "The cargo-semver-checks target path is too long for reliable Windows baseline builds ($($targetDir.Length) characters): '$targetDir'. Set OXIDIZER_SEMVER_TARGET_DIR to a shorter absolute path."
+        }
```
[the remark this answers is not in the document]
kate-shine: Yep, that was a nonsense.

scripts/release-packages.ps1
kate-shine: For now, I added a better explanation and a possible workaround, we can change it anytime and it's fully customizable for user.

kate-shine:
GHCP: Fixed the Ubuntu script-test failure in 7a7ce9b5. The regression test used SystemDrive, which is empty on Linux and produced an unwritable /ox-sv-* path; it now uses Pester's cross-platform TestDrive. Full script suite: 433 passed.

sgalkin (dismissed):

kate-shine:
GHCP: Fixed in 2147d0e6.

The finding was half right, and worth acting on either way. Empirically `$Matches` *does* propagate out of the `Where-Object` block on PowerShell 7.6.5 — that is why the existing test passed rather than throwing — so it was not broken as described. But relying on implicit leakage of an automatic variable is fragile, and there is a subtler hazard nearby: a failed `-match` leaves `$Matches` untouched rather than clearing it, so the early-return guard was the only thing standing between a missing `host:` line and a stale triple.

Switched to `[regex]::Match` and read the capture off the returned Match object, so nothing depends on automatic-variable scoping and the failure case cannot inherit an earlier value.

Added two tests that pin the parse: one mocks `rustc` reporting an `aarch64` host and asserts the resulting variable name, which proves the triple is read from the `host:` line rather than picked up from the ambient machine; the other asserts output with no `host:` line returns nothing instead of a stale or malformed name.

kate-shine:
GHCP: Addressed both points from the latest review in 77948bed.

**1. `$LASTEXITCODE` after `rustc` — real defect, confirmed by experiment.**
I tested it rather than assuming: with a PowerShell-function shim for `rustc`, `$LASTEXITCODE` retains its previous value (set it to `7`, call the shim, it is still `7`), so the guard rejected a perfectly good `host:` line. My earlier `aarch64` test was only passing because `$LASTEXITCODE` happened to be `0` at that point.

I did not switch to `$?`, which has its own pitfalls with the `2>$null` redirection. The exit-code check was simply redundant: the `host:` regex is the real success signal, and a missing or failing `rustc` produces nothing that can match. So the guard is gone, and the invocation is wrapped in `try/catch` to also cover `rustc` not being installed (verified that a command-resolution failure is catchable).

Two regression tests pin this. I verified they are genuine by temporarily reverting the fix: both fail on the old code and pass on the new.

**2. `$pathHint` wording.** Correct — `ConvertFrom-SemverChecksOutput` is reachable when no override was applied (`Get-SemverChecksLinkerEnvName` returns `$null`, or a caller bypasses `Invoke-SemverChecksCli`), so asserting "despite the rust-lld override" could be false. Reworded to state only what is known.

Linker tests 10 passed / 1 skipped, `PureFunctions` 118/118, `just spellcheck` clean; full Pester suite running.

scripts/lib/releasing.ps1
```diff
@@ -958,6 +958,99 @@ function Test-PackageAllowlistNamesTarget {
...
+# RUSTFLAGS would not -- that replaces target.<triple>.rustflags wholesale.
+function Invoke-SemverChecksCli {
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory = $true)][string]$PackageName,
+        [Parameter(Mandatory = $true)][string]$BaselineSha,
+        [Parameter(Mandatory = $true)][string]$RepoRoot
+    )
+
+    $linkerVar = Get-SemverChecksLinkerEnvName
```
martin-kolinek: 🤖: `Get-SemverChecksLinkerEnvName` runs before `Push-Location $RepoRoot`, but rustup resolves directory-specific toolchain overrides from the current working directory. If this release entry point is invoked from a directory whose override has a different host triple (for example GNU outside the repo and MSVC inside it), this computes the wrong `CARGO_TARGET_<TRIPLE>_LINKER`; Cargo then runs in `RepoRoot`, ignores that variable, and the baseline still uses its default linker — i.e. the fix silently doesn't apply.

Please perform the rustc probe / environment setup after entering `RepoRoot` (while retaining the restoration `finally`), so the probe and semver-checks see the same active toolchain.

scripts/lib/releasing.ps1
```diff
@@ -958,6 +958,99 @@ function Test-PackageAllowlistNamesTarget {
...
+# toolchain default should stand.
+#
+# MSVC link.exe is not long-path aware, so baseline builds under a deep target
+# directory fail with LNK1104; rust-lld handles those paths. Only Windows needs
+# this -- other platforms already default to long-path-safe linkers.
+function Get-SemverChecksLinkerEnvName {
+    [CmdletBinding()]
+    param()
+
+    if (-not $IsWindows) {
```
martin-kolinek, line 974: 🤖: Non-blocking — the gate here is `$IsWindows`, but the justification (and the comment three lines above) is MSVC-specific: `link.exe` is what is not long-path aware. On a `*-pc-windows-gnu` host this still fires and sets the linker to `rust-lld.exe` — and rustc's flavor inference maps stem `rust-lld` to `(Cc::No, Lld::Yes)`, so on a gnu target that switches linking from the gcc driver to lld invoked directly: a materially different link path that this fix doesn't need and that nothing here has exercised.

The triple is already parsed a few lines down, so the narrower gate is nearly free: compute the triple first, then return `$null` unless it ends in `-msvc`. That also makes the function say exactly what the comment says.

scripts/lib/releasing.ps1
```diff
@@ -958,6 +958,99 @@ function Test-PackageAllowlistNamesTarget {
...
+    # already rejects; $LASTEXITCODE would be no help here anyway, since it is
+    # only updated for native executables and would hold a stale value whenever
+    # rustc resolves to a shim.
+    $version = try { & rustc -vV 2>$null | Out-String } catch { '' }
+
+    # Read the capture off the Match object rather than $Matches, which -match
+    # leaves untouched when it fails and so can hold a stale value.
+    $hostMatch = [regex]::Match($version, '(?m)^host:\s*(\S+)')
+    if (-not $hostMatch.Success) {
+        return $null
```
martin-kolinek, line 992: 🤖: Non-blocking — failing open here is the right policy (a rustc probe must not break releases), but the skip is completely silent: on Windows, if `rustc -vV` is missing, throws, or prints no `host:` line, the override this whole PR exists for is quietly not applied, the baseline build fails with the original `LNK1104`, and the only thing the user sees is `ConvertFrom-SemverChecksOutput`'s new hint telling them to shorten the repository path — advice for a problem the tooling believes it already fixed, with no indication that the intended remedy never ran.

A single `Write-Warning` on this Windows-and-no-triple path ("could not determine the host triple from `rustc -vV`; leaving the toolchain default linker in place") makes the two failure modes distinguishable without changing the fail-open behaviour.

scripts/lib/releasing.ps1
```diff
@@ -958,6 +958,99 @@ function Test-PackageAllowlistNamesTarget {
...
+
+    $linkerVar = Get-SemverChecksLinkerEnvName
+    $hadPrevious = $false
+    $previous = $null
+
+    if ($linkerVar) {
+        $path = "Env:\$linkerVar"
+        $hadPrevious = Test-Path $path
+        $previous = if ($hadPrevious) { (Get-Item $path).Value } else { $null }
+        Set-Item -Path $path -Value 'rust-lld.exe'
```
martin-kolinek: 🤖: Non-blocking — worth deciding explicitly: this overwrites a *pre-existing* `CARGO_TARGET_<HOST>_LINKER` unconditionally, so a developer who has deliberately configured their own linker for the host target (a wrapper, a different lld, a custom driver) has that choice silently discarded for the duration of the semver run — restored afterwards, but ignored while it matters. The stated premise is "MSVC `link.exe` can't do long paths"; if the user has already told cargo not to use `link.exe`, the premise doesn't hold and we're overriding for nothing.

`$hadPrevious` is already computed just above, so honouring an explicit setting is a two-line change (skip the `Set-Item` when `$hadPrevious`, ideally with a note on stdout). If stomping it is the intended answer, a one-line comment saying so would keep the next reader from re-asking.

scripts/lib/releasing.ps1
```diff
@@ -958,6 +958,99 @@ function Test-PackageAllowlistNamesTarget {
...
+    }
+
+    try {
+        Push-Location $RepoRoot
+        try {
+            # A required version bump produces an expected non-zero exit code.
+            $PSNativeCommandUseErrorActionPreference = $false
+            $output = & cargo semver-checks --package $PackageName --baseline-rev $BaselineSha --all-features --color never 2>&1 | Out-String
```
martin-kolinek, line 1052: 🤖: Non-blocking — could we add a Windows-only integration test around this real invocation? Every new test mocks `cargo`, so they prove the environment-variable lifecycle but not the load-bearing claim of the PR: that cargo-semver-checks forwards this per-target linker setting to its inner baseline build, and that rustc resolves the bare `rust-lld.exe` there. If either assumption is wrong, the whole change is a no-op and the suite stays green.

A focused test under `scripts/tests/Pester/integration/` (or a dedicated Windows workflow step) could install the pinned cargo-semver-checks version, create a tiny versioned crate under a deliberately deep path, call `Invoke-SemverChecksCli`, and assert a parseable result with no `LNK1104`.

scripts/lib/releasing.ps1
```diff
@@ -1006,18 +1099,9 @@ function Invoke-CrateSemverCheck {
...
-    } finally {
-        Pop-Location
-    }
+    $result = Invoke-SemverChecksCli -PackageName $PackageName -BaselineSha $bump.Sha -RepoRoot $RepoRoot
 
-    return ConvertFrom-SemverChecksOutput -Output $output -ExitCode $exitCode -PackageName $PackageName
+    return ConvertFrom-SemverChecksOutput -Output $result.Output -ExitCode $result.ExitCode -PackageName $PackageName
 }
 
 # Parses `cargo semver-checks` combined output into a change type. Pure (no I/O)
```
martin-kolinek, line 1122: 🤖: Non-blocking — this header still says the function is "Pure (no I/O) so it can be unit-tested against captured tool output", but after this change its output also depends on `$IsWindows`, so the thrown message is no longer a function of the arguments alone — which is exactly why the new case in `PureFunctions.Tests.ps1` has to carry `-Skip:(-not $IsWindows)`. Worth a word in the comment (e.g. noting the message carries a platform-conditional hint) so the stated contract matches the code.

martin-kolinek (approved):

scripts/lib/releasing.ps1
kate-shine: GHCP: Fixed in 50e98117 — good catch, this was the one that could silently defeat the whole change. The probe now runs inside `Push-Location $RepoRoot`, so it and cargo see the same rustup-resolved toolchain. Added a unit test asserting the probe observes `$RepoRoot` as its working directory.

scripts/lib/releasing.ps1
kate-shine: GHCP: Agreed, and taken. The gate is now the parsed triple rather than `$IsWindows`: anything not ending in `-msvc` returns `$null`, so a `*-windows-gnu` host keeps its gcc driver instead of being switched to lld invoked directly. The comment and the code now say the same thing. Covered by a new `windows-gnu` test.

scripts/lib/releasing.ps1
kate-shine: GHCP: Fixed — this was the failure mode where the user gets advice for a problem the tooling thinks it solved. A `Write-Warning` now fires on the Windows-and-no-triple path, naming LNK1104, while the fail-open behaviour is unchanged. The non-msvc path stays silent, since skipping there is intended rather than a degradation. Test asserts the warning.

scripts/lib/releasing.ps1
kate-shine: GHCP: Taken — honouring it. If `CARGO_TARGET_<HOST>_LINKER` is already set we leave it alone (with a `Write-Verbose`), exactly on your reasoning: the premise is that link.exe cannot do long paths, and an explicit setting has already opted out of link.exe. As a result nothing is saved or restored any more, only removed when we set it. The restore-on-throw test became a remove-on-throw test.

scripts/lib/releasing.ps1
kate-shine: GHCP: Added `scripts/tests/Pester/integration/SemverChecksLinker.Tests.ps1`. It builds a real binary crate with `CARGO_TARGET_DIR` forced past 230 chars: first with the default linker, then with the override. It self-skips if the control build unexpectedly succeeds, so it cannot fail spuriously on a host that already links long paths. On my machine it does not skip — the control fails with LNK1104 and the override links clean, so the mechanism is now pinned end to end rather than asserted.

I left cargo-semver-checks itself out deliberately: requiring it would add a network install to the suite, and the part it contributes is plain process-environment inheritance to the build it spawns. What was genuinely unproven was that a bare `rust-lld.exe` resolves and links under a long path, and that is what the test covers. Runtime ~46s.

scripts/lib/releasing.ps1
kate-shine: GHCP: Fixed — the header now records that the message is a function of the arguments *and* `$IsWindows`, because of the platform-conditional hint, which is what forces the `-Skip:(-not $IsWindows)` on that case.

kate-shine:
GHCP: Copilot review addressed in 5c47b3c2 — two of the four suppressed comments were valid, one was not.

**Fixed — permissive assertion (integration test).** `Should -Match 'LNK1104|path'` was close to vacuous: the crate path appears in almost any cargo failure, so an unrelated error would have satisfied the premise and made the real assertion meaningless. Tightened to `LNK1104` alone. The test still passes, which confirms the control build genuinely produces LNK1104 rather than merely failing.

**Fixed — `$global:LASTEXITCODE` leak (unit test).** Correct, that one is explicitly global and did persist. Now saved and restored in a `finally`.

**Not applied — `$PSNativeCommandUseErrorActionPreference` "leaks across tests".** This is incorrect, and I checked rather than assuming: assigning a preference variable inside a scriptblock creates a *local* that shadows the outer one, so it is discarded when the block exits.

```powershell
$PSNativeCommandUseErrorActionPreference = $true
& { $PSNativeCommandUseErrorActionPreference = $false }
$PSNativeCommandUseErrorActionPreference   # -> True
```

Since each `It` body is its own scriptblock, nothing escapes into later tests and there is no value to save or restore. The same reasoning applies to the existing use in `Invoke-SemverChecksCli`. Adding save/restore here would be ceremony against a problem that does not exist, so I left both alone.

Verification: linker unit tests 13 passed / 1 skipped, the integration test passes, and the full Pester suite was green at 529 passed / 0 failed / 1 skipped on the previous commit (this change touches only the two test files re-run above).

martin-kolinek (approved):

kate-shine:
GHCP: Addressed in ab209810 — the point was fair: the help text described the outcome without saying who causes it, so it read as though cargo-semver-checks selects rust-lld by itself.

`scripts/release-packages.ps1` now names the mechanism and its scope explicitly: these scripts set `CARGO_TARGET_<HOST>_LINKER=rust-lld.exe` for the duration of the call, and invoking cargo-semver-checks directly keeps the default linker.

I made the same correction in `docs/releasing.md`. It already credited "the release scripts", but was equally vague about *how*, and the two would have drifted apart otherwise.

Docs and comments only — no behaviour change. `just spellcheck` clean; I also confirmed no test asserts against this help text, so the previously green suite still stands.

martin-kolinek (approved):

Vaiz (approved):

PR #647, opened 2026-08-06
feat(arty_executor): migrate executor from ox-sdk

martintmk:
## Summary

- migrate `oxidizer_executor` from the O365 Core `ox-sdk` repository at commit `47aaa9f9fc60e95de9b0c3aeb7742e7cd883f9ce`
- rename the crate, examples, benchmarks, imports, metrics, and documentation to `arty_executor`
- adapt workspace dependencies, testing helpers, package metadata, generated README, and spelling dictionary
- replace the direct `fast_time` dependency with `tick` using its `fast-instant` feature

## Hardening included during migration

Review feedback led to a small set of behavioral fixes on top of the source migration:

- unrepresentable shutdown deadlines fail explicitly when shutdown begins
- dropping the executor validates that shutdown actually reached `CycleOutcome::Shutdown`
- panics from future destructors are contained consistently with polling panics, including during completion and abort
- the awakened-task queue lock is released before invoking the owner waker, avoiding reentrant lock contention
- wake synchronization documentation now explains the relaxed RMW plus acquire-fence relationship
- pointer hashing accepts both 32-bit and 64-bit pointer widths

## Benchmark comparison

Callgrind instruction counts from the migrated Arty crate and the source SDK crate show no material performance regression:

| Scenario | Arty | SDK | Delta |
|---|---:|---:|---:|
| `cycle_empty` | 6,873 | 6,726 | +2.2% |
| `task_add` | 5,254 | 5,254 | 0.0% |
| `cycle_one_ready` | 15,493 | 15,387 | +0.7% |
| `poll_completed` | 78 | 78 | 0.0% |
| `yield_round_trip` | 31,160 | 31,001 | +0.5% |

## Validation

- full workspace build and 6,012-test suite
- workspace doctests and Clippy
- `arty_executor` all-feature/all-target tests
- package verification, formatting, README, spelling, license, Cargo sorting, and dependency policy checks


crates/arty_executor/Cargo.toml
```diff
@@ -0,0 +1,67 @@
...
+readme = "README.md"
+homepage = { workspace = true }
+include = { workspace = true }
+repository = "https://github.com/microsoft/oxidizer/tree/main/crates/arty_executor"
+keywords = ["arty", "executor", "async", "runtime"]
+categories = ["asynchronous"]
+
+[features]
+default = []
+unstable-testing = ["dep:scopeguard"]
```
martintmk: why not just test-utils

martintmk: &#x1F916; Agreed - `test-util` matches the feature naming convention already used throughout this repository. I am updating the migrated crate now.

---

martintmk: &#x1F916; Fixed: renamed the feature to `test-util` and updated both benchmark requirements and the public testing module gates.

---

crates/arty_executor/Cargo.toml
```diff
@@ -0,0 +1,67 @@
...
+
+[features]
+default = []
+test-util = ["dep:scopeguard"]
+
+[dependencies]
+events_once = { workspace = true }
+fast_time = { workspace = true }
+foldhash = { workspace = true }
+infinity_pool = { workspace = true }
```
martintmk, line 30: how difficult would be to switch to plurality, do not make the change just investigate

martintmk: &#x1F916; I investigated this without changing code. This would be a medium-sized refactor rather than a dependency swap.

`arty_executor` currently relies on `RawLocalEventLake` for three things that `plurality` does not provide directly: one-shot sender/receiver semantics (including waker and disconnect handling), pooling across heterogeneous task result types via `TypeId`, and shutdown diagnostics/counting for outstanding join handles. `plurality::Pool<T>` provides the reusable stable storage, but it is homogeneous and intentionally has no event state machine.

A direct conversion would therefore require an internal result-event module: a `LocalEvent<R>` state machine, sender/receiver endpoint types, and a type-erased map of `plurality::Pool<LocalEvent<R>>` instances. It would touch `executor_core.rs`, `task.rs`, and `join_handle.rs`, plus substantial lifecycle/drop tests. I would estimate roughly 1-3 engineering days, with most risk in endpoint cleanup and executor shutdown safety rather than the pool allocation itself.

The lower-risk design would be to add a plurality-backed pool/lake implementation to `events_once` while preserving its endpoint API, then update `arty_executor` once that is released. That centralizes the unsafe event lifecycle logic instead of duplicating it here.

---

scripts/run-examples.rs
```diff
@@ -33,6 +33,12 @@ const TIMEOUT: Duration = Duration::from_secs(30);
 const EXCLUDED_EXAMPLES: &[&str] = &[
     // Interactive - requires user input from stdin.
     "employees",
+    // Demonstrates process termination after an unhandled task panic.
+    "ae_panic",
```
martintmk: I think these examples should be rather updated to assert the failure but not fail the execution

martintmk: &#x1F916; Agreed - I’ll make the examples verify their intended outcomes under the example runner instead of skipping them.

---

martintmk: &#x1F916; Fixed in `2049844d`: the runner now treats `ae_panic` and `ae_shutdown_failure` as expected failures, while `ae_x_multimillionaire` runs a smaller representative workload under `IS_TESTING`. All four executor examples now execute and validate successfully.

---

scripts/run-examples.rs
```diff
@@ -28,13 +28,18 @@ use automation::{Outcome, run_with_timeout};
...
-/// Examples that are expected to panic, hang, or require user interaction
-/// and so must be skipped by this runner.
+/// Examples that require user interaction and so must be skipped by this runner.
 const EXCLUDED_EXAMPLES: &[&str] = &[
     // Interactive - requires user input from stdin.
     "employees",
 ];
 
+/// Examples that demonstrate process termination and must exit unsuccessfully.
+const EXPECTED_FAILURE_EXAMPLES: &[&str] = &[
```
martintmk: I want you to update examples itselt to not produce error code, this script must stay untouched in this PR

martintmk: &#x1F916; Agreed - I’ll revert all `scripts/run-examples.rs` changes and make each intentional-failure example assert its child process failure while returning success itself under `IS_TESTING`.

---

martintmk: &#x1F916; Fixed: `ae_panic` and `ae_shutdown_failure` now supervise a child copy under `IS_TESTING`, assert that the child terminates unsuccessfully, and return success from the parent. `scripts/run-examples.rs` is unchanged from the PR base.

---

martintmk:
review this PR

crates/arty_executor/src/executor_core.rs
```diff
@@ -0,0 +1,776 @@
...
+
+            self.drop_inert_tasks(&mut state_exclusive, &mut state_reentrant);
+
+            let outcome = if self.evaluate_shutdown_completion(&mut state_exclusive, &state_reentrant) {
+                CycleOutcome::Shutdown
+            } else if self.has_work_to_do(&state_exclusive, &state_reentrant) {
+                // We want to be immediately called again because we may have more work to do.
+                CYCLE_OUTCOME_CONTINUE.with(Event::observe_once);
+                CycleOutcome::Continue
+            } else {
```
martintmk, line 288: This can return `Suspend` while shutdown is still waiting on completed tasks or result resources. Since dropping a `JoinHandle` or the last cloned task waker does not wake `owner_waker`, a caller following `Suspend` can stop cycling and never re-enter the only code that checks the shutdown timeout; return `Continue` while `shutdown_deadline` is set, or wake the owner whenever a blocking resource is released.

crates/arty_executor/src/executor_core.rs
```diff
@@ -0,0 +1,776 @@
...
+
+impl Drop for ExecutorCore {
+    fn drop(&mut self) {
+        if thread::panicking() {
+            // We skip the assertions if we are already panicking because a double panic more often
+            // does not help anything and may even obscure the initial panic in test runs.
+            return;
+        }
+
+        assert!(
```
martintmk, line 676: This checks only that shutdown started, not the `unsafe build()` precondition that a cycle returned `Shutdown`. Dropping after `begin_shutdown()` but before completion passes this guard and can free result-event storage under live `JoinHandle`s; assert the same completed/result-event emptiness used by `evaluate_shutdown_completion()`.

crates/arty_executor/src/task.rs
```diff
@@ -0,0 +1,506 @@
...
+        let poll_result = match catch_unwind(AssertUnwindSafe(|| future_as_mut_pinned.poll(&mut cx))) {
+            Ok(x) => x,
+            Err(panic) => on_unhandled_task_panic(panic),
+        };
+
+        match poll_result {
+            task::Poll::Ready(result) => {
+                // We need to send the result through the result transmitter. This also drops
+                // any state still held by the state machine, as it cannot be polled any more.
+                let tx = maybe_payload
```
martintmk: The panic trap covers `Future::poll`, but taking and dropping the completed future here—and dropping it in `abort()`—runs arbitrary user `Drop` code outside `catch_unwind`. A panicking destructor can unwind after the task left its executor queue; include payload destruction in the trap and route it to `on_unhandled_task_panic`.

crates/arty_executor/src/wake.rs
```diff
@@ -0,0 +1,482 @@
...
+        if let Ok(mut awakened_set) = self.awakened_queue.try_lock() {
+            // We only add if we can do so without increasing capacity, because increasing capacity
+            // from an arbitrary thread may require reallocation, which we do not want to do on a
+            // different thread than the one that owns the set.
+            if awakened_set.len() < awakened_set.capacity() {
+                // If we experienced spurious awakenings, we might push the same task multiple
+                // times. That is fine - it is up to the receiver of the notifications to deal
+                // with spurious notifications (which may arrive anyway through other means).
+                awakened_set.push_back(self.task_ref);
+                self.parent_waker.wake_by_ref();
```
martintmk, line 178: **Non-blocking:** `parent_waker` is arbitrary external code and is invoked while the awakened-queue mutex is still held, widening the window in which concurrent wakes fall back to the O(n) scan. Drop `awakened_set` after the push and before calling `wake_by_ref()`.

crates/arty_executor/src/join_handle.rs
```diff
@@ -0,0 +1,61 @@
...
+
+/// A handle that can be used to await the completion of a task and to receive its result `R`.
+///
+/// If the join handle becomes disconnected because the task it references failed to reach
+/// completion, the join handle will panic when polled.
+///
+/// # Resource management
+///
+/// Join handles must be dropped before the executor they came from shuts down or the executor
+/// shutdown will result in a timeout and panic. You can think of it as there existing an imaginary
```
martintmk: **Non-blocking:** consumers cannot select the test-only `Panic` timeout behavior; the production path calls `std::process::abort()`. Document process termination rather than a panic, because this cannot be caught or unwound.

crates/arty_executor/benches/ae_basic_operations_cg.rs
```diff
@@ -0,0 +1,215 @@
...
+    );
+
+    // -----------------------------------------------------------------------------
+    // Composite group: parity with the wall-clock `_one` Criterion benches.
+    // -----------------------------------------------------------------------------
+
+    #[library_benchmark]
+    #[bench::async_round_trip(setup = make_executor_only)]
+    fn spawn_and_complete(mut state: State) -> State {
+        let mut join_handle = Box::pin(state.tasks.add(async move {}));
```
martintmk: **Non-blocking:** this allocation is inside the measured Callgrind region, while the paired Criterion benchmark stack-pins the handle. Use `pin!` and let the handle drop in the measured body so the claimed wall-clock/Callgrind parity is real.

martintmk (commented):
[Copilot speaking]

This migration has shutdown-state contract gaps and a task-destruction panic gap that can leave executor-owned storage inconsistent. **Verdict: changes requested.**

crates/arty_executor/src/executor_core.rs
martintmk, line 288: This can return `Suspend` while shutdown is still waiting on completed tasks or result resources. Since dropping a `JoinHandle` or the last cloned task waker does not wake `owner_waker`, a caller following `Suspend` can stop cycling and never re-enter the only code that checks the shutdown timeout; return `Continue` while `shutdown_deadline` is set, or wake the owner whenever a blocking resource is released.

crates/arty_executor/src/executor_core.rs
```diff
@@ -0,0 +1,776 @@
...
+            state_reentrant.shutdown_deadline.is_none(),
+            "Cannot start shutdown process when it has already been started"
+        );
+
+        let shutdown_start_time = state_exclusive.clock.now();
+        state_reentrant.shutdown_deadline = Some(
+            shutdown_start_time
+                .checked_add(self.shared.shutdown_timeout)
```
martintmk, line 517: `shutdown_timeout()` accepts every `Duration`, but `Duration::MAX` reaches this branch and panics in `begin_shutdown()` (verified with a focused test). Treat the unrepresentable deadline as no deadline, reject it when configuring the builder, or document and test the panic; the current `expect` incorrectly claims caller-controlled input is impossible.

crates/arty_executor/src/task.rs
martintmk: The panic trap covers `Future::poll`, but taking and dropping the completed future here—and dropping it in `abort()`—runs arbitrary user `Drop` code outside `catch_unwind`. A panicking destructor can unwind after the task left its executor queue; include payload destruction in the trap and route it to `on_unhandled_task_panic`.

crates/arty_executor/src/wake.rs
martintmk, line 178: **Non-blocking:** `parent_waker` is arbitrary external code and is invoked while the awakened-queue mutex is still held, widening the window in which concurrent wakes fall back to the O(n) scan. Drop `awakened_set` after the push and before calling `wake_by_ref()`.

crates/arty_executor/src/join_handle.rs
martintmk: **Non-blocking:** consumers cannot select the test-only `Panic` timeout behavior; the production path calls `std::process::abort()`. Document process termination rather than a panic, because this cannot be caught or unwound.

crates/arty_executor/benches/ae_basic_operations_cg.rs
martintmk: **Non-blocking:** this allocation is inside the measured Callgrind region, while the paired Criterion benchmark stack-pins the handle. Use `pin!` and let the handle drop in the measured body so the claimed wall-clock/Callgrind parity is real.

martintmk (commented):
[Copilot speaking]

This migration has shutdown-progress, timeout-boundary, and task-destruction contract defects that can stall shutdown or leave executor state inconsistent. **Verdict: changes requested.**

crates/arty_executor/src/executor_core.rs
martintmk, line 288: This returns `Suspend` while shutdown is still waiting on completed tasks or result resources, but releasing those resources does not wake `owner_waker`; verified: a live `JoinHandle` makes the first shutdown cycle suspend, and dropping it provides no signal to cycle again. Return `Continue` while `shutdown_deadline` is set, or wake the owner whenever a shutdown-blocking resource is released.

crates/arty_executor/src/executor_core.rs
```diff
@@ -0,0 +1,776 @@
...
+                .expect("impossible for shutdown timeout to be so high we cross the end of the universe"),
```
martintmk: `shutdown_timeout()` accepts every `Duration`, but verified: `Duration::MAX` panics here when shutdown begins. Reject unrepresentable timeouts in the builder, treat them as no deadline, or document and test this panic instead of claiming caller-controlled input is impossible.

crates/arty_executor/src/task.rs
martintmk: Taking and dropping the completed future here, and dropping it in `abort()`, runs arbitrary user `Drop` code outside the panic trap; verified: a future that returns `Ready` and panics in `Drop` unwinds from `execute_cycle()`. Include payload destruction in the trap and route it to `on_unhandled_task_panic` so executor-owned queues cannot be left inconsistent.

crates/arty_executor/src/join_handle.rs
martintmk: **Non-blocking:** consumers cannot select the test-only `Panic` timeout behavior; the production path aborts the process. Document process termination rather than a panic, because this cannot be caught or unwound.

crates/arty_executor/src/wake.rs
martintmk, line 178: **Non-blocking:** `parent_waker` is arbitrary external code and is invoked while the awakened-queue mutex remains held, widening the window in which concurrent wakes fall back to the O(n) scan. Drop `awakened_set` after the push and before calling `wake_by_ref()`.

crates/arty_executor/benches/ae_basic_operations_cg.rs
martintmk: **Non-blocking:** this heap allocation is inside the measured Callgrind region, while the paired Criterion benchmark stack-pins the handle. Use `pin!` and let the handle drop in the measured body so the two benchmarks measure the same operation.

martintmk (commented):
[Copilot speaking]

Shutdown progress and timeout validation still violate the public lifecycle contract, and a completed future's destructor can escape the executor's panic trap and leave executor-owned state inconsistent. **Verdict: changes requested.**

crates/arty_executor/Cargo.toml
```diff
@@ -0,0 +1,70 @@
...
+[package.metadata.cargo_check_external_types]
+allowed_external_types = ["scopeguard::*"]
+
+[features]
+default = []
+test-util = ["dep:scopeguard"]
+
+[dependencies]
+events_once = { workspace = true }
+fast_time = { workspace = true }
```
martintmk: use `tick` with fast time feature instead and drop `fast_time`

crates/arty_executor/src/executor_core.rs
martintmk: &#x1F916; Agreed - implementing this now.

---

crates/arty_executor/src/executor_core.rs
martintmk: &#x1F916; Agreed - implementing this now.

---

crates/arty_executor/src/task.rs
martintmk: &#x1F916; Agreed - implementing this now.

---

crates/arty_executor/src/wake.rs
martintmk: &#x1F916; Agreed - implementing this now.

---

crates/arty_executor/src/join_handle.rs
martintmk: &#x1F916; Agreed - implementing this now.

---

crates/arty_executor/benches/ae_basic_operations_cg.rs
martintmk: &#x1F916; Agreed - implementing this now.

---

crates/arty_executor/src/executor_core.rs
martintmk: &#x1F916; Agreed - implementing this now.

---

crates/arty_executor/src/executor_core.rs
martintmk: &#x1F916; Agreed - implementing this now.

---

crates/arty_executor/src/task.rs
martintmk: &#x1F916; Agreed - implementing this now.

---

crates/arty_executor/src/wake.rs
martintmk: &#x1F916; Agreed - implementing this now.

---

crates/arty_executor/src/join_handle.rs
martintmk: &#x1F916; Agreed - implementing this now.

---

crates/arty_executor/benches/ae_basic_operations_cg.rs
martintmk: &#x1F916; Agreed - implementing this now.

---

crates/arty_executor/src/executor_core.rs
martintmk: &#x1F916; Agreed - implementing this now.

---

crates/arty_executor/src/executor_core.rs
martintmk: &#x1F916; Agreed - implementing this now.

---

crates/arty_executor/src/task.rs
martintmk: &#x1F916; Agreed - implementing this now.

---

crates/arty_executor/src/join_handle.rs
martintmk: &#x1F916; Agreed - implementing this now.

---

crates/arty_executor/src/wake.rs
martintmk: &#x1F916; Agreed - implementing this now.

---

crates/arty_executor/benches/ae_basic_operations_cg.rs
martintmk: &#x1F916; Agreed - implementing this now.

---

crates/arty_executor/Cargo.toml
martintmk: &#x1F916; Agreed - implementing this now.

---

crates/arty_executor/src/executor_core.rs
martintmk: &#x1F916; Fixed: shutdown cycles now return Continue until cleanup completes or the timeout is evaluated.

---

crates/arty_executor/src/executor_core.rs
martintmk: &#x1F916; Fixed: ExecutorCore drop now asserts that completed tasks and result-event storage are empty.

---

crates/arty_executor/src/task.rs
martintmk: &#x1F916; Fixed: completed and aborted task payload destruction now runs inside the panic trap.

---

crates/arty_executor/src/wake.rs
martintmk: &#x1F916; Fixed: the awakened-queue mutex is released before invoking the parent waker.

---

crates/arty_executor/src/join_handle.rs
martintmk: &#x1F916; Fixed: JoinHandle documentation now states that a shutdown timeout terminates the process.

---

crates/arty_executor/benches/ae_basic_operations_cg.rs
martintmk: &#x1F916; Fixed: the composite Callgrind benchmarks now stack-pin and drop join handles inside the measured body.

---

crates/arty_executor/src/executor_core.rs
martintmk: &#x1F916; Fixed: shutdown cycles now return Continue until cleanup completes or the timeout is evaluated.

---

crates/arty_executor/src/executor_core.rs
martintmk: &#x1F916; Fixed: unrepresentable shutdown deadlines are documented and covered by a focused Duration::MAX panic test.

---

crates/arty_executor/src/task.rs
martintmk: &#x1F916; Fixed: completed and aborted task payload destruction now runs inside the panic trap.

---

crates/arty_executor/src/wake.rs
martintmk: &#x1F916; Fixed: the awakened-queue mutex is released before invoking the parent waker.

---

crates/arty_executor/src/join_handle.rs
martintmk: &#x1F916; Fixed: JoinHandle documentation now states that a shutdown timeout terminates the process.

---

crates/arty_executor/benches/ae_basic_operations_cg.rs
martintmk: &#x1F916; Fixed: the composite Callgrind benchmarks now stack-pin and drop join handles inside the measured body.

---

crates/arty_executor/src/executor_core.rs
martintmk: &#x1F916; Fixed: shutdown cycles now return Continue until cleanup completes or the timeout is evaluated.

---

crates/arty_executor/src/executor_core.rs
martintmk: &#x1F916; Fixed: unrepresentable shutdown deadlines are documented and covered by a focused Duration::MAX panic test.

---

crates/arty_executor/src/task.rs
martintmk: &#x1F916; Fixed: completed and aborted task payload destruction now runs inside the panic trap.

---

crates/arty_executor/src/join_handle.rs
martintmk: &#x1F916; Fixed: JoinHandle documentation now states that a shutdown timeout terminates the process.

---

crates/arty_executor/src/wake.rs
martintmk: &#x1F916; Fixed: the awakened-queue mutex is released before invoking the parent waker.

---

crates/arty_executor/benches/ae_basic_operations_cg.rs
martintmk: &#x1F916; Fixed: the composite Callgrind benchmarks now stack-pin and drop join handles inside the measured body.

---

crates/arty_executor/Cargo.toml
martintmk: &#x1F916; Fixed: arty_executor now uses tick SimpleClock with the fast-instant feature and no longer directly depends on fast_time.

---

crates/arty_executor/Cargo.toml
```diff
@@ -0,0 +1,70 @@
...
+allowed_external_types = ["scopeguard::*"]
+
+[features]
+default = []
+test-util = ["dep:scopeguard"]
+
+[dependencies]
+events_once = { workspace = true }
+foldhash = { workspace = true }
+infinity_pool = { workspace = true }
```
martintmk, line 29: could we use multitude rather than infinity_pool in arty_executor?

evaluate whether this is possible

martintmk: &#x1F916; I evaluated this, and `multitude` is not currently a suitable direct replacement for `infinity_pool` here.

`arty_executor` needs stable pinned addresses for heterogeneous task types, a cheap `Copy` type-erased `TaskRef`, and individual task removal/reuse as each task becomes inert. `RawBlindPool` provides that ownership ticket and dyn-cast model directly.

`multitude::Box` can pin and unsize trait objects, but the owning handle is not `Copy` and its raw reconstruction APIs are crate-private. Keeping ownership elsewhere would require adding a separate stable-key slab/map and raw-pointer layer, effectively rebuilding the pool abstraction. Using `multitude::Rc` would also create a self-reference cycle because each task contains its own wake signal.

It could become viable if `multitude` gains a public stable ticket/raw-owner API with individual removal, but with its current API the migration would add complexity and likely worsen reclamation for mixed-lifetime tasks. I recommend retaining `infinity_pool`.

---

martintmk: investigate if we could use `plurality`

crates/arty_executor/src/wake.rs
```diff
@@ -0,0 +1,483 @@
...
+        // led to the task being awakened), so we add an Acquire fence.
+        if self.awakened.swap(false, atomic::Ordering::Relaxed) {
+            // This does nothing on x84 but on weaker memory models, the visibility of
+            // writes to arbitrary locations may be delayed without this fence.
+            atomic::fence(atomic::Ordering::Acquire);
+            true
+        } else {
+            false
+        }
+    }
```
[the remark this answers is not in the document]
martintmk: &#x1F916; Agreed - implementing this now.

---

crates/arty_executor/src/testing/functions.rs
```diff
@@ -0,0 +1,56 @@
...
+
+use scopeguard::{Always, ScopeGuard};
+
+use crate::{CycleOutcome, Executor};
+
+/// We do not want the timeout panic to occur under mutation testing because that makes for slow
+/// mutation tests. Instead, we want the mutation test harness itself to time out! Therefore, this
+/// is a very high value to ensure that under mutation testing the executor timeout logic will
+/// never trigger.
+const MUTATION_TESTING_SHUTDOWN_TIMEOUT: Duration = Duration::from_mins(15);
```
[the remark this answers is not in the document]
martintmk: &#x1F916; Agreed - implementing this now.

---

crates/arty_executor/benches/ae_basic_operations.rs
```diff
@@ -0,0 +1,337 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Benchmarks for basic operations of the executor.
```
sandersaares: Would be good to have benchmark results before/after migration, as the PR description suggests there were functional changes done to the library in addition to migrating it.

crates/arty_executor/Cargo.toml
martintmk: &#x1F916; I investigated `plurality`. It is closer than `multitude` because `plurality::Box` supports trait-object unsizing plus public `into_raw`/`from_raw`, but its `Pool<T>` is homogeneous while `arty_executor` stores heterogeneous `Task<F>` values (each future may have a different concrete type). Using it directly would require either creating a separate pool/chunk per spawned task, which defeats slot reuse, or maintaining a `TypeId`-keyed registry of concrete pools and type-erased ownership, effectively rebuilding the heterogeneous pool layer. I recommend retaining `infinity_pool` for this executor model.

---

crates/arty_executor/src/wake.rs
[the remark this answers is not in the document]
martintmk: &#x1F916; I checked the Rust atomic memory model. The existing relaxed RMW followed by an Acquire fence is valid: when the RMW reads the value written by the Release store, the fence sequenced after that read establishes the synchronization. Making the `swap` itself Acquire would unnecessarily impose acquire semantics on the common false fast path. I retained the ordering, fixed the `x84` typo to `x86`, and clarified the synchronization comment.

---

crates/arty_executor/src/testing/functions.rs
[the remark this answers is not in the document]
martintmk: &#x1F916; I verified this against the workspace MSRV: `rustc 1.93.1` compiles `Duration::from_mins(15)`. The replacement also fails this workspace's Clippy configuration with `duration_suboptimal_units`, which specifically requires `from_mins`. I am retaining the existing constructor.

---

crates/arty_executor/src/ptr_hash.rs
```diff
@@ -0,0 +1,63 @@
...
+    fn write(&mut self, bytes: &[u8]) {
+        let input_raw = u64::from_ne_bytes(
+            bytes
+                .try_into()
+                .expect("PointerHasher only supports 64-bit input data, so the slice must be exactly 8 bytes long."),
+        );
+
+        self.value ^= input_raw;
+        self.value ^= input_raw.reverse_bits();
+    }
```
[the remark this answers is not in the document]
martintmk: &#x1F916; Agreed - implementing this now.

---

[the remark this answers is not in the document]
martintmk: &#x1F916; Fixed: PointerHasher now accepts both 32-bit and 64-bit pointer encodings, zero-extending 32-bit inputs before mixing, with focused tests for the 32-bit and invalid-width paths.

---

crates/arty_executor/src/cycle_outcome.rs
```diff
@@ -0,0 +1,36 @@
...
+    /// periodically, merely because there may (in theory) be events that create more work for the
+    /// executor but which fail to be detected via any of the above signals. Therefore, a time-based
+    /// periodic execution cycle may serve as a useful fail-safe.
+    ///
+    /// [1]: crate::ExecutorBuilder::owner_waker
+    Suspend,
+
+    /// The executor has completed shutdown and is ready to be dropped.
+    ///
+    /// Attempting additional executor processing cycles in this state will panic.
```
psandana: 🤖 Pato's Pull Request Agent: This doc says "Attempting additional executor processing cycles in this state will panic," but that isn't enforced. After `execute_cycle()` first returns `Shutdown`, `shutdown_deadline` is still `Some` and `completed`/`result_events` are empty, so a subsequent `execute_cycle()` runs the (no-op) pipeline and `evaluate_shutdown_completion()` returns `true` again — it returns `Shutdown` rather than panicking. Either enforce the contract (a terminal flag that panics on re-entry) or soften the doc to state that further cycles are harmless no-ops that keep returning `Shutdown`.

crates/arty_executor/src/executor.rs
```diff
@@ -0,0 +1,722 @@
...
+///
+/// The executor will terminate the process if the shutdown process times out. This is typically
+/// a sign of a resource leak that prevented resource manager finalization.
+///
+/// Potential causes include:
+///
+/// * Some future awaited by a task failed to cancel an ongoing `await` operation when the future
+///   was dropped. This suggests a resource management defect in the future.
+/// * A [`JoinHandle`][1] remains alive somewhere with an independent lifetime (e.g. in
+///   a `thread_static!` variable). This suggests a resource management defect in whatever logic
```
psandana: 🤖 Pato's Pull Request Agent: `thread_static!` isn't a real macro; the std macro is `thread_local!` (as used in `executor_core.rs`).
```suggestion
///   a `thread_local!` variable). This suggests a resource management defect in whatever logic
```

crates/arty_executor/src/executor.rs
```diff
@@ -0,0 +1,722 @@
...
+    pub fn builder() -> ExecutorBuilder {
+        ExecutorBuilder::new()
+    }
+
+    #[must_use]
+    pub(crate) fn new(core: ExecutorCore) -> Self {
+        Self { core: Rc::new(core) }
+    }
+
+    /// Creates a new handle to the set of tasks registered with with the executor. You can use
```
psandana: 🤖 Pato's Pull Request Agent: Doubled word "with with".
```suggestion
    /// Creates a new handle to the set of tasks registered with the executor. You can use
```

crates/arty_executor/src/executor_core.rs
```diff
@@ -0,0 +1,783 @@
...
+            // creates references to the tasks and it only ever creates temporary non-overlapping
+            // references narrowly bounded to individual code blocks, ensuring that aliasing rules
+            // are upheld. Anything outside `ExecutorCore` only passes `TaskRef` by value, never
+            // dereferencing it. Reentrant logic for registering new tasks cannot touch existing tasks.
+            let task = unsafe { task_ref.as_task() };
+
+            match task.poll() {
+                task::Poll::Ready(()) => {
+                    // The task has completed, so we can move it to the completed list.
+                    // It will sit there until it signals `is_inert()` at which point is is dropped.
```
psandana: 🤖 Pato's Pull Request Agent: Typo "is is dropped".
```suggestion
                    // It will sit there until it signals `is_inert()` at which point it is dropped.
```

crates/arty_executor/src/executor_core.rs
```diff
@@ -0,0 +1,783 @@
...
+
+    /// Storage for all the tasks registered with the executor. Once a task has been registered,
+    /// the executor only sees it as a [`TypeErasedTask`][crate::TypeErasedTask], with only the
+    /// task itself knowing the specific type of the future and the type of the result it produces.
+    task_storage: RawBlindPool,
+
+    /// In shutdown mode (`Some`), all tasks are considered completed and the only thing we do is
+    /// wait for them to become inert (which may be driven by uncontrollable actions of foreign
+    /// threads). New tasks can no longer be scheduled in this mode (a panic will occur). If the
+    /// deadline is reached without a successful shutdown, we terminate the process and try report
```
psandana: 🤖 Pato's Pull Request Agent: "try report" → "try to report".
```suggestion
    /// deadline is reached without a successful shutdown, we terminate the process and try to report
```

crates/arty_executor/src/task.rs
```diff
@@ -0,0 +1,581 @@
...
+    fn consume_awakened(&self) -> bool;
+
+    /// Clears the inner state of the task, entering a form where no further polling is possible
+    /// and any future-specific state is dropped.
+    ///
+    /// The resources owned by the task may still remain in use - this has no implications with
+    /// regard to what `is_inert()` is expected to return.
+    fn abort(self: Pin<&Self>);
+
+    /// Initializes the the task, providing it the wake signal that it needs to enable polling
```
psandana: 🤖 Pato's Pull Request Agent: Doubled word "the the task".
```suggestion
    /// Initializes the task, providing it the wake signal that it needs to enable polling
```

crates/arty_executor/src/executor_core.rs
```diff
@@ -0,0 +1,783 @@
...
+        );
+
+        let wake_signal = WakeSignal::new(
+            Arc::clone(&self.shared.awakened),
+            Arc::clone(&self.shared.probe_embedded_wake_signals),
+            self.shared.owner_waker.clone(),
+            task_ref,
+        );
+
+        // SAFETY: The task is alive (we own it and just created it) and we are on the the thread
```
psandana: 🤖 Pato's Pull Request Agent: The SAFETY comment has a doubled word "on the the thread". This same typo is repeated across the copied SAFETY blocks at lines 239, 369, 392, 449, 542, and 651 — should be "on the thread" in all of them.

crates/arty_executor/src/cycle_outcome.rs
martintmk: &#x1F916; The comment contains only a reference to a local temporary file, which is not accessible from GitHub. Please paste the intended feedback into the thread and unresolve it so I can address it.

---

crates/arty_executor/src/executor.rs
martintmk: &#x1F916; The comment contains only a reference to a local temporary file, which is not accessible from GitHub. Please paste the intended feedback into the thread and unresolve it so I can address it.

---

crates/arty_executor/src/executor.rs
martintmk: &#x1F916; The comment contains only a reference to a local temporary file, which is not accessible from GitHub. Please paste the intended feedback into the thread and unresolve it so I can address it.

---

crates/arty_executor/src/executor_core.rs
martintmk: &#x1F916; The comment contains only a reference to a local temporary file, which is not accessible from GitHub. Please paste the intended feedback into the thread and unresolve it so I can address it.

---

crates/arty_executor/src/executor_core.rs
martintmk: &#x1F916; The comment contains only a reference to a local temporary file, which is not accessible from GitHub. Please paste the intended feedback into the thread and unresolve it so I can address it.

---

crates/arty_executor/src/task.rs
martintmk: &#x1F916; The comment contains only a reference to a local temporary file, which is not accessible from GitHub. Please paste the intended feedback into the thread and unresolve it so I can address it.

---

crates/arty_executor/src/executor_core.rs
martintmk: &#x1F916; The comment contains only a reference to a local temporary file, which is not accessible from GitHub. Please paste the intended feedback into the thread and unresolve it so I can address it.

---

crates/arty_executor/benches/ae_basic_operations.rs
sandersaares, line 4: Would be desirable to have a benchmark comparison in the PR description, so we have a clear record that this was not a performance-affecting migration.

crates/arty_executor/examples/ae_x_multimillionaire.rs
```diff
@@ -0,0 +1,87 @@
...
+
+// Two layers multiplied together is 10 million.
+const FIRST_LAYER_TASK_COUNT: usize = 1_000;
+const SECOND_LAYER_TASK_COUNT: usize = 10_000;
+const TEST_FIRST_LAYER_TASK_COUNT: usize = 10;
+const TEST_SECOND_LAYER_TASK_COUNT: usize = 100;
+
+fn main() {
+    // Preserve the behavior while keeping automated example validation fast.
+    let (first_layer_task_count, second_layer_task_count) = if env::var_os("IS_TESTING").is_some() {
```
sandersaares, line 21: `IS_TESTING` relies on our dev-env automation. Is it correctly setting that variable in GitHub? I ask because we have reinvented the wheel on the dev-env automation many times, so I cannot be sure that GH==ADO. Let's make sure (and maybe document here what makes it so we can expect this var to be present).

crates/arty_executor/src/testing/functions.rs
```diff
@@ -0,0 +1,56 @@
...
+const TEST_TIMEOUT: Duration = Duration::from_secs(10);
+
+#[cfg_attr(test, mutants::skip)] // Test-harness configuration is not production behavior.
+fn is_mutation_testing() -> bool {
```
sandersaares, line 20: This stuff in the upper part of the file should be in `testing_aids` or whatever it was called, as timeouts and mutation testing flags are not functionality specific to this crate.

crates/arty_executor/src/testing/functions.rs
```diff
@@ -0,0 +1,56 @@
...
+    env::var("MUTATION_TESTING").as_deref() == Ok("1")
```
sandersaares, line 21: Same comment as `IS_TESTING` - do we have the proper automation wired up in GitHub to actually set that variable? If not, let's correct that and leave a comment here to explain what ensures we can rely on this.

https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7765832

crates/arty_executor/src/lib.rs
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) Microsoft Corporation.
```
sandersaares, line 1: The PR description says "harden shutdown progression, drop validation, panic containment, wake locking, and timeout documentation based on review feedback" but this part of the change is impossible to review as the entire crate is new in the diff - hard to know what changed. It would be better reviewable if this stuff were a separate PR. That may be too much hassle, so at least a description of this hardening in the PR description would be helpful to facilitate a review.

crates/arty_executor/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+#![cfg_attr(coverage_nightly, feature(coverage_attribute))]
+
+//! Async task executor for the Arty Runtime.
+//!
+//! The purpose of the executor is to ensure that async tasks registered with the executor make
+//! progress, quickly and efficiently reacting to `await` statements that complete.
+//!
+//! The executor is a building block for an application runtime that provides foundational
+//! capabilities like task execution, multithreaded task management, I/O, timers, and more.
+//! Application logic may encounter [`JoinHandle`]s but otherwise has no direct interaction
```
sandersaares, line 15: The readme generator is stupid and thinks this link is for the `JoinHandle` crate. Suggested workaround attempt (unverified):

```rust
//! [`JoinHandle`]: crate::JoinHandle
```

https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7765826

Cargo.toml
sandersaares, line 53: https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7765788

crates/arty_executor/benches/ae_basic_operations.rs
martintmk: I already ran the benchmarks, He are the results of old vs new: (minimal impact)

┌──────────────────┬────────┬────────┐
│ Scenario         │   Arty │    SDK │
├──────────────────┼────────┼────────┤
│ cycle_empty      │  6,873 │  6,726 │
├──────────────────┼────────┼────────┤
│ task_add         │  5,254 │  5,254 │
├──────────────────┼────────┼────────┤
│ cycle_one_ready  │ 15,493 │ 15,387 │
├──────────────────┼────────┼────────┤
│ poll_completed   │     78 │     78 │
├──────────────────┼────────┼────────┤
│ yield_round_trip │ 31,160 │ 31,001 │
└──────────────────┴────────┴────────┘

crates/arty_executor/src/lib.rs
martintmk: (changes are minimal, but let's also mention them in the PR)

crates/arty_executor/src/cycle_outcome.rs
martintmk: &#x1F916; Agreed - addressing this now.

---

crates/arty_executor/src/executor.rs
martintmk: &#x1F916; Agreed - addressing this now.

---

crates/arty_executor/src/executor.rs
martintmk: &#x1F916; Agreed - addressing this now.

---

crates/arty_executor/src/executor_core.rs
martintmk: &#x1F916; Agreed - addressing this now.

---

crates/arty_executor/src/executor_core.rs
martintmk: &#x1F916; Agreed - addressing this now.

---

crates/arty_executor/src/task.rs
martintmk: &#x1F916; Agreed - addressing this now.

---

crates/arty_executor/src/executor_core.rs
martintmk: &#x1F916; Agreed - addressing this now.

---

crates/arty_executor/benches/ae_basic_operations.rs
martintmk: &#x1F916; Agreed - addressing this now.

---

crates/arty_executor/examples/ae_x_multimillionaire.rs
martintmk: &#x1F916; Agreed - addressing this now.

---

crates/arty_executor/src/testing/functions.rs
martintmk: &#x1F916; Agreed - addressing this now.

---

crates/arty_executor/src/lib.rs
martintmk: &#x1F916; Agreed - addressing this now.

---

crates/arty_executor/src/cycle_outcome.rs
martintmk: &#x1F916; Fixed: the shutdown outcome docs now match the implementation - additional cycles are harmless and keep returning `Shutdown`.

---

crates/arty_executor/src/executor.rs
martintmk: &#x1F916; Fixed: corrected `thread_static!` to `thread_local!`.

---

crates/arty_executor/src/executor.rs
martintmk: &#x1F916; Fixed: removed the duplicated "with".

---

crates/arty_executor/src/executor_core.rs
martintmk: &#x1F916; Fixed: corrected "is is dropped" to "it is dropped".

---

crates/arty_executor/src/executor_core.rs
martintmk: &#x1F916; Fixed: corrected "try report" to "try to report".

---

crates/arty_executor/src/task.rs
martintmk: &#x1F916; Fixed: removed the duplicated "the".

---

crates/arty_executor/src/executor_core.rs
martintmk: &#x1F916; Fixed: corrected all six repeated "on the the thread" SAFETY comments.

---

crates/arty_executor/benches/ae_basic_operations.rs
martintmk: &#x1F916; Fixed: added the Arty versus SDK Callgrind comparison and percentage deltas to the PR description.

---

crates/arty_executor/examples/ae_x_multimillionaire.rs
martintmk: &#x1F916; Fixed: documented that `scripts/run-examples.rs` sets `IS_TESTING` and that GitHub Actions runs examples through that script.

---

crates/arty_executor/src/testing/functions.rs
martintmk: &#x1F916; I kept this local because `testing_aids` is restricted to dev-dependencies, while this helper is compiled into the library through the public `test-util` feature. Moving the configuration would require making `testing_aids` a normal optional dependency, which would violate that dependency boundary.

---

crates/arty_executor/src/lib.rs
martintmk: &#x1F916; Fixed: expanded the PR description with the specific shutdown, drop-validation, panic-containment, wake-locking, synchronization, and pointer-width hardening changes.

---

crates/arty_executor/src/testing/functions.rs
martintmk: &#x1F916; Thanks for the feedback. This will be addressed in a follow-up.

---

crates/arty_executor/src/lib.rs
martintmk: &#x1F916; Thanks for the feedback. This will be addressed in a follow-up.

---

Cargo.toml
martintmk: &#x1F916; Thanks for the feedback. This will be addressed in a follow-up.

---

crates/arty_executor/src/join_handle.rs
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use std::marker::PhantomData;
+use std::pin::Pin;
+use std::task;
```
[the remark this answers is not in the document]
martintmk: &#x1F916; Thanks for the feedback. This will be addressed in a follow-up.

---

crates/arty_executor/src/task.rs
```diff
@@ -0,0 +1,581 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use std::any::Any;
+use std::cell::UnsafeCell;
+use std::marker::{PhantomData, PhantomPinned};
+use std::panic::{AssertUnwindSafe, catch_unwind};
+use std::pin::Pin;
+use std::task;
```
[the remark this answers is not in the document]
martintmk: &#x1F916; Thanks for the feedback. This will be addressed in a follow-up.

---

crates/arty_executor/src/testing/test_subject_future.rs
```diff
@@ -0,0 +1,81 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use std::cell::{Cell, RefCell};
+use std::pin::Pin;
+use std::rc::Rc;
+use std::task::{self, Waker};
```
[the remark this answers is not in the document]
martintmk: &#x1F916; Thanks for the feedback. This will be addressed in a follow-up.

---

crates/arty_executor/src/task.rs
```diff
@@ -0,0 +1,581 @@
...
+    #[cfg(debug_assertions)]
+    fn inspect_waker_backtraces(&self, f: &mut dyn FnMut(&Backtrace));
+}
+
+#[cfg(test)]
+mod tests {
+    use std::future::Ready;
+    use std::pin::pin;
+    use std::rc::Rc;
+    use std::task::Waker;
```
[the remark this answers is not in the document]
martintmk: &#x1F916; Thanks for the feedback. This will be addressed in a follow-up.

---

crates/arty_executor/src/executor.rs
```diff
@@ -0,0 +1,722 @@
...
+    pub fn begin_shutdown(&self) {
+        self.core.begin_shutdown();
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::cell::{Cell, RefCell};
+    use std::future::poll_fn;
+    use std::pin::Pin;
```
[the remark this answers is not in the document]
martintmk: &#x1F916; Thanks for the feedback. This will be addressed in a follow-up.

---

crates/arty_executor/src/executor_core.rs
```diff
@@ -0,0 +1,783 @@
...
+        self.poll_active_tasks(&mut state_exclusive);
+
+        {
+            let mut state_reentrant = self.reentrancy_safe.borrow_mut();
+
+            self.drop_inert_tasks(&mut state_exclusive, &mut state_reentrant);
+
+            let outcome = if self.evaluate_shutdown_completion(&state_exclusive, &state_reentrant) {
+                CycleOutcome::Shutdown
+            } else if state_reentrant.shutdown_deadline.is_some() || self.has_work_to_do(&state_exclusive, &state_reentrant) {
```
sandersaares: The change here (adding `state_reentrant.shutdown_deadline.is_some() ||` to the condition) is invalid. It is not OK to return `Continue` merely because we are in shutdown. `Continue` means the executor has more work to do and merely being in a shutting-down state is not grounds to signal that. Please revert to `else if self.has_work_to_do(..) {` as it was originally (and related test/tests).

If there is some API contract issue that creates the impression this change is justified, the API contract should be revised to remove that impression (but this does not need to be in this PR).

martintmk: agree, let's revert this change

martintmk: &#x1F916; Agreed - reverting the shutdown-only `Continue` behavior and updating the related test now.

---

martintmk: &#x1F916; Fixed: restored `Suspend` when shutdown has no immediately actionable work, reverted the related test expectation, and updated the PR description.

---

sandersaares (approved):

Vaiz (approved):

psandana (approved):

PR #650, opened 2026-08-07
fix(multitude): gate over-aligned doctest behind utc_backend cfg

kate-shine:
## Problem

`cargo test -p multitude --doc` fails on codegen backends that support a lower maximum type alignment than the default one.

The doctest on `AllocError::is_alignment_too_large` has to declare an over-aligned type in order to demonstrate the error it documents:

```rust
#[repr(align(32768))]
struct OverAligned;

let arena = multitude::Arena::new();
let Some(error) = arena.try_alloc(OverAligned).err() else {
    panic!("over-aligned values must be rejected");
};
assert!(error.is_alignment_too_large());
```

On a backend that caps alignment below that value, the type cannot be compiled at all, so the doctest fails.

#501 gated the equivalent *integration* tests behind `cfg(utc_backend)` and registered that cfg in the workspace `check-cfg` list, but this doctest was missed.

## Fix

Wrap the body in the hidden `#[cfg(...)]` shim documented in `AGENTS-feature-gated-doctests.md` (Pattern C — no `?`, no user `main`):

```rust
/// # fn main() {
/// # #[cfg(not(utc_backend))] {
/// ... existing body, unchanged ...
/// # }
/// # }
```

The `#` prefix keeps the shim out of the rendered documentation, so the published example is unchanged.

## Note for reviewers: `RUSTDOCFLAGS` is required

Doctests are compiled by `rustdoc`, and cargo passes `RUSTFLAGS` to `rustc` but **not** to `rustdoc`. Enabling the gate for doctests therefore needs both:

```
RUSTFLAGS=--cfg utc_backend
RUSTDOCFLAGS=--cfg utc_backend
```

Verified: with `RUSTFLAGS` alone the gate compiles in but the doctest still fails, because rustdoc never sees the cfg. Anything driving this build (for example a publish pipeline that currently sets only `RUSTFLAGS`) needs `RUSTDOCFLAGS` as well, otherwise this fix will appear to have no effect.

## Verification

- Doctests **with** the gate set (both env vars): **301 passed, 0 failed** — previously 1 failed.
- Doctests **without** the gate, on a normal toolchain: **301 passed, 0 failed** — the example still genuinely compiles and asserts, it is not silently skipped.
- `cargo clippy -p multitude --profile dev --all-targets --all-features --locked` — clean, zero warnings.
- `just package=multitude format-check` — clean.
- `just spellcheck` — clean.

Confirmed separately that the `multitude` **library** builds cleanly on such a backend and its 46 unit tests pass. The library declares no over-aligned types; alignment is handled at runtime via `align_of::<T>()`. Only test and doc code that must *construct* an over-aligned type is affected.

AB#7707893

### DIFF
--- a/crates/multitude/src/error.rs
+++ b/crates/multitude/src/error.rs
@@ -95,6 +95,8 @@ impl AllocError {
     /// succeed, regardless of how much memory is available.
     ///
     /// ```
+    /// # fn main() {
+    /// # #[cfg(not(utc_backend))] {
     /// #[repr(align(32768))]
     /// struct OverAligned;
     ///
@@ -103,6 +105,8 @@ impl AllocError {
     ///     panic!("over-aligned values must be rejected");
     /// };
     /// assert!(error.is_alignment_too_large());
+    /// # }
+    /// # }
     /// ```
     #[must_use]
     pub fn is_alignment_too_large(self) -> bool {


sgalkin (approved):

rust-toolchain.toml
```diff
@@ -0,0 +1,28 @@
...
+# `utc_backend` cfg, so the pin below is what keeps `git clone && cargo test`
+# working out of the box.
+#
+# CI does not rely on this pin: `actions-rust-lang/setup-rust-toolchain` exports
+# `RUSTUP_TOOLCHAIN`, which takes precedence, so the MSRV and nightly jobs still
+# get the toolchain they ask for.
+
+[toolchain]
+# This needs to match RUST_LATEST in constants.env.
+channel = "1.96.1"
```
ralfbiedert: 🤖 This new pin is required to match `RUST_LATEST`, but `scripts/update_rust_toolchain.ps1` only reads and rewrites `constants.env`. The next stable update will advance CI while leaving clean local checkouts on the old toolchain. Please update `rust-toolchain.toml` atomically in that script and adjust its README description.

martintmk (approved):

ralfbiedert (approved):
🤖 Approved.

ralfbiedert (approved):
🤖 Reviewed exact head d25a7f5. The cfg polarity is correct, normal backends continue executing the doctest, and the UTC backend avoids compiling the unsupported over-aligned declaration. I found no blocking correctness, performance, or unsafe-context issues.

sandersaares (approved):

geeknoid (commented):
Static-only five-pass review of exact head `d25a7f5ca33ddeafa2231c486b5dc282a192e5bd`; nothing was executed.

I found no inline defect in the isolated four-line change. The cfg polarity is correct, the hidden wrapper follows the repository’s Pattern C, the ordinary-backend path still contains the substantive assertion, and the capped-backend path removes the unsupported `#[repr(align(32768))]` declaration. The cfg is also registered in the workspace `check-cfg` list.

This PR should not be merged or rebased, however: it is currently `CONFLICTING`/`DIRTY` and has been superseded on `main`. An equivalent hidden-cfg fix landed in [`8c179d28`](https://github.com/microsoft/oxidizer/commit/8c179d28f17d87d0379ccab2110249c74e7c7e26), the cfg was subsequently generalized, and current `main` has removed this non-portable doctest in favor of alignment-guard tests ([current source](https://github.com/microsoft/oxidizer/blob/7a00180f9911c3158f8837bbd3f157f87f7f17e5/crates/multitude/src/error.rs#L92-L104)). Closing this stale PR is cleaner than reconstructing it atop the replacement design.

One historical testing gap was identified but not emitted inline: repository CI at this PR’s base runs doctests only in ordinary configurations and does not set both `RUSTFLAGS` and `RUSTDOCFLAGS` for `utc_backend`, so removal or inversion of the shim would not have a persistent repository-owned regression gate. That implementation has since been replaced on `main`.

PR #651, opened 2026-08-07
fix(cachet): report rejected insertions accurately

codingsnoop:
## Summary
- make `insert` return `InsertOutcome` directly across `CacheTier`, `Cache`, service extensions, and adapters; this is an intentional breaking API change
- avoid stamping computed entries when every configured tier rejects insertion
- propagate aggregate acceptance through fallback, transform, dynamic, wrapper, and service layers
- report measured latency when an inner tier rejects insertion; immediate local-policy rejection remains zero-duration
- document that composite outcomes answer whether any tier accepted the write, while `CacheEventHandler` provides per-tier detail

## Versioning
- `cachet_tier`: 0.4.0
- `cachet_service`: 0.4.0
- `cachet_memory`: 0.6.1
- `cachet`: 0.12.0

## Validation
- affected tests and doctests
- Clippy across affected targets with warnings denied
- dual-configuration coverage gate at 100% for affected packages
- targeted insertion and telemetry mutants

Azure DevOps bug: 7438084


crates/cachet/src/transform/tier.rs
```diff
@@ -71,9 +71,13 @@ where
     }
 
     async fn insert(&self, key: K, entry: CacheEntry<V>) -> Result<(), Error> {
+        self.insert_with_outcome(key, entry).await.map(drop)
+    }
+
+    async fn insert_with_outcome(&self, key: K, entry: CacheEntry<V>) -> Result<InsertOutcome, Error> {
```
schgoo, line 77: I think if this is really needed, we should just make the breaking change now, rather than adding this secondary API. Alternatively (maybe preferably), if this is for telemetry purposes, we can surface it in the telemetry handler.

ralfbiedert:
🤖 CacheResponse::Insert changes from a unit variant to Insert(InsertOutcome), which breaks existing service implementations and pattern matches. This PR itself does not bump cachet_service from 0.2.9. PR #652 separately prepares cachet_service 0.3.0, but its current head does not contain this change, so merge/release ordering is important: please either incorporate this change before 0.3.0 is published or arrange a subsequent breaking minor bump.

crates/cachet/src/transform/tier.rs
sgalkin: Could we make the outcome-aware operation the single primitive in this breaking change - have `insert` return `InsertOutcome` across `CacheTier`, `CacheServiceExt`, and the adapters - and remove the optimistic compatibility default? `CacheResponse::Insert` is already source-breaking, so the parallel lossy method adds surface area and lets rejecting implementations accidentally report `Accepted`.

🤖 Authored by Copilot.

crates/cachet/src/cache.rs
```diff
@@ -292,6 +312,17 @@ where
...
+    async fn insert_computed_entry(&self, key: &K, entry: CacheEntry<V>) -> Result<CacheEntry<V>, Error> {
+        let original = entry;
+        let mut candidate = original.clone();
+        candidate.ensure_cached_at(self.clock.system_time());
+
+        match self.storage.insert_with_outcome(key.clone(), candidate.clone()).await? {
+            InsertOutcome::Accepted => Ok(candidate),
+            InsertOutcome::Rejected => Ok(original),
+        }
+    }
```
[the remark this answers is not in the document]
sgalkin: Confirmed: `CacheEntry<V>` derives `Clone`, so both calls duplicate `V`. Could we retain the original entry, clone and stamp only the candidate sent to storage, then stamp the original with the same captured timestamp only on `Accepted`? That preserves rejection semantics with one value clone instead of two.

```rust
let cached_at = self.clock.system_time();
let mut candidate = entry.clone();
candidate.ensure_cached_at(cached_at);

match self.storage.insert_with_outcome(key.clone(), candidate).await? {
    InsertOutcome::Accepted => {
        let mut entry = entry;
        entry.ensure_cached_at(cached_at);
        Ok(entry)
    }
    InsertOutcome::Rejected => Ok(entry),
}
```

🤖 Authored by Copilot.

crates/cachet_tier/src/tier.rs
```diff
@@ -11,25 +11,28 @@ use std::future::Future;
...
-/// written. It may return `None` (e.g., if the entry was evicted or invalidated)
-/// or a newer value, but never a stale one that predates the most recent write.
-///
-/// This guarantee is scoped to a **single instance** - if the same backing store
-/// is accessed through multiple `CacheTier` instances (e.g., separate processes
-/// connected to the same Redis cluster), replication lag or network partitions
-/// may cause one instance to observe stale data written through another. The
-/// monotonicity guarantee only applies to reads and writes through the same
-/// Rust object.
+/// [`Accepted`](InsertOutcome::Accepted) means that this tier, or at least one
```
sgalkin, line 30: Confirmed: the composite behavior already existed, but this change resolves that mismatch by weakening the base `CacheTier` contract from same-instance read-after-write monotonicity to allowing stale reads after `Accepted`. Could we preserve monotonicity for ordinary tiers and represent mixed composite acceptance explicitly (for example, `PartiallyAccepted`), scoping the weaker guarantee to composites?

🤖 Authored by Copilot.

crates/cachet/src/wrapper.rs
```diff
@@ -139,17 +139,22 @@ where
...
+            return Ok(InsertOutcome::Rejected);
         }
 
         let watch = self.clock.stopwatch();
-        let result = self.inner.insert(key, entry).await;
+        let result = self.inner.insert_with_outcome(key, entry).await;
         match &result {
-            Ok(()) => self.telemetry.record_inserted(self.name, watch.elapsed(), self.fallback),
+            Ok(InsertOutcome::Accepted) => self.telemetry.record_inserted(self.name, watch.elapsed(), self.fallback),
+            Ok(InsertOutcome::Rejected) => self.telemetry.record_insert_rejected(self.name, self.fallback),
```
sgalkin: Could we keep `record_insert_rejected` for immediate local-policy rejection and add a timed downstream-rejection path here? This branch runs after awaiting `inner`, but the helper emits `Duration::ZERO`, so the public `CacheTierEvent::duration` loses the actual tier latency.

🤖 Authored by Copilot.

crates/cachet/src/transform/tier.rs
schgoo: Yes, that's what I'm suggesting. There's also the less breaking change of adding this information to the insert telemetry event.

crates/cachet_tier/src/tier.rs
codingsnoop: Good point, but I don't think PartiallyAccepted is warranted here. Accepted describes write acceptance, not read-after-write visibility, which was never guaranteed due to policy rejection and immediate eviction. For composites, it explicitly means at least one tier accepted the write. Adding another public outcome would significantly expand the API without affecting this fix, so I'd keep the documented and tested composite semantics and revisit per-tier outcomes if callers need them.

crates/cachet/src/fallback.rs
```diff
@@ -138,13 +138,19 @@ where
...
         );
-        primary_result?;
-        fallback_result
+        let primary_outcome = primary_result?;
+        let fallback_outcome = fallback_result?;
+
+        if primary_outcome == InsertOutcome::Accepted || fallback_outcome == InsertOutcome::Accepted {
+            Ok(InsertOutcome::Accepted)
+        } else {
+            Ok(InsertOutcome::Rejected)
```
schgoo, line 152: This signal is a little muddy. Does it still solve for your use case if you can't distinguish between tiers? I guess it might if you only have a single tier structure, but for multi-tier caches this is less helpful. Maybe `InsertOutcome` contains a list of `Accepted`/`Rejected`?

crates/cachet/src/wrapper.rs
```diff
@@ -182,9 +186,76 @@ where
...
+    use std::sync::{Arc, Mutex};
+
     use cachet_tier::MockCache;
+    use tick::ClockControl;
 
     use super::*;
+    use crate::{CacheEventHandler, CacheOperationEvent, CacheTierEvent};
+
+    #[derive(Debug)]
+    struct RejectingCache;
```
schgoo: Could the existing MockCache be used instead?

schgoo: Potentially it could be extended to support DelayedRejectingCache, as well.

crates/cachet/CHANGELOG.md
```diff
@@ -1,5 +1,71 @@
 # Changelog
 
+## [0.12.0] - 2026-08-10
```
schgoo: Most of these are not relevant for the cachet changelog

crates/cachet_memory/CHANGELOG.md
```diff
@@ -1,5 +1,59 @@
 # Changelog
 
+## [0.6.1] - 2026-08-10
```
schgoo: Same here

crates/cachet_service/CHANGELOG.md
```diff
@@ -1,5 +1,48 @@
 # Changelog
 
+## [0.4.0] - 2026-08-10
```
schgoo: And here

crates/cachet_tier/CHANGELOG.md
```diff
@@ -1,5 +1,46 @@
 # Changelog
 
+## [0.4.0] - 2026-08-10
```
schgoo: Also here

crates/cachet/src/transform/tier.rs
codingsnoop: I selected the breaking option and updated the PR description accordingly. Telemetry alone cannot fix the computed-entry bug because `insert_computed_entry` needs the rejection result synchronously to decide whether to stamp `cached_at`; the direct `InsertOutcome` return provides that control-flow signal.

crates/cachet/src/fallback.rs
codingsnoop: Yes, the aggregate signal solves the functional use case: the computed helper only needs to know whether any tier accepted the entry before stamping `cached_at`. Returning a per-tier list would allocate, expose composite topology, and become ambiguous for nested composites. Commit `4faf7765` clarifies that the return value is aggregate and points callers to `CacheEventHandler` for per-tier outcomes.

crates/cachet/src/wrapper.rs
codingsnoop: Updated in `4faf7765`: both bespoke rejecting cache implementations were removed. The tests now compose the existing `MockCache` with an inner `CacheWrapper` and rejecting policy, including the controlled-clock delay needed to verify downstream rejection latency.

crates/cachet/CHANGELOG.md
codingsnoop: Removed the generated changelog section in `4faf7765`; release tooling can regenerate the relevant entry at release time.

crates/cachet_memory/CHANGELOG.md
codingsnoop: Removed in `4faf7765` along with the other unrelated generated changelog sections.

crates/cachet_service/CHANGELOG.md
codingsnoop: Removed in `4faf7765` along with the other unrelated generated changelog sections.

crates/cachet_tier/CHANGELOG.md
codingsnoop: Removed in `4faf7765` along with the other unrelated generated changelog sections.

schgoo (approved):

Vaiz (approved):

PR #634, opened 2026-08-03
fix(ohno)!: improve `#[display]` and `#[error]` diagnostics

Vaiz:
🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

## What this changes

The macro reported several of its own mistakes through the code it generates. Users saw errors about fields they never declared, spanned at a derive rather than at anything they wrote. This PR moves that reporting into the macro, adds the validation it needs, and pins the spans with a trybuild UI suite.

Implicit `self` scoping stays the syntax model for `#[display(...)]` arguments. The new validation rejects some inputs that the macro previously accepted and expanded into wrong or unbuildable code.

## Details

- `#[display(...)]` arguments are validated before expansion. The macro walks each argument to its leftmost term and checks that term names a field. A misspelled root, a `self.` prefix, or a root that cannot follow a dot is reported at the term itself.
- The `OhnoCore` field that `#[ohno::error]` injects is marked with a reserved doc string, not with `#[error]`. `#[error]` therefore takes no arguments at all, and the injected field is excluded from every "available fields" list. The marker carries a nonce, because a doc string can only fail to match and a marker that read as prose would let an ordinary doc comment hide the field it documents.
- A core field the user declares is data. It is passed to the generated constructors, appears in `Debug`, and can be interpolated. It is never read for `source()`, the backtrace, or enrichment.
- `#[ohno::error]` rejects a field marked `#[error]`. The attribute always generates the error representation from the field it injects. The diagnostic names both remedies.
- A second `#[error]` marker is reported instead of being resolved by declaration order.
- Tuple indices resolve. `{0}` works, and the index check shares its source of truth with the available-fields list.
- Raw identifiers keep their prefix. A field named `r#type` is rebuilt with `Ident::new_raw` instead of panicking, and is echoed back as `r#type`.
- Multi-token diagnostics use a new `bail_spanned!` helper built on `Error::new_spanned`. A node span collapses to its first token where `Span::join` is absent, so one snapshot cannot match both toolchains.
- Crate docs and README gain a "Format Arguments" section with an accepted/rejected table.

## Effects

- A positional tuple index that addressed the injected core is now rejected. On an error with one declared tuple field, `#[display("bad: {}", 1)]` expanded to `&self.1`, compiled, and formatted the error's own chain. It now fails as an unknown field.
- `#[error(generated)]` is rejected. That spelling was the macro's own handoff and is no longer part of the attribute's grammar.
[...]

PR #652, opened 2026-08-10
chore: release bytesbuf v0.8.0 and 27 additional packages

sandersaares:
[Copilot speaking]

## Motivation

Five unrelated breaking changes have accumulated since the last release and none of them have shipped. Because each affected crate is named in the `allowed_external_types` allowlist of its dependents, every dependent that re-exposes one of them must also go out as a breaking release for the published graph to stay coherent.

## Changes

Six crates break on their own merit:

- **`bytesbuf` 0.8.0** and **`thread_aware` 0.9.0** became `no_std`. Their hosted API moved behind a new `std` feature. `std` is on by default, so `default-features = true` consumers are unaffected, but builds using `default-features = false` must now add `std` to keep `bytesbuf::mem::GlobalPool`, `BytesBufWriter` and the `std::io` adapters, or `thread_aware`'s `Arc`, `PerCore`, `PerNuma`, `PerProcess` and `storage`. `thread_aware`'s `threads` feature now implies `std`.
- **`multitude` 0.8.0** gained trait-object coercion: `Coercion`, the `coerce!` macro, `Box::unsize`, `Rc::unsize`, `Rc::unsize_pin`, `Arc::unsize` and `Arc::unsize_pin` convert an owner into a trait-object owner without moving or reallocating the value. Vtable metadata now lives in the handle rather than an allocation prefix, so trait-object handles are one word wider and the former "8 bytes for every `T`" guarantee holds only for sized values, `str` and `[T]`. The pointee bound on `Box`, `Rc`, `Arc`, `Cow` and the `Arena::alloc_dst_*` families tightened from `ptr_meta::Pointee` to the sealed `SmartPointerPointee`.
- **`tick` 0.5.0** validates format values at construction and drops cross-format conversions, and adds the EcmaScript timestamp format.
- **`ohno` 0.4.0** and **`ohno_macros` 0.4.0** tightened the `#[display]` and `#[error]` diagnostics. `#[display(...)]` arguments are validated before expansion, `#[error(generated)]` is rejected, combining `#[ohno::error]` with an `#[error]` marker is now a hard error, and a tuple index that addresses the injected core is refused. Inputs that previously expanded now fail to compile. `ohno`'s own library API is unchanged, but it re-exports the macros, so its effective contract moves with them.

**`templated_uri` 0.4.0** and **`recoverable` 0.2.0** ship the documentation-module removal from #635, which took `_documentation` out of the normal build in four crates. `fetch` and `http_extensions` carry the same change, so releasing all four together keeps one change under one classification; deferring the other two would have left an unlabelled public-API removal waiting to reach consumers in a later patch release.

**`templated_uri_macros` 0.4.0** and **`templated_uri_macros_impl` 0.4.0** move in lockstep with the runtime crate they generate code against. `templated_uri` 0.3.6 introduced a version floor on the macro crate precisely so Cargo cannot unify macros across incompatible runtime lines; leaving the macros on 0.3 while `templated_uri` moves to 0.4 would defeat that, and the published `_impl` crate would still resolve `ohno` 0.3, giving consumers two incompatible `ohno` copies.

`plurality` 0.2.1 is additive only: unwind-safety impls for the pooled handles, plus a `coerce!` fix for trait objects that name an enclosing generic type parameter. `multitude_macros` and `multitude_macros_impl` 0.1.1 pick up the generated README switch to absolute asset URLs so crates.io renders them.

The remaining increments are pure cascade — a dependency bump and nothing else.

```
bytesbuf 0.8.0 ─┬─> bytesbuf_io 0.8.0
                ├─> multitude 0.8.0
                ├─> cachet 0.11.0
                ├─> http_extensions 0.9.0 ─┬─> fetch_hyper 0.6.0 ─┐
                ├─> fetch_azure 0.5.0      └─> seatbelt_http 0.7.0┤
                └─> fetch 0.15.0 <─────────────────────────────────┘

thread_aware 0.9.0 ─┬─> bytesbuf 0.8.0
                    ├─> anyspawn 0.7.0 ──> anyspawn_azure 0.2.0
                    ├─> cachet_memory 0.6.0
                    ├─> seatbelt 0.7.0
                    ├─> uniflight 0.4.0
                    └─> tick 0.5.0

ohno 0.4.0 ─┬─> ohno_macros 0.4.0 (dependency; bumped first)
            ├─> fetch_tls 0.3.0 ──> fetch 0.15.0, fetch_hyper 0.6.0
            ├─> templated_uri_macros_impl 0.4.0 ──> templated_uri_macros 0.4.0 ─┐
            ├─> bytesbuf_io, cachet_memory, cachet_tier, fetch, fetch_hyper     │
            ├─> http_extensions, recoverable, seatbelt, seatbelt_http, tick     │
            └─> templated_uri 0.4.0 <───────────────────────────────────────────┘

templated_uri 0.4.0 ──> fetch, fetch_hyper, http_extensions, seatbelt_http

recoverable 0.2.0 ─┬─> seatbelt 0.7.0
                   ├─> http_extensions 0.9.0
                   └─> cachet_tier 0.3.0 ─┬─> cachet_memory 0.6.0
                                          ├─> cachet_service 0.3.0 ──> cachet 0.11.0
                                          └─> cachet 0.11.0

plurality 0.2.1 ──> bytesbuf 0.8.0        (compatible; no cascade)
```

## Minimum supported Rust version

The workspace floor moved to 1.93.1 in #629, pulled in by `serial_test` 4.0.1. Every released changelog now records it. The requirement is inherited from `[workspace.package]`, so it applies to all crates here including the patch-level releases — a consumer on 1.93.0 cannot resolve them. No consumer-visible dependency needs 1.93.1; the floor comes from a test-only crate.

## Publishing order

`bytesbuf` must be published before `multitude`. `multitude`'s `bytesbuf` feature enables `bytesbuf/std`, which does not exist in the published `bytesbuf` 0.7.0, so `multitude` is unpublishable until `bytesbuf` 0.8.0 is on crates.io. Likewise `ohno_macros` precedes `ohno`, and `templated_uri_macros_impl` precedes `templated_uri_macros`.

## Notes for the reviewer

`crates/plurality/CHANGELOG.md` carried an `## [Unreleased]` heading whose contents had in fact shipped. It is retitled to `## [0.2.0] - 2026-07-24`, matching the `plurality-v0.2.0` tag, and the new 0.2.1 section sits above it.

`recoverable` 0.1.7, `ohno_macros` 0.3.5 and the crates listed below are published on crates.io but have no Git tag. Release tooling that derives its changelog range from tags will pick up a stale baseline for them; both needed their baseline established from the version-bump commit instead. Worth reconciling separately.

`Cargo.lock` legitimately contains a second `ohno` 0.3.5 entry alongside the workspace's 0.4.0. It is pulled from crates.io by the third-party `cpulist` crate and predates this change; no workspace member resolves to it.

`thread_aware_macros` 0.7.5, `thread_aware_macros_impl` 0.7.4, `data_privacy_core` 0.1.2, `data_privacy_macros` 0.10.3, `data_privacy_macros_impl` 0.10.3 and `fetch_options` 0.2.3 are left untouched.


sandersaares:
[Copilot speaking]

Manual proc-macro SemVer review for `multitude_macros` 0.1.0 -> 0.1.1: **compatible**.

The crate has no source changes since its baseline commit `8ae7af6`. The only differences are the `[package] version` bump and the regenerated `README.md`:

```
crates/multitude_macros/Cargo.toml      | 2 +-
crates/multitude_macros/README.md       | 8 ++++----
crates/multitude_macros_impl/Cargo.toml | 2 +-
crates/multitude_macros_impl/README.md  | 8 ++++----
```

Exported macro names, accepted input syntax, emitted diagnostics and generated output are therefore byte-identical to 0.1.0. The release exists solely so the published README carries absolute asset URLs that crates.io can render.

crates/plurality/CHANGELOG.md
```diff
@@ -5,7 +5,20 @@ All notable changes to this project are documented here.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [0.2.1] - 2026-08-09
```
wukchung, line 8: 🤖 **This release silently raises the MSRV from 1.93 to 1.93.1, and no changelog records it.**

`rust-version` moved `1.93` → `1.93.1` in #629. It lives in `[workspace.package]`, so all 19 crates here publish declaring `rust-version = "1.93.1"` and Cargo hard-refuses them on a 1.93.0 toolchain. Verified against the previous tags — `plurality-v0.2.0`, `bytesbuf-v0.7.0`, `multitude-v0.7.1` and `tick-v0.4.0` all declare `rust-version = "1.93"`.

It bites hardest on the patch/minor releases that otherwise look safe to take: `plurality` 0.2.1, `multitude_macros` 0.1.1 and `multitude_macros_impl` 0.1.1 are advertised as additive or README-only, but a consumer on Rust 1.93.0 cannot resolve them.

Worth noting where the floor comes from: `serial_test` is referenced from exactly two manifests — `crates/cachet/Cargo.toml:82` (under `[dev-dependencies]`) and `crates/testing_aids/Cargo.toml:40` (`publish = false`). So no consumer-visible dependency needs 1.93.1; the floor is imposed on downstream users by a test-only crate.

Repo precedent is to record this explicitly — see `crates/multitude/CHANGELOG.md:118` ("release all packages for MSRV increment (#463)") and `:134` ("bump MSRV to 1.91 and refresh dependencies (#457)"). Prior MSRV increments were also released workspace-wide; this one covers 19 of ~44 publishable crates, so the rest stay on crates.io advertising 1.93 while the source requires 1.93.1.

Suggest at minimum adding an MSRV line to the released crates' changelog entries; if the increment is meant to be workspace-wide, `docs/releasing.md:15-20` explicitly warns that the `-Changed` scan cannot surface it because the edit is to the workspace-level `Cargo.toml`, and recommends `-All`.

crates/fetch/CHANGELOG.md
```diff
@@ -1,5 +1,22 @@
...
 
+## [0.15.0] - 2026-08-09
+
+- 🔧 Maintenance
+
+  - Now requires `0.3.6` of `templated_uri`
+
+- ⚠️ Breaking
+
+  - hide documentation modules from public API ([#635](https://github.com/microsoft/oxidizer/pull/635))
```
wukchung, line 11: 🤖 **#635 is shipped as breaking here, but stranded unreleased in two other crates it touched.**

#635 (`5638f2bb`) added `#[cfg(any(doc, test))]` above `pub mod _documentation;` in four crates: `fetch`, `http_extensions`, `recoverable` and `templated_uri`. This release ships it in the first two, both labelling it ⚠️ Breaking (here and `crates/http_extensions/CHANGELOG.md:11`). The other two are not in the release set:

- **`templated_uri`** — `crates/templated_uri/src/lib.rs:188`. Last tag `templated_uri-v0.3.6`, published to crates.io 2026-07-31; #635 landed 2026-08-04, so it is not in any published version. `templated_uri` is also absent from the "left untouched" list in the PR description, unlike the other deferred crates.
- **`recoverable`** — `crates/recoverable/src/lib.rs:81`. `recoverable` 0.1.7 was published 2026-06-29, again before #635.

The deferral is legal and nothing breaks today — `fetch` / `fetch_hyper` / `http_extensions` / `seatbelt_http` pin `templated_uri` `^0.3.6`, which resolves fine — so this is a release-plan question rather than a bug in the diff. The reason it seems worth raising: the *next* release of these crates now carries an unlabelled public-API removal. `recoverable` is the riskier one, since per the increment table in `docs/releasing.md` a breaking change on `0.1.7` needs `0.2.0`; if it goes out as `0.1.8`, every consumer on `^0.1` picks up the removal silently.

Either pull `templated_uri` (0.4.0) and `recoverable` (0.2.0) into this release, or add them to the deferred list so the next releaser classifies them correctly. Was leaving them out deliberate?

wukchung (commented):
🤖 Reviewed the release mechanics. The mechanical parts check out: for all 19 crates the `[package] version`, the root `[workspace.dependencies]` pin, `Cargo.lock` and the newest `CHANGELOG.md` heading agree; every `Now requires X of Y` line matches the real non-dev dependency set; the cascade over `bytesbuf` / `thread_aware` / `tick` is complete (`cachet_tier` and `layered` are correctly excluded — dev-dependency and patch-compatible respectively); and the `multitude` → `bytesbuf/std` publish-order claim holds (`crates/multitude/Cargo.toml`: `bytesbuf = ["dep:bytesbuf", "std", "bytesbuf/std"]`).

Two things about the *release set* rather than the diff itself, inline below.

crates/plurality/CHANGELOG.md
sandersaares: [Copilot speaking]

Addressed in `e7638b3f` — verified and recorded.

`rust-version` moved `1.93` → `1.93.1` in `9513482e` (#629), alongside the `serial_test` 3.5.0 → 4.0.1 bump whose own MSRV is 1.93.1. Confirmed against the previous tags (`plurality-v0.2.0`, `bytesbuf-v0.7.0`, `multitude-v0.7.1`, `tick-v0.4.0`, `thread_aware-v0.8.0` all declare `1.93`) and against crates.io, where the last published `templated_uri` 0.3.6 and `recoverable` 0.1.7 both report `rust_version=1.93`. The observation that the floor is imposed by a test-only crate also holds: `serial_test` is referenced only from `crates/cachet/Cargo.toml` under `[dev-dependencies]` and from `crates/testing_aids` (`publish = false`).

Every changelog in the release set now carries the requirement, matching each crate's existing format:

- generated-style crates get `- Now requires Rust ` + "`1.93.1`" + ` ([#629](https://github.com/microsoft/oxidizer/pull/629))` under `🔧 Maintenance`
- `thread_aware` gets an equivalent `### 🔧 Maintenance` entry
- `plurality` gets a Keep a Changelog `### Changed` entry

The PR description now has an MSRV section stating that the requirement is inherited from `[workspace.package]` and therefore applies to the patch-level releases too.

The increment is deliberately not being turned into a workspace-wide release; scope stays as-is.

crates/fetch/CHANGELOG.md
sandersaares: [Copilot speaking]

Addressed in `e7638b3f` — both crates pulled into the release set, along with two cascade crates the analysis surfaced.

The deferral was not deliberate. Confirmed the finding: `#[cfg(any(doc, test))]` sits above `pub mod _documentation;` at `crates/templated_uri/src/lib.rs:188` and `crates/recoverable/src/lib.rs:81`; #635 landed 2026-08-04 while `templated_uri` 0.3.6 was published 2026-07-31 and `recoverable` 0.1.7 on 2026-06-29, so the removal is in neither published version.

Released as breaking:

- `templated_uri` 0.3.6 → **0.4.0**
- `recoverable` 0.1.7 → **0.2.0**

Tracing the `recoverable` cascade found two further crates that were missing from the set, both of which expose the crate below them in their public API per their `allowed_external_types` allowlists:

- `cachet_tier` 0.2.7 → **0.3.0** — exposes `recoverable::Recovery` and `recoverable::RecoveryInfo`
- `cachet_service` 0.2.9 → **0.3.0** — exposes `cachet_tier::{entry::CacheEntry, error::Error, tier::CacheTier}`

The closure terminates at `cachet`, already in the set. Neither cascade crate has any source change since its tag, so their entries carry only the dependency bump.

Consequent edits to entries already in the set: the `templated_uri` requirement moved out of `🔧 Maintenance` into `⚠️ Breaking` at `0.4.0` for `fetch`, `fetch_hyper`, `http_extensions` and `seatbelt_http`; `recoverable` `0.2.0` added under `⚠️ Breaking` for `seatbelt` and `http_extensions`; `cachet_tier` `0.3.0` for `cachet` and `cachet_memory`, plus `cachet_service` `0.3.0` for `cachet`.

The set is now 23 crates. Version, workspace pin, `Cargo.lock` entry and newest changelog heading agree for all 23; `just clippy`, `just format`, `just spellcheck` and `just readme-check` all pass.

On the missing tag: there is no `recoverable-v0.1.7`, so the baseline was established from the version-bump commit `542a32c0` rather than from `git tag --list`. That crate and the other untagged-but-published ones are called out in the PR description as needing separate reconciliation.

sgalkin (approved):

wukchung (approved):

ralfbiedert:
🤖 crates/recoverable/CHANGELOG.md currently attributes REST-over-gRPC transcoding to recoverable 0.2.0. The referenced change modified the crate documentation but did not add that implementation to recoverable, so this would publish a misleading feature claim. Please remove or correct that changelog entry before release.

sandersaares:
[Copilot speaking]

Scope has grown since @sgalkin's and @wukchung's approvals — please re-review.

`origin/main` merged 257c5f72 (#634), which tightened the `#[display]` and `#[error]` diagnostics in `ohno_macros`. That is a breaking change to the proc-macro contract, and `ohno` re-exports those macros, so `ohno`'s effective contract breaks with it. Releasing it now avoids an immediate second breaking release of every crate that exposes `ohno`.

Five crates were added, taking the set from 23 to 28:

| Crate | From | To | Reason |
|---|---|---|---|
| `ohno` | 0.3.9 | 0.4.0 | breaking, own merit (#634) |
| `ohno_macros` | 0.3.5 | 0.4.0 | breaking, own merit (#634) |
| `fetch_tls` | 0.2.6 | 0.3.0 | cascade; exposes `ohno` in its public API |
| `templated_uri_macros_impl` | 0.3.0 | 0.4.0 | depends on `ohno`; also carries the unreleased `syn` 3.0.2 / `darling` 0.24.0 realignment |
| `templated_uri_macros` | 0.3.0 | 0.4.0 | keeps the macro line aligned with `templated_uri` 0.4.0 |

The macro crates move to 0.4.0 rather than taking a minimal patch bump because `templated_uri` 0.3.6 introduced a version floor on `templated_uri_macros` specifically to stop Cargo unifying macros across incompatible runtime lines. Leaving them on 0.3 would defeat that, and the published `_impl` crate would still resolve `ohno` 0.3, handing consumers two incompatible `ohno` copies.

Every other direct dependent of `ohno` was already releasing breaking in this set and gains a cascade bullet only.

Also in this push: the changelog files are normalised so every release heading is preceded by a blank line, and the ones I had written with CRLF are corrected to LF per `.gitattributes`.

sandersaares:
[Copilot speaking]

Confirmed and fixed in eaf5809b.

`5b39c892` (#600) touched exactly one file in each of the two crates:

```
crates/recoverable/README.md  | 10 +++++-----
crates/ohno_macros/README.md  |  8 ++++----
```

Both are pure asset/link rewrites from relative paths to absolute URLs, part of the packaging work that keeps LFS-tracked images out of published crates. No transcoding was added to either crate, so the `✨ Features` bullet has been dropped from both new sections rather than reworded — a regenerated README is not a library change and does not warrant a changelog entry on its own. Both crates are being released on other merit anyway (#635 for `recoverable`, #634 for `ohno_macros`).

I swept the remaining new sections for the same defect by resolving every referenced PR back to its commit and checking whether it touched anything beyond `README.md`/`CHANGELOG.md` in that crate. These two were the only occurrences.

Worth noting the root cause is systemic rather than local: the changelog generator attributes a commit's subject line to every crate whose directory that commit touched, so a workspace-wide README regeneration inherits whatever feature the commit was actually about. The same inaccurate bullet is already published in several 0.7.0/0.5.1-era entries. I have left the published history alone.

sandersaares:
[Copilot speaking]

Manual proc-macro SemVer review for the two macro crates added since the previous bot run. (`multitude_macros` 0.1.0 -> 0.1.1 was reviewed earlier in https://github.com/microsoft/oxidizer/pull/652#issuecomment-5236556684 and is unchanged since.)

## `ohno_macros` 0.3.5 -> 0.4.0 — **breaking**, bump is correct

Baseline `542a32c`. Source delta:

```
src/derive_error/attributes.rs        |   2 +
src/derive_error/display.rs           | 556 +++++++++++-------------
src/derive_error/field_detection.rs   | 431 ++++++++++++--------
src/derive_error/mod.rs               |   2 +-
src/enrich_err/mod.rs                 |   2 +-
src/error_type_attr.rs                | 147 ++++++-
src/lib.rs                            |   8 +-
src/utils.rs                          |  38 ++
```

The exported entry points are unchanged — `derive_error`, `enrich_err` and `error`, same names and same signatures. Nothing was added or removed from the macro surface.

The break is behavioural: input that previously expanded is now rejected at compile time. `257c5f72` (#634) added a trybuild suite under `crates/ohno/tests/ui/` that pins each newly-refused form, and the filenames enumerate the contract change precisely:

- `#[display]` arguments are validated before expansion rather than being passed through — unbalanced braces, unknown fields, raw identifiers naming unknown fields, binary-expression roots and other unsupported argument roots are now errors.
- `#[display]` positional arguments are implicitly scoped to `self`, so a `self.`-prefixed argument is now rejected. Code written as `#[display("{}", self.path.display())]` must become `#[display("{}", path.display())]`.
- A tuple index that addresses the injected core is refused.
- `#[error(...)]` with arguments is rejected, as is `#[error]` on more than one field.
- Combining `#[ohno::error]` with an `#[error]` marker is a hard error, and the marker name is reserved.
- With no `#[error]` marker the macro now requires a single field whose type is literally named `OhnoCore`. A core reached through a type alias or a renamed import must be marked explicitly.

Every one of these turns a previously-compiling crate into a failing one, so a major-line bump is required. `0.4.0` is correct.

## `templated_uri_macros` 0.3.0 -> 0.4.0 — **compatible contract**, intentional over-bump

Baseline `aae9c5e`. Complete file delta since:

```
crates/templated_uri_macros/CHANGELOG.md
crates/templated_uri_macros/Cargo.toml
```

No source or test file changed. The generated code and the accepted input are byte-for-byte identical to 0.3.0, so the macro contract itself is unbroken and `0.3.1` would have satisfied SemVer.

The major-line bump is deliberate. This crate is a thin re-export of `templated_uri_macros_impl`, which does move breaking: it depends on `ohno`, now 0.4.0. `templated_uri` 0.3.6 introduced a floor on this crate specifically to stop Cargo unifying macros across incompatible runtime lines. Had the macro crate stayed on 0.3, a consumer on `templated_uri` 0.4.0 could still resolve `templated_uri_macros` 0.3.0 -> `templated_uri_macros_impl` 0.3.0 -> `ohno` ^0.3.9 and end up with two incompatible `ohno` copies in one graph. Moving the whole macro line to 0.4 keeps that impossible.

Over-bumping is always SemVer-safe; the cost is that consumers pinning `templated_uri_macros` directly must update, which the floor in `templated_uri` already forces.

## Chain propagation

The 14 crates flagged `automated check ok; manual proc-macro chain review required` are all downstream of `ohno` being manually classified breaking. Each already carries a `Now requires 0.4.0 of ohno` bullet and a major-line bump in this PR, so the chain is consistent. No further action.

wukchung (approved):

Vaiz (approved):

PR #653, opened 2026-08-10
fix(ohno)!: document pub(crate) constructors, reject #[no_constructors] under #[ohno::error]

Vaiz:

Two related facts about the constructors `#[derive(Error)]` generates were undocumented, and the
second turned out to be a design problem rather than a documentation one.

## 1. The generated constructors are `pub(crate)`

`#[derive(Error)]` generates `new()` and `caused_by()` as `pub(crate)`
(`ohno_macros/src/derive_error/constructors.rs`), but the documentation never said so. The
generated-signature comments read `ConfigError::new(path: String) -> Self`, and the
`#[no_constructors]` counter-example directly below writes `pub fn new` — both cues point the reader
towards a public constructor.

The ambiguity has a measured cost. On 2026-08-07, on the ox-sdk PR adopting `ohno` errors in
`observed`, three independent AI reviewers posted the *same* incorrect finding: that
`#[ohno::error]` generates public `new()` and `caused_by()` constructors. Three separate rebuttals
for one missing sentence. Three models converging on the same wrong claim is not three coincidences
— the reviewers were reading the docs correctly, and the docs were ambiguous.

The docs now state the visibility, give the rationale (the owning crate keeps control of how its
errors are built, so adding a field is not a breaking change for callers), and show the real
expansion as the single `impl` block it is.

## 2. `#[no_constructors]` is now rejected under `#[ohno::error]` (breaking)

Writing the first half surfaced the second. `#[ohno::error]` adds the `OhnoCore` field itself.
`#[no_constructors]` removes the generated constructors, so the author has to write the struct
literal by hand — which means naming that added field.

That name is not theirs to rely on. `generate_unique_field_name` picks `ohno_core`, then
`ohno_core_1`, `ohno_core_2` and so on if the struct already declares one. A hand-written
constructor therefore names a field the macro chose, and an unrelated later field declaration can
rename it out from under the constructor.

**The one existing user is the evidence.** `cachet_tier::Error` was the only place in the repo
combining the two:

```rust
#[ohno::error]
#[no_constructors]
#[derive(Clone)]
pub struct Error {
    recovery_info: RecoveryInfo,
}

impl Error {
    fn caused_by(cause: impl Into<Box<dyn StdError + Send + Sync>>) -> Self {
        Self {
            ohno_core: OhnoCore::from(cause),   // a field it never declared
            recovery_info: RecoveryInfo::never(),
        }
    }
}
```

Every other `#[no_constructors]` in `oxidizer` and `ox-sdk` already pairs with
`#[derive(ohno::Error)]` (6 sites in ox-sdk, plus `http_extensions`, tests and examples here) and is
unaffected.

This is the rule the attribute already applies to `#[error]`, whose rejection tells the author to
use `#[derive(ohno::Error)]` to place the core themselves. From `docs/error_error.md`: *"This
attribute always adds the OhnoCore field itself. It always builds the error from the field it
adds."* Building that value by hand is the same request, so it gets the same answer.

The alternative — documenting the field name so manual construction is legitimate — was rejected: it
would promise a name the macro reserves the right to change, turning an implementation detail into
public API.

```
error: `#[no_constructors]` is not supported under `#[ohno::error]`. Writing a constructor by hand
means writing the struct literal, which needs the name of the OhnoCore field this attribute adds --
a name it chooses, and changes if you later declare a field of your own called `ohno_core`. Use
`#[derive(ohno::Error)]` and declare the OhnoCore field yourself, so the field your constructor
names is one you can see
  --> tests/ui/ohno_error_no_constructors.rs:11:1
   |
11 | #[no_constructors]
   | ^^^^^^^^^^^^^^^^^^
```

## Changes

- `ohno/src/lib.rs` — constructor visibility stated and rationalised; generated signatures shown as
  the `impl` block they actually are; `#[no_constructors]` documented as derive-only.
- `ohno_macros/src/error_type_attr.rs` — `reject_no_constructors`, alongside the existing `reject_*`
  checks.
- `ohno_macros/docs/error_error.md` — the rule under `#[ohno::error]`, plus an entry in "What gets
  rejected".
- `ohno_macros/src/lib.rs` — the same notes on the `#[no_constructors]` bullet, since a reviewer may
  read the macro docs rather than the crate docs.
- `cachet_tier/src/error.rs` — migrated to `#[derive(ohno::Error)]` with a declared `ohno_core`, so
  the field its constructors name is one a reader can see.
- Tests: a unit test and a `trybuild` UI case, matching the sibling rejections.
- Both READMEs regenerated with `cargo doc2readme`.

## Effects

- A reader of either the crate docs or the macro docs learns the constructor visibility at the point
  where they would otherwise guess it, which should stop that class of review finding recurring.
- Breaking for any out-of-tree crate combining `#[ohno::error]` with `#[no_constructors]`. The
  migration is mechanical — swap to `#[derive(ohno::Error)]` and declare the `OhnoCore` field — and
  the error message states it.
- A struct that opts out of the generated constructors now declares its core, so the field a manual
  constructor names is visible at the definition instead of supplied by macro expansion.
- `#[ohno::error]` covers one case fewer, and authors who want both behaviours write one extra
  field.

## Validation

- `cargo +1.96.1 test -p ohno -p ohno_macros -p cachet_tier --all-features` — all pass, including
  the new UI case.
- `just anvil-clippy` — clean across the workspace.
- `rustfmt +nightly-2026-05-30 --config-path ./unstable-rustfmt.toml --check` — 0 would be
  reformatted.
- `cargo doc2readme --check` — up to date for both `ohno` and `ohno_macros`.


crates/ohno/src/lib.rs
```diff
@@ -145,16 +145,29 @@
...
 //! // The derive macro automatically generates:
-//! // - ConfigError::new(path: String) -> Self
-//! // - ConfigError::caused_by(path: String, error: impl Into<Box<dyn Error...>>) -> Self
+//! // - pub(crate) fn ConfigError::new(path: String) -> Self
+//! // - pub(crate) fn ConfigError::caused_by(path: String, error: impl Into<Box<dyn Error...>>) -> Self
 //!
 //! let error = ConfigError::new("/etc/config.toml");
 //! let error_with_cause = ConfigError::caused_by("/etc/config.toml", "File not found");
 //! ```
 //!
```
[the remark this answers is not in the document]
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Agreed, and fixed in `ce09fbd7`. `pub(crate) fn ConfigError::new(path: String) -> Self` is not a form Rust ever writes — it splices a path-qualified name onto a `fn` signature — so it doesn't map onto the expansion a reader is trying to match it against. Adding `pub(crate) fn` in this PR made that worse, since visibility is precisely the detail the reader now has to locate in the real code.

I took the `impl` block option rather than just dropping the `ConfigError::` prefix. The generated methods do share a single inherent impl (`ohno_macros/src/derive_error/constructors.rs:135`), so the block is a faithful picture rather than a convenient fiction — and it makes the name collision described just below self-evident: the hand-written constructor lands in a *second* impl, which is why reusing `new` or `caused_by` is `error[E0592]`.

```rust
// The derive macro automatically generates:
//
// impl ConfigError {
//     pub(crate) fn new(path: String) -> Self { ... }
//     pub(crate) fn caused_by(path: String, error: impl Into<Box<dyn Error...>>) -> Self { ... }
// }
```

On doc2readme: `crates/ohno/README.md` is updated in the same commit and its wrapping matches the doc comment line for line, so the CI `Readmes` check (`cargo doc2readme --check`) stays green. Doctests (36 passed, 0 failed, `--all-features`) and the nightly fmt check with `unstable-rustfmt.toml` were re-run after the edit.</parameter>
<parameter name="pr_url">https://github.com/microsoft/oxidizer/pull/653

crates/ohno/src/lib.rs
```diff
@@ -145,16 +145,35 @@
...
+//! **The generated constructors are `pub(crate)`, regardless of the visibility of the error type
+//! itself.** They are an implementation convenience for the crate that defines the error, not part
+//! of its public API, so a `pub struct` error exported from a library cannot be constructed with
+//! `new()` or `caused_by()` by a downstream crate. This is deliberate: it keeps the set of ways an
+//! error can be built under the control of the crate that owns it, so adding a field is not a
+//! breaking change for callers.
+//!
+//! To expose a constructor to other crates, write your own and declare it `pub`. A constructor
+//! under any other name coexists with the generated ones; one named `new` or `caused_by` collides
+//! with them (`error[E0592]: duplicate definitions`), so it also needs `#[no_constructors]`.
```
Vaiz: > //! To expose a constructor to other crates, write your own and declare it `pub`. A constructor
//! under any other name coexists with the generated ones; one named `new` or `caused_by` collides
//! with them (`error[E0592]: duplicate definitions`), so it also needs `#[no_constructors]`.

this whole part can be removed

crates/ohno/src/lib.rs
```diff
@@ -145,16 +145,35 @@
...
+//!
 //! **Disabling Automatic Constructors:**
 //!
-//! Use `#[no_constructors]` to disable automatic generation when you need custom constructors:
+//! `#[no_constructors]` disables the generated constructors when you need custom ones. It belongs
+//! to `#[derive(Error)]` and is rejected under `#[ohno::error]`: writing a constructor by hand
+//! means writing the struct literal, which needs the name of the `OhnoCore` field — and
+//! `#[ohno::error]` adds that field itself. Declare the core yourself so your constructor names a
+//! field you can see. A hand-written constructor is reachable outside the defining crate only if
+//! you declare it `pub`, as in the `pub fn new` below:
```
Vaiz: > //! `#[no_constructors]` disables the generated constructors when you need custom ones. It belongs
//! to `#[derive(Error)]` and is rejected under `#[ohno::error]`: writing a constructor by hand
//! means writing the struct literal, which needs the name of the `OhnoCore` field — and
//! `#[ohno::error]` adds that field itself. Declare the core yourself so your constructor names a
//! field you can see. A hand-written constructor is reachable outside the defining crate only if
//! you declare it `pub`, as in the `pub fn new` below:

replace with something along those lines

> `#[no_constructors]` disables the generated constructors and allow to define your own version of `new` and `caused_by` methods. It works only with `derive(ohno::Error)` macro, forcing users to define `OhnoCore` field manually.

crates/ohno/tests/ui/ohno_error_no_constructors.stderr
```diff
@@ -0,0 +1,5 @@
+error: `#[no_constructors]` is not supported under `#[ohno::error]`. Writing a constructor by hand means writing the struct literal, which needs the name of the OhnoCore field this attribute adds -- a name it chooses, and changes if you later declare a field of your own called `ohno_core`. Use `#[derive(ohno::Error)]` and declare the OhnoCore field yourself, so the field your constructor names is one you can see
```
Vaiz: > error: `#[no_constructors]` is not supported under `#[ohno::error]`. Writing a constructor by hand means writing the struct literal, which needs the name of the OhnoCore field this attribute adds -- a name it chooses, and changes if you later declare a field of your own called `ohno_core`. Use `#[derive(ohno::Error)]` and declare the OhnoCore field yourself, so the field your constructor names is one you can see

replace the error with something along those lines:

> error: `#[no_constructors]` is not supported under `#[ohno::error]`. Constructor must initialize OhnoCore field, and field inserted by `#[ohno::error]` doesn't have stable form and should not be used in the code.

crates/ohno/src/lib.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Removed in `6b0a4b51`.

Agreed on reflection: both facts in that paragraph are things a reader meets at the compiler rather than things the reference has to carry, and the section is about what the macro generates. The `pub(crate)` paragraph above it already carries the fact that motivated this PR.</parameter>
<parameter name="pr_url">https://github.com/microsoft/oxidizer/pull/653

crates/ohno/src/lib.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Applied in `6b0a4b51`, following your wording with the second person dropped:

```rust
//! `#[no_constructors]` disables the generated constructors, leaving the names `new` and
//! `caused_by` free for hand-written versions. It works only with `#[derive(Error)]`, which
//! requires the `OhnoCore` field to be declared explicitly — and that field is the one such a
//! constructor has to initialize:
```

Kept the closing clause about the field a hand-written constructor initializes, because it is what connects "declare `OhnoCore` manually" to the reason for it, and the example directly below shows exactly that field being set.</parameter>
<parameter name="pr_url">https://github.com/microsoft/oxidizer/pull/653

crates/ohno/tests/ui/ohno_error_no_constructors.stderr
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Applied in `6b0a4b51`. The message is now:

```
error: `#[no_constructors]` is not supported under `#[ohno::error]`. A constructor has to
initialize the OhnoCore field, and the field inserted by `#[ohno::error]` has no stable name, so it
must not be referred to in code. Use `#[derive(ohno::Error)]` and declare the OhnoCore field
explicitly
```

Your version with two small keeps: the trailing pointer to `#[derive(ohno::Error)]`, matching the sibling `ALREADY_MARKED` and `MARKED_FIELD_WITH_GENERATED` messages, which both end by naming the alternative; and "has no stable name" rather than "doesn't have stable form", since the instability is specifically in the name (`ohno_core` → `ohno_core_1`) and that is the part a constructor would otherwise depend on.

The `ohno_core_1` walkthrough is dropped from the diagnostic and kept in `docs/error_error.md`, where a reader who wants the reasoning is already looking. trybuild snapshot regenerated.</parameter>
<parameter name="pr_url">https://github.com/microsoft/oxidizer/pull/653

Vaiz:

One more fix in `6b0a4b51`, from a **suppressed** comment in the Copilot review (review body, not an inline thread) — and it was a real bug:

> The documented "generated" signatures show `path: String`, but the macro expansion actually uses `impl Into<String>` for `String` fields (so calls like `ConfigError::new("/etc/config.toml")` type-check).

Correct. `constructors.rs:174` generates `#(#param_names: impl Into<#field_types>),*`, so the comment contradicted the two example calls immediately below it, which pass `&str`. Worse, this PR is what put `pub(crate) fn` on those lines — so it drew attention to a signature that was wrong in its other half. Now:

```rust
//! // impl ConfigError {
//! //     pub(crate) fn new(path: impl Into<String>) -> Self { ... }
//! //     pub(crate) fn caused_by(path: impl Into<String>, error: impl Into<Box<dyn Error...>>) -> Self { ... }
//! // }
```

Also, on the broader style point about second person: `your own` / `yourself` / `you can see` are gone from every line this PR adds. The rest of this crate's reference documentation describes the macro rather than addressing the reader, and these additions did not match it.

**Validation after the changes** — `ohno` + `ohno_macros` + `cachet_tier` with `--all-features` pass including the trybuild case, `just anvil-clippy` clean, nightly rustfmt with `unstable-rustfmt.toml` clean (36 files, 0 reformatted), and `cargo doc2readme --check` up to date for both crates.

Vaiz:

Swept the rest of the crate in `0f1582b4`. Fixing only the lines this PR added had left it *less* consistent, not more — the untouched documentation still addressed the reader, and two of the three rejection messages still ended with "yourself" while the third no longer did.

| Location | Before | After |
|---|---|---|
| `lib.rs:62` | adds a `OhnoCore` field to **your** struct | …to **the** struct |
| `lib.rs:89` | attribute **allows you to customize** the main error message | attribute **customizes** the main error message |
| `lib.rs:183` | `// Your custom constructor logic here` | `// Custom constructor logic here` |
| `app/error.rs:20` | applications **where you need** a simple, catch-all error type | applications **that need** … |
| `app/error.rs:28` | catch-all error type in **your** application | …in **an** application |
| `error_display.md:49` | ## Which fields **you can** name | ## Which fields **can be** named |
| `ALREADY_MARKED` | …to place the core **yourself** | …to place the core **explicitly** |
| `MARKED_FIELD_WITH_GENERATED` | …to place the OhnoCore **yourself** | …to place the OhnoCore **explicitly** |

The `#[display]` one was worth more than a pronoun swap: "allows you to customize" was a periphrasis wrapped around the verb the sentence actually needed, so removing the second person made it shorter and more direct at the same time.

**Deliberately left:** two plain `//` comments in `tests/enrich_err.rs:176` and `tests/enrich_err_async.rs:114` ("you can't use message as it's consumed in the function"). Those are notes to someone reading the test, never rendered as documentation — different register, and rewriting them would be churn.

READMEs regenerated, trybuild snapshots updated for the two message changes.

**Validation** — `ohno` + `ohno_macros` + `cachet_tier` `--all-features` pass including all 12 trybuild cases; `just anvil-clippy` clean; nightly rustfmt with `unstable-rustfmt.toml` clean (30 files, 0 reformatted); `cargo doc2readme --check` up to date for both crates.

Vaiz:

Answering the suppressed comment in [review 4896789519](https://github.com/microsoft/oxidizer/pull/653#pullrequestreview-4896789519) (`crates/ohno_macros/src/lib.rs:36`).

**Rejected — the premise is incorrect.** The comment says the `#[no_constructors]` bullet is "a multi-sentence bullet unlike the others" and is alone in lacking terminal punctuation. Neither half holds. The full list as it stands:

```rust
/// - `#[error]` - Mark the field containing the `OhnoCore`. At most one field may be marked. With
///   no marker the macro looks for a single field whose type is named `OhnoCore`, so a core reached
///   through a type alias or a renamed import has to be marked
/// - `#[display("...")]` - Custom display message with field interpolation. Positional arguments
///   are implicitly scoped to `self`, so fields are referenced by their bare name
///   (`path.display()`, not `self.path.display()`)
/// - `#[no_constructors]` - Disable automatic constructor generation. The `new()` and `caused_by()`
///   constructors generated by default are `pub(crate)` even when the error type is `pub`, so they
///   are usable only inside the crate that defines the error. Rejected under `#[ohno::error]`,
///   which adds the `OhnoCore` field a hand-written constructor would have to initialize
/// - `#[no_debug]` - Disable automatic Debug trait implementation
/// - `#[from(Type1, Type2, ...)]` - Generate From implementations for specified types
```

The `#[error]` and `#[display]` bullets are also multi-sentence, and **not one bullet in the list ends with a period** — that is the existing house style here, and it predates this PR. Adding a trailing period to this bullet alone would introduce the inconsistency the comment is trying to remove. Normalising all five is a cosmetic change to lines this PR does not otherwise touch, so it is out of scope for a change about constructor visibility.

The two earlier suppressed comments on this PR (reviews [4895981591](https://github.com/microsoft/oxidizer/pull/653#pullrequestreview-4895981591) and [4896001801](https://github.com/microsoft/oxidizer/pull/653#pullrequestreview-4896001801)), which correctly flagged the documented signatures as `String` where `generate_complex_named_constructors` emits `impl Into<FieldType>`, were **accepted and are already fixed** at head `0f1582b` — `crates/ohno/src/lib.rs` now documents `pub(crate) fn new(path: impl Into<String>) -> Self` inside a real `impl` block. Thanks for catching that one; it was a genuine bug.

sandersaares (approved):

crates/ohno_macros/src/error_type_attr.rs
```diff
@@ -56,6 +56,9 @@ fn error_impl(input: &mut DeriveInput) -> proc_macro2::TokenStream {
     if let Err(err) = reject_generated_marker(input) {
         return err.to_compile_error();
     }
+    if let Err(err) = reject_no_constructors(input) {
```
sgalkin: Could we reconsider this unconditional `reject_no_constructors` check?

**Problem.** The check equates disabling `new` and `caused_by` with writing a struct literal that names the injected field. Those are separate concerns: `impl_error_derive` generates `From` implementations independently, using the actual selected error field. This makes conversion-only errors safe without exposing the field name.

**Examples.** This case is safe and currently valid:

```rust
#[ohno::error]
#[no_constructors]
#[from(std::io::Error)]
pub struct IoError;
```

Conversely, field-name coupling remains possible even without `#[no_constructors]`:

```rust
#[ohno::error]
pub struct IoError {
    context: String,
}

impl IoError {
    fn custom(context: String, error: std::io::Error) -> Self {
        Self {
            context,
            ohno_core: OhnoCore::from(error),
        }
    }
}
```

**Judgment and potential fix.** Removing the rejection preserves conversion-only APIs but still allows code in the defining module to depend on the generated field name. That field remains private, so this is not a cross-crate leak; it is compile-time coupling to a macro implementation detail. The proposed rejection blocks both safe and unsafe `#[no_constructors]` uses while not preventing equivalent coupling through another constructor name.

Could we remove this blanket check, retain the `cachet_tier` migration, document direct references to the generated field as unsupported, and replace the compile-fail test with a compile-pass conversion-only case? If field-name coupling must be technically impossible, that requires a deeper design such as a stable macro-generated internal initializer.

🤖 Co-authored with Copilot.

sgalkin (approved):

crates/ohno_macros/src/error_type_attr.rs
```diff
@@ -64,8 +67,19 @@ fn error_impl(input: &mut DeriveInput) -> proc_macro2::TokenStream {
...
+const ALREADY_MARKED: &str = "`#[ohno::error]` adds the OhnoCore field itself and generates the error representation from it, so no field may be marked with `#[error]`. Remove the marker to keep the field as data, or use `#[derive(ohno::Error)]` to place the core explicitly";
 const RESERVED_MARKER: &str = "This doc comment is reserved for `#[ohno::error]`, which puts it on the OhnoCore field it adds. Remove it; if this is the field holding the OhnoCore, use `#[derive(ohno::Error)]` and mark it with `#[error]`";
+const NO_CONSTRUCTORS: &str = "`#[no_constructors]` is not supported under `#[ohno::error]`. A constructor has to initialize the OhnoCore field, and the field inserted by `#[ohno::error]` has no stable name, so it must not be referred to in code. Use `#[derive(ohno::Error)]` and declare the OhnoCore field explicitly";
+
+/// Reject a struct that opts out of the generated constructors
+///
+/// See `docs/error_error.md`.
```
psandana, line 75: **The rejection is broader than the rationale supports.** The premise is that opting out of the generated constructors implies hand-writing a struct literal, which needs the injected field's name. That is not true in general: `#[from(...)]` builds the literal *inside* the macro (`derive_error/from_impls.rs`, `generate_from_implementations_named`), including the error field, so

```rust
#[ohno::error]
#[from(std::io::Error)]
#[no_constructors]
pub struct Error { pub kind: Kind }
```

is fully constructible today, never names the injected field, and is rejected by this change. `#[no_constructors]` is also used simply to free the `new`/`caused_by` names — which is in fact why `cachet_tier` used it: a name collision with its own `caused_by`, not an inability to name the core.

A cheaper, non-breaking alternative: keep accepting the combination and make the injected name a stable contract — always `ohno_core` — by *rejecting* a user-declared field named `ohno_core` under `#[ohno::error]` rather than silently renaming to `ohno_core_1` (`generate_unique_field_name`). The instability the message cites is entirely self-inflicted by that collision fallback, and the collision can only be created deliberately by the author. If publishing the name is unwanted, a narrower rule (reject only when there is no other construction path, i.e. no `#[from(...)]`) or a generated `pub(crate) fn from_core(core: OhnoCore, ..)` seam would be more targeted than an outright ban.

crates/ohno_macros/docs/error_error.md
```diff
@@ -88,6 +88,25 @@ On a unit struct, the attribute rewrites the struct as a tuple struct holding
 the added field. `#[derive(ohno::Error)]` alone rejects a unit struct, because
 there is no room for a core.
 
+**`#[no_constructors]` is not accepted.** Opting out of the generated
+constructors means writing the struct literal by hand, and that needs the name
+of the field this attribute adds. The name is not the author's to rely on: it is
+`ohno_core`, unless the struct already declares one, in which case the added
+field becomes `ohno_core_1`. A constructor would then name a field chosen for it,
```
psandana: Two things in this sentence.

**The `ohno_core_1` claim understates the generator.** `generate_unique_field_name` (`src/utils.rs`) loops until the candidate is free: with `ohno_core` and `ohno_core_1` both declared it produces `ohno_core_2`, and its own `test_generate_unique_field_name_multiple_collisions` pins `ohno_core_3`. The doc describes the single-collision case as if it were the rule. Also, the name-based scheme applies only to named structs — on a tuple or unit struct the core is appended positionally and has no name at all, so "the name of the field this attribute adds" doesn't literally hold there.

**"a later, unrelated field declaration can rename it out from under it" overstates the instability.** The only declaration that triggers the rename is a field named literally `ohno_core` — hardly unrelated, and something the author writes knowingly. As written the rationale argues a technical impossibility that isn't quite one.

```suggestion
of the field this attribute adds. The name is not the author's to rely on: it is
`ohno_core` for a named struct, or `ohno_core_N` for the first `N` no declared
field already uses, and on a tuple or unit struct the core is appended
positionally and has no name at all. A constructor would then name a field
chosen for it, which a later declaration of a field called `ohno_core` can
rename out from under it.
```

If the real reason is "the injected field is not part of the type's source-level contract" — which is a perfectly good reason — saying that plainly would be stronger than the mechanism argument.

crates/cachet_tier/src/error.rs
```diff
@@ -40,11 +40,11 @@ use recoverable::{Recovery, RecoveryInfo};
 ///     Ok(v) => // Success
 /// }
 /// ```
-#[ohno::error]
+#[derive(ohno::Error, Clone)]
 #[no_constructors]
-#[derive(Clone)]
 pub struct Error {
     recovery_info: RecoveryInfo,
```
psandana, line 46: The migrated core is now selected by step 3 of field detection — matching on the last path segment being literally `OhnoCore` — which is the most fragile of the three selection paths. A future `use ohno::OhnoCore as Core;`, a type alias, or a second `OhnoCore`-typed field silently changes or breaks selection, and the diagnostic (`Multiple OhnoCore fields found` / no field found) surfaces far from the edit that caused it.

Since this PR deliberately steers authors from the attribute to the derive, this migration is the canonical example others will copy, so it is worth making it the robust one — and worth using the same spelling in the new `docs/error_error.md` and README examples:

```suggestion
    #[error]
    ohno_core: OhnoCore,
```

The rest of the migration looks faithful: field order, the single core, `Clone`, and the generated `Debug`/`Display` are unchanged. A one-line comment on why the field exists would also help the next editor, since the invariant is now maintained by hand rather than by the macro.

crates/ohno_macros/src/error_type_attr.rs
psandana: Two small things about the check itself.

1. **Not gated on `Data::Struct`,** unlike `reject_marked_field` and `reject_generated_marker`. Because it runs before `add_ohno_core_field`, `#[ohno::error] #[no_constructors] enum E { A }` now reports "use `#[derive(ohno::Error)]` and declare the OhnoCore field explicitly" instead of the accurate "`#[ohno::error]` can only be applied to structs". Returning `Ok(())` for non-structs and letting `add_ohno_core_field` speak fixes it, and the unit test already has a unit-struct case that would extend naturally to an enum.
2. **Error accumulation.** All three rejections are independent whole-item validations, so a struct that trips two reports only the first; the author fixes it and hits a second unrelated hard error. `syn::Error::combine` would report everything in one compile. If the sequencing is deliberate, ignore this — the placement before `add_ohno_core_field` is right either way.

Also, could the UI fixture cover `#[no_constructors]` written *above* `#[ohno::error]`, and a tuple struct? The unit test covers the shapes, but `tests/ui/ohno_error_no_constructors.rs` pins only the named-struct, attribute-below case — which is the one ordering obviously present in `input.attrs`.

crates/ohno/tests/ui/ohno_error_no_constructors.stderr
```diff
@@ -0,0 +1,5 @@
+error: `#[no_constructors]` is not supported under `#[ohno::error]`. A constructor has to initialize the OhnoCore field, and the field inserted by `#[ohno::error]` has no stable name, so it must not be referred to in code. Use `#[derive(ohno::Error)]` and declare the OhnoCore field explicitly
```
psandana: The message explains the constraint well, but leaves the reader to derive the fix. Since this fires on code that compiled yesterday, this diagnostic is the entire migration guide for most people who hit it — a literal replacement shape would make it a copy-paste fix rather than a doc hunt. Something like: `... explicitly: replace #[ohno::error] with #[derive(ohno::Error)] and add a private `#[error] ohno_core: ohno::OhnoCore` field.` Attaching it as a secondary `syn::Error` so it renders as a separate `note:` would keep the primary message short.

crates/ohno/src/lib.rs
```diff
@@ -145,16 +145,29 @@
...
+//! // impl ConfigError {
+//! //     pub(crate) fn new(path: impl Into<String>) -> Self { ... }
+//! //     pub(crate) fn caused_by(path: impl Into<String>, error: impl Into<Box<dyn Error...>>) -> Self { ... }
+//! // }
 //!
 //! let error = ConfigError::new("/etc/config.toml");
 //! let error_with_cause = ConfigError::caused_by("/etc/config.toml", "File not found");
 //! ```
 //!
+//! **The generated constructors are `pub(crate)`, regardless of the visibility of the error type
```
psandana, line 158: This paragraph is accurate and well placed. Two additions worth making.

**Mention that `pub(crate)` also excludes integration tests and benches.** `tests/*.rs` and benches compile as separate crates, so they cannot call the generated constructors either. That surprises people and is the natural next question after this paragraph.

**The derive's own rustdoc buries the same fact.** In `ohno_macros/src/lib.rs` the visibility is stated only inside the `#[no_constructors]` bullet — the attribute that *removes* those constructors — rather than in the lead prose next to "automatically implements ...", which is where the `ohno` crate docs put it. Relatedly, the `#[ohno::error]` attribute macro's own rustdoc lists no accepted/rejected attributes at all, so a reader on docs.rs learns nothing about the new hard error; `docs/error_error.md` has the full explanation but is an internal design doc. A one-line note on `pub fn error` would keep the two in sync.

crates/ohno_macros/src/lib.rs
```diff
@@ -30,7 +30,10 @@ mod utils;
 /// - `#[display("...")]` - Custom display message with field interpolation. Positional arguments
 ///   are implicitly scoped to `self`, so fields are referenced by their bare name
 ///   (`path.display()`, not `self.path.display()`)
-/// - `#[no_constructors]` - Disable automatic constructor generation
+/// - `#[no_constructors]` - Disable automatic constructor generation. The `new()` and `caused_by()`
+///   constructors generated by default are `pub(crate)` even when the error type is `pub`, so they
+///   are usable only inside the crate that defines the error. Rejected under `#[ohno::error]`,
+///   which adds the `OhnoCore` field a hand-written constructor would have to initialize
```
psandana: Wording nit: "the `OhnoCore` field a hand-written constructor would have to initialize" is a garden path — the reader parses "the `OhnoCore` field a hand-written constructor" before backtracking. Adding "that" fixes it.

```suggestion
///   which adds the `OhnoCore` field that a hand-written constructor would have to initialize
```

Same class of thing in `crates/ohno/src/lib.rs` ("and that field is the one **such a** constructor has to initialize") — "the hand-written constructor" reads better, since the antecedent is two clauses back. And in `docs/error_error.md`, "Declaring the core is what makes a hand-written constructor **honest**" is vague next to the rest of that file; the concrete property is that a declared core gives the constructor a field name the author owns. Note the `README.md` copies are doc2readme output — fix the `//!` docs and regenerate.

psandana (commented):
Reviewed the docs half and the new rejection, and read `constructors.rs`, `utils.rs`, `from_impls.rs` and `field_detection.rs` at head to check the claims. The `pub(crate)` documentation is accurate and a clear improvement — the generated signature shown as the `impl` block it actually is, with `impl Into<FieldTy>` parameters, matches `generate_complex_named_constructors` exactly. Two non-issues I checked so nobody re-raises them: `ohno_macros/README.md` correctly needs no regeneration (doc2readme embeds only the crate-level `//!` block, and the change is on item docs), and `ohno/README.md` is in sync with `ohno/src/lib.rs`.

My main concerns are on the breaking half, filed inline. One that has no natural line:

**Breaking change is not surfaced to downstream consumers.** This turns previously-compiling code into a hard error, but the diff touches no `CHANGELOG.md` and no `Cargo.toml` version. At `ohno` 0.3.9 / `ohno_macros` 0.3.5, a 0.3.x bump is a semver-compatible upgrade for every consumer, so downstreams pick this up on a routine `cargo update` and fail to build with no migration runway. Could you confirm the release tooling derives the breaking marker from the commit message? Note the scope is `fix(ohno)!` while the behavioral change lives in `ohno_macros`, so the `⚠️ Breaking` entry may land in the wrong crate's changelog. It would also be worth stating in the description that a `no_constructors` sweep was run across the consuming repos (ox-sdk, Substrate, 3srust), since the PR body only establishes the in-repo count.

psandana (approved):
Left some comments, but we are not bumping a new version, so issues can be fixed in subsequent PR

crates/ohno_macros/src/error_type_attr.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Thanks — I read `derive_error/from_impls.rs` before answering, and your reading of the mechanism is correct: `generate_from_implementations_named` builds the struct literal itself and fills the core via `error_field.to_field_access()`, so a `#[from(...)]`-only type genuinely never names the injected field. The counterexample is real and I'm not disputing it.

**The blanket check stays anyway.** `#[from(...)]` is syntax sugar over a construction the macro happens to write for the author today; it is not a guarantee about what the author will write tomorrow. The rule this check enforces is not "this particular type currently has no code that names the injected field" — it is "a type whose core is injected has no author-written construction path at all". Once `#[no_constructors]` is accepted under `#[ohno::error]`, the door is open: the overwhelmingly likely next step for someone who reached for `#[no_constructors]` is a hand-written constructor, and the only way to write one is to name a field the macro reserves the right to rename. A rule that admits the safe case and relies on authors not taking the obvious next step is not a rule.

Narrowing to "reject only when there is no `#[from(...)]`" was considered and rejected for the same reason: it makes the presence of an unrelated attribute decide whether a construction path is legal, so adding or removing a `#[from]` silently flips the type between supported and unsupported. That is a worse contract than a flat no.

The supported path for everything under discussion — conversion-only errors included — is `#[derive(ohno::Error)]` with the core declared explicitly. It costs one line, and it gives the author a field name they own, which is exactly the property that makes a hand-written constructor legitimate. Nothing is lost; only the injected-field variant is closed off.

On `cachet_tier`: agreed that it used `#[no_constructors]` to free the `caused_by` *name*, not because it needed to name the core. Its migration to the derive is in this PR and is the canonical example.

Keeping this as-is. Your other threads on this PR are separate points and I'm working through them now.

crates/ohno_macros/src/error_type_attr.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

You and @sgalkin arrived at the same counterexample independently, so I checked it in the source rather than argue from the PR description: `generate_from_implementations_named` in `derive_error/from_impls.rs` does build the literal inside the macro, filling the core through `error_field.to_field_access()`. Your `#[from(std::io::Error)] #[no_constructors]` example is constructible today and never names the injected field. Correct on the facts.

**The check stays blanket regardless.** `#[from(...)]` is sugar over a construction the macro writes; it says nothing about what the author adds next. Someone who reaches for `#[no_constructors]` is overwhelmingly likely to be heading for a hand-written constructor, and under `#[ohno::error]` the only way to write one is to name a field the macro owns. Admitting the safe case and trusting authors not to take the obvious next step isn't a rule, it's a hope. Your narrower variant — reject only when there's no `#[from(...)]` — has the specific defect that adding or deleting an unrelated `#[from]` would silently flip the type between supported and unsupported, which is a worse contract than a flat no.

**On making `ohno_core` a stable contract instead:** that is the one option I can't take, and it's the basis of the change rather than a detail. Promising the name turns a macro implementation detail into public API that can never be changed, and it legitimizes exactly the manual construction the attribute exists to avoid. Rejecting an author-declared `ohno_core` to remove the collision fallback would also break compiling code for a field name that is currently perfectly legal — a second breaking change to avoid the first.

**Your `generate_unique_field_name` correction is right and I'm fixing it.** I read `utils.rs`: the loop increments until the candidate is free, so it's `ohno_core`, `ohno_core_1`, `ohno_core_2`, … — `error_error.md` describes the single-collision case as if it were the rule, and it also doesn't hold at all for tuple/unit structs where the core is appended positionally with no name. I'm rewriting that sentence to lead with the point you make at the end — the injected field is not part of the type's source-level contract — with the naming scheme as supporting detail rather than the argument. That's a better rationale than the mechanism one, thank you.

I'm working through your other threads (the `Data::Struct` gating, the `#[error]` marker on the migrated `cachet_tier` core, the docs.rs gap on `pub fn error`, and the wording nits) in this same round.

crates/ohno_macros/docs/error_error.md
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Both corrections accepted, and they pushed me further than the suggestion did. Fixed in `9b83a054`.

You're right on the mechanism: `generate_unique_field_name` loops until the candidate is free, so it's `ohno_core_N`, not `ohno_core_1` — and on a tuple or unit struct the core is appended positionally with no name at all, so "the name of the field this attribute adds" doesn't hold there either.

Rather than correct the walkthrough, I deleted it. Your closing line is the right argument, so it is now the only argument:

> **`#[no_constructors]` is not accepted.** Opting out of the generated constructors means writing the struct literal by hand, and that needs the name of the field this attribute adds. That field is not part of the type's source-level contract, so it has no name the author may rely on.

Describing the naming scheme was documentation the macro doesn't owe anyone: it was wrong in two ways you caught, and being right about it would have been worse — a precise description of `ohno_core_N` is a promise about a name the whole change exists to *not* promise. Less prose here is strictly less prose to go stale.

For the same reason I removed the "Into:" expansion example from the attribute's own rustdoc in `error_type_attr.rs`, which showed `ohno_core: OhnoCore` outright. That one directly contradicted the rule this PR adds.

The `#[derive(ohno::Error)]` example still names `ohno_core`, and that stays — there the author declares the field, so the name is theirs.

crates/cachet_tier/src/error.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Agreed on the marker — applied in `9b83a054`. The migrated core now reads:

```rust
#[derive(ohno::Error, Clone)]
#[no_constructors]
pub struct Error {
    recovery_info: RecoveryInfo,
    #[error]
    ohno_core: OhnoCore,
}
```

Your reasoning is the deciding one: this migration is the example others will copy, so it should not be the variant that leans on step 3 of field detection matching the last path segment. I made the same change to the `docs/error_error.md` example so the two agree.

Declining the comment on the field. What it would say — "this field holds the OhnoCore" — is already said by the type and by `#[error]`, so it is a sentence that can only rot.

crates/ohno_macros/src/error_type_attr.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

**(1) accepted** — good catch, and it was a real misdiagnosis. `#[ohno::error] #[no_constructors] enum E { A }` was being told to declare an `OhnoCore` field, which an enum cannot do. `reject_no_constructors` now returns `Ok(())` for non-structs and lets `add_ohno_core_field` report "`#[ohno::error]` can only be applied to structs", matching `reject_marked_field` and `reject_generated_marker`. Extended the unit test with the `#[no_constructors] enum` case. In `9b83a054`.

**(2) declined** — the sequencing is deliberate. Each of the three rejections says "this construct is not supported here, use the derive instead", so the second one is not new information; combining them would print two spellings of the same instruction. I'd rather keep one hard error per compile than pay a `combine` for it.

**On the fixture coverage:** I added the attribute-above-`#[ohno::error]` ordering as a unit-test case rather than a UI fixture. You're right that it is worth pinning — `input.attrs` is order-insensitive here and nothing enforces that — but a `.stderr` fixture per ordering buys the same assertion at the cost of another generated file to keep in sync. The tuple-struct shape is likewise covered in the unit test. The UI fixture stays as the one end-to-end proof that the diagnostic reaches a real compiler.

crates/ohno/tests/ui/ohno_error_no_constructors.stderr
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Declining, though the reasoning is sound and I want to be clear it isn't being waved away.

The message already names the replacement attribute and what to do with it: "Use `#[derive(ohno::Error)] and declare the OhnoCore field explicitly`". Spelling out the literal field line would make it marginally more copy-pasteable, at the cost of putting a concrete field name into a compiler diagnostic — and a diagnostic is the most expensive place to keep a name current, because it is pinned by a `.stderr` snapshot and duplicated into every downstream build log. Given the whole point of this change is that the injected field's name is not something to hand people, a diagnostic that hands them a name to type is the wrong artifact for it.

The length is also deliberate: this message is already the longest of the three siblings, and it was shortened once on review. `docs/error_error.md` carries the shape, including the full example.

crates/ohno/src/lib.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Thanks for the check on the paragraph. Declining all three additions, on a consistent principle rather than case by case: every sentence added here is a sentence that can later be wrong, and none of these three carries information the reader cannot get more reliably elsewhere.

**Integration tests and benches.** True, and it does surprise people — but it is a fact about what `pub(crate)` means in Rust, not about `ohno`. Restating the language's own rule in a crate's docs is exactly the kind of sentence that ages badly while the compiler says it accurately and immediately.

**Lead prose on the derive.** The `ohno` crate docs state the visibility in full, in the Automatic Constructors section where the constructors are introduced. Repeating it in the derive's lead prose creates a second copy to keep in sync — and this PR exists precisely because a stale claim about constructor visibility was believed. One authoritative statement beats two agreeing ones.

**A note on `pub fn error`.** This is the strongest of the three and I still won't take it: an attribute macro's rustdoc listing its accepted and rejected attributes is a list that silently rots every time the accept/reject set changes. The compiler already tells a reader that `#[no_constructors]` is rejected, at the moment they try it, with the alternative in the message. That is a better delivery mechanism than a docs.rs list, and it cannot go stale.

Related: I removed the "Into:" expansion example from `#[ohno::error]`'s own rustdoc in `9b83a054`, because it showed the injected field by name and so contradicted the rule this PR adds. That is the direction I'd rather take these docs — fewer claims, each load-bearing.

crates/ohno_macros/src/lib.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Both wording fixes taken in `9b83a054` — the garden path was real:

- `ohno_macros/src/lib.rs`: "…which adds the `OhnoCore` field **that** a hand-written constructor would have to initialize"
- `ohno/src/lib.rs`: "…and that field is the one **the hand-written** constructor has to initialize"

On "honest" in `error_error.md`: agreed it was vague, and the concrete property you name is the right one. The sentence now reads "Declaring the core gives a hand-written constructor a field name the author owns, so that is the supported path" — which is the same point in the same number of words.

And noted on the READMEs — `crates/ohno/README.md` was regenerated with `cargo doc2readme`, not hand-edited, and I re-ran with `--check` afterwards to confirm (`INFO: Readme is up to date`). `ohno_macros`'s is unaffected, as you'd already worked out, since only item docs changed there; I verified that with `--check` too rather than assume it.

Vaiz:

Answering the part of [review 4897434158](https://github.com/microsoft/oxidizer/pull/653#pullrequestreview-4897434158) that has no natural line, and summarizing this round.

**On the design objection** — you and @sgalkin independently found the same counterexample, so I read `derive_error/from_impls.rs` rather than argue from the PR body. You are right on the mechanism: `generate_from_implementations_named` builds the literal inside the macro and fills the core via `error_field.to_field_access()`, so a `#[from(...)]`-only type never names the injected field.

**The blanket rejection stays anyway**, and I've replied in detail on both threads. Short version: `#[from(...)]` is sugar over a construction the macro writes today; it is not a statement about what the author writes next. Someone reaching for `#[no_constructors]` under `#[ohno::error]` is overwhelmingly likely to be heading for a hand-written constructor, and the only way to write one is to name a field the macro owns. Narrowing to "reject only when there is no `#[from]`" would make an unrelated attribute decide whether a construction path is legal, so adding or deleting a `#[from]` silently flips the type between supported and unsupported — a worse contract than a flat no. Making `ohno_core` a stable name is the one option that is off the table, since it converts the implementation detail into permanent public API and legitimizes exactly the manual construction being closed off.

**On the breaking change and semver** — a fair challenge, and I can only answer half of it:

- **Changelog and version:** entries here are produced by release tooling from the commit messages with PR links, not hand-written, so there is deliberately no `CHANGELOG.md` edit in this diff. Your specific worry is the sharper one: the subject is scoped `fix(ohno)!` while the behavioral change lives in `ohno_macros`, so the `⚠️ Breaking` marker may land in the wrong crate's changelog. I have **not** verified how the tooling attributes the scope, so I am recording that as unresolved rather than guessing. Worth confirming before release.
- **Downstream sweep:** the PR body establishes only the in-repo count. I have not run a `no_constructors` sweep across ox-sdk / Substrate / 3srust and will not claim one — also unresolved.

Both are release-gating rather than code-gating, and given no version bump is happening in this PR they can be settled before the next one.

**Changed in `9b83a054`:**
- Removed the `ohno_core` / `ohno_core_1` naming walkthrough from `docs/error_error.md`. It was wrong twice over (the generator loops to `ohno_core_N`; a tuple or unit struct has no named core), and being *right* about it would have been worse — a precise description of the scheme is a promise about the very name this change exists not to promise. The rationale now rests on your own better formulation: the injected field is not part of the type's source-level contract.
- Removed the "Into:" expansion example from `#[ohno::error]`'s rustdoc, which printed `ohno_core: OhnoCore` and so contradicted the rule being added.
- Gated `reject_no_constructors` on `Data::Struct`, so an enum gets the accurate "can only be applied to structs" diagnostic again.
- Marked the migrated `cachet_tier` core with `#[error]`, and the matching `error_error.md` example, so the canonical example does not rely on type-name matching.
- Two wording fixes (the garden-path sentence, and "honest" → the concrete property).

**Declined, with reasons on each thread:** adding the integration-tests/benches consequence of `pub(crate)`, duplicating the visibility statement into the derive's lead prose, adding an accepted/rejected attribute list to `pub fn error`, putting a literal field line into the diagnostic, `syn::Error::combine`, and extra `.stderr` fixtures for the attribute ordering and tuple shape (covered by unit tests instead). The common thread: each adds a sentence or a generated file that can later be wrong, and in every case the compiler or an existing single statement already conveys it.

Gates run before pushing: `just anvil-clippy` clean, nightly rustfmt with `unstable-rustfmt.toml` clean, `cargo test --all-features` green for `ohno` / `ohno_macros` / `cachet_tier` including the trybuild UI snapshots, and both READMEs verified with `doc2readme --check`.

crates/ohno_macros/src/error_type_attr.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Correcting my previous reply on this thread: the fix I described — returning `Ok(())` for non-structs and letting `add_ohno_core_field` speak — was not robust, and I have replaced it. Now in `320b4c5c` (branch rebased onto the 0.4.0 release, so the earlier SHAs are gone).

The problem with that first attempt is that it left the real precondition being established *last*, as a side effect of `add_ohno_core_field` finding no fields to add to. Each of the three rejections then re-derived "is this a struct?" for itself, and whichever fired first decided the diagnostic. Making `reject_no_constructors` defer politely fixed the symptom you reported while leaving the ordering dependency, and four copies of the same check, in place.

The shape is now settled once, up front:

```rust
fn validate(input: &DeriveInput) -> syn::Result<()> {
    let Data::Struct(data_struct) = &input.data else {
        return Err(syn::Error::new_spanned(&input.ident, NOT_A_STRUCT));
    };

    reject_marked_field(data_struct)?;
    reject_generated_marker(data_struct)?;
    reject_no_constructors(input)
}
```

Consequences: `reject_marked_field` and `reject_generated_marker` take a `DataStruct` instead of guessing at one, `reject_no_constructors` needs no shape check at all, and `add_ohno_core_field` takes `&mut DataStruct` and is infallible — its "can only be applied to structs" branch is now unreachable, so it is deleted rather than left to rot. Correctness no longer depends on the order the checks run in.

Per the note in this repo about tests that pin a fix: I ran both new tests against the previous ordering first and watched them fail with the wrong diagnostic (the `#[no_constructors]` message on an enum), then reverted. They cover `enum`, `#[no_constructors] enum`, and `#[no_constructors] union`.

My answer to your point (2) is unchanged — the sequencing is deliberate and I am still not adding `syn::Error::combine`.

Vaiz:

Answering the suppressed comment in [review 4898230758](https://github.com/microsoft/oxidizer/pull/653#pullrequestreview-4898230758) (`crates/ohno_macros/src/error_type_attr.rs`, the `NO_CONSTRUCTORS` message).

**Accepted on substance — the rationale was written for one of the three shapes it covers.** `add_ohno_core_field` appends the core positionally on a tuple struct, and rewrites a unit struct into a tuple struct holding it, so on two of the three shapes there is no name involved at all. Arguing from "no stable name" describes the named-struct case as if it were the rule. This is the same defect @psandana caught in the prose earlier, and I had fixed it there only for named structs.

Fixed in `1ba9aa08`, in `docs/error_error.md`:

> **`#[no_constructors]` is not accepted.** Opting out of the generated constructors means writing the struct literal by hand, and that has to reach the field this attribute adds. That field is an implementation detail rather than part of the type's source-level contract, so nothing about it — its name on a named struct, its position on a tuple or unit struct — is the author's to rely on.

**The diagnostic string itself is unchanged, deliberately, and I want to be transparent about why.** Two reasons:

1. It is not false as written. On a tuple or unit struct the injected field has no name, so "has no stable name" holds trivially; it is imprecise about the mechanism rather than wrong about the outcome. The message already points at the supported alternative, which is the part a reader acts on.
2. Changing it invalidates `crates/ohno/tests/ui/ohno_error_no_constructors.stderr`, which must be regenerated with `TRYBUILD=overwrite` and never hand-written. That command is not available to this automation, and writing the snapshot by hand is exactly the failure mode this PR already hit once with a generated README.

So the mechanism argument now lives in one place, stated correctly for all three shapes, and the compiler error stays short and points to the fix — which is where this PR's earlier review rounds landed on diagnostic length. If the wording in the diagnostic is worth the extra round to you as well, say so and I will make the change with a properly regenerated snapshot rather than a hand-edited one.

**CI on `96570463` (the previous head):** 37 passing / 0 failing / 6 pending. The `anvil-llvm-cov` coverage failure from the last round — `ohno_macros` at 99.8% against a 100% threshold, caused by an unreachable branch my own restructure left behind — is resolved; I deleted the dead code rather than touching the threshold.

psandana (approved):

Vaiz:

Answering [review 4898460720](https://github.com/microsoft/oxidizer/pull/653#pullrequestreview-4898460720), which raises the `NO_CONSTRUCTORS` wording a second time and correctly points out that the UI snapshot pins it.

**Both accepted, and done properly this time — `2d5ef37d`.** In my previous reply I said the diagnostic would stay as-is because rewording it needs the trybuild snapshot regenerated with `TRYBUILD=overwrite`, and that command was not available to this automation. I found a way to run the generator without it: the repo's own `.cargo/config.toml` accepts an `[env]` table, so trybuild produced the snapshot itself, and I reverted the config immediately afterwards. `git status` confirmed only the source file and the `.stderr` were modified before staging, and the suite was then re-run **without** the override to confirm the committed snapshot matches. Nothing was hand-written.

The message now reads:

> `#[no_constructors]` is not supported under `#[ohno::error]`. A constructor has to initialize the OhnoCore field, and the field inserted by `#[ohno::error]` **is an implementation detail with no stable name or position**, so it must not be referred to in code. Use `#[derive(ohno::Error)]` and declare the OhnoCore field explicitly

That is the property that actually holds across all three shapes the rejection covers: `add_ohno_core_field` appends the core positionally to a tuple struct and rewrites a unit struct into a tuple struct holding it, so on two of the three there is no name to be unstable. The wording also now matches the rationale in `docs/error_error.md`, which I corrected in `1ba9aa08`, so the compiler error and the reference doc make the same claim.

Worth noting for the record: you and @psandana reached this independently, from different directions — psandana from the prose, you from the diagnostic and its snapshot. Two reviewers converging on the same imprecision is a good signal it was worth the round, and it is now fixed in all three places it appeared.

Gates re-run before pushing: `anvil-clippy` clean, nightly rustfmt clean, `cargo test --all-features` green for `ohno` / `ohno_macros` / `cachet_tier` including the regenerated trybuild snapshot. CI on the previous head `1ba9aa08` was 37 passing / 0 failing / 6 pending; the new head will be checked next pass rather than assumed.

sandersaares (approved):

PR #654, opened 2026-08-10
perf(tick): avoid allocations when deserializing timestamps

martintmk:
## Summary

- parse textual `tick::fmt` serde inputs directly from borrowed strings instead of allocating an owned `String`
- preserve valid UTF-8 byte input compatibility without introducing allocations
- keep `UnixSeconds` on its direct numeric deserialization path
- add allocation regression coverage for all four formats

## Work item

- [AB#7667382](https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7667382)


crates/tick/src/fmt/mod.rs
```diff
@@ -136,13 +136,58 @@ mod iso_8601;
...
+
 pub use ecmascript::EcmaScript;
 pub use iso_8601::Iso8601;
 pub use rfc_2822::Rfc2822;
 pub use unix_seconds::UnixSeconds;
 
 use crate::Error;
 
+#[cfg(any(feature = "serde", test))]
+fn deserialize_from_str<'de, D, T>(deserializer: D) -> Result<T, D::Error>
```
martintmk, line 154: is this really necessary, some simpler built-in way to deserialize from str?

crates/tick/src/fmt/mod.rs
```diff
@@ -136,13 +136,58 @@ mod iso_8601;
...
+        T: FromStr,
+        T::Err: Display,
+    {
+        type Value = T;
+
+        fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+            formatter.write_str("a timestamp string")
+        }
+
+        fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
```
sgalkin: Could we preserve the previous behavior for deserializers that provide strings through `visit_bytes` or `visit_byte_buf`? `String::deserialize` accepts valid UTF-8 byte input, while this visitor inherits Serde's rejecting default; implementing `visit_bytes` using `str::from_utf8` would retain compatibility without allocating.

🤖 Authored by Copilot.

crates/tick/src/fmt/mod.rs
martintmk: &#x1F916; Serde does not provide a built-in generic `deserialize_from_str` helper. The visitor is the minimal dependency-free mechanism that lets the deserializer borrow its input; using `String::deserialize` necessarily allocates, while `serde_with::DisplayFromStr` would add a dependency and still needs equivalent visitor behavior.

---

crates/tick/src/fmt/mod.rs
martintmk: &#x1F916; Agreed - preserving valid UTF-8 byte input keeps the previous `String::deserialize` behavior without reintroducing allocations. I am implementing this now.

---

martintmk: &#x1F916; Fixed in `38d0ab06`: the shared visitor now accepts valid UTF-8 byte input through `visit_bytes`, which also preserves `visit_byte_buf` behavior via Serde's default forwarding, with regression tests for valid and invalid UTF-8.

---

crates/tick/src/fmt/mod.rs
```diff
@@ -136,13 +136,65 @@ mod iso_8601;
...
+
 pub use ecmascript::EcmaScript;
 pub use iso_8601::Iso8601;
 pub use rfc_2822::Rfc2822;
 pub use unix_seconds::UnixSeconds;
 
 use crate::Error;
 
+#[cfg(any(feature = "serde", test))]
+fn deserialize_from_str<'de, D, T>(deserializer: D) -> Result<T, D::Error>
```
martintmk: can you move this to serde module?

martintmk: &#x1F916; Agreed - I am moving the shared deserialization visitor into a private `fmt::serde` module now.

---

martintmk: &#x1F916; Fixed in `983ed359`: the shared visitor now lives in private `fmt::serde`, and each textual format delegates to `serde::deserialize_from_str`.

---

crates/tick/benches/tick_serde_deserialize.rs
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) Microsoft Corporation.
```
martintmk: drop the new benchmarks

martintmk: &#x1F916; Agreed - I am removing both new benchmark targets and their benchmark-only dependencies while keeping the allocation regression tests.

---

martintmk: &#x1F916; Fixed in `798e9575`: both serde benchmark targets and their benchmark-only dependencies have been removed; the zero-allocation regression tests remain.

---

crates/tick/src/fmt/serde.rs
```diff
@@ -0,0 +1,50 @@
...
+        fn visit_bytes<E>(self, v: &[u8]) -> Result<Self::Value, E>
+        where
+            E: serde_core::de::Error,
+        {
+            std::str::from_utf8(v).map_err(E::custom)?.parse().map_err(E::custom)
+        }
+    }
+
+    deserializer.deserialize_str(FromStrVisitor(PhantomData))
+}
```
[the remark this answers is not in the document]
martintmk: &#x1F916; I do not think `deserialize_any` is the right tradeoff here. Non-self-describing formats are allowed not to support `deserialize_any`, so switching would regress formats that can deserialize strings directly. The `textual_format_deserializes_utf8_bytes` regression test uses Serde's `BytesDeserializer` and confirms the current `deserialize_str` path reaches `visit_bytes`, preserving byte input without narrowing format compatibility.

---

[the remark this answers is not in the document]
wukchung: &#x1F916; Different root cause on this same line, worth reopening: `deserialize_any` is the wrong fix for the reason you gave, but `deserialize_str` is still a compatibility narrowing — just not because of `visit_bytes`.

The visitor method set is fine. Serde's defaults forward `visit_string` → `visit_str` and `visit_byte_buf` / `visit_borrowed_bytes` → `visit_bytes`, so no callback is lost. The regression is in the **hint**: the old code went through `String::deserialize`, which calls `deserialize_string`; this calls `deserialize_str`. Those are not interchangeable in ciborium 0.2.2 (`src/de/mod.rs`):

- `deserialize_str` (L320-339) accepts only `Header::Text(Some(len)) if len <= self.scratch.len()`, and the catch-all arm errors on everything else.
- `deserialize_string` (L341-362) matches `Header::Text(len)` for *any* `len` and streams the segments into a `String`.

ciborium's own `deserialize_any` (L144-147) routes `Text(None)` and oversized text to `deserialize_string` rather than `deserialize_str`, so this is the intended split, not an accident.

So CBOR indefinite-length text strings (`0x7F <chunk>… 0xFF`, RFC 8949 §3.2.3) used to deserialize and now fail. Verified on this branch by adding a `ciborium` dev-dependency and hand-encoding `2024-08-06T21:30:00Z` as two segments:

- with `main`'s `String::deserialize` body: `Ok(2024-08-06T21:30:00Z)`
- with this branch as-is: `Err(Semantic(None, "invalid type: string, expected str"))`

Definite-length text longer than the scratch buffer regresses identically, which is reachable for `from_reader_with_buffer` callers passing a small buffer, even for a 20-byte timestamp.

Switching the hint to `deserialize_string` restores exactly the pre-PR format compatibility and costs nothing for the goal of this PR — `serde_json`'s `deserialize_string` is a straight forward to `deserialize_str` (`serde_json/src/de.rs` L1551-1556), so the borrowed JSON path stays allocation-free:

```rust
deserializer.deserialize_string(FromStrVisitor(PhantomData))
```

I confirmed both properties with that one-line change applied: `borrowed_json_deserialization_is_allocation_free` still passes for all four formats, and the indefinite-length CBOR input deserializes again.

Worth noting the current test suite cannot catch this — `StrDeserializer` and `BytesDeserializer` invoke the visitor directly and never exercise a real deserializer's `deserialize_str` implementation, so a `ciborium` (or similar) round-trip case would be the regression guard here.

Vaiz (approved):

ralfbiedert (approved):
🤖 Reviewed at head 7b22745b66b003967f5a3e97adcf979cb8b55628. No blocking correctness, performance, API, or safety issues found.

wukchung (approved):

crates/tick/src/fmt/serde.rs
```diff
@@ -0,0 +1,50 @@
...
+use std::marker::PhantomData;
+use std::str::FromStr;
+
+pub(super) fn deserialize_from_str<'de, D, T>(deserializer: D) -> Result<T, D::Error>
+where
+    D: serde_core::Deserializer<'de>,
+    T: FromStr,
+    T::Err: Display,
+{
+    struct FromStrVisitor<T>(PhantomData<T>);
```
sandersaares, line 18: [Copilot speaking]

# Explain why textual Serde deserialization uses a custom visitor

`deserialize_from_str` replaces the previous `String::deserialize(...).parse()` path with a custom `Visitor`, but the new module does not state why the extra mechanism is necessary. This is a hand-written alternative to the obvious owned-string implementation, which `docs/performance.md` requires to be justified beside the code. The refreshed discussion separately identifies the current `deserialize_str` hint as a compatibility regression, so the rationale requested here is limited to the custom visitor and must not defend that hint.

**Reproducible reasoning**: The three textual timestamp wrappers now call this helper. `String::deserialize` invokes Serde's owned `String` visitor; when a deserializer supplies borrowed text, that visitor copies the slice into a new `String`, while this visitor parses the slice delivered to `visit_str` or `visit_borrowed_str` directly. The allocation regression test confirms that the resulting JSON path allocates zero bytes. `visit_bytes` converts valid UTF-8 bytes to `&str`, retaining an input form accepted by the former `String` visitor. These are the non-obvious benefits of the custom visitor that the adjacent comment needs to preserve. The refreshed agent comment demonstrates that the old `String::deserialize` path requested `deserialize_string`, while the new helper requests `deserialize_str`; those hints differ for deserializers such as ciborium, so the earlier approved text's claim that `deserialize_str` preserves format compatibility is not valid and has been removed.

**Consequence**: Without the rationale beside the helper, a cleanup may reintroduce a per-value allocation through `String::deserialize` or remove the valid UTF-8 byte handling that preserves previously accepted input. Either change would undo behavior this pull request deliberately tests.

**Recommended action**: After resolving the separately reported deserializer-hint regression, add a concise module- or function-level comment above `deserialize_from_str` that names the rejected owned-`String` path, states that direct slice parsing avoids its allocation, and explains `visit_bytes` as compatibility with previously accepted valid UTF-8 byte input. Keep the comment scoped to the custom visitor; do not present the current `deserialize_str` hint as compatibility-preserving.

**References**:

- docs/performance.md, Justify deviations from standard patterns
- crates/tick/tests/serde_allocations.rs
- [Discussion of the missing built-in FromStr helper](https://github.com/microsoft/oxidizer/pull/654#discussion_r3749541632)
- [Agent report of the deserialize_str compatibility regression](https://github.com/microsoft/oxidizer/pull/654#discussion_r3765544916)

**Impacted locations**:

- `crates/tick/src/fmt/serde.rs:8-50`

sandersaares (commented):
[Copilot speaking]

| Diagnostic | Value |
|------------|-------|
| Cache | Miss |

</details>

crates/tick/src/fmt/serde.rs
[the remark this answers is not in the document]
martintmk: &#x1F916; Agreed - the `deserialize_string` versus `deserialize_str` distinction is a real compatibility issue for streamed text formats. I am restoring the original hint and adding a regression test.

---

crates/tick/src/fmt/serde.rs
martintmk: &#x1F916; Agreed - I am adding an adjacent rationale that explains why the custom visitor avoids the owned-`String` allocation and why valid UTF-8 byte handling is retained, without claiming the current hint is compatibility-preserving.

---

crates/tick/src/fmt/serde.rs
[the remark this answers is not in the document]
martintmk: &#x1F916; Fixed in `7aa90aff`: the helper now requests `deserialize_string`, preserving streamed and oversized text behavior while the visitor still parses borrowed JSON without allocating. A string-hint regression test prevents narrowing this back to `deserialize_str`.

---

crates/tick/src/fmt/serde.rs
martintmk: &#x1F916; Fixed in `7aa90aff`: the helper now documents why it avoids the owned-`String` path, why the `deserialize_string` hint is retained, and why valid UTF-8 byte handling remains.

---

Vaiz (approved):

sandersaares (approved):

martintmk (commented):
[Copilot speaking]

No blocking correctness, public API, compatibility, dependency, or test issues found. The private visitor preserves the prior deserializer hint and UTF-8 byte behavior while removing the borrowed-JSON allocation, with focused regression coverage.

Verdict: approve.

crates/tick/tests/serde_allocations.rs
```diff
@@ -0,0 +1,114 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Allocation contracts for timestamp deserialization.
+
+#![cfg(all(feature = "fmt", feature = "serde", not(miri)))]
+#![allow(clippy::unwrap_used, reason = "test code uses unwrap for concise failure backtraces")]
```
psandana: Could we use `#![expect(clippy::unwrap_used, reason = "test code uses unwrap for concise failure backtraces")]` here? M-LINT-OVERRIDE-EXPECT requires scoped lint overrides to use `#[expect]` rather than `#[allow]`, so the override becomes visible if these unwraps are later removed.

martintmk: &#x1F916; Agreed - I am replacing the scoped `allow` with `expect` so the override becomes stale-visible when the test no longer uses `unwrap`.

---

martintmk: &#x1F916; Fixed in `1b49eb85`: the test now uses `#![expect(clippy::unwrap_used, ...)]`, so the override is reported if it becomes unnecessary.

---

sandersaares (approved):

PR #655, opened 2026-08-11
feat(testing_aids): add manual future-polling test helpers

Vaiz:
Ports the manual future-polling test helpers from the ox-sdk `oxidizer_futures_test` crate into this repo's `testing_aids` crate, and applies them in two places as a proof of concept.

## What is added

`testing_aids::poll` (re-exported from the crate root):

- `poll_once(&mut fut)` — polls an `Unpin` future exactly once with a noop waker. For tests that need to inspect external state between polls.
- `FutureTestExt` — a blanket-implemented extension trait with consuming assertions:
  - `unwrap_ready()` — poll once, return the output, panic if `Pending`
  - `unwrap_pending()` — poll once, assert `Pending`, drop
  - `unwrap_ready_within(max_polls, msg)` — poll up to N times, return the first `Ready`
  - `unwrap_ready_after(n_pending, msg)` — assert exactly N `Pending` polls then `Ready`, which catches off-by-one timing bugs in state machines
  - `unwrap_pending_for(n, msg)` — assert N consecutive `Pending` polls

Every `FutureTestExt` method takes the future **by value** and pins it internally, so they work directly on `!Unpin` futures such as `async` blocks with no caller-side `pin!`.

## Proof of concept

- `bytesbuf_io/src/testing/pending.rs` — four `*_returns_pending_on_first_poll` tests each drop a `pin!` + `Waker::noop()` + `Context::from_waker` + `matches!` block down to a single call.
- `cachet/src/telemetry/cache.rs` — `nested_with_request_id_restores_outer_id` had two nested hand-built poll harnesses; both collapse to `.unwrap_ready()`.

Net effect on the two call sites: 59 lines removed, 33 added.

## Effects

- Tests that drive a future by hand no longer repeat waker and pinning setup, so the assertion is what the reader sees. The same tests now depend on a shared helper rather than on locally visible `std::task` calls.
- `unwrap_ready_after` and `unwrap_pending_for` make poll-count expectations explicit, which turns a silent change in polling behaviour into a failure. Tests that state a poll count are correspondingly tighter and will need updating if the polling schedule legitimately changes.
- The helpers panic rather than return `Result`, which keeps call sites to one line and confines them to test code.
- `testing_aids` grows a fifth module and no new dependencies.
- The port duplicates code that also exists in ox-sdk. The two copies can drift; `oxidizer_futures_test` remains the upstream reference.

## Related

I filed a proposal to add this into future_tests crate https://github.com/rust-lang/futures-rs/issues/3018, but haven't gotten any reaction yet.

### DIFF
--- a/crates/testing_aids/src/poll.rs
+++ b/crates/testing_aids/src/poll.rs
@@ -283,8 +283,10 @@ mod tests {
         assert_eq!(polls.get(), 3);
     }
 
-    /// An output type deliberately without a [`Debug`](std::fmt::Debug) impl,
-    /// pinning that the helpers stay usable with opaque outputs.
+    /// An output type deliberately without a [`Debug`](std::fmt::Debug) impl.
+    ///
+    /// Any helper that requires one would fail to compile against this type,
+    /// so it keeps the assertions usable with opaque outputs.
     struct NotDebug(u32);
 
     /// Returns `Pending` on its first poll, then `Ready`.
--- a/crates/testing_aids/src/poll.rs
+++ b/crates/testing_aids/src/poll.rs
@@ -10,7 +10,6 @@
 //! lower-level [`poll_once`] primitive performs exactly one poll of an unpinned
 //! future.
 
-use std::fmt::Debug;
 use std::task::{Context, Poll, Waker};
 
 /// Polls a future exactly once with a no-op waker.
@@ -103,18 +102,12 @@ pub trait FutureTestExt: Future + Sized {
     /// Panics (using `message_if_not_pending`) if any of the first `n_pending`
     /// polls returns `Ready`, or if the final poll is still `Pending`.
     #[track_caller]
-    fn unwrap_ready_after(self, n_pending: usize, message_if_not_pending: &str) -> Self::Output
-    where
-        Self::Output: Debug,
-    {
+    fn unwrap_ready_after(self, n_pending: usize, message_if_not_pending: &str) -> Self::Output {
         let mut fut = std::pin::pin!(self);
         for i in 0..n_pending {
             match poll_once(&mut fut) {
                 Poll::Pending => {}
-                Poll::Ready(value) => panic!(
-                    "{message_if_not_pending}: got Ready({value:?}) after {} polls, expected Pending",
-                    i + 1
-                ),
+                Poll::Ready(_) => panic!("{message_if_not_pending}: got Ready after {} polls, expected Pending", i + 1),
             }
         }
 
@@ -288,6 +281,34 @@ mod tests {
         let (fut, polls) = ScriptedFuture::with_poll_count(vec![Poll::Pending, Poll::Pending, Poll::Ready(9)]);
         assert_eq!(fut.unwrap_ready_after(2, "should be pending"), 9);
         assert_eq!(polls.get(), 3);
+    }
+
+    /// An output type deliberately without a [`Debug`](std::fmt::Debug) impl,
+    /// pinning that the helpers stay usable with opaque outputs.
+    struct NotDebug(u32);
+
+    /// Returns `Pending` on its first poll, then `Ready`.
+    struct PendingOnce {
+        polled: bool,
+    }
+
+    impl Future for PendingOnce {
+        type Output = NotDebug;
+
+        fn poll(mut self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<NotDebug> {
+            if self.polled {
+                Poll::Ready(NotDebug(4))
+            } else {
+                self.polled = true;
+                Poll::Pending
+            }
+        }
+    }
+
+    #[test]
+    fn unwrap_ready_after_accepts_non_debug_output() {
+        let value = PendingOnce { polled: false }.unwrap_ready_after(1, "should be pending once");
+        assert_eq!(value.0, 4);
     }
 
     #[test]
--- a/crates/testing_aids/src/poll.rs
+++ b/crates/testing_aids/src/poll.rs
@@ -47,6 +47,7 @@ pub trait FutureTestExt: Future + Sized {
     /// # Panics
     ///
     /// Panics if the future returns [`Poll::Pending`].
+    #[track_caller]
     fn unwrap_ready(self) -> Self::Output {
         let mut fut = std::pin::pin!(self);
         match poll_once(&mut fut) {
@@ -60,6 +61,7 @@ pub trait FutureTestExt: Future + Sized {
     /// # Panics
     ///
     /// Panics if the future returns [`Poll::Ready`].
+    #[track_caller]
     fn unwrap_pending(self) {
         let mut fut = std::pin::pin!(self);
         assert!(
@@ -73,6 +75,7 @@ pub trait FutureTestExt: Future + Sized {
     /// # Panics
     ///
     /// Panics with `timeout_msg` if the future never completes within `max_polls` polls.
+    #[track_caller]
     fn unwrap_ready_within(self, max_polls: usize, timeout_msg: &str) -> Self::Output {
         let mut fut = std::pin::pin!(self);
         for _ in 0..max_polls {
@@ -93,6 +96,7 @@ pub trait FutureTestExt: Future + Sized {
     ///
     /// Panics (using `message_if_not_pending`) if any of the first `n_pending`
     /// polls returns `Ready`, or if the final poll is still `Pending`.
+    #[track_caller]
     fn unwrap_ready_after(self, n_pending: usize, message_if_not_pending: &str) -> Self::Output
     where
         Self::Output: Debug,
@@ -110,7 +114,10 @@ pub trait FutureTestExt: Future + Sized {
 
         match poll_once(&mut fut) {
             Poll::Ready(value) => value,
-            Poll::Pending => panic!("expected Ready after {} polls, but got Pending", n_pending + 1),
+            Poll::Pending => panic!(
+                "{message_if_not_pending}: expected Ready after {} polls, but got Pending",
+                n_pending + 1
+            ),
         }
     }
 
@@ -118,8 +125,11 @@ pub trait FutureTestExt: Future + Sized {
     ///
     /// # Panics
     ///
-    /// Panics (using `message`) if any poll returns `Ready`.
+    /// Panics (using `message`) if any poll returns `Ready`, and panics if `n`
+    /// is zero, because zero polls would assert nothing.
+    #[track_caller]
     fn unwrap_pending_for(self, n: usize, message: &str) {
+        assert!(n > 0, "unwrap_pending_for requires n > 0; zero polls would assert nothing");
         let mut fut = std::pin::pin!(self);
         for _ in 0..n {
             assert!(poll_once(&mut fut).is_pending(), "{message}");
@@ -131,22 +141,38 @@ impl<F: Future> FutureTestExt for F {}
 
 #[cfg(test)]
 mod tests {
+    use std::cell::Cell;
     use std::collections::VecDeque;
     use std::pin::Pin;
+    use std::rc::Rc;
 
     use super::*;
 
-    /// A future returning a scripted sequence of polls, defaulting to `Pending`
-    /// once the script is exhausted.
+    /// A future returning a scripted sequence of polls.
+    ///
+    /// It counts every poll it receives and panics once the script is
+    /// exhausted, so a helper that polls more times than it documents fails
+    /// loudly instead of silently observing `Pending` forever.
     struct ScriptedFuture {
         steps: VecDeque<Poll<u32>>,
+        polls: Rc<Cell<usize>>,
     }
 
     impl ScriptedFuture {
         fn new(steps: Vec<Poll<u32>>) -> Self {
             Self {
                 steps: VecDeque::from(steps),
+                polls: Rc::new(Cell::new(0)),
             }
+        }
+
+        /// Returns the future together with a handle that observes how many
+        /// times it has been polled, so a test can assert the exact poll count
+        /// after the helper has consumed the future.
+        fn with_poll_count(steps: Vec<Poll<u32>>) -> (Self, Rc<Cell<usize>>) {
+            let future = Self::new(steps);
+            let polls = Rc::clone(&future.polls);
+            (future, polls)
         }
     }
 
@@ -154,7 +180,10 @@ mod tests {
         type Output = u32;
 
         fn poll(mut self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<u32> {
-            self.steps.pop_front().unwrap_or(Poll::Pending)
+            self.polls.set(self.polls.get() + 1);
+            self.steps
+                .pop_front()
+                .unwrap_or_else(|| panic!("ScriptedFuture polled after its script was exhausted"))
         }
     }
 
@@ -165,8 +194,23 @@ mod tests {
     }
 
     #[test]
+    fn poll_once_returns_pending_and_can_be_repeated() {
+        let (mut fut, polls) = ScriptedFuture::with_poll_count(vec![Poll::Pending, Poll::Ready(5)]);
+        assert_eq!(poll_once(&mut fut), Poll::Pending);
+        assert_eq!(poll_once(&mut fut), Poll::Ready(5));
+        assert_eq!(polls.get(), 2, "each call must poll the future exactly once");
+    }
+
+    #[test]
     fn unwrap_ready_returns_value() {
         assert_eq!(ScriptedFuture::new(vec![Poll::Ready(42)]).unwrap_ready(), 42);
+    }
+
+    #[test]
+    fn unwrap_ready_polls_exactly_once() {
+        let (fut, polls) = ScriptedFuture::with_poll_count(vec![Poll::Ready(42)]);
+        assert_eq!(fut.unwrap_ready(), 42);
+        assert_eq!(polls.get(), 1);
     }
 
     #[test]
@@ -187,6 +231,13 @@ mod tests {
     }
 
     #[test]
+    fn unwrap_pending_polls_exactly_once() {
+        let (fut, polls) = ScriptedFuture::with_poll_count(vec![Poll::Pending]);
+        fut.unwrap_pending();
+        assert_eq!(polls.get(), 1);
+    }
+
+    #[test]
     #[should_panic(expected = "Pending after one poll")]
     fn unwrap_pending_panics_when_ready() {
         ScriptedFuture::new(vec![Poll::Ready(3)]).unwrap_pending();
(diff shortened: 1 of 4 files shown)


crates/testing_aids/src/poll.rs
```diff
@@ -0,0 +1,238 @@
...
+        }
+    }
+
+    /// Polls `n` times, asserting each poll returns `Pending`, then drops the future.
+    ///
+    /// # Panics
+    ///
+    /// Panics (using `message`) if any poll returns `Ready`.
+    fn unwrap_pending_for(self, n: usize, message: &str) {
+        let mut fut = std::pin::pin!(self);
```
wukchung, line 132: 🤖 **`unwrap_pending_for(0, ..)` fails open — it passes without polling at all.**

`for _ in 0..n` is empty when `n == 0`, so the future is constructed and dropped having never been polled, and the helper returns successfully. This is the only method in the file that can report "assertion held" while asserting nothing; `unwrap_ready_within(0, ..)` has the mirror-image degenerate case but fails *safe* (panics with `timeout_msg`).

Reproduced against this branch — an immediately-`Ready` future passes the "stay pending" assertion:

```rust
#[test]
fn zero_pending_polls_passes_on_ready_future() {
    std::future::ready(1_u32).unwrap_pending_for(0, "should stay pending");
    // reached: the assertion was vacuous
}
```

No current call site is affected — all four in `bytesbuf_io` pass a literal `1`. But this is a helper being introduced for repo-wide use, and any caller deriving `n` from a computed value (`iterations - 1`, a config constant) that degrades to `0` gets a permanently green test with nothing to flag it.

```suggestion
    fn unwrap_pending_for(self, n: usize, message: &str) {
        assert!(n > 0, "unwrap_pending_for requires n > 0; zero polls would assert nothing");
        let mut fut = std::pin::pin!(self);
```

Worth a `#[should_panic]` regression test for the `n = 0` path too.

crates/testing_aids/src/poll.rs
```diff
@@ -0,0 +1,238 @@
...
+                Poll::Ready(value) => panic!(
+                    "{message_if_not_pending}: got Ready({value:?}) after {} polls, expected Pending",
+                    i + 1
+                ),
+            }
+        }
+
+        match poll_once(&mut fut) {
+            Poll::Ready(value) => value,
+            Poll::Pending => panic!("expected Ready after {} polls, but got Pending", n_pending + 1),
```
wukchung: 🤖 **`unwrap_ready_after` never uses `message_if_not_pending` on this branch, contradicting its own `# Panics` docs.**

The doc comment on lines 94-95 says it panics *"(using `message_if_not_pending`) if any of the first `n_pending` polls returns `Ready`, **or if the final poll is still `Pending`**"*. The first clause is true (line 104), but this branch panics with a hardcoded string and drops the caller's message entirely.

Reproduced on this branch:

```
thread '...' panicked at crates/testing_aids/src/poll.rs:113:30:
expected Ready after 2 polls, but got Pending
note: panic did not contain expected string
      panic message: "expected Ready after 2 polls, but got Pending"
 expected substring: "custom message"
```

"never became ready" is arguably the more common failure mode of this helper, so it is the one that most needs the caller's context. The existing test `unwrap_ready_after_panics_when_never_ready` (line 222) passes `"should be pending"` but then asserts only on `"but got Pending"`, so it silently confirms the message is dropped rather than catching it.

Either include the message (fits within `max_width = 140`):

```suggestion
            Poll::Pending => panic!("{message_if_not_pending}: expected Ready after {} polls, but got Pending", n_pending + 1),
```

…or narrow the doc on line 94-95 to scope the parenthetical to the early-`Ready` branch only.

wukchung (commented):
🤖 Reviewed the new `testing_aids::poll` helpers and both call sites.

I confirmed the parts that matter most are sound: the pinning chain (`pin!(self)` → `&mut Pin<&mut Self>` → `poll_once`) is correct and works on `!Unpin` futures as documented, no helper polls a future after it returned `Ready`, and both rewritten call sites are semantically equivalent to the hand-written harnesses they replace — nothing was silently weakened. `cargo test -p testing_aids`, `-p bytesbuf_io --lib` and `-p cachet --lib nested_with_request_id` all pass, and `cargo doc` with `-D warnings` is clean.

Four points below, two of which I reproduced locally.

crates/testing_aids/src/poll.rs
```diff
@@ -0,0 +1,238 @@
...
+/// assert_eq!(async { 7 }.unwrap_ready(), 7);
+/// ```
+pub trait FutureTestExt: Future + Sized {
+    /// Polls the future exactly once and returns its output, panicking if it is
+    /// still `Pending`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the future returns [`Poll::Pending`].
+    fn unwrap_ready(self) -> Self::Output {
```
wukchung, line 56: 🤖 **Missing `#[track_caller]` on the five panicking helpers — failures now point into `poll.rs` instead of the failing test.**

None of `unwrap_ready` (line 50), `unwrap_pending` (63), `unwrap_ready_within` (76), `unwrap_ready_after` (96) or `unwrap_pending_for` (122) carry `#[track_caller]`, so every panic they raise is attributed to a line inside this file rather than to the assertion's call site. Confirmed on this branch:

```
thread '...' panicked at crates/testing_aids/src/poll.rs:125:13:
should stay pending
```

This is a regression against the harnesses being replaced: those used `assert!` inline in the test body and reported the test's own file and line. It bites hardest in `bytesbuf_io/src/testing/pending.rs`, where four tests now call `unwrap_pending_for` and would all report the same `poll.rs:125` location on failure.

The repo already applies this pattern to its assertion helpers — see `crates/ohno/src/test_util.rs:49`. `#[track_caller]` propagates through default trait method bodies, so adding it to all five declarations is sufficient.

crates/testing_aids/src/poll.rs
```diff
@@ -0,0 +1,238 @@
...
+                steps: VecDeque::from(steps),
+            }
+        }
+    }
+
+    impl Future for ScriptedFuture {
+        type Output = u32;
+
+        fn poll(mut self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<u32> {
+            self.steps.pop_front().unwrap_or(Poll::Pending)
```
wukchung: 🤖 **`ScriptedFuture` cannot detect over-polling, so the helpers' central "exactly N polls" contract is untested.**

`self.steps.pop_front().unwrap_or(Poll::Pending)` makes an exhausted script return `Pending` forever, so every happy-path test in this module still passes if the helper polls *more* times than documented:

- `unwrap_pending_passes_when_pending` (line 185) passes even if `unwrap_pending` polled twice.
- `unwrap_pending_for_passes` (line 229) passes even if `unwrap_pending_for` polled `n + 1` times.

Over-polling is the failure mode that actually matters here, because polling an `async fn` or `async {}` future after it returns `Ready` panics with `` `async fn` resumed after completion `` — precisely the class of bug these helpers exist to make impossible for their users. `unwrap_ready`, documented as "polls the future exactly once", has no test that would fail if it polled twice.

Having `ScriptedFuture` count its polls (or panic once the script is exhausted) and asserting the exact count after each helper returns would close this. Related smaller gaps: `poll_once`'s `Pending` return path is untested (line 162 covers only `Ready`), as is its documented "poll the same future from several places" repeated-poll use.

crates/testing_aids/src/poll.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Agreed and fixed. `unwrap_pending_for` now rejects `n == 0` up front:

```rust
assert!(n > 0, "unwrap_pending_for requires n > 0; zero polls would assert nothing");
```

The `# Panics` doc now states that condition, and there is a `#[should_panic(expected = "requires n > 0")]` regression test (`unwrap_pending_for_panics_on_zero_polls`) that calls it on an immediately-`Ready` future — exactly your reproducer, which passed before this change and now panics.

I left `unwrap_ready_within(0, ..)` alone deliberately: it fails safe with the caller's `timeout_msg`, so a degenerate `0` there is already reported rather than hidden.

Pushed in `fb8b4f9b`.

crates/testing_aids/src/poll.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Correct, and I took the first option — the docs were right and the code was wrong. "Never became ready" is the failure mode that most needs the caller's context, so the message is now included:

```rust
Poll::Pending => panic!(
    "{message_if_not_pending}: expected Ready after {} polls, but got Pending",
    n_pending + 1
),
```

You were also right that the existing test was confirming the bug rather than catching it. `unwrap_ready_after_panics_when_never_ready` now passes `"custom context"` and asserts on the full string `"custom context: expected Ready after 2 polls, but got Pending"`, so a future regression that drops the message fails the test.

Pushed in `fb8b4f9b`.

crates/testing_aids/src/poll.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Accepted — this was a genuine regression against the hand-written harnesses, and the `bytesbuf_io` case you point at (four tests all reporting the same `poll.rs` line) is the one that would have hurt most.

`#[track_caller]` is now on all five declarations (`unwrap_ready`, `unwrap_pending`, `unwrap_ready_within`, `unwrap_ready_after`, `unwrap_pending_for`). As you note, it propagates through the default trait method bodies, so the attribute on the declaration is sufficient. `poll_once` is left without it: it returns `Poll` and never panics.

Pushed in `fb8b4f9b`.

crates/testing_aids/src/poll.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

This was the most valuable of the four — the "exactly N polls" contract was genuinely unpinned, and I have closed it both ways you suggested.

`ScriptedFuture` now:
- counts every poll it receives, exposed via `with_poll_count(steps) -> (Self, Rc<Cell<usize>>)` so a test can assert the count *after* the helper has consumed the future;
- panics (`"ScriptedFuture polled after its script was exhausted"`) instead of returning `Pending` forever, so over-polling fails loudly even in tests that do not assert a count.

New exact-count tests: `unwrap_ready_polls_exactly_once`, `unwrap_pending_polls_exactly_once`, `unwrap_ready_within_stops_polling_once_ready` (3 polls, not the `max_polls` of 10), `unwrap_ready_after_polls_exactly_n_plus_one_times`, `unwrap_pending_for_polls_exactly_n_times`. Plus `poll_once_returns_pending_and_can_be_repeated`, which covers both gaps you flagged on `poll_once`: the `Pending` return path and the documented repeated-poll use.

Pushed in `fb8b4f9b`.

Vaiz:

Answering the two suppressed comments from review [#pullrequestreview-4904526673](https://github.com/microsoft/oxidizer/pull/655#pullrequestreview-4904526673).

**1. `poll_once` doc comment does not mention `Unpin` — accepted, fixed in `dc30cba3`.**

Fair, and this is exactly the case that trips test authors up, since the futures they reach for first (`async fn`, `async {}`) are the ones that do not qualify. The doc now states the requirement, explains *why* it exists (the future is polled through a mutable reference the caller keeps across polls), and names both ways out — pin it with `std::pin::pin!`, or use the `FutureTestExt` methods, which take the future by value and pin internally. I also added `poll_once_accepts_a_pinned_async_block`, so the workaround the doc recommends is compiled and exercised rather than merely asserted in prose.

**2. `mod poll;` should be `pub mod poll;` — declining the code change.**

The private-module-plus-root-re-export shape is this crate's existing convention, not an oversight: `io`, `metrics` and `yielding` are all declared `mod` and re-exported with `pub use ...::*` from `lib.rs`. Making `poll` the sole `pub mod` would give its items two public paths while its siblings have one, for no benefit — the helpers are already reachable as `testing_aids::poll_once` and `testing_aids::FutureTestExt`, which is how both converted call sites import them.

If the PR description does say `testing_aids::poll` is the public module path, then the description is what is wrong, and it should read `testing_aids::{poll_once, FutureTestExt}`. I could not verify or correct that here: reading and editing the PR body needs a tool this automated run does not have, so I am flagging it rather than guessing. @eshutov, please check the description wording.

crates/testing_aids/src/poll.rs
```diff
@@ -0,0 +1,330 @@
...
+//!
+//! [`FutureTestExt`] adds consuming manual-polling assertions (`unwrap_ready`,
+//! `unwrap_pending`, `unwrap_ready_within`, `unwrap_ready_after`,
+//! `unwrap_pending_for`) that let a test poll a future a fixed number of times
+//! and assert the outcome, without spinning up an async runtime. The
+//! lower-level [`poll_once`] primitive performs exactly one poll of an unpinned
+//! future.
+
+use std::fmt::Debug;
+use std::task::{Context, Poll, Waker};
```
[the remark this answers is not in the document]
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

This one is incorrect — no change needed.

`Future` **is** in scope: it was added to the standard prelude in the 2024 edition (alongside `IntoFuture`), and this crate is on edition 2024. So `F: Future + Unpin` and `pub trait FutureTestExt: Future + Sized` resolve without any import.

The empirical check agrees: on this exact head the crate compiles clean under `cargo test -p testing_aids` (25 tests pass), the repo's `just anvil-clippy` gate, and `cargo doc -D warnings`. A missing trait import would be a hard `E0405`/`E0404`, not something any of those would let through.

The same pattern already exists elsewhere in this crate — `lib.rs` writes `FF: Future<Output = ()>` in `async_test` with no `use std::future::Future;` either.

crates/testing_aids/src/poll.rs
```diff
@@ -0,0 +1,330 @@
...
+    /// bugs in state machines.
+    ///
+    /// # Panics
+    ///
+    /// Panics (using `message_if_not_pending`) if any of the first `n_pending`
+    /// polls returns `Ready`, or if the final poll is still `Pending`.
+    #[track_caller]
+    fn unwrap_ready_after(self, n_pending: usize, message_if_not_pending: &str) -> Self::Output
+    where
+        Self::Output: Debug,
```
sgalkin: Could we avoid requiring `Self::Output: Debug` here? The bound is only used to format an unexpectedly early value, so it prevents this timing assertion from working with opaque outputs even though the successful path simply returns them. Matching `Poll::Ready(_)` and omitting the value from the panic would keep the API generic.

🤖 Co-authored with Copilot.

sgalkin (approved):

crates/testing_aids/src/poll.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Agreed — done in `ce777c77`.

You are right that the bound bought very little: it only formatted a value on the early-`Ready` path, while the success path returns that same value opaquely. The panic now matches `Poll::Ready(_)` and reads:

```
{message_if_not_pending}: got Ready after {n} polls, expected Pending
```

The poll count plus the caller's message are what actually identify the failure, so nothing diagnostic is lost that matters. `use std::fmt::Debug` is gone with it.

I also added `unwrap_ready_after_accepts_non_debug_output`, which drives a future whose `Output` has no `Debug` impl — it would not compile under the old bound, so it pins the generality rather than just asserting it in prose.

Gates re-run after a rebase onto `main`: `anvil-clippy`, nightly rustfmt `--check`, `cargo doc -D warnings`, `cargo-mutants` on `poll.rs`, and tests for `testing_aids` / `bytesbuf_io` / `cachet` are all clean.

crates/testing_aids/src/poll.rs
```diff
@@ -0,0 +1,353 @@
...
+    fn unwrap_ready_within(self, max_polls: usize, timeout_msg: &str) -> Self::Output {
+        let mut fut = std::pin::pin!(self);
+        for _ in 0..max_polls {
+            if let Poll::Ready(value) = poll_once(&mut fut) {
+                return value;
+            }
+        }
+        panic!("{timeout_msg}");
+    }
+
```
sandersaares, line 93: [Copilot speaking]

# Remove the wake-blind repeated-poll helpers

Keep the single-poll APIs, but remove `unwrap_ready_within`, `unwrap_ready_after`, and `unwrap_pending_for`. A waker is the notification handle a future stores when it returns `Pending`; the executor polls that future again after the waker signals that progress may be possible. These helpers instead create a no-op waker, discard any notification, and immediately poll again. None of this pull request's consumers needs that behavior: the `cachet` test uses the single-poll `unwrap_ready`, and each `bytesbuf_io` test expresses a one-poll assertion that can use `unwrap_pending`.

**Reproducible reasoning**: `poll_once` builds every `Context` from `Waker::noop()`. Each repeated-poll helper calls `poll_once` in a loop, so after one call returns `Pending`, the next iteration polls again without observing a wake. Rust's `Future::poll` contract says a pending future arranges a wake when it can make progress and specifically warns clients not to poll repeatedly in a tight loop. Because a no-op waker discards that signal, these helpers cannot distinguish a future that is ready for another poll from one still waiting for an external event. Their tests do not close the gap: `ScriptedFuture` advances to its next result merely because it is polled again, without waking, so it models the helper's tight loop rather than executor behavior. Removing the three methods also eliminates the need to define edge semantics for zero limits, diagnostic wording, and exact poll schedules. The four changed `bytesbuf_io` calls preserve their stated behavior by changing `unwrap_pending_for(1, ...)` to `unwrap_pending()`.

**Consequence**: A hot-polled test can pass even when a future returns `Pending` without arranging the wake that a real executor needs, allowing production code to stall despite a green test. Conversely, a valid timer-, channel-, or I/O-backed future can exhaust `max_polls` before its external event occurs and fail even though it would complete normally under an executor.

**Recommended action**: Remove `unwrap_ready_within`, `unwrap_ready_after`, and `unwrap_pending_for`, their tests, and their module-level documentation. Replace the four `unwrap_pending_for(1, ...)` call sites with `unwrap_pending()`, retaining `poll_once`, `unwrap_ready`, and `unwrap_pending` for deliberate single-poll inspection. Add a multi-poll API later only for a concrete use case, using an executor or a harness that observes wake notifications before repolling.

**References**:

- [Rust `Future::poll` contract and runtime characteristics](https://doc.rust-lang.org/std/future/trait.Future.html)
- [Rust `Waker` contract](https://doc.rust-lang.org/std/task/struct.Waker.html)

**Impacted locations**:

- `crates/testing_aids/src/poll.rs:4-11`
- `crates/testing_aids/src/poll.rs:79-143`
- `crates/testing_aids/src/poll.rs:261-329`
- `crates/bytesbuf_io/src/testing/pending.rs:256-258`
- `crates/bytesbuf_io/src/testing/pending.rs:266-268`
- `crates/bytesbuf_io/src/testing/pending.rs:275-277`
- `crates/bytesbuf_io/src/testing/pending.rs:285-287`

sandersaares (commented):
[Copilot speaking]

| Diagnostic | Value |
|------------|-------|
| Cache | Hit |

</details>

crates/testing_aids/src/poll.rs
Vaiz: technically, you're right. but current implementation is good enough for most tests because they don't use wakers. If you find a need for a correct implementation, feel free to adjust the code.

sandersaares (approved):

PR #657, opened 2026-08-11
feat(internity): export frozen readers and guarantee dense handles

geeknoid:
Addresses a field report from a production XML stack that adopted `internity` as its string interner, replacing `lasso`. Each change below pairs the fix with the problem it solves.

## Export the frozen readers as concrete types

`freeze` returned an unnameable `impl Reader`, so a struct wanting to *hold* a frozen table had no way to write the field type and had to fall back to `Box<dyn Reader>` — an allocation plus virtual dispatch on the hottest read path, with no inlining through it.

Both readers are now public and returned concretely, renamed to match their lexicons:

| Before | After |
|---|---|
| `pub(crate) struct FlatReader` | `pub struct LocalReader` |
| `pub(crate) struct ShardedReader` | `pub struct ThreadedReader` |
| `LocalLexicon::freeze(self) -> impl Reader` | `LocalLexicon::freeze(self) -> LocalReader` |
| `ThreadedLexicon::freeze(self) -> impl Reader` | `ThreadedLexicon::freeze(self) -> ThreadedReader` |

**Not a breaking change**: a concrete return type still satisfies code written against `impl Reader`, and the old names were never public. Renaming was free now and would have been breaking later. It also removes a near-collision with the existing internal `ShardReader` (one shard, versus all of them).

`Reader` remains sealed, so downstream crates still cannot implement it. That is the only guarantee sealing gives us: publishing these types *does* widen the supported surface, since their names, inherent methods, and auto-trait bounds all become compatibility commitments. That is the intended trade — being nameable is the point.

## Guarantee dense handles, and add safe conversions

Downstream code was already deriving a dense index from `as_u32() - 1` — an implementation detail we never promised to keep. Handles being consecutive is what makes a `Vec<T>` side table possible instead of a hash map keyed by the handle.

Adds `index_of` and `sym_at` to `LocalLexicon` and `LocalReader`, both returning `Option`, and documents consecutive insertion-order numbering as a guarantee that `freeze` preserves.

They live on the lexicon and reader rather than on `Sym` deliberately: `Sym` does not know which interner produced it, so a method there could not range-check, and `ThreadedLexicon` handles — which pack a shard index into their high bits and are **not** dense — could reach it by mistake. Putting the pair here means the conversion simply is not reachable when you only hold a `ThreadedReader`. Note this is about which types expose the method, not about which handles can reach it: `Sym` is one shared type, so a foreign handle that happens to be in range still returns a valid-but-wrong index. The docs say so, and say to keep side tables with the lexicon they describe.

`as_u32` / `from_u32` are unchanged: they are the opaque round-trip that works for *both* interners, which `index_of` cannot replace.

## Document the trade-offs on the types

The CSR layout costs 4 bytes per string where a `Vec<&str>` design costs 16, in exchange for one extra dependent load per resolve. That is a good trade, but a reader comparing against `lasso` should see it stated plainly rather than discover it.

The hash-DoS warning also moves onto the types themselves. It previously lived only in module-level prose, and there was nothing anywhere noting that `lasso` defaults to a collision-resistant hasher — so a type-for-type migration silently lost that protection.

## Document rebuilding a string lookup over a frozen reader

Freezing drops the dedup hash map, which *is* the memory saving, so a frozen reader can resolve but not look strings up. A sorted handle list costs ~4 bytes per string against the ~7 the hash map costs and answers lookups in `O(log n)`.

This needs no new API — `Reader::iter` and `Reader::resolve` are enough — so it ships as a doctest on `freeze` rather than as crate code. Building it in would have added permanent public API surface to save callers about 18 lines, with an unmeasured chance that bisection loses to simply keeping the hash map.

## Testing

- 5 new integration tests: index round-trip through lexicon *and* reader, out-of-range handling, a dense side table, storing both readers by value in a struct, and the external lookup pattern end to end.
- 4 new doctests, so the documented patterns cannot rot.
- `cargo clippy --all-targets --all-features`, tests under both `--all-features` and `--no-default-features`, format check, regenerated README, and spellcheck all pass locally. `just` itself could not run in this environment (its recipes require `pwsh`), so the underlying commands were invoked directly — CI is the real confirmation.


crates/internity/src/local_lexicon.rs
```diff
@@ -311,6 +341,85 @@ impl<S: BuildHasher> LocalLexicon<S> {
         self.offsets.len() - 1
     }
 
+    /// Returns the 0-based position of `sym` in insertion order, or `None` if it
+    /// is out of range for this lexicon.
+    ///
+    /// Handles are assigned consecutively from zero, so this index is a *dense*
+    /// key: per-symbol data can live in a `Vec<T>` indexed by it, rather than a
+    /// hash map keyed by the handle. That avoids hashing and probing entirely, and
```
wukchung: 🤖 **Docs say handles are numbered from zero, but `Sym` values are 1-based.**

`Sym::pack_dense` stores `index + 1` (`sym.rs:103-113`), so `as_u32()` returns 1 for the first interned string while `index_of` returns `Some(0)`. Verified on this head: after `intern("a")` then `intern("b")`, `as_u32()` is 1 and 2 while `index_of` is `Some(0)` and `Some(1)`.

It is the *positions* that start at zero, not the handles. That matters here specifically because this section formalises the guarantee meant to replace the downstream `as_u32() - 1` workaround the PR describes: a migrating reader who takes "handles are assigned consecutively from zero" literally and indexes a side table with `as_u32()` reproduces exactly the off-by-one this API exists to prevent, and runs off the end on the last entry.

The same claim appears at `src/lib.rs:86` ("numbers its handles consecutively from zero in insertion order"), which is the more prominent of the two and worth the same correction; the README picks it up on the next `just readme`.

```suggestion
    /// Positions are assigned consecutively from zero in insertion order, so this
    /// index is a *dense* key: per-symbol data can live in a `Vec<T>` indexed by
    /// it, rather than a hash map keyed by the handle. The raw handle value is
    /// 1-based, so `index_of`/`sym_at` are the supported conversions.
```

crates/internity/src/local_lexicon.rs
```diff
@@ -334,14 +443,60 @@ impl<S: BuildHasher> LocalLexicon<S> {
...
+    /// This trades interning for a smaller, read-optimized table: the dedup hash
+    /// map is dropped and the storage is shrunk to fit, which is where the memory
+    /// saving comes from. The consequence is that a frozen reader can
+    /// [`resolve`](Reader::resolve) handles but has no
+    /// [`get`](LocalLexicon::get) — looking a string up is exactly the hash map
+    /// that was just freed.
+    ///
+    /// If you need string lookup on a frozen table, keep the lexicon live instead,
+    /// or rebuild a smaller index from [`iter`](Reader::iter). A sorted handle
+    /// list costs 4 bytes per string against the roughly 18 the hash map costs,
```
wukchung: 🤖 **The "roughly 18" bytes per string for the dedup map looks about 2.5x too high.**

`dedup` is a `HashTable<Sym>` and `Sym` is a `NonZeroU32`, so hashbrown holds 4 bytes of value plus 1 control byte per *bucket*, with the bucket count a power of two between `8/7x` and `2x` the item count. That bounds the real cost at roughly 5.7 to 11.4 bytes per string; there is no occupancy at which it reaches 18.

The crate's own `docs/PERF.md` agrees. For the 6000-string corpus it reports `LocalLexicon` at 172.0 KiB filled and 96.5 KiB frozen. The frozen figure reconciles almost exactly to the boxed `offsets` (6001 x 4 = 24004 B) plus the 73.1 KiB buffer, and the remainder of the filled figure leaves 40960 B for the dedup table: exactly 8192 buckets x 5 bytes, or 6.8 bytes per string at that occupancy.

Since the figure exists to size a memory trade-off, an inflated value overstates the saving the sorted-handle-list recipe actually buys.

```suggestion
    /// list costs 4 bytes per string against the roughly 7 the hash map costs,
```

wukchung (commented):
🤖 Automated review of the `internity` changes.

Verified locally in a worktree at `151a4be`: `cargo test -p internity --all-features --test basic` (50 pass) and `--doc` (23 pass). The `compile_fail` trybuild target fails on this machine with `cannot find module or crate serde` inside the trybuild sandbox, an environment artifact in fixtures this PR does not touch; CI on this head is green.

No correctness, safety or API-compatibility defects found. The `index_of`/`sym_at` arithmetic holds at 0, `len - 1`, `len` and `u32::MAX`; the density guarantee survives dedup hits, the `intern_bytes` UTF-8 error path and a `StorageRollback` unwind; the CSR sentinel invariant is intact at every `offsets.len() - 1` site; and the concrete-return-type change is source-compatible with existing `impl Reader` callers, with `Reader` still sealed.

Two documentation-accuracy issues inline.

crates/internity/src/local_lexicon.rs
geeknoid: Confirmed and fixed — thanks, this was a real footgun.

I reproduced it on this head: after `intern("a")`/`intern("b")`, `as_u32()` returns 1 and 2 while `index_of` returns `Some(0)` and `Some(1)`. `pack_dense` stores `index + 1` to keep the handle non-zero, exactly as you describe, so a side table indexed with `as_u32()` runs off the end on the last entry — the very bug `index_of`/`sym_at` exist to prevent.

Applied your suggestion at `local_lexicon.rs`, reworded to talk about *positions* rather than handles, and made the 1-based nature of the raw value explicit. I corrected the more prominent claim at `lib.rs:86` too, and regenerated the README so it now reads "The raw `Sym::as_u32` value is 1-based and must not be used as an index directly."

I also added a regression test (`dense_index_is_zero_based_while_raw_handle_is_one_based`) that pins both numberings and asserts the last handle's raw value equals `len` — i.e. out of bounds for a `len`-sized side table — so this cannot silently drift back.

crates/internity/src/local_lexicon.rs
geeknoid: Agreed — the 18 was wrong, and I have taken your figure of 7.

I checked the arithmetic rather than just accepting it: `dedup` is a `HashTable<Sym>` with `Sym` a `NonZeroU32`, so hashbrown holds 4 bytes of value plus 1 control byte per *bucket*. With power-of-two buckets at a 7/8 load factor, the per-string cost ranges from 5/(7/8) = 5.7 bytes just before a resize to 5/(7/16) = 11.4 bytes just after one. There is indeed no occupancy that reaches 18. For the 6000-string corpus the bucket count is 8192, giving 8192 x 5 = 40960 B, or 6.8 B/string — matching your reading of `docs/PERF.md`.

Since the number exists to size a trade-off, overstating it inflated the apparent saving of the sorted-handle-list recipe. Now reads "4 bytes per string against the roughly 7 the hash map costs".

(One small note on the reconciliation: filled minus frozen for that row is 75.5 KiB rather than 40 KiB, since the live form also carries growth slack in the `offsets` and buffer `Vec`s. That does not affect your conclusion — the per-bucket bound is what rules out 18.)

geeknoid:
Addressed the review feedback in `a663f86f` (amended, force-pushed).

**@wukchung — both documentation-accuracy issues confirmed and fixed.** I verified each against the source rather than taking them at face value; both held up:

1. **1-based handles vs. 0-based positions.** Reproduced on this head: `as_u32()` gives 1 and 2 where `index_of` gives `Some(0)` and `Some(1)`, because `pack_dense` stores `index + 1` to keep the handle non-zero. Fixed in `local_lexicon.rs` and the more prominent `lib.rs` claim, plus the regenerated README, and pinned with a new regression test asserting the last handle's raw value equals `len` — out of bounds for a `len`-sized side table.
2. **Dedup map at ~18 B/string.** The hashbrown per-bucket layout (4-byte `Sym` value + 1 control byte, power-of-two buckets at 7/8 load) bounds the real cost to 5.7–11.4 B/string, so 18 was unreachable at any occupancy. Now states ~7, matching the 6.8 B/string that `docs/PERF.md` implies for the 6000-string corpus.

**Copilot reviewer — also took the suppressed suggestion** on `LocalReader`'s `Debug` impl: it now calls `Reader::len(self)` instead of recomputing `self.offsets.len() - 1`, so the CSR sentinel invariant lives in one place and matches `ThreadedReader`.

Validation on the new head: 51 integration tests and 23 doctests pass, clippy clean, rustdoc clean under `-D warnings`, rustfmt and spellcheck clean, README regenerated via `doc2readme`.

crates/internity/src/local_lexicon.rs
```diff
@@ -311,6 +341,90 @@ impl<S: BuildHasher> LocalLexicon<S> {
...
+    ///     let i = lexicon
+    ///         .index_of(sym)
+    ///         .expect("handle came from this lexicon");
+    ///     lengths[i as usize] = s.len();
+    /// }
+    /// assert_eq!(lengths, vec![1, 1]);
+    /// ```
+    #[inline]
+    #[must_use]
+    pub fn index_of(&self, sym: Sym) -> Option<u32> {
```
sandersaares: [Copilot speaking]

# Remove unsupported inline hints from generic conversions

The new `LocalLexicon<S>::index_of` and `sym_at` methods are defined in a generic impl and carry `#[inline]`, but the repository performance policy requires benchmark or disassembly evidence before adding that hint to generic functions. No `index_of` or `sym_at` benchmark exists. The equivalent `LocalReader` methods are non-generic exported functions and are permitted by the policy's separate hot-path rule.

**Reproducible reasoning**: The workspace defines a method in a generic type or impl as generic because it is monomorphized for each `S`. Such methods are already inlining candidates, so `#[inline]` is an additional code-generation hint that the policy reserves for cases where `bench-cg` or disassembly shows the compiler made the wrong default choice. The pull request adds the attributes but no benchmark exercises either conversion, leaving the two generic annotations unsupported while the adjacent non-generic reader annotations remain in scope of the policy's first rule.

**Consequence**: The extra hint can encourage call-site duplication across downstream monomorphizations without measured benefit and makes future maintainers unable to distinguish deliberate tuning from decorative annotations.

**Recommended action**: Remove `#[inline]` only from `LocalLexicon<S>::index_of` and `LocalLexicon<S>::sym_at`. Leave the non-generic `LocalReader` annotations unchanged. If measurement later shows that either generic conversion is not inlined on a relevant hot path, retain the hint with the required `just package=internity bench-cg` before-and-after evidence.

**References**:

- docs/performance.md, #[inline] annotations

**Impacted locations**:

- `crates/internity/src/local_lexicon.rs:393-395`
- `crates/internity/src/local_lexicon.rs:421-423`

crates/internity/src/lib.rs
sandersaares, line 1: [Copilot speaking]

# Treat the exported readers as public API throughout the documentation

The pull request makes `LocalReader` and `ThreadedReader` public, nameable return types, but the surrounding documentation does not consistently recognize that API boundary. The Cargo feature summary omits both readers, `docs/COMPARISON.md` and `docs/TODO.md` still describe hidden concrete types and `impl Reader` returns, and the pull request description incorrectly suggests that sealing `Reader` means exporting the reader types does not widen the supported public surface.

**Reproducible reasoning**: `LocalReader` is re-exported unconditionally, `ThreadedReader` is re-exported when `std` is enabled, and the inherent `freeze` methods return those concrete types. Downstream crates can therefore name the types, store them by value, and call their public methods. Sealing only prevents downstream crates from implementing `Reader`; it does not hide the crate's own public implementors or remove compatibility expectations for their public methods. The feature omission, stale opaque-return descriptions, and sealing claim are separate symptoms of the same incomplete propagation of the new public API contract.

**Consequence**: Consumers can receive contradictory guidance about whether the readers are available in `no_std` builds, whether they can be named and stored by value, and whether their identities and methods are supported API. Contributors may then make compatibility decisions from documentation that still models the pre-change opaque-reader design.

**Recommended action**: Align every affected API description with the exported-reader contract: list `LocalReader` in the `no_std` + `alloc` surface and `ThreadedReader` in the `std` surface, update `docs/COMPARISON.md` and `docs/TODO.md` to use the concrete `freeze` signatures, and narrow the sealing statement to say only that downstream crates cannot implement `Reader`. Regenerate the README with `just readme` after changing crate-level documentation.

**References**:

- crates/internity/docs/COMPARISON.md
- crates/internity/docs/TODO.md
- [Rust API Guidelines — sealed traits](https://rust-lang.github.io/api-guidelines/future-proofing.html#c-sealed)
- [Pull request #657](https://github.com/microsoft/oxidizer/pull/657)

**Impacted locations**:

- `crates/internity/src/lib.rs:129-137`
- `crates/internity/src/lib.rs:204-213`
- `crates/internity/docs/COMPARISON.md:25`
- `crates/internity/docs/TODO.md:56-60`
- `crates/internity/src/local_reader.rs:42-98`
- `crates/internity/src/local_lexicon.rs:499`
- `crates/internity/src/threaded_lexicon.rs:235`

sandersaares (commented):
[Copilot speaking]

Published 2 findings. No finding follows up on an existing discussion thread.

| Diagnostic | Value |
|------------|-------|
| Cache | Miss |

</details>

Vaiz (approved):

ralfbiedert (approved):
🤖 Two non-blocking suggestions: local_lexicon.rs:53 describes &str entries as 16 bytes, but that is only true on 64-bit targets; describing them as two machine words would be portable. Also, LocalReader::iter and ThreadedReader::iter still return Box<dyn Iterator>, so concrete-reader scans retain an allocation and dynamic dispatch. Consider inherent impl Iterator methods while keeping boxing at the erased Reader boundary.

crates/internity/src/local_lexicon.rs
geeknoid: Denied, this is overly pedantic.

crates/internity/src/lib.rs
geeknoid: Confirmed and fixed in `639ba8d7`. I checked each impacted location rather than applying the list wholesale; four of the five were genuinely stale, one was not.

**Fixed:**

- **Cargo feature summary** (`lib.rs`) — now names `ThreadedReader` alongside `ThreadedLexicon` in the `std` bullet, and `LocalReader` in the `no_std` + `alloc` list.
- **`COMPARISON.md:25`** — the frozen-`Reader` row still read "`freeze` returns `impl Reader` (static dispatch, concrete types hidden)". Now describes the concrete `LocalReader` / `ThreadedReader` returns as nameable and storable by value.
- **`TODO.md`** — the baseline signatures said `-> impl Reader`; corrected to the concrete types. Also updated two bullets that still said "Flat reader" / "Sharded reader", which the rename had missed.
- **PR description** — you are right that the sealing claim was wrong, and it was the load-bearing error here. Sealing only stops downstream crates from *implementing* `Reader`; it says nothing about the crate's own public implementors. Exporting these types does widen the supported surface — names, inherent methods, and auto-trait bounds all become commitments. Reworded to say exactly that, and to note the widening is the intended trade rather than an accident.

**Not changed:** `DESIGN.md:34-36`. Those are mermaid graph edges labelled `impl Reader`, meaning "implements `Reader`" — a relationship that is still accurate, not a stale return type.

README regenerated. Docs build clean under `-D warnings`, clippy/rustfmt/spellcheck clean, 51 integration tests and 23 doctests pass.

Separately, this rev also picks up @ralfbiedert's portability note: the `Vec<&str>` comparison described entries as "16-byte", now "two machine words: 16 bytes on a 64-bit target, 8 on a 32-bit one", along with the dependent "quarter of the per-string overhead" claim in the same paragraph.

geeknoid:
Rebased onto `main` (picking up #659) and pushed `639ba8d7` addressing the latest round. Thanks @Vaiz and @ralfbiedert for the approvals, and @sandersaares for the audit.

## @ralfbiedert — both suggestions

**1. `&str` entries described as 16 bytes — fixed.** You are right that this only holds on 64-bit. It now reads "two machine words: 16 bytes on a 64-bit target, 8 on a 32-bit one". I also fixed the sentence two lines below, which claimed "a quarter of the per-string overhead" — same 64-bit assumption baked in, now qualified as a quarter on 64-bit and half on 32-bit. The crate supports `no_std` and 32-bit targets, so this mattered more than a wording nit.

**2. Boxed `iter` — captured as a TODO rather than implemented here.** I agree with the diagnosis and prototyped it to confirm the mechanics: an inherent `impl Iterator` method takes precedence at concrete call sites, generic `R: Reader` code keeps the boxed trait method, and `Reader::iter(&r)` stays reachable. It works cleanly, and the trait method can just box the inherent iterator so the traversal logic exists once.

I have written it up as a new section in `crates/internity/docs/TODO.md` instead of landing it, for two reasons. It is net-new public API beyond what this PR proposes, and it carries a narrow source break: code binding a concrete `reader.iter()` straight to a `Box<dyn Iterator<...>>` would stop compiling once resolution picks the inherent method. That deserves its own PR with the `semver` check as the gate, rather than riding along with a docs-correction rev. The TODO records the motivation, the sketch, the `impl Iterator` vs. named-type trade, and that compatibility note.

Worth noting the boxing is not incidental at the trait level: `Lexicon::freeze` returns `Box<dyn Reader>`, so `Reader` must stay dyn-compatible and cannot use `impl Trait` in return position. The win is specifically for callers holding a concrete reader — which is a category this PR created.

## @sandersaares — documentation propagation

Addressed in the inline thread. Four of the five locations were genuinely stale and are fixed (feature list, `COMPARISON.md`, `TODO.md`, and the sealing claim in the PR description, which was the substantive error). `DESIGN.md:34-36` I left alone — those mermaid edges labelled `impl Reader` denote "implements `Reader`", which is still true.

The `#[inline]` finding was declined by @geeknoid as overly pedantic; that thread is resolved.

## Validation

Docs build clean under `-D warnings`, clippy clean, rustfmt and spellcheck clean, README regenerated via `doc2readme`, 51 integration tests and 23 doctests pass, plus `--no-default-features`.

sandersaares (approved):

psandana (commented):
Reviewed with the Oxidizer PR-review skill set (correctness/guidelines, design, docs/consistency). Two general points below; the remaining 11 are inline.

Overall this is a well-motivated change and the core decision — returning concrete `LocalReader`/`ThreadedReader` instead of `impl Reader` — is the right fix for the boxing/vtable problem. A type alias wouldn't have helped (the type still had to become nameable), and a supertrait change would have widened the sealed contract for no reason. The non-breaking claim holds: the old names were `pub(crate)`, fields stay private, the trait stays sealed, and `ThreadedReader` is correctly `#[cfg(feature = "std")]`-gated.

---

**1. `src/lib.rs` (general) · `[design]` — Promoting dense numbering to a permanent guarantee forecloses implementation freedom**

Observation: consecutive insertion-order numbering moves from an incidental detail (`as_u32() - 1`) to a stated public guarantee on `LocalLexicon`, `LocalReader`, and the crate root.

Concern: density is one of the strongest constraints an interner can commit to, and it permanently rules out (a) removal/tombstoning, (b) compaction or GC of unused entries, (c) dedup-merging two lexicons, and (d) ever making `LocalLexicon` internally sharded for larger capacity — the very thing that makes `ThreadedLexicon` non-dense. Once callers build and persist `Vec<T>` side tables keyed by dense index, any of those becomes a silent breaking change with an unchanged type signature. The motivation (downstream already relied on `as_u32() - 1`) fully justifies giving them a *supported conversion*; it doesn't by itself require promising density is permanent.

Alternative: keep `index_of`/`sym_at` as the supported API, but consider wording density as a property these accessors currently expose — explicitly reserving the right to change the mapping across versions, so callers don't persist dense indices across builds or crate upgrades. If permanence *is* intended, capture it in `DESIGN.md` along with the list of things it rules out, so a future maintainer can see the cost was accepted knowingly rather than inherited by accident.

---

**2. `docs/DESIGN.md` (general) · `[doc-consistency]` — The design doc never records the new invariant**

`DESIGN.md` is this crate's design record, and it's thorough — but it documents neither `index_of`/`sym_at` nor the dense-numbering guarantee that `freeze` is now required to preserve, which is the most consequential thing this PR commits to. The 1-based encoding rationale at line 72 and the `Reader` trait section at line 277 are the natural homes. A short subsection stating the invariant, which types expose it, and why `ThreadedLexicon` is excluded would keep the design doc authoritative as the API grows.

---

Verified clean, so not raised: README ↔ `lib.rs` are in sync including the renumbered `__link*` targets; no remaining rustdoc links to private items after `a663f86`; `.spelling` addition matches actual usage. No FFI or `.cargo/config.toml` changes here, so the interop and cargo-config checks yielded nothing.

crates/internity/docs/DESIGN.md
```diff
@@ -283,7 +283,7 @@ crate add methods without a breaking change. Its core operations are
 `(Sym, &str)` pairs).
 
 Resolution is a pure CSR lookup with a range check, so it needs no locks or
-atomics. `FlatReader` resolves a dense index directly; `ShardedReader` first
+atomics. `LocalReader` resolves a dense index directly; `ThreadedReader` first
 indexes the shard (always in range, since the shard bits can't exceed the shard
 count) then does a checked local lookup within that shard. Iteration order is
 handle order for the flat reader and shard-grouped for the sharded reader.
```
psandana: **[doc-consistency] Stale reader names after the rename**

Every other reference in this file was updated for the rename — the mermaid nodes at lines 34–35, plus lines 252, 286, and 353–354 — but this sentence still uses the old prose spellings:

> Iteration order is handle order for the **flat reader** and shard-grouped for the **sharded reader**.

It sits directly under line 286, which already says `LocalReader`/`ThreadedReader`, so it now reads as if it's describing two different types.

```suggestion
handle order for the local reader and shard-grouped for the threaded reader.
```

crates/internity/src/local_reader.rs
```diff
@@ -0,0 +1,126 @@
...
+impl core::fmt::Debug for LocalReader {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_struct("LocalReader")
+            .field("len", &Reader::len(self))
+            .finish_non_exhaustive()
+    }
+}
+
+impl crate::reader::Sealed for LocalReader {}
+impl Reader for LocalReader {
```
psandana, line 105: **[guidelines · M-ESSENTIAL-FN-INHERENT] Core reader operations are reachable only through the sealed trait**

Now that `LocalReader` and `ThreadedReader` are public, nameable types, their essential operations — `len`, `is_empty`, `resolve`/`try_resolve`, and `iter` — live *only* on the sealed `Reader` trait. A user who stores a `LocalReader` in a struct by value (the headline feature of this PR) still can't call `reader.len()` or `reader.resolve(sym)` without also writing `use internity::Reader`, for a trait that by construction nobody downstream can implement.

The friction is visible in the PR itself: every new doctest carries `use internity::{..., Reader}`, and both `Debug` impls have to spell it `Reader::len(self)`.

Per [M-ESSENTIAL-FN-INHERENT](https://microsoft.github.io/rust-guidelines/guidelines/libraries/#M-ESSENTIAL-FN-INHERENT), a concrete type's essential functionality should be available as inherent methods, with the trait there to *abstract over* readers rather than being the only door to core behavior. Consider adding thin inherent forwards on both readers so the trait import becomes opt-in.

crates/internity/src/local_lexicon.rs
psandana: **[design] `index_of`/`sym_at` use `u32` while `len()` returns `usize`**

Observation: `index_of` returns `Option<u32>` and `sym_at` takes `u32`, but the documented and tested purpose of the value is indexing a `Vec<T>` side table, and `len()`/`iter()` already work in `usize`.

Concern: the primary consumer of this value requires `usize`, so callers cast at every use site — the doctest just below writes `lengths[i as usize]`, and the new integration test does the same. The API is also internally inconsistent with `len()`, and `u32` re-exposes the 4-byte handle width at exactly the boundary where callers only want an index.

Alternative: return `Option<usize>` and accept `usize`. It matches `len()`/`iter()`, removes the cast everywhere, and loses nothing — the handle is 32-bit internally, so the domain bound is already enforced. If the compact width is a deliberate choice (e.g. for callers who want to *store* indices), a one-line rationale in the doc would help; as written the asymmetry with `len()` reads as accidental.

crates/internity/src/local_reader.rs
```diff
@@ -0,0 +1,126 @@
...
+    /// ```
+    #[inline]
+    #[must_use]
+    pub fn index_of(&self, sym: Sym) -> Option<u32> {
+        let index = sym.dense();
+        if index >= self.len() {
+            return None;
+        }
+        // `dense` decodes a `u32` handle, so this conversion cannot fail.
+        u32::try_from(index).ok()
```
psandana: **[guidelines · M-PANIC-ON-BUG] A self-described impossible case degrades into a plausible wrong answer**

The comment is correct that this can't fail: `Sym::dense()` is `(u32 - 1) as usize`, and the `index >= self.len()` guard above bounds it further. But `u32::try_from(index).ok()` then maps that impossible failure onto `None` — which in this API means "symbol not present". If an invariant break ever did make the cast fail, the method would silently return the wrong answer (a valid, in-range `Sym` reported as absent) rather than surfacing the bug.

Per [M-PANIC-ON-BUG](https://microsoft.github.io/rust-guidelines/guidelines/correctness/#M-PANIC-ON-BUG), a genuinely impossible condition should panic or be expressed so it can't arise, not quietly become a legitimate-looking result. Since `index < len()` is bounded by the `u32` handle space, a plain cast says what you mean:

```suggestion
        // `index < len()`, which is bounded by the u32 handle space, so this
        // narrowing is lossless.
        Some(index as u32)
```

The same pattern is duplicated in `LocalLexicon::index_of`.

crates/internity/src/local_reader.rs
```diff
@@ -0,0 +1,126 @@
...
+    /// use internity::LocalLexicon;
+    ///
+    /// let mut lexicon = LocalLexicon::new();
+    /// let a = lexicon.intern("a");
+    /// let reader = lexicon.freeze();
+    /// assert_eq!(reader.index_of(a), Some(0));
+    /// ```
+    #[inline]
+    #[must_use]
+    pub fn index_of(&self, sym: Sym) -> Option<u32> {
```
psandana: **[design] `index_of`/`sym_at` bodies are duplicated verbatim across two types**

These bodies are byte-for-byte identical to `LocalLexicon::index_of`/`sym_at` — same `sym.dense()` decode, same `len()` range check, same `Sym::pack_dense`, same `u32::try_from(...).ok()` idiom. What they encode is the correctness-critical mapping that `freeze` is now *guaranteed* to preserve, so the two copies must stay bit-identical forever; that's exactly the setup where a later fix lands on one and misses the other.

`Reader` is sealed and shouldn't carry these (they must not appear on `ThreadedReader`), so a trait default isn't the vehicle — but a crate-private free function is. Something like `fn dense_index_of(len: usize, sym: Sym) -> Option<u32>` and `fn dense_sym_at(len: usize, index: u32) -> Option<Sym>` in `storage`/`sym`, with both public methods delegating, keeps the two public surfaces while giving the mapping a single source of truth.

crates/internity/src/local_lexicon.rs
```diff
@@ -311,6 +341,90 @@ impl<S: BuildHasher> LocalLexicon<S> {
...
+    /// [`freeze`](Self::freeze) preserves it — see
+    /// [`LocalReader::index_of`](crate::LocalReader::index_of).
+    ///
+    /// The range check is against the *current* length: a handle from a different
+    /// `LocalLexicon` that happens to be in range returns an index here, and a
+    /// side table indexed with it would silently read the wrong row. This is not a
+    /// memory-safety problem, but it is silent, so keep side tables with the
+    /// lexicon they describe.
+    ///
+    /// `ThreadedLexicon` handles are **not** dense — they encode a shard index in
```
psandana, line 370: **[design] The "structural, not a documentation note" claim promises more than is enforced**

The PR description argues that putting these on the lexicon/reader instead of on `Sym` "makes the restriction structural instead of a documentation note," because a `ThreadedLexicon` handle "could reach it by mistake" if the method lived on `Sym`.

`Sym` is a single shared type across both interners, though, so what's actually structural is only *which types expose the method*. A `ThreadedLexicon`-minted `Sym` still compiles when passed to `LocalReader::index_of` — it usually falls out of range and returns `None`, but returns a valid-but-wrong index when it happens to be in range. And the doc a few lines up already concedes the two-`LocalLexicon` case is silently wrong.

So the real protection is "the conversion isn't available when you only hold a `ThreadedReader`," not "the wrong `Sym` can't reach it." Worth softening the wording to match, and leaning on the honest "keep side tables with the lexicon they describe" caveat as the actual safety story. Docs-only — no code change implied.

For the record on [M-STRONG-TYPES-GUARD](https://microsoft.github.io/rust-guidelines/guidelines/resilience/#M-STRONG-TYPES-GUARD): the mitigation here is documentation rather than the type system, which is consistent with existing `resolve` behavior, so this is a deliberate trade-off and not a regression.

crates/internity/src/local_lexicon.rs
```diff
@@ -334,14 +448,60 @@ impl<S: BuildHasher> LocalLexicon<S> {
...
+    /// This trades interning for a smaller, read-optimized table: the dedup hash
+    /// map is dropped and the storage is shrunk to fit, which is where the memory
+    /// saving comes from. The consequence is that a frozen reader can
+    /// [`resolve`](Reader::resolve) handles but has no
+    /// [`get`](LocalLexicon::get) — looking a string up is exactly the hash map
+    /// that was just freed.
+    ///
+    /// If you need string lookup on a frozen table, keep the lexicon live instead,
+    /// or rebuild a smaller index from [`iter`](Reader::iter). A sorted handle
+    /// list costs 4 bytes per string against the roughly 7 the hash map costs,
```
psandana, line 458: **[doc-consistency] Hash-map cost disagrees with the PR description (roughly 7 vs ~18)**

This line says a sorted handle list "costs 4 bytes per string against the **roughly 7** the hash map costs." The PR description's *Document rebuilding a string lookup over a frozen reader* section says "~4 bytes per string against the **~18** the hash map costs" — a >2× disagreement.

The code backs the doc: the dedup structure is `dedup: HashTable<Sym>` (line 76), i.e. hashbrown storing 4-byte `Sym` entries plus ~1 control byte per slot at ≤7/8 load, so roughly 5–7 bytes per string. The shipped "roughly 7" looks right and the PR body's "~18" looks like the outlier. Worth reconciling so the description matches what actually shipped (or correcting the doc, if 18 was meant to count something else).

crates/internity/src/threaded_reader.rs
```diff
@@ -1,30 +1,43 @@
...
-pub(crate) struct ShardedReader {
+/// A frozen [`ThreadedLexicon`](crate::ThreadedLexicon): one flat blob per shard,
+/// addressed by the `Sym`'s `[shard|local]` partition.
+///
+/// Returned by [`ThreadedLexicon::freeze`](crate::ThreadedLexicon::freeze).
+///
+/// Unlike [`LocalReader`](crate::LocalReader), handles here are **not** densely
+/// numbered — they encode a shard index in their high bits — so this reader offers
+/// no index conversions.
+pub struct ThreadedReader {
```
psandana, line 30: **[guidelines] Newly public readers implement no common traits beyond `Debug`**

With `LocalReader`/`ThreadedReader` now public and storable by value, users will put them in their own structs. `Debug` is correctly provided ([M-PUBLIC-DEBUG](https://microsoft.github.io/rust-guidelines/guidelines/universal/#M-PUBLIC-DEBUG) ✅) and both are `Send + Sync` ([M-TYPES-SEND](https://microsoft.github.io/rust-guidelines/guidelines/interop/#M-TYPES-SEND) ✅), and omitting `Display` is right — these aren't user-facing display types.

But neither derives `Clone`. The readers are immutable and own two boxed slices (or an array of them), so `Clone` is well-defined and straightforward, and its absence pushes anyone wanting two owners into `Arc<LocalReader>` even when no trait object is involved. Worth considering `Clone` on both — and `PartialEq`/`Eq` if value comparison is meaningful here.

crates/internity/src/threaded_reader.rs
```diff
@@ -1,30 +1,43 @@
...
-/// A frozen concurrent interner: one flat [`ShardReader`] per shard, addressed by
-/// the `Sym`'s `[shard|local]` partition. Returned as `impl Reader` from
-/// [`ThreadedLexicon::freeze`](crate::ThreadedLexicon::freeze).
-pub(crate) struct ShardedReader {
+/// A frozen [`ThreadedLexicon`](crate::ThreadedLexicon): one flat blob per shard,
+/// addressed by the `Sym`'s `[shard|local]` partition.
+///
+/// Returned by [`ThreadedLexicon::freeze`](crate::ThreadedLexicon::freeze).
+///
+/// Unlike [`LocalReader`](crate::LocalReader), handles here are **not** densely
```
psandana: **[design] Make the local/threaded asymmetry discoverable at the point of absence**

After this change the two halves of the public surface diverge: the local types gain `index_of`/`sym_at` plus a density guarantee, the threaded types get neither. That's *justified* — threaded handles genuinely pack a shard index and aren't dense — so this isn't a request for symmetric methods.

The issue is discoverability: someone on the threaded path goes looking for the same conversions, doesn't find them, and has to read the local-side prose or the crate root to learn why. This doc comment cross-references `LocalReader` but frames it as "unlike" rather than telling the reader what to do instead. Per M-SINGLE-ITEM-PATH / M-SIMPLE-ABSTRACTIONS, the "why not here" should be visible at the point of absence — e.g. mirroring the explicit note added to `ThreadedLexicon`: no dense index conversion, handles are shard-partitioned, use `resolve`/`iter`. It then reads as an intentional design boundary rather than a gap.

crates/internity/src/local_lexicon.rs
```diff
@@ -334,14 +448,60 @@ impl<S: BuildHasher> LocalLexicon<S> {
...
+    ///
+    /// This trades interning for a smaller, read-optimized table: the dedup hash
+    /// map is dropped and the storage is shrunk to fit, which is where the memory
+    /// saving comes from. The consequence is that a frozen reader can
+    /// [`resolve`](Reader::resolve) handles but has no
+    /// [`get`](LocalLexicon::get) — looking a string up is exactly the hash map
+    /// that was just freed.
+    ///
+    /// If you need string lookup on a frozen table, keep the lexicon live instead,
+    /// or rebuild a smaller index from [`iter`](Reader::iter). A sorted handle
```
psandana, line 457: **[design] Frozen-reader string lookup ships as a doctest rather than API — worth recording the decision**

Observation: the "rebuild a string→`Sym` lookup over a frozen reader" pattern ships as ~18 lines of doctest boilerplate, on the reasoning that it needs no new primitives and that binary-search-over-indirect-strings may not beat simply keeping the hash map.

That's a defensible call and the reasoning is sound. The counterweight: the pattern came from a real field report, every consumer will re-derive it, and it's easy to get subtly wrong — duplicates, resolving twice per comparison, or silently reintroducing the weak default hasher on the way back. The binary search also costs `O(log n)` dependent `resolve` loads per lookup, which is worth measuring before recommending it as *the* frozen-lookup path.

Alternative: either a feature-gated, opt-in helper (`build_string_index()` / `FrozenLookup`) that owns the sorted table and the resolve-on-compare logic in one tested place, or — if the doctest-only decision stands — a line in `DESIGN.md` recording that the helper was considered and declined, so the trade-off is captured rather than implicit.

crates/internity/src/local_lexicon.rs
```diff
@@ -334,14 +448,60 @@ impl<S: BuildHasher> LocalLexicon<S> {
...
+    ///
+    /// struct Frozen {
+    ///     reader: LocalReader,
+    ///     by_string: Box<[Sym]>,
+    /// }
+    ///
+    /// impl Frozen {
+    ///     fn new(reader: LocalReader) -> Self {
+    ///         let mut by_string: Vec<Sym> = reader.iter().map(|(sym, _)| sym).collect();
+    ///         by_string.sort_unstable_by(|&a, &b| reader.resolve(a).cmp(reader.resolve(b)));
```
psandana: **[perf] Minor: two resolves per comparison in the recommended sort** *(non-blocking nit)*

`sort_unstable_by(|&a, &b| reader.resolve(a).cmp(reader.resolve(b)))` performs two `resolve` calls per comparison, so an index build does `O(n log n)` string reconstructions. `resolve` is O(1) and allocation-free and this is a one-time build cost, so the current form is a perfectly reasonable default — I wouldn't block on it.

Mentioning it only because the docs steer every consumer toward this exact shape: `sort_by_cached_key(|&sym| reader.resolve(sym))` resolves each string once, trading a temporary key `Vec` for far fewer reconstructions. Confirmed there's no hidden allocation or hot-path cost in the pattern otherwise.

ralfbiedert (approved):

crates/internity/src/local_lexicon.rs
geeknoid: You are right, and I have reworded the PR description in `531674fd`.

The claim overreached. `Sym` is a single shared type, so a `ThreadedLexicon`-minted handle still compiles when passed to `LocalReader::index_of` — usually out of range and `None`, but valid-but-wrong when it happens to land in range. What is actually structural is only *which types expose the conversion*: it is unreachable when you hold a `ThreadedReader`, which is a real and useful property, just a narrower one than "the wrong `Sym` cannot reach it."

The description now says that explicitly and points at the "keep side tables with the lexicon they describe" caveat as the honest safety story, which the doc already carried a few lines up.

Noted too on M-STRONG-TYPES-GUARD — agreed this is a deliberate trade consistent with `resolve`, not a regression. A per-lexicon branded handle would fix it properly but at the cost of the 4-byte `Copy` handle, which is the crate's whole point.

crates/internity/src/local_lexicon.rs
geeknoid: You are right, and the doc is the correct one. Fixed the PR description in `531674fd`.

The "~18" was an error from the first revision that @wukchung caught with the same hashbrown reasoning you give — 4-byte `Sym` plus ~1 control byte per bucket, power-of-two buckets at 7/8 load, so 5.7–11.4 B/string depending on where occupancy sits relative to the last resize. I corrected the doc to "roughly 7" then but missed that the PR body repeated the old figure.

For anyone reading the description later: the other "18" in that section refers to 18 *lines* of doctest, not bytes, and is unrelated.

crates/internity/src/local_lexicon.rs
geeknoid: Taken — changed in `531674fd`.

You are right that the docs steer every consumer toward this exact shape, which makes it worth getting right even as a one-time build cost. It is now `sort_by_cached_key(|&sym| reader.resolve(sym))`, which resolves each string once instead of twice per comparison.

I verified the borrow works before recommending it: the cached key is a `&str` borrowed from the reader while the handle vector is mutably borrowed, which is fine as they are separate objects, and the reader is still movable into the struct afterwards.

crates/internity/src/threaded_reader.rs
geeknoid: Agreed and rewritten in `531674fd`.

The doc now has a short "No dense index conversions" section that says what the reader should do instead — `resolve` for a single handle, `iter` to walk the corpus, or intern with `LocalLexicon` if positions for a `Vec<T>` side table are actually what is wanted — rather than only noting that `LocalReader` differs.

Your framing was the useful bit: the absence is a design boundary, and the place to explain it is where someone hits it, not three types away. I also spelled out *why* the handles cannot yield a dense position (the high bits select a shard, so live values are spread across the handle space) so it does not read as an unimplemented feature.

crates/internity/src/local_lexicon.rs
geeknoid: Recorded in `DESIGN.md` in `531674fd` — the doctest-only decision stands, but the reasoning is now written down instead of implicit.

The new subsection states that a `build_string_index()` / `FrozenLookup` helper was considered and declined, why (`iter` + `resolve` are sufficient, and `O(log n)` dependent loads over indirectly-stored strings may well lose to keeping the lexicon live), and the condition for revisiting: field reports showing the pattern is both common and being got wrong.

Your counterweight is fair, and two of the failure modes you list are worth flagging — resolving twice per comparison, and reintroducing a weak hasher on the way back. The first I have now fixed in the example itself via your `sort_by_cached_key` suggestion. The second does not arise in this shape, since the rebuilt index is a sorted `Box<[Sym]>` with no hashing at all, which was part of why I preferred it over handing back a `HashMap`.

Agreed the `O(log n)` cost wants measuring before this is recommended as *the* frozen-lookup path; the doc presents it as one option against keeping the lexicon live rather than as the blessed answer.

crates/internity/src/local_reader.rs
geeknoid: Good catch, and fixed in `531674fd` — though not the way you suggested.

You are right about the defect: `None` means "not present" in this API, so mapping an impossible conversion failure onto it would report a valid, in-range symbol as absent and hide the invariant break behind a legitimate-looking answer.

Rather than panic on it, I removed the possibility. Switching `index_of`/`sym_at` to `usize` (your other finding) means there is no narrowing conversion at all — the body is now a bounds check and nothing else. M-PANIC-ON-BUG's first option is to express the condition so it cannot arise, which is strictly better than an `.expect()` that can never fire.

I avoided the `as` cast in your suggestion, since the workspace lints `clippy::cast_possible_truncation` and it would have needed an `#[expect(...)]` to sit alongside — more machinery than the impossible case deserved.

crates/internity/src/local_reader.rs
geeknoid: Agreed on the substance, but deferred rather than implemented here — @geeknoid's call, and I think the right one.

The finding is correct: `len`, `is_empty`, `resolve`/`try_resolve` and `iter` are reachable only through a trait nobody downstream can implement, and the friction shows up in this PR's own code exactly as you describe — every doctest imports `Reader`, and both `Debug` impls spell it `Reader::len(self)`.

The reason to hold it: @ralfbiedert separately asked for `iter` to stop returning `Box<dyn Iterator>` now that the concrete types are nameable. That lands on the same methods as your request, and adding *boxed* inherent forwards now would mean touching this surface twice, since the source break arrives whenever `iter` switches to `impl Iterator` regardless of when the forwards appear.

So both are captured as one item in `crates/internity/docs/TODO.md`, rewritten in `531674fd` to cover the ergonomics motivation (citing M-ESSENTIAL-FN-INHERENT) alongside the performance one, with the full method set, the trait-forwarding scheme that keeps the traversal logic single-sourced, and the compatibility note. It wants its own rev with `cargo-semver-checks` as the gate rather than riding along with a docs-correction PR.

Worth noting for that future change: the boxing cannot be removed from the trait itself, since `Lexicon::freeze` returns `Box<dyn Reader>` and the trait must stay dyn-compatible. Inherent methods on the concrete types are the only route, which makes your point and ralfbiedert's the same change.

crates/internity/docs/DESIGN.md
geeknoid: Fixed in `531674fd` — applied your suggestion.

Same class of miss as the two `TODO.md` bullets caught in the previous round; the rename swept the code and the mermaid nodes but left prose spellings behind in a few places. This one was the last of them.

crates/internity/src/local_reader.rs
geeknoid: Agreed and fixed in `531674fd`.

Both types now delegate to crate-private `dense_index_of(len, sym)` and `dense_sym_at(len, index)` in `sym.rs`, which is where the rest of the handle-encoding logic lives. `Reader` stays untouched, so nothing leaks onto `ThreadedReader`.

Your framing of the risk was the persuasive part: this mapping is precisely what `freeze` is now *guaranteed* to preserve, so the live and frozen copies had to stay bit-identical forever. That is a poor thing to enforce by vigilance. It also paid off immediately — the `usize` change and the `None`-swallowing fix from your other two comments each landed once instead of twice.

crates/internity/src/threaded_reader.rs
geeknoid: Added `Clone` to both readers in `531674fd` (and to the crate-private `ShardReader` it needs). Declined `PartialEq`/`Eq` for now.

`Clone` was clearly right: the readers are immutable and own plain boxed slices, so it is a derive, and C-COMMON-TRAITS favours providing it. Your point about `Arc` was the convincing one — reaching for it purely because the type is not `Clone` is the wrong reason to pay for an allocation and a refcount. `Arc` remains the better choice for *sharing* a large corpus, but that should be the caller's decision rather than something the API forces.

On `PartialEq`/`Eq` I would rather wait for a concrete use case. The semantics are ambiguous — two lexicons holding the same strings interned in different orders produce different handle mappings, so structural equality and "same corpus" are not the same question, and the cheap-looking `==` would be an `O(n)` byte comparison. Easy to add later, impossible to narrow once shipped.

There is a test covering both clones, including that the original can be dropped while the clone stays valid.

crates/internity/src/local_lexicon.rs
geeknoid: Agreed and changed in `531674fd`: `index_of` now returns `Option<usize>` and `sym_at` takes `usize`.

Your read of the use sites was right — both the doctest and the integration test were writing `lengths[i as usize]`, and those casts are now gone. The `u32` width was not a deliberate choice for storage; it leaked out of the handle representation, which is exactly the layer this API is supposed to hide.

One thing worth calling out: the change also dissolved the M-PANIC-ON-BUG problem you raised separately. With `usize` on both sides there is no narrowing left, so `index_of` is now just a bounds check (`(index < len).then_some(index)`) with no fallible conversion to mis-map onto `None`. Two findings, one fix.

geeknoid:
Pushed `531674fd` addressing @psandana's review. Thanks — this was a genuinely useful pass; nothing in it turned out to be wrong, and two of the findings collapsed into a single better fix.

All 11 inline threads are answered and resolved. The two general points below.

## 1. Density as a permanent guarantee `[design]`

This is the sharpest question in the review, and I have taken your **second** option: keep the guarantee, but record the cost in `DESIGN.md` so it is knowingly accepted rather than inherited.

The reason for not softening the wording: a conversion whose mapping may change across versions cannot be used for the thing this API exists to enable. Persisting a `Vec<T>` side table keyed by position — or even building one and holding it across an upgrade — needs the numbering to be stable, so "these accessors currently expose density, but do not rely on it" would leave callers exactly where the field report found them, reaching for `as_u32() - 1` because nothing supported was on offer.

But you are right that the cost should be explicit. `DESIGN.md` now has a **What the guarantee costs** subsection listing what it forecloses — removal/tombstoning, compaction or GC, dedup-merging two lexicons, and ever making `LocalLexicon` internally sharded — and states that any of those now needs a new type or a major version rather than a quiet relaxation.

I did not add wording telling callers not to persist indices across crate upgrades, because that would contradict the guarantee. If we ever want that freedom back, the honest move is a version bump, not a hedge in the docs.

## 2. `DESIGN.md` never records the new invariant `[doc-consistency]`

Fixed, at both the locations you identified.

A new **Dense numbering: `index_of` / `sym_at`** section after the `Reader` trait section states the invariant, which types expose it, that `freeze` is required to preserve it, why `ThreadedLexicon` is excluded (shard bits scatter live values across the handle space), and that both types delegate to one crate-private implementation. It also carries the cost subsection above and the record of the declined frozen-lookup helper.

The 1-based encoding rationale at line 72 now cross-references it, since that offset is precisely why `as_u32` is not a side-table index.

## Everything else

- **`usize` conversions** — `index_of` → `Option<usize>`, `sym_at(usize)`. This also dissolved your M-PANIC-ON-BUG finding: with no narrowing left there is no fallible conversion to swallow into a misleading `None`.
- **Deduplicated** the dense mapping into `dense_index_of`/`dense_sym_at` in `sym.rs`; both public surfaces delegate.
- **`Clone`** on both readers (plus crate-private `ShardReader`), with a test. Declined `PartialEq`/`Eq` pending a use case.
- **`sort_by_cached_key`** in the recommended doctest; **`ThreadedReader`** doc now says what to use instead of the absent conversions.
- **Inherent methods (M-ESSENTIAL-FN-INHERENT)** deferred to `TODO.md`, merged with @ralfbiedert's unboxing request since they touch the same methods — details in that thread.
- **PR description** corrected on both the "structural" overclaim and the stale ~18 figure.

Validation: 52 integration tests and 23 doctests pass, `--no-default-features` clean, clippy clean, rustdoc clean under `-D warnings`, rustfmt and spellcheck clean, README regenerated. `cargo-semver-checks` is not installed locally so CI is the gate there; the signature change is to methods that do not exist in the published 0.1.0, and `Clone` is purely additive, so I expect it quiet.

geeknoid:
Follow-up in `3a789c83`: `pr-mutants` caught a surviving mutant in the helper I extracted, and it was a fair hit.

Replacing `<` with `<=` in `dense_index_of` survived on all three platforms — no test distinguished them, meaning the exact `index == len` boundary was unpinned. The existing out-of-range test used `Sym::from_u32(u32::MAX)`, which is so far past the end that it passes either way.

That is exactly the off-by-one class this API exists to prevent, so the gap mattered more than a coverage statistic: under the mutant, the first handle past the end reports `Some(len)` — an in-range-looking index that would read one row off the end of a caller's side table.

The test now also asserts the first handle past the end (`Sym::from_u32(2)` against a one-string lexicon, whose dense position is exactly `len`) resolves to `None`, on both the lexicon and the frozen reader. I verified it kills the mutant by applying the `<=` change locally and confirming the test fails, then reverting.

psandana (approved):
Approving. All eleven points from my earlier review are resolved, and I verified each against the code rather than the replies.

Three of the fixes are better than what I proposed:

- **`u32` → `usize`** dissolved the `u32::try_from(index).ok()` bug instead of patching it. There is no narrowing left to get wrong, and declining the `as` cast was right given `clippy::cast_possible_truncation` is denied here.
- **Declining `PartialEq`/`Eq`** with the reasoning that two lexicons interned in different orders hold the same strings under different handles, so `==` would be both ambiguous and O(n). That is the correct call; a derived impl would have been a trap.
- **`dense_index_of`/`dense_sym_at` in `sym.rs`** put the invariant `freeze` must preserve in one place, without leaking anything onto `Reader` or `ThreadedReader`.

The documentation changes do the thing I actually cared about: `DESIGN.md` now records *why* the density guarantee exists, what it costs, and that relaxing it "needs a new type or a major version, not a quiet relaxation." The declined alternatives are written down with revisit conditions, so the next person to ask these questions finds the answers instead of re-deriving them.

One thing worth calling out for the record, since it is easy to miss under a force-push amend. Deduplicating the bounds check concentrated the `index < len` boundary into a single definition, and `cargo-mutants` then caught that the test suite only exercised it from one side:

```
MISSED  crates/internity/src/sym.rs:153:12: replace < with <= in dense_index_of
```

`dense_sym_at` survived the same mutation because `sym_at(1)` on a one-element lexicon covers it, but nothing called `index_of` with a handle whose dense position was exactly `len` — the one input that separates `<` from `<=`. The `first_past_end` case added in `basic.rs` kills it on both the lexicon and reader paths. The pre-existing `u32::MAX` case could not, since both operators reject it. Good catch by the mutation gate, and a nice illustration of why it earns its runtime.

Mutants are green on all four platforms and the rest of CI is clean.

The deferred item (`M-ESSENTIAL-FN-INHERENT`) is tracked in `docs/TODO.md`, correctly merged with @ralfbiedert's `iter` de-boxing request. They are the same change — `Reader` has to stay dyn-compatible because `Lexicon::freeze` returns `Box<dyn Reader>`, so both can only land as inherent methods on the concrete types, and one of them carries a source break. Landing that behind a `cargo-semver-checks` gate in its own PR is the right sequencing.

No remaining blockers from me.

PR #658, opened 2026-08-12
fix: declare optional dependencies as dev-dependencies

sandersaares:
[Copilot speaking]

## Motivation

Feature-dependent code in this workspace is gated behind `cfg(any(test, feature = "foo"))` rather than `cfg(feature = "foo")`, so that a crate's own test build compiles the feature-dependent code without the test target having to enumerate features.

`cfg(test)` and Cargo features are resolved by different mechanisms. A test build turns on `cfg(test)`, which pulls the gated code into the compilation, but it does not turn the feature on, so nothing the feature would have activated is activated:

```
  cfg(test)  ──────────────► gated code compiled in
  feature    ──── OFF ─────► optional dependency NOT linked
                             dependency features NOT enabled
                                        │
                                        ▼
                       error[E0432]: unresolved import
```

`layered` demonstrated the failure: `cargo check -p layered --tests` failed with E0432/E0433 because `mod dynamic` is gated on the `dynamic-service` feature while `plurality` was declared only as an optional dependency. Every check in CI that builds test targets passes `--all-features`, which enables the feature, links the optional dependency, and hides the problem.

Auditing the workspace found the same latent gap in 12 crates, plus a related class in 7 more: a dev-dependency present but carrying fewer features than a feature build would give it, which lets gated code compile against a smaller API surface than it really has.

## Changes

Every optional dependency is now also declared as a non-optional dev-dependency, so it is present in test builds. Same-workspace members are referenced path-only so that publish ordering still resolves; external crates inherit from the workspace.

Every dev-dependency that mirrors a regular dependency now carries the features a feature build gives that dependency, drawn from both the `[dependencies]` declaration and the `dep/feature` entries in the `[features]` table.

`cargo check -p layered --tests` now succeeds. Feature resolution for non-test builds is unaffected: the workspace uses resolver 2, which excludes dev-dependencies from feature unification when test targets are not being built, and `Cargo.lock` is unchanged.

The convention is documented in `docs/optional-deps-in-test-builds.md` and referenced from `AGENTS.md`. It covers the requirement that features be additive, the mirroring of both dependencies and their features, the limitation that a feature implying another feature of the same crate cannot be reproduced this way, and that `cargo hack --feature-powerset` rather than the test build is the authority on whether feature-gated code compiles in a real feature build.

This invariant currently has no automated enforcement — every existing check passes `--all-features`, which is exactly what masks it. A manifest-level check would fit naturally in `cargo-anvil`, which is where the workspace's checks are generated from.

### DIFF
--- a/docs/optional-deps-in-test-builds.md
+++ b/docs/optional-deps-in-test-builds.md
@@ -58,15 +58,22 @@ not weaken the feature gate for them.
 
 ## Dependency features are also mirrored
 
-A feature can activate features of a dependency as well as the dependency
-itself. The dev-dependency carries those features, otherwise the gated code
-compiles against a smaller API surface than a feature build gives it:
+Reproducing the dependency by name is not enough. The dev-dependency also
+carries every feature that a feature build would give it, otherwise the gated
+code compiles against a smaller API surface than the feature build provides.
+
+Two sources contribute. The `[dependencies]` entry declares features directly,
+and the `[features]` table activates further ones through `dep/feature` entries:
 
 ```toml
 [features]
-test-util = ["bytesbuf/test-util"]
+codegen = ["dep:syn", "bytesbuf/test-util"]
+
+[dependencies]
+syn = { workspace = true, optional = true, features = ["full", "parsing"] }
 
 [dev-dependencies]
+syn = { workspace = true, features = ["full", "parsing"] }
 bytesbuf = { path = "../bytesbuf", features = ["test-util"] }
 ```
 
--- a/docs/optional-deps-in-test-builds.md
+++ b/docs/optional-deps-in-test-builds.md
@@ -1,0 +1,106 @@
+# Optional dependencies in test builds
+
+Feature-dependent code is gated behind `cfg(any(test, feature = "foo"))` rather
+than `cfg(feature = "foo")`, so that a crate's own test build compiles the
+feature-dependent code without the test target having to enumerate features.
+
+```rust
+#[cfg(any(test, feature = "dynamic-service"))]
+mod dynamic;
+
+#[cfg(any(test, feature = "dynamic-service"))]
+pub use dynamic::{DynamicService, DynamicServiceExt};
+```
+
+`cfg(test)` and Cargo features are resolved by different mechanisms. A test
+build turns on `cfg(test)`, which pulls the gated code into the compilation. It
+does not turn the feature on, so nothing the feature would have activated is
+activated. The manifest has to supply that separately.
+
+## Features are additive
+
+The pattern depends on features being additive: enabling one never removes or
+changes behaviour that was available without it, and enabling all of them at
+once is a valid configuration.
+
+Mutually exclusive features are an invalid pattern. Two features that cannot be
+enabled together cannot both be on in a test build, and a consumer combining two
+unrelated dependents of the crate would break through no fault of their own.
+
+## Optional dependencies are also dev-dependencies
+
+A feature commonly activates an optional dependency. That dependency is not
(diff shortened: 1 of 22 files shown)


sandersaares:
[Copilot speaking]

The `static-analysis` failure was a latent `clippy::unused_self` in `seatbelt` (`report_state_change`, `create_guard`), not something this branch introduced. Both methods use `self` only inside `#[cfg(any(feature = "metrics", ...))]` / `#[cfg(any(feature = "logs", ...))]` blocks, so `self` is genuinely unused when neither feature is on.

It surfaces here because `crates/seatbelt_http/Cargo.toml` is touched, which brings `seatbelt_http` into the affected set; clippy does not apply `--cap-lints allow` to workspace path dependencies, so it then lints `seatbelt` in a feature configuration that main never lints. Confirmed reproducible on a clean `origin/main` worktree via `cargo clippy -p seatbelt_http --all-targets --all-features --locked -- -D warnings`.

Fixed by scoping an `expect(clippy::unused_self)` to exactly the configuration where the lint is correct. Verified clean for the no-feature, `metrics`-only, `logs`-only, both-features and `--all-features` configurations, plus full-workspace clippy with `-D warnings`.

ralfbiedert (commented):
🤖 The required codecov/project check currently fails at 99.99%. Codecov identifies crates/http_path_template/src/error.rs:84 as the sole uncovered line. Although this exceeds the M365 95% standard, it does not satisfy this repository's required 100% gate; please restore coverage for that branch before merging.

sandersaares:
[Copilot speaking]

Addressed in b4c7f3f (fix commit 73c9789).

`crates/http_path_template/src/error.rs:84` was the `Self::Disabled` arm of `MaybeBacktrace::capture()`, selected only when the `std` feature is off. It is not restorable by a test: mutation testing runs with default features, and in the coverage run's `--no-default-features` leg Cargo feature unification across the selected dependency graph re-enables `std` through the crate's dependents, so the arm is frequently absent from the measured build entirely.

We do not test `no_std` configurations, so such code is now exempt from the coverage gate and from mutation testing rather than being held to an obligation the harness cannot discharge. `capture()` is split into per-configuration definitions and the `no_std` one carries `#[cfg_attr(coverage_nightly, coverage(off))]` and `#[cfg_attr(test, mutants::skip)]`; the `std` arm keeps its full obligations. The policy is declared in `AGENTS.md` under "`no_std` Support".

`codecov/project` and `codecov/patch` are both green on the current head.

docs/optional-deps-in-test-builds.md
```diff
@@ -0,0 +1,113 @@
+# Optional dependencies in test builds
```
Vaiz, line 1: in the spirit of saving AI tokens, can we reduce this doc? On a glance, it goes too deep into weeds when issue can be explained in a couple sentences.

Vaiz (approved):

st-dev-gh (approved):

docs/optional-deps-in-test-builds.md
sandersaares: Seems OK to me but feel free to make a PR to shape it better to your vision if you feel a different shape is more preferable.

martintmk (approved):

docs/optional-deps-in-test-builds.md
martintmk: And why not just do it directly in the PR that introduces this? If you agree that doc is verbose, we can just instruct agent to shorten it and avoid extra overhead.

sandersaares: Sorry for ambiguity, by "seems OK to me" I meant the doc seems fine - no need to change.

PR #659, opened 2026-08-12
fix: declare dependency features where the API is used

sandersaares:
[Copilot speaking]

## Motivation

Cargo unifies dependency features across the whole graph. A crate that uses feature-gated API from a dependency will therefore compile even when it never declares the feature that gates it, provided some *other* crate in the graph happens to enable it. The dependency edge is silently incomplete, and the build breaks the moment the unrelated supplier changes its own feature choices.

The workspace root declares external dependencies with `default-features = false`, which makes this failure mode readily available: nothing is enabled unless somebody asks for it, and "somebody" is frequently the wrong crate.

`bytesbuf` was the clearest case. Its Windows-only test helper calls `GlobalMemoryStatusEx` and `GetLastError`, but the `windows-sys` dev-dependency declared no features at all. Both required Win32 modules were supplied entirely by `criterion → walkdir → same-file → winapi-util`:

```
windows-sys feature "Win32_System_SystemInformation"
  └── winapi-util          <-- sole supplier; bytesbuf contributes nothing
windows-sys feature "Win32_Foundation"
  └── winapi-util          <-- sole supplier
```

Dropping or changing the benchmark harness would have broken the Windows test build for reasons visible nowhere near the code that failed.

An audit of every dependency declaration in the workspace found the pattern was not isolated.

## Changes

Ten dependency declarations now name the features whose API the declaring crate actually uses, so each edge stands on its own:

| Crate | Dependency | Features | Gated API in use |
|---|---|---|---|
| `bytesbuf` | `windows-sys` (dev) | `Win32_Foundation`, `Win32_System_SystemInformation` | `GlobalMemoryStatusEx`, `MEMORYSTATUSEX`, `GetLastError` |
| `bytesbuf_io` | `futures` (dev) | `executor` | `futures::executor::block_on` |
| `fetch` | `futures` (dev) | `executor`, `std` | `futures::executor::block_on` |
| `fetch` | `serde_json` (dev) | `std` | `serde_json` value/error API |
| `http_extensions` | `bytes` | `std` | `std::io` integration |
| `multitude` | `zerocopy` | `derive` | `#[derive(FromBytes, IntoBytes, ...)]` |
| `ohno_macros` | `proc-macro2` | `proc-macro` | conversions to/from `proc_macro` types |
| `seatbelt` | `serde_json` (dev) | `std` | `serde_json` value/error API |
| `tick` | `serde_json` (dev) | `std` | `serde_json` value/error API |

`fetch`'s `tracing` dependency moves rather than gains a feature. Its only `std`-gated use is `tracing::subscriber::set_default` inside `#[cfg(test)]` modules, so `std` belongs in `[dev-dependencies]`. Declaring it under `[dependencies]` would have propagated `tracing/std` to every downstream consumer of the published crate to serve a test-only need.

Each entry was verified against the resolved feature graph, confirming the declaring crate is now a direct contributor of the feature rather than a passenger:

```
cargo tree -p <crate> --all-features -e features -i <dep>
```

The remaining declarations examined during the audit resolve extra features through the graph but do not depend on them: the crates either use only ungated API, or propagate the feature conditionally through their own `[features]` table.

`Cargo.lock` is unchanged. These features were already enabled in every build; the change is which crate is responsible for asking.

## Note on detection

`cargo hack --feature-powerset` does not surface this class of defect. It varies the features of workspace crates, while the incomplete declarations concern features of external dependencies, which stay fixed across the powerset. A build that omits the unrelated supplier is the only thing that exposes them.

### DIFF
--- a/crates/http_extensions/Cargo.toml
+++ b/crates/http_extensions/Cargo.toml
@@ -63,7 +63,7 @@ json = ["dep:serde_core", "dep:serde_json"]
 test-util = ["tick/test-util"]
 
 [dependencies]
-bytes = { workspace = true }
+bytes = { workspace = true, features = ["std"] }
 bytesbuf = { workspace = true, features = ["bytes-compat", "std"] }
 futures = { workspace = true, features = ["std"] }
 http = { workspace = true }
--- a/crates/multitude/Cargo.toml
+++ b/crates/multitude/Cargo.toml
@@ -70,7 +70,7 @@ ptr_meta = { workspace = true }
 serde = { workspace = true, features = ["alloc"], optional = true }
 serde_json = { workspace = true, features = ["alloc", "raw_value"], optional = true }
 widestring = { workspace = true, features = ["alloc"], optional = true }
-zerocopy = { workspace = true, optional = true }
+zerocopy = { workspace = true, features = ["derive"], optional = true }
 
 # Loom is only used under `--cfg loom` for the model-checked tests.
 # `target.'cfg(loom)'` is matched by Cargo when `RUSTFLAGS="--cfg loom"` is set.
--- a/crates/seatbelt/Cargo.toml
+++ b/crates/seatbelt/Cargo.toml
@@ -90,7 +90,7 @@ opentelemetry-stdout = { workspace = true, default-features = false, features = 
 opentelemetry_sdk = { workspace = true, default-features = false, features = ["metrics", "testing", "experimental_metrics_custom_reader"] }
 rstest = { workspace = true }
 serde = { workspace = true, features = ["std", "derive"] }
-serde_json = { workspace = true }
+serde_json = { workspace = true, features = ["std"] }
 static_assertions = { workspace = true }
 testing_aids = { path = "../testing_aids" }
 tick = { path = "../tick", features = ["test-util", "tokio"] }
--- a/crates/bytesbuf/Cargo.toml
+++ b/crates/bytesbuf/Cargo.toml
@@ -64,7 +64,11 @@ gungraun = { workspace = true, features = ["default"] }
 libc = { workspace = true }
 
 [target.'cfg(target_os = "windows")'.dev-dependencies]
-windows-sys = { workspace = true }
+# `windows-sys` enables no Win32 modules by default; each module used must be listed.
+windows-sys = { workspace = true, features = [
+    "Win32_Foundation",
+    "Win32_System_SystemInformation",
+] }
 
 [[bench]]
 name = "buf"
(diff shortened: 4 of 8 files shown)


martintmk (approved):

wukchung (approved):

ralfbiedert (approved):
🤖 Reviewed at head 0f439c29edec6d383a29aaca9f1c30c20919c16c. The dependency features are declared at their consuming edges, and I found no blocking correctness, performance, compatibility, or safety issue.

Vaiz (approved):

crates/multitude/Cargo.toml
```diff
@@ -70,7 +70,7 @@ ptr_meta = { workspace = true }
 serde = { workspace = true, features = ["alloc"], optional = true }
 serde_json = { workspace = true, features = ["alloc", "raw_value"], optional = true }
 widestring = { workspace = true, features = ["alloc"], optional = true }
-zerocopy = { workspace = true, optional = true }
+zerocopy = { workspace = true, features = ["derive"], optional = true }
```
psandana, line 73: The library only uses the `FromZeros` trait; `#[derive(FromZeros)]` appears in doctests and integration tests. Enabling `zerocopy/derive` on the normal optional dependency makes every downstream `multitude/zerocopy` consumer compile the proc-macro crate. Keep the normal dependency feature-free and add a dev-dependency enabling `derive`. Also update the PR table, which currently names `FromBytes`/`IntoBytes` rather than `FromZeros`.

crates/http_extensions/Cargo.toml
```diff
@@ -63,7 +63,7 @@ json = ["dep:serde_core", "dep:serde_json"]
 test-util = ["tick/test-util"]
 
 [dependencies]
-bytes = { workspace = true }
+bytes = { workspace = true, features = ["std"] }
```
psandana, line 66: The only production `bytes/std` use is `Bytes::reader()` in the `json` module. Enabling it unconditionally propagates `std` to consumers that do not enable JSON. Keep the dependency feature-free, add `bytes/std` to the crate's `json` feature, and enable it through a dev-dependency for tests.

crates/seatbelt/Cargo.toml
```diff
@@ -90,7 +90,7 @@ opentelemetry-stdout = { workspace = true, default-features = false, features =
 opentelemetry_sdk = { workspace = true, default-features = false, features = ["metrics", "testing", "experimental_metrics_custom_reader"] }
 rstest = { workspace = true }
 serde = { workspace = true, features = ["std", "derive"] }
-serde_json = { workspace = true }
+serde_json = { workspace = true, features = ["std"] }
```
psandana, line 93: The PR table attributes `serde_json/std` to “value/error API,” but this crate's direct uses are `from_str`/`to_string` and no `Value` use was found. Please identify the actual `std`-gated API requiring this feature or use the narrower feature that gates the serialization/deserialization APIs, then update the description accordingly.

PR #662, opened 2026-08-12
feat(release): AI-agentic release skill with multi-model consensus

martintmk:
## What

Repackages the release workflow as a unified repository skill at `.github/skills/release-packages/`, with model judgment limited to source compatibility and proc-macro contracts while deterministic scripts own version arithmetic, cascades, ordering, and atomic application.

## Architecture

- `SKILL.md` orchestrates targeted, changed, and all-package releases plus multi-model consensus.
- `release-facts.ps1` emits schema-v2 workspace facts: dependency/type exposure edges, proc-macro publication edges, implementation closures, runtime façades, baselines, and modification state.
- `resolve-plan.ps1` owns token parsing, Cargo-aware SemVer arithmetic, exact pins, fixed-point cascades, macro-contract attestations, ambiguity blocking, warnings, and dependency-first ordering.
- `apply-plan.ps1` rejects blocked or stale plans and performs manifest edits, changelog/README generation, validation, and rollback atomically.
- `scripts/ci/semver-report.ps1` remains the CI entry point over the shared release library.

## Important behavior

- Ordinary Rust dependency breaks propagate through public type-exposure edges; every direct released dependency still gives published dependents a patch pickup floor.
- Proc-macro implementation dependency versions do **not** automatically imply a breaking consumer contract. Macro releases require reviewed attestations covering exports, accepted syntax, compile behavior, generated API/runtime paths, and hygiene.
- Missing, incomplete, contradictory, or ambiguous macro reviews block resolution instead of guessing.
- Public macro façades are inferred from concrete allowlist roots. Wildcards and unpublished consumers are not treated as publication evidence; optional `macro_runtime` metadata is reserved for exceptional generated-runtime relationships.
- Unpublished implementation helpers participate in macro review scope without becoming release candidates.
- Exact proc-macro pins and cascade-reached non-patch macro classifications still require contract review.
- First releases, pins, force warnings, downstream effective types, dependency ordering, stale-plan checks, and rollback remain deterministic.

## Validation

- **541 Pester tests pass**, covering stable, `0.x`, and `0.0.x` arithmetic; pins; first releases; fixed-point chains and diamonds; ordinary exposure; compatible/breaking proc-macro contracts; public/private macro use; renamed and wildcard edges; generated-runtime ambiguity; unpublished helpers; stale schemas; malformed attestations; changelogs; and atomic rollback.
- Workspace formatting, README generation, spelling, and Clippy pass.
- The live `templated_uri` oracle keeps `templated_uri_macros_impl` and `templated_uri_macros` at patch for a compatible macro contract while preserving `templated_uri`'s independent breaking `ohno` type exposure.
- The last **10 distinct historical release commits** were replayed in isolated worktrees, with the audit posted at https://github.com/microsoft/oxidizer/pull/662#issuecomment-5281783303.
- Independent Claude and GPT reviews were applied, including fail-closed fixes for stale facts, exact pins, cascade-reached macro classifications, wildcard/renamed publication evidence, registry dependencies, and unpublished implementation helpers.

The retired prompt and top-level helper locations are removed; release documentation now points to the unified skill bundle.


.github/prompts/release-packages.prompt.md
```diff
@@ -0,0 +1,435 @@
...
+`cargo semver-checks` verdicts) and the documented rules below -- NOT of which
+model runs it. Two different reasoning models given the same repository state must
+produce the identical plan. To guarantee this:
+
+1. Do the mechanical, error-prone work with the deterministic helper scripts, not
+   by hand (see "Deterministic helpers").
+2. Apply the classification and cascade rules verbatim. Where a rule says
+   "objective floor", never go below it and never elevate above it without
+   concrete, cited evidence from a diff.
+3. Before writing anything, run the Multi-model consensus gate. If independent
```
martintmk: > the rules were ambiguous for that case

can we also report back the ambiguity and to improve this script

.github/prompts/release-packages.prompt.md
```diff
@@ -0,0 +1,435 @@
...
+
+Everything else -- token parsing, change-type classification decisions, cascade
+resolution, version arithmetic, elevation decisions, and the consensus gate -- is
+your reasoning, done under the rules below.
+
+## Inputs
+
+You are invoked in one of three modes (mirrors the old `-Packages`/`-Changed`/`-All`):
+
+- Targeted: an explicit list of `name@change-spec` tokens.
```
martintmk:
```suggestion
- Targeted: an explicit list of `name` tokens.
```

do we really need change spec? Can't the model determine the change?

.github/prompts/release-packages.prompt.md
martintmk: &#x1F916; Agreed - on consensus divergence I'll have the gate emit a structured ambiguity report (which package/rule diverged, each model's decision, and a concrete proposed edit to tighten this skill) instead of just halting. Implementing now.

---

.github/prompts/release-packages.prompt.md
martintmk: &#x1F916; Good point - the model already classifies each package objectively (cargo-semver-checks for libraries, diff review for proc-macros), so the change type is redundant for the common case. I'll make the change-spec optional: a bare `name` token lets the model determine the change, while `name@type` (lower-bound override) and `name@version` (explicit pin) stay available. Implementing now.

---

.github/prompts/release-packages.prompt.md
martintmk: &#x1F916; Fixed in 30809f10: on divergence the consensus gate now emits a structured ambiguity report - the diverging package/rule, each model's decision, a quote of the under-specified sentence, and a concrete proposed edit to this skill plus a recommended next action - so each divergence tightens the skill instead of only blocking.

---

.github/prompts/release-packages.prompt.md
martintmk: &#x1F916; Fixed in 30809f10: the change-spec is now optional. A bare `name` token lets the skill determine the change type objectively (cargo-semver-checks + diff review, as it already does for cascade dependents); `name@type` stays a lower-bound override and `name@version` an explicit pin. Updated Inputs, Step 1, and Step 2.

---

martintmk:
can you simulate a release of `tick` crate according to change history and output the affected versions according to new prompt?

martintmk:
review this PR

.github/prompts/release-packages.prompt.md
```diff
@@ -0,0 +1,469 @@
+---
```
martintmk: can you simulate a release of tick crate according to change history and output the affected versions according to new prompt?

.github/prompts/release-packages.prompt.md
```diff
@@ -0,0 +1,469 @@
...
+  Regenerates one `crates/<folder>/CHANGELOG.md` (conventional-commit sections,
+  PR links, `## Unreleased` folding, and cascade "Now requires `X` of `Y`"
+  bullets). `CascadeReasonsJson` is `[{"Target","Version","Breaking"}]`.
+
+- `just readme` (or `just package=<name> readme`)
+  Regenerates `README.md` files from crate docs. Never hand-edit generated READMEs.
+
+Everything else -- token parsing, change-type classification decisions, cascade
+resolution, version arithmetic, elevation decisions, and the consensus gate -- is
+your reasoning, done under the rules below.
```
martintmk: The most safety-critical mechanical work is the part that was *not* forwarded. `Resolve-ReleaseSet` (`scripts/lib/release-flow.ps1:381`) already implements Steps 1-2, 4, 6 and 7 — it takes `-ParsedTokens`, `-WorkspaceBaseline`, a `-GetRequiredChangeType` classifier scriptblock and `-Force`, does the transitive dependent closure, applies `Get-StrongerChangeType` for elevation and `Get-NextVersion` for the version-bump table, and enforces the "pin must be strictly greater" rule — and it is covered by the golden Pester scenarios this prompt cites as its oracle.

So the skill forwards the lowest-risk sub-task (changelog prose) and re-derives the highest-risk one (which packages, which numbers) in prose, then compensates with a model vote. A third helper in the same shape as the other two would invert that:

```
scripts/release-plan.ps1 -Tokens <tokens> [-Force]   # -> canonical plan JSON
```

calling `Parse-ReleaseTokens` + `Resolve-ReleaseSet` with the `cargo semver-checks` classifier. That makes the affected-package set and the bump sequence deterministic *by construction* rather than by agreement, deletes roughly 150 lines of this prompt (Steps 1-2, 4, 6-7 and the version-bump table), and removes the duplicate spec that can drift from the library. The agent's remaining job is Step 5, which is where the judgment actually is — and the consensus gate then covers exactly that.

.github/skills/release-packages/scripts/release-facts.ps1
```diff
@@ -0,0 +1,107 @@
...
+        name              = $package.Name
+        version           = $package.Version
+        published         = [bool]$package.Published
+        procMacroOnly     = [bool]$package.IsProcMacroOnly
+        hasLibraryTarget  = [bool]$package.HasLibraryTarget
+        deps              = @($package.Deps)
+        baselineSha       = $baselineSha
+        hasBaseline       = ($null -ne $baselineSha)
+        modified          = $modified.ContainsKey($package.Folder)
+        modifiedFileCount = if ($modified.ContainsKey($package.Folder)) { [int]$modified[$package.Folder] } else { 0 }
```
martintmk: `hasBaseline` cannot distinguish the two cases Step 3 branches on, so Correction #2 guards a branch that in practice never fires.

`Get-PreviousVersionBumpCommit` treats the `null -> 0.1.0` transition at the crate-introducing commit as a version bump (`releasing.ps1:449-451`), so a crate that has never been released still gets a baseline. Verified on this PR head: every never-tagged crate in the workspace (`testing_aids`, `automation`, `benchmarking`, `rest_over_grpc_examples`, `rest_over_grpc_tests`) reports `hasBaseline = true` with `baselineSha` equal to the commit that *added* its `Cargo.toml`. No package in the workspace reports `hasBaseline = false`.

The live path is the one that misbehaves: a genuinely brand-new published crate on its first release gets `cargo semver-checks --baseline-rev <its own introducing commit>`, so ordinary pre-publication API churn classifies as `breaking` and the crate ships its *first* release as `0.2.0` instead of `0.1.0`. Several in-flight PRs add new published crates, so this is reachable soon.

The fact base should answer the question directly rather than leaving the agent to hand-check tags and changelogs (which "Deterministic helpers" forbids). Something like:

```suggestion
        deps              = @($package.Deps)
        baselineSha       = $baselineSha
        hasBaseline       = ($null -ne $baselineSha)
        # Whether the crate has ever been published. A crate's introducing commit
        # counts as a version bump, so hasBaseline alone cannot tell a first-ever
        # release from a real one - and cargo-semver-checks against an unpublished
        # baseline yields a floor with no published API behind it.
        everReleased      = [bool](Invoke-Git -Arguments @('tag', '--list', "$($package.Name)-v*") -RepoRoot $RepoRoot)
        modified          = $modified.ContainsKey($package.Folder)
        modifiedFileCount = if ($modified.ContainsKey($package.Folder)) { [int]$modified[$package.Folder] } else { 0 }
```

with Step 3 branching on `everReleased` instead of `hasBaseline`.

.github/prompts/release-packages.prompt.md
```diff
@@ -0,0 +1,469 @@
...
+  - a required major bump -> `breaking`
+  - only a required minor bump -> `nonbreaking`
+  - compatible / no update required -> `patch`
+- Ordinary library crate with `hasBaseline = false`:
+  - If it is genuinely brand-new (never released: no `<name>-vX.Y.Z` git tag AND
+    no prior released version in its `CHANGELOG.md`), it imposes NO floor
+    (`none`).
+  - Otherwise (it has released before but no baseline commit could be resolved),
+    do NOT treat it as unconstrained. Set floor `patch` and flag
+    `manualReview = true`. [Corrected defect -- see Corrections #2.]
```
martintmk: Follow-on from the `release-facts.ps1` comment: this branch is unreachable in practice (`hasBaseline` is true even for a never-released crate, because its introducing commit counts as a version bump), and it asks the agent to hand-check git tags and parse `CHANGELOG.md` — which line 44 explicitly forbids and which is exactly what the fact base exists to prevent. Branch on an `everReleased` fact instead, and give the first-ever-release case its own rule (no floor; release at the declared version).

.github/prompts/release-packages.prompt.md
```diff
@@ -0,0 +1,469 @@
...
+workspace dependencies via `<dep>.workspace = true`, the ONLY place a dependency
+version requirement lives is the root `[workspace.dependencies]` table -- you do
+not edit consumers' `[dependencies]`.
+
+For each released package, in order:
+1. Set `[package].version` in `crates/<folder>/Cargo.toml` to
+   `effectiveTargetVersion` (edit the `version = "..."` line only).
+2. Set that package's entry in the root `Cargo.toml` `[workspace.dependencies]`
+   table to the new version (plain `version = "x.y.z"`, no `^`, no `=` -- match
+   the existing format exactly).
```
martintmk: "Set that package's entry ... to the new version (plain `version = "x.y.z"`)" reads as replacing the entry. Real entries are inline tables:

```
anyspawn = { path = "crates/anyspawn", default-features = false, version = "0.7.0" }
```

Dropping `path` would fail `cargo check` and be caught by the rollback, but dropping `default-features = false` would not — it silently enables default features workspace-wide and the validation step still passes. Say what the tested `Update-PackageVersion` actually does (`release-flow.ps1:844` replaces only the version *value* inside the existing table):

```suggestion
2. In the root `Cargo.toml` `[workspace.dependencies]` table, edit ONLY the
   `version = "..."` value inside that package's existing inline table; leave
   `path`, `default-features` and every other key byte-identical. Plain
   `x.y.z`, no `^`, no `=`. Do not rewrite the entry as a bare version string --
   dropping `default-features = false` changes feature resolution workspace-wide
   and `cargo check` will NOT catch it.
```

.github/prompts/release-packages.prompt.md
```diff
@@ -0,0 +1,469 @@
...
+  repo root. Version pinned in `constants.env` (`CARGO_SEMVER_CHECKS_VERSION`).
+
+- `scripts/release-changelog.ps1 -RepoRoot <path> -PackageFolder <folder>
+  -NewVersion <x.y.z> -PrBaseUrl <url> [-CascadeReasonsJson <json>]`
+  Regenerates one `crates/<folder>/CHANGELOG.md` (conventional-commit sections,
+  PR links, `## Unreleased` folding, and cascade "Now requires `X` of `Y`"
+  bullets). `CascadeReasonsJson` is `[{"Target","Version","Breaking"}]`.
+
+- `just readme` (or `just package=<name> readme`)
+  Regenerates `README.md` files from crate docs. Never hand-edit generated READMEs.
```
martintmk: nit: `just package=<name> readme` does not narrow anything. The `readme` recipe (`justfiles/basic.just:135`) ignores the `package` variable and always runs `cargo workspaces exec --ignore-private cargo doc2readme ...` across the workspace. An agent following this hint would believe it scoped the regeneration and instead drag unrelated README churn into the release commit.

```suggestion
- `just readme`
  Regenerates every crate's `README.md` from its docs (workspace-wide; the
  `package=` variable does not narrow this recipe). Never hand-edit generated READMEs.
```

scripts/release-facts.ps1
```diff
@@ -0,0 +1,107 @@
...
+
+# Validate BaseRef once, loudly. An unresolvable or un-fetched ref must fail here
+# rather than being swallowed per-package and disguised as "no baseline", which
+# would silently drop each crate's cargo-semver-checks floor and let an
+# under-incremented release pass. Test-GitRef is the same check
+# Get-PreviousVersionBumpCommit uses internally; doing it once up front gives a
+# single clear failure instead of one per package.
+if (-not (Test-GitRef -Ref $BaseRef -RepoRoot $RepoRoot)) {
+    throw "Base ref '$BaseRef' could not be resolved in '$RepoRoot'. Ensure it is fetched (CI should checkout with fetch-depth: 0) and spelled correctly."
+}
```
martintmk: nit: the justification describes behaviour the library does not have. `Get-PreviousVersionBumpCommit` already runs the same `Test-GitRef` and *throws* on an unresolvable ref (`releasing.ps1:433-435`), with a near-identical message — it never swallows it per-package or disguises it as "no baseline". The check is still worth keeping for the single clear failure, but the comment should say only that, otherwise it reads as guarding against a defect that was fixed.

docs/releasing.md
```diff
@@ -35,6 +35,48 @@ This document is the reference for the human-driven release tooling in
...
+floor plus mandatory manual review instead of silently dropping to "no
+constraint"; and dependency-requirement edits are validated with
+`cargo check` / `cargo metadata` rather than trusted. The intentional
+Cargo `0.x` / `0.0.x` conservatism in the version-bump table is preserved
+exactly.
+
+The PowerShell driver and `scripts/ci/semver-report.ps1` remain in place:
+CI still consumes the report, and the tested library is the source of the
+helpers' behavior. The rest of this document (glossary, invariants,
+version-component rules) is the shared specification both paths implement.
```
martintmk: The two paths no longer implement the same input grammar, so this sentence overstates it. `Parse-ReleaseTokens` (`release-flow.ps1:167-170`) rejects any token without `@` — `bytesbuf` alone is a hard error for the CLI driver — while the skill now accepts a bare `name` and derives the change type itself. The plan JSON also spells the change type `nonbreaking`, whereas the library's internal vocabulary (and this document) use `non-breaking`; `Get-NextVersion`'s `ValidateSet` would reject the prompt's spelling if a plan were ever handed back to the library.

Both are defensible, but worth one sentence here so a reader does not assume tokens are interchangeable between the two paths.

martintmk (commented):
[Copilot speaking]

Reviewed the AI-agentic release skill and its two helpers against the existing tested release library, and verified the claims below against the real workspace (53 packages) on the PR head. The helpers are clean thin shells and the new Pester tests are green (13/13), but the split of "what gets forwarded to deterministic code" is inverted, and one of the headline corrections guards a branch that cannot fire.

**Verdict: changes requested.** (`REQUEST_CHANGES` is not accepted on your own PR, so this is posted as a comment.)

### Design note, not attached to a single line

The determinism argument leans on the consensus gate, but a 3-sample vote is evidence of *agreement*, not of determinism: the same model can diverge run-to-run, so unanimity on four cases does not establish the property the PR claims. It also cannot run in CI and multiplies the cost and latency of every release. That is acceptable for the genuinely subjective part — the Step 5 diff judgments that replace the operator's TTY call. It is the wrong tool for the cascade closure and the version arithmetic, which are already deterministic code in `scripts/lib/`. Forwarding those (finding 1) would shrink the gate to the part that actually needs it and let the version-bump table, the elevation ranks and the pin rules stop being restated in prose where they can drift from `Get-NextVersion` / `Get-StrongerChangeType`.

Verified while reviewing: `scripts/release-facts.ps1` runs clean on the real workspace (exit 0, ~34s, 53 packages); `scripts/release-changelog.ps1` produces the expected `🔧 Maintenance` / `⚠️ Breaking` sections; `Invoke-Pester` on both new files passes 13/13. I also checked whether the breaking-cascade test was coverage-shaped and it is not — a non-breaking cascade renders no literal `Breaking`, so the assertion does discriminate.

.github/prompts/release-packages.prompt.md
martintmk: &#x1F916; Sure - running the skill's algorithm now: gathering facts via `release-facts.ps1`, classifying `tick` with `cargo semver-checks` against its baseline commit, cascading to its published dependents, and I'll post the resulting affected-package + version-bump plan.

---

.github/prompts/release-packages.prompt.md
martintmk: the whole point of the release promt is eventually get rid of the script-heavy infra around releasing

.github/prompts/release-packages.prompt.md
martintmk: &#x1F916; Done - I ran the new skill's algorithm end to end against the current tree.

**Method (all per the prompt):**
1. `scripts/release-facts.ps1` for the fact base (graph, versions, baseline commits, modified set).
2. `cargo semver-checks --package tick --baseline-rev 7c185b44 --all-features --color never` to classify `tick` objectively.
3. Cascade to `tick`'s transitive published dependents (floor at `patch`), version math via the 0.x table.

`tick` was passed as a bare `name` (no change-spec), so the skill determined the change type itself. cargo-semver-checks verdict: **`no semver update required`** - `tick`'s one unreleased file changed internal code, not its public API - so `tick` classifies as **patch**.

**Resulting plan - 9 affected packages:**

| package | from -> to | change | source |
|---|---|---|---|
| tick | 0.5.0 -> 0.5.1 | patch | user |
| anyspawn_azure | 0.2.0 -> 0.2.1 | patch | cascade |
| cachet | 0.11.0 -> 0.11.1 | patch | cascade |
| fetch | 0.15.0 -> 0.15.1 | patch | cascade |
| fetch_azure | 0.5.0 -> 0.5.1 | patch | cascade |
| fetch_hyper | 0.6.0 -> 0.6.1 | patch | cascade |
| http_extensions | 0.9.0 -> 0.9.1 | patch | cascade |
| seatbelt | 0.7.0 -> 0.7.1 | patch | cascade |
| seatbelt_http | 0.7.0 -> 0.7.1 | patch | cascade |

**Notes:**
- The cascade is the patch floor: each dependent re-releases only to pick up `tick 0.5.1`. Because `tick`'s change is API-compatible, nothing breaks from it, so none are raised above patch. (The skill still runs cargo-semver-checks per dependent; a dependent is raised only if its OWN public API changed.)
- Some dependents also have their own unreleased edits. Those are a separate decision - Step 5 elevation review would surface them for a broader release, but "release `tick`" alone yields the patch cascade above.
- `fetch_azure` is included transitively (it depends on `fetch`, which depends on `tick`).

---

.github/prompts/release-packages.prompt.md
martintmk: &#x1F916; Done - took direction 2 and simplified the scripts. Retired the interactive driver (`release-packages.ps1`) and the whole plan engine (`release-flow.ps1` + the plan/cascade/elevation half of `releasing.ps1`) plus their scenario / PromptFlow / Resolve tests. The prompt is now the single source of truth for planning; only small tested helpers remain (`release-facts.ps1`, `release-changelog.ps1` -> new `changelog.ps1`, and the unchanged CI `semver-report.ps1`). The deletion set was verified via an AST call-graph closure; all 145 remaining Pester tests and `semver-report.ps1` stay green. Net -9284 lines.

---

.github/skills/release-packages/scripts/release-facts.ps1
martintmk: &#x1F916; Fixed: `release-facts.ps1` now emits an `everReleased` fact (`[bool]` of the crate's `<name>-v*` release tags), as you suggested. New test asserts a never-released crate reports `everReleased=false` even though `hasBaseline=true`.

---

.github/prompts/release-packages.prompt.md
martintmk: &#x1F916; Fixed: Step 3 now branches on `everReleased` instead of `hasBaseline`. A never-released crate imposes no floor and ships its first release at its declared version - no `cargo semver-checks` against its own introducing commit, and no hand-checking of tags/changelogs.

---

.github/prompts/release-packages.prompt.md
martintmk: &#x1F916; Fixed: the apply step now says to edit ONLY the `version` value inside the existing `[workspace.dependencies]` inline table, leaving `path` / `default-features` / every other key byte-identical, with your exact warning that dropping `default-features = false` is not caught by `cargo check`.

---

.github/prompts/release-packages.prompt.md
martintmk: &#x1F916; Fixed: adopted your suggestion - `just readme` regenerates every crate's README workspace-wide and the note now says the `package=` variable does not narrow it.

---

scripts/release-facts.ps1
martintmk: &#x1F916; Fixed: softened the comment - it now states only that `Get-PreviousVersionBumpCommit` already runs `Test-GitRef` and throws on a bad ref, so the up-front check just fails once here rather than once per package. No claim of guarding a defect.

---

docs/releasing.md
martintmk: &#x1F916; Addressed: retiring the CLI driver removes the divergent input grammar entirely - the prompt is now the only input path (bare `name`, or `name@type` / `name@version`). The library's `Get-NextVersion` (and its `non-breaking` ValidateSet) is deleted too, so the plan JSON's `nonbreaking` spelling can no longer be rejected by the library. Docs note the token spelling.

---

martintmk:
PR description is broken, try to update it

.github/prompts/release-packages.prompt.md
```diff
@@ -0,0 +1,489 @@
+---
```
martintmk: simulate the release of bytesbuf with a new major version

martintmk:
PR description is wrong, update and refresh it according to latest changes

.github/prompts/release-packages.prompt.md
martintmk: &#x1F916; Agreed - I’m simulating `bytesbuf@major` against the current PR head now, including the full dependent cascade, elevation rules, and resulting version plan.

---

martintmk: &#x1F916; Simulation complete for targeted token `bytesbuf@major` at PR head `7631363e`.

I ran `release-facts.ps1`, then `cargo semver-checks` for every crate in the transitive dependent closure. All nine objective API verdicts were compatible (`patch` floor). The explicit `major` lower bound therefore raises only `bytesbuf`; every dependent re-releases at the cascade patch floor.

| package | from -> to | change | source |
|---|---|---|---|
| bytesbuf | 0.8.0 -> 0.9.0 | breaking | user |
| bytesbuf_io | 0.8.0 -> 0.8.1 | patch | cascade |
| cachet | 0.11.0 -> 0.11.1 | patch | cascade |
| http_extensions | 0.9.0 -> 0.9.1 | patch | cascade |
| multitude | 0.8.0 -> 0.8.1 | patch | cascade |
| fetch_hyper | 0.6.0 -> 0.6.1 | patch | cascade |
| seatbelt_http | 0.7.0 -> 0.7.1 | patch | cascade |
| fetch | 0.15.0 -> 0.15.1 | patch | cascade |
| fetch_azure | 0.5.0 -> 0.5.1 | patch | cascade |

Apply order is dependency-first as shown (the independent `fetch_hyper` / `seatbelt_http` entries may swap without changing the plan). `fetch` cascades after `fetch_hyper`, `http_extensions`, and `seatbelt_http`; `fetch_azure` follows `fetch`. No proc-macro manual review is required.

**Consensus:** unanimous across GPT-5.6, Claude Opus, and Gemini. All agreed on the same normalized `folder + target + changeType + source` tuples.

No files were changed; this was simulation only.

---

martintmk: Not this evaluation is wrong. If bytesbuf has a breaking change and the public APIs of bytesbuf  are used in bytesbuf_io then bytesbuf_io has a breaking change too. 

This applies across multiple crates. http_extensions exposes public types from bytesbuf  so this is breaking change too. 

The prompt and scripts must be aware of this. The best way to determine whether a breaking change crate dependency would cause breaking change version could be retrieved from Cargo.toml and `[package.metadata.cargo_check_external_types]`.

For example:

[package.metadata.cargo_check_external_types]
allowed_external_types = ["bytesbuf::*", "ohno::*", "futures_core::stream::Stream"]


use multiple reasoning models to come with the best solution and also appropriately fix the promt and scripts

martintmk: &#x1F916; Agreed - the simulation under-classified dependents that expose `bytesbuf` types in their public API. I’m updating the deterministic fact base to capture external-type exposure metadata, revising the cascade rule so breaking dependencies propagate through those public API edges, and adding focused tests. I’ll use independent reasoning models to compare the design before implementing it.

---

martintmk: &#x1F916; Fixed: the release skill now models public dependency exposure explicitly and propagates breaking changes to a fixed point.

The corrected `bytesbuf@breaking` plan is:

| package | from -> to | change |
|---|---|---|
| bytesbuf | 0.8.0 -> 0.9.0 | breaking |
| bytesbuf_io | 0.8.0 -> 0.9.0 | breaking |
| cachet | 0.11.0 -> 0.12.0 | breaking |
| http_extensions | 0.9.0 -> 0.10.0 | breaking |
| multitude | 0.8.0 -> 0.9.0 | breaking |
| fetch_hyper | 0.6.0 -> 0.7.0 | breaking |
| seatbelt_http | 0.7.0 -> 0.8.0 | breaking |
| fetch | 0.15.0 -> 0.16.0 | breaking |
| fetch_azure | 0.5.0 -> 0.6.0 | breaking |

Exposure facts are derived from checked external-type metadata for ordinary libraries. Proc-macro-only and other unchecked targets conservatively expose every dependency. The resolver applies Cargo compatibility rules, including every `0.0.z` transition, across the full dependent graph. The behavior is covered by synthetic graph, proc-macro, and real Cargo API tests.

---

README.md
```diff
@@ -107,10 +106,10 @@ this simple process:
 
 2. Create a branch off of main.
 
-3. Run `./scripts/release-packages.ps1 -Packages '<crate_name>@<change_type>'` to update versions and changelogs.
-   The change type for each package is one of `breaking`, `nonbreaking`, `patch`, or an explicit version like
-   `1.0.0`. To release several crates together, list them all in the same `-Packages` argument
-   (for example, `'foo@nonbreaking','bar@patch'`); the script plans the entire release up-front.
+3. Invoke the repository's `release-packages` skill with `<crate_name>@<change_type>`.
```
martintmk:
```suggestion
3. Invoke the repository's `release-packages` skill with `<crate_name>`.
```

the skill should determine what kind of change it is based on GitHub history

martintmk: &#x1F916; Agreed - the default invocation should be a bare crate name, with the skill deriving the change type from repository history and source review. I’m updating the README now while retaining explicit lower bounds and pins as optional advanced inputs.

---

martintmk:
> [!NOTE]
> 🤖 Historical release audit generated by GitHub Copilot using the release skill in this PR, `cargo-semver-checks 0.48.0`, direct source review, and independent GPT/Claude model-family reviews.

## Historical validation: last 10 distinct release commits

I replayed the last 10 distinct commits referenced by package release tags in a separate clone. A single release commit can own many tags, so auditing the last 10 tag names would only sample part of the latest batch.

For each commit the replay:

1. checked out the released source in an isolated worktree;
2. restored only package and workspace dependency versions to their parent values;
3. rebuilt release facts as of the parent, including historically reachable tags;
4. ran objective SemVer checks where a prior published baseline existed;
5. resolved bare package tokens with the current deterministic resolver;
6. reviewed source, public API, behavior, proc-macro contracts, and cascade edges independently.

The sample covers **10 release commits, 65 package tags, first releases, patch/nonbreaking/breaking changes, proc-macro bundles, stable `0.x` cascades, and staged batch releases**. The synthetic validation suite remains **476 passing tests**, including the 304-case resolver oracle.

## Release-by-release result

| # | Release | Actual packages | Result |
|---|---|---:|---|
| 1 | [`7c185b44`](https://github.com/microsoft/oxidizer/commit/7c185b447c5c1c94db36c6176d42093aa67b83a2) — bytesbuf 0.8 batch | 28 | Most direct and exposed-cascade breaking bumps were correct. Found a real `plurality` under-bump and two omitted dependency-floor releases. |
| 2 | [`aae9c5eb`](https://github.com/microsoft/oxidizer/commit/aae9c5eb8b5d30df486ae3b8b43e8dbbc26b0822) — templated-uri macro boundary | 3 | Macro 0.3 elevations were intentional compatibility-boundary repair. Strict skill policy would also patch-release five dependents; existing Cargo ranges remained compatible. |
| 3 | [`b6509226`](https://github.com/microsoft/oxidizer/commit/b6509226b0dffd8b3c2ec737c87516b3781cd620) — rest-over-grpc 0.2 | 1 | First recorded publication. The 0.2 choice was defensible because the diff contained breaking behavioral/generated-contract changes and there were no published dependents. |
| 4 | [`0de29c65`](https://github.com/microsoft/oxidizer/commit/0de29c656d919c6a3fce94ad7df8c2de9557550e) — internity bundle | 3 | Correct first releases at 0.1.0, including proc-macro review boundaries. |
| 5 | [`f9fcf506`](https://github.com/microsoft/oxidizer/commit/f9fcf5069412ca47dfa9f35c35a21c1992ebe150) — fetch-winhttp | 1 | Correct first release; design-only crate with no dependent cascade. |
| 6 | [`94229b87`](https://github.com/microsoft/oxidizer/commit/94229b8778baab215e5bf9b3af4edd5f92178b0a) — bytesbuf 0.7 batch | 22 | Released bumps were consistent. The skill would additionally require `multitude 0.8.0` and `rest_over_grpc 0.2.1`; `multitude` is the material omitted exposed-breaking cascade. |
| 7 | [`78aa4470`](https://github.com/microsoft/oxidizer/commit/78aa447029b311716860b74cb3870e7b3643ed36) — multitude streaming/stats | 1 | **Under-versioned:** shipped 0.7.1, minimum valid target 0.8.0. |
| 8 | [`fa49a453`](https://github.com/microsoft/oxidizer/commit/fa49a453bab5a6a8a6f105b8907f5ca14b70035f) — arena pinning/deserialization | 2 | Both direct breaking bumps were correct. Resolver-proposed consumers were staged and then released at the same predicted targets in later commits. |
| 9 | [`958d222a`](https://github.com/microsoft/oxidizer/commit/958d222a01fd780f8d6d5c2fd9965c4431176a1b) — plurality 0.1.1 | 1 | Valid first release; skipping 0.1.0 cannot be reconstructed from a bare token but is not a SemVer defect. Deferred consumers were later released at the resolver's exact targets. |
| 10 | [`8ae7af65`](https://github.com/microsoft/oxidizer/commit/8ae7af6590bc415d84ba9abaf4aa239577da975a) — multitude Serde/Cow | 3 | Numeric versions were valid. `multitude` should have been classified nonbreaking rather than patch, but both classifications produce 0.6.2 under repository rules. Proc-macro first releases were correct. |

## Confirmed historical defects

### 1. `plurality 0.2.1` should have been `0.3.0`

The latest batch added manual `UnwindSafe`/`RefUnwindSafe` implementations for public `Box`, `Rc`, and `Arc` types. They replaced structural auto-trait derivation and narrowed allocator bounds.

A cross-review compiled the real source on both sides and reproduced:

```text
plurality 0.2.0: Box<u64, InteriorMutableAllocator>: UnwindSafe
plurality 0.2.1: Box<u64, InteriorMutableAllocator>: !UnwindSafe
```

The same regression exists for `Rc` and `Arc`. This is a compiler-visible public compatibility break, not only a classification-label issue. The correct transition was **0.2.0 -> 0.3.0**. `cargo-semver-checks` did not detect the narrowed generic auto-trait bounds.

Evidence: [`boxed.rs`](https://github.com/microsoft/oxidizer/blob/7c185b447c5c1c94db36c6176d42093aa67b83a2/crates/plurality/src/boxed.rs), [`rc.rs`](https://github.com/microsoft/oxidizer/blob/7c185b447c5c1c94db36c6176d42093aa67b83a2/crates/plurality/src/rc.rs), [`sync.rs`](https://github.com/microsoft/oxidizer/blob/7c185b447c5c1c94db36c6176d42093aa67b83a2/crates/plurality/src/sync.rs).

### 2. `multitude 0.7.1` should have been `0.8.0`

Release `78aa4470` added feature-accessible public API:

- `Arena::try_deserialize_json_each`;
- `Arena::try_deserialize_json_each_with_limits`;
- public, non-exhaustive `JsonEachError`.

These are backward-compatible public additions, so the source-review elevation rule requires `nonbreaking`. For a `0.7.x` crate, the exact target is **0.8.0**, not the shipped 0.7.1. Two independent reviews reached the same result.

Evidence: [`json.rs`](https://github.com/microsoft/oxidizer/blob/78aa447029b311716860b74cb3870e7b3643ed36/crates/multitude/src/de/json.rs), [`json_each_error.rs`](https://github.com/microsoft/oxidizer/blob/78aa447029b311716860b74cb3870e7b3643ed36/crates/multitude/src/de/json_each_error.rs).

### 3. Latest batch omitted `layered 0.3.7` and consequently `rest_over_grpc 0.2.1`

The latest batch released `plurality 0.2.1` but left previously released direct dependent `layered` at 0.3.6. Under the skill's dependency-floor invariant, every released dependency gives a publishable, previously released direct dependent a patch floor even when the dependency transition is compatible and private.

The missing `layered 0.3.7` then creates a patch floor for exposed dependent `rest_over_grpc`, requiring 0.2.1. This is one primary omission plus one derivative omission.

Evidence: [`layered/Cargo.toml`](https://github.com/microsoft/oxidizer/blob/7c185b447c5c1c94db36c6176d42093aa67b83a2/crates/layered/Cargo.toml), [`rest_over_grpc/Cargo.toml`](https://github.com/microsoft/oxidizer/blob/7c185b447c5c1c94db36c6176d42093aa67b83a2/crates/rest_over_grpc/Cargo.toml).

### 4. Bytesbuf 0.7 batch omitted the exposed `multitude 0.8.0` cascade

Release `94229b87` moved `bytesbuf` from 0.6.0 to 0.7.0. `multitude` publicly exposes bytesbuf types, so the breaking dependency transition required a breaking dependent release. The deterministic resolver produces **multitude 0.8.0**. The historical batch omitted it.

Existing crates.io artifacts were not retroactively broken because `multitude 0.7.1` retained its old published requirement; this is a release-set consistency defect under the repository's new cascade rules rather than proof that an already downloaded artifact stopped compiling.

The same strict policy also produces `rest_over_grpc 0.2.1` after `layered 0.3.6`, but that omission is only a patch-floor alignment issue.

## Policy deviations without demonstrated consumer breakage

- Release 2's `templated_uri 0.3.6` would, under the new dependency-floor rule, also release `http_extensions`, `seatbelt_http`, `fetch_hyper`, `fetch`, and `fetch_azure` at patch targets. Existing caret requirements already admitted 0.3.6, so this is historical process drift rather than a breaking-version defect.
- Releases 8 and 9 staged broad dependent batches for several days. The later bytesbuf 0.7 batch published those crates at the resolver's exact predicted targets, so no lasting omission remained.
- Release 10's public `Cow` and Serde additions were labeled patch instead of nonbreaking, but repository arithmetic maps both to 0.6.2. This is signaling/changelog fidelity only.
- The macro compatibility repair in release 2 intentionally elevated the macro crates to 0.3.0 despite patch-only source changes; that elevation prevented incompatible resolver combinations and was justified.

## Skill changes resulting from the audit

- Source review now explicitly treats narrowed generic or auto-trait implementation bounds as breaking.
- The documented `cargo-semver-checks` blind-spot guidance now calls out manual auto-trait implementations replacing structural derivation.
- Bare package tokens remain the recommended interface; exact version pins are reserved for deliberate first-version or exceptional historical choices.

## Conclusion

The replay validates the resolver's arithmetic, fixed-point ordering, first-release behavior, proc-macro handling, and exposure-aware cascades across real history. It also demonstrates why the skill cannot delegate classification entirely to `cargo-semver-checks`: the two strongest defects were a missed public addition and a missed generic auto-trait-bound narrowing.

The audited skill would have prevented both numeric under-bumps and would have made the omitted dependency-floor releases explicit before application.

README.md
martintmk: &#x1F916; Fixed: the README now recommends a bare crate name as the default invocation. The skill derives the change type from SemVer tooling and source review; suffixes remain optional lower bounds or exact pins.

---

.github/skills/release-packages/SKILL.md
```diff
@@ -0,0 +1,83 @@
+---
```
martintmk, line 1: Use this skill to release all packages with some changes and post overview and version bumps.

martintmk: &#x1F916; Completed this as a separate review-only simulation so PR #662 remains focused on the release skill itself.

Draft PR: https://github.com/microsoft/oxidizer/pull/669

The skill selected **37 releases**: **16 direct** changes and **21 dependency cascades**, with **0 resolver warnings**. The draft includes the complete package-by-package version table, direct/cascade classification, breaking proc-macro findings, generated changelogs, lockfile updates, and READMEs. It is explicitly not intended to publish packages.

---

.github/skills/release-packages/SKILL.md
```diff
@@ -0,0 +1,83 @@
...
+   - Review modified packages according to the elevation rules in
+     `references/planning.md`.
+   - If resolution reports missing classifications, classify the complete
+     dependent closure it names and rerun with the same frozen facts.
+
+4. **Resolve mechanically**
+   - Write request JSON containing `mode`, accepted `tokens`, `classifications`,
+     and optional `force`.
+   - Run `scripts/resolve-plan.ps1 -FactsPath <facts.json> -RequestPath
+     <request.json>`.
```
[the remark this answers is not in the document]
martintmk: &#x1F916; Agreed - I’m correcting the resolver paths now.

---

.github/skills/release-packages/references/planning.md
```diff
@@ -0,0 +1,239 @@
+# Release planning
+
+This reference contains the model-owned parts of an Oxidizer release. Mechanical
+token parsing, version arithmetic, dependency closure, pin reconciliation, and
+topological ordering belong to `scripts/resolve-plan.ps1`.
```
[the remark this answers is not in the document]
martintmk: &#x1F916; Agreed - I’m updating this reference to the actual skill script path.

---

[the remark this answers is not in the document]
martintmk: &#x1F916; Fixed: the planning reference now points to `.github/skills/release-packages/scripts/resolve-plan.ps1`.

---

.github/skills/release-packages/SKILL.md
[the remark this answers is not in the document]
martintmk: &#x1F916; Fixed: the workflow now points to `.github/skills/release-packages/scripts/resolve-plan.ps1` and the apply step uses the matching full skill path.

---

martintmk (commented):
**Posted by an AI agent**

Reviewed the complete release-skill rewrite at the pinned current head with all ten Review Lens areas. I found seven blocking issues in plan validation, evidence binding, dependency classification, changelog generation, test coverage, and fact determinism, plus two documentation inconsistencies and two nits. Targeted Pester suites passed (up to 464 focused cases per area), and cargo check for templated_uri_macros_impl passed; CI currently exposes only the license/CLA check.

One finding cannot be attached to a specific changed hunk without implying that the nearby assertion itself is defective, so it is included here:

**Posted by an AI agent**

**Indirect cascade changelogs lost direct-dependency coverage**

**Problem**
The deleted ADO 7536096 integration case asserted that a to b to c makes a name direct dependency b, not transitive root c. The replacement checks resolver target core but does not apply the plan or assert the published changelog reason.

**Why this matters**
The release flow can regress to misleading changelogs for indirect cascades without any test detecting it.

**Suggested fix**
Restore an end-to-end three-package resolve-and-apply test asserting the top package names its newly released direct dependency and excludes the transitive root.

.github/skills/release-packages/scripts/release-facts.ps1
```diff
@@ -0,0 +1,995 @@
...
+            if ($published) { $targetName }
+        }
+    )
+
+    # baselineSha is the crate's previous version-bump commit, or null if none can
+    # be found. Note a crate's introducing commit counts as a bump, so in practice
+    # even a never-released crate gets a baselineSha (its own first commit) -- use
+    # the everReleased fact below, NOT hasBaseline, to tell a first-ever release
+    # apart from a real one.
+    $bump = Get-PreviousVersionBumpCommit -RepoRoot $RepoRoot -BaseRef $BaseRef -PackageFolder $package.Folder
```
martintmk, line 762: **Posted by an AI agent**

**release-facts.ps1 reuses the preceding package's baseline**

**Problem**
baselineSha is assigned only when a bump commit exists, so a package with no history inherits the previous loop iteration SHA. A synthetic workspace reproduced hasBaseline=true and the same unrelated SHA for the new package.

**Why this matters**
Fact extraction becomes package-order-dependent and can compare a new crate against another crate's history, corrupting manifest, dependency, and release classification inputs.

**Suggested fix**
Assign baselineSha on every iteration, explicitly null when no bump exists, and add a new-crate regression test.

.github/skills/release-packages/scripts/apply-plan.ps1
```diff
@@ -0,0 +1,289 @@
...
+}
+
+$changelogScript = Join-Path $PSScriptRoot 'release-changelog.ps1'
+$rootManifest = Join-Path $RepoRoot 'Cargo.toml'
+if (-not (Test-Path -LiteralPath $rootManifest)) {
+    throw "Repository root '$RepoRoot' does not contain Cargo.toml."
+}
+
+$plan = Get-Content -LiteralPath (Resolve-Path $PlanPath) -Raw | ConvertFrom-Json
+if (
```
martintmk, line 50: **Posted by an AI agent**

**apply-plan.ps1 accepts plans that bypass resolver approval**

**Problem**
The status check is conditional on the property existing, and the successful apply fixture omits status. A hand-authored or truncated plan with releases is therefore accepted without the canonical resolved envelope.

**Why this matters**
Such a plan can bypass blocked-plan, ambiguity, classification, cascade, and consensus gates before mutating versions and changelogs.

**Suggested fix**
Require a versioned plan envelope with status=resolved and required identity/fields; reject missing or unsupported envelopes before any mutation.

.github/skills/release-packages/scripts/resolve-plan.ps1
```diff
@@ -0,0 +1,1885 @@
...
+    )
+
+    # The only mechanical reading of a compile fixture: what the same consumer
+    # program did before the change versus after it.
+    if ($Baseline -eq 'pass' -and $Current -eq 'fail') { return 'breaking' }
+    if ($Baseline -eq 'fail' -and $Current -eq 'pass') { return 'non-breaking' }
+    return 'patch'
+}
+
+function ConvertTo-CompileEvidenceSide {
```
martintmk, line 545: **Posted by an AI agent**

**Macro compile evidence accepts contradictory exit codes**

**Problem**
ConvertTo-CompileEvidenceSide validates field presence but not that pass means exitCode 0 and fail means nonzero. A focused probe recorded fail/0 then fail/1 and the resolver emitted compatible with a patch release.

**Why this matters**
A newly failing proc-macro consumer can be transcribed as compatible and evade the breaking floor.

**Suggested fix**
Reject contradictory result/exitCode pairs before deriving the compile outcome and add both contradiction cases to resolver tests.

.github/skills/release-packages/scripts/resolve-plan.ps1
```diff
@@ -0,0 +1,1885 @@
...
+            }
+        }
+
+        $outcome = $null
+        if (
+            $baseline.Complete -and $current.Complete -and
+            ($baseline.Result -eq 'pass') -eq ($baseline.ExitCode -eq 0) -and
+            ($current.Result -eq 'pass') -eq ($current.ExitCode -eq 0)
+        ) {
+            if ($baseline.Revision -eq $current.Revision) {
```
martintmk, line 255: **Posted by an AI agent**

**Behavior evidence is not bound to the release snapshots**

**Problem**
Get-RegressionEvidence only requires different revision strings. A probe used unrelated-old/unrelated-new despite a different fact baseline, and the resolver accepted fail-to-pass as a patch release.

**Why this matters**
Evidence from unrelated commits can mechanically prove a behavior fix that is absent at the package baseline or current head.

**Suggested fix**
Require baseline evidence to match Fact.baselineSha and current evidence to match the frozen current snapshot; block mismatches.

scripts/lib/releasing.ps1
```diff
@@ -300,17 +284,412 @@ function Test-IsBreakingChange {
...
+
+# The set of compatibility lines a Cargo requirement admits, or $null when the
+# requirement is not understood well enough to answer. $null is a deliberate
+# "unknown" that callers must fail closed on: guessing here would ship a
+# dependency major bump as a compatible release.
+#
+# Comma-separated comparators are an AND, so they describe one line only when
+# every comparator names it. A wildcard, an unbounded range, or a mix of lines
+# is unknown.
+function Get-CargoRequirementLines {
```
martintmk, line 322: **Posted by an AI agent**

**Disjoint exact and tilde requirements are treated as compatible**

**Problem**
Test-CargoRequirementBreaking returns false for =1.2.0 to =1.3.0 and ~1.2.0 to ~1.3.0 because these forms are reduced to the same broad compatibility line, even though their accepted version sets are disjoint.

**Why this matters**
When dependency types are exposed publicly, the resolver can permit a patch release although old and new consumers cannot resolve the same dependency type identity.

**Suggested fix**
Compare admitted version-set overlap, or conservatively treat changed exact/tilde requirements as breaking when overlap cannot be proven.

.github/skills/release-packages/scripts/apply-plan.ps1
```diff
@@ -0,0 +1,289 @@
...
+            -ExpectedVersion $release.from `
+            -Version $release.to
+
+        $reasonsJson = @($release.cascadeReasons) | ConvertTo-Json -Depth 5 -Compress
+        & $changelogScript `
+            -RepoRoot $RepoRoot `
+            -PackageFolder $release.folder `
+            -PackageName $release.name `
+            -NewVersion $release.to `
+            -PrBaseUrl 'https://github.com/microsoft/oxidizer' `
```
martintmk, line 222: **Posted by an AI agent**

**Generated changelog PR links omit /pull**

**Problem**
apply-plan.ps1 passes https://github.com/microsoft/oxidizer as PrBaseUrl, while Format-ConventionalCommits appends only /<number>. A focused call generated https://github.com/microsoft/oxidizer/662.

**Why this matters**
Every PR reference written by the new release flow points to a nonexistent repository path instead of the pull request.

**Suggested fix**
Pass https://github.com/microsoft/oxidizer/pull or construct the complete /pull/<number> URL in the formatter.

.github/skills/release-packages/references/planning.md
```diff
@@ -0,0 +1,690 @@
...
+
+```powershell
+./.github/skills/release-packages/scripts/apply-plan.ps1 -PlanPath plan.json
+```
+
+The helper edits only package and workspace dependency version values, generates
+changelogs and READMEs, validates with Cargo, verifies applied versions, and
+restores every file it touched if any operation fails.
+
+## Canonical output
```
martintmk, line 630: **Posted by an AI agent · Non-blocking**

**The canonical plan example includes consensus data the resolver never emits**

**Problem**
The canonical plan.json example includes consensus, but resolve-plan.ps1 emits no consensus field and consensus runs only after the plan is frozen.

**Why this matters**
Operators and models cannot tell whether to mutate resolver output, keep a separate report, or treat the missing field as incomplete.

**Suggested fix**
Remove consensus from the resolver-output example and document separate report metadata, or define a deterministic augmentation step.

docs/releasing.md
```diff
@@ -1,604 +1,431 @@
...
+
+| Component | Responsibility |
+|---|---|
+| `release-facts.ps1` | Workspace packages, versions, dependencies, type exposure, macro relationships, release baselines, and modifications |
+| `resolve-plan.ps1` | Tokens, SemVer arithmetic, pins, type/macro contract cascades, ambiguities, and topological ordering |
+| `apply-plan.ps1` | Version writes, changelogs, README generation, Cargo validation, and rollback |
+| `release-changelog.ps1` | One deterministic changelog |
+| `scripts/ci/semver-report.ps1` | CI report for version changes already present in a PR |
+
+Shared functions remain in `scripts/lib/releasing.ps1` and
```
martintmk, line 17: **Posted by an AI agent · Non-blocking**

**Release-tooling inventories still describe the removed architecture**

**Problem**
The docs say both shared libraries remain because CI consumes them, although CI imports only releasing.ps1; an unchanged workflow comment also names deleted release-packages.ps1 and release-flow.ps1.

**Why this matters**
Maintainers are directed to nonexistent entry points and incorrect ownership/validation boundaries.

**Suggested fix**
Describe each library's actual consumer and update the workflow comment to the new skill scripts and shared libraries.

.github/skills/release-packages/scripts/resolve-plan.ps1
```diff
@@ -0,0 +1,1885 @@
...
+
+    $exposed = @($Fact.externalExposedDeps)
+    return @(
+        @($Fact.externalDepChanges) |
+            Where-Object { [bool]$_.breaking -and $exposed -contains $_.name } |
+            Sort-Object -Property name
+    )
+}
+
+function Format-ExternalExposureProbe {
```
martintmk, line 1027: **Posted by an AI agent · Nit**

**The new Format helpers return structured records, not text**

**Problem**
Format-ExternalExposureProbe, Format-SelectionDecisionOutput, and Format-MacroContractOutput return objects, unlike established Format helpers that render strings.

**Why this matters**
The names obscure whether serialization remains required and establish a conflicting helper family.

**Suggested fix**
Rename them to ConvertTo-*Records (or equivalent structured-projection names) and update call sites.

.github/skills/release-packages/scripts/apply-plan.ps1
```diff
@@ -0,0 +1,289 @@
...
+if ([string]::IsNullOrWhiteSpace($RepoRoot)) {
+    $RepoRoot = $skillRepoRoot
+} else {
+    $RepoRoot = (Resolve-Path $RepoRoot).Path
+}
+
+$changelogScript = Join-Path $PSScriptRoot 'release-changelog.ps1'
+$rootManifest = Join-Path $RepoRoot 'Cargo.toml'
+if (-not (Test-Path -LiteralPath $rootManifest)) {
+    throw "Repository root '$RepoRoot' does not contain Cargo.toml."
```
martintmk, line 46: **Posted by an AI agent · Nit**

**New release-script errors diverge from the lowercase message convention**

**Problem**
New throw and warning strings start with capitals across the four entry points instead of the repository's lowercase error-message convention.

**Why this matters**
The replacement workflow establishes inconsistent shared diagnostic vocabulary that future validation is likely to copy.

**Suggested fix**
Lowercase leading words while preserving package names, paths, and acronyms.

.github/skills/release-packages/scripts/apply-plan.ps1
```diff
@@ -0,0 +1,289 @@
...
+}
+
+$plan = Get-Content -LiteralPath (Resolve-Path $PlanPath) -Raw | ConvertFrom-Json
```
geeknoid: **Resolved plans are not bound to the source snapshot they classified** — *Correctness · High · High confidence*

The plan carries package versions but no facts digest, HEAD SHA, or working-tree fingerprint, and apply only rechecks those versions. A source, feature, dependency, or proc-macro-contract edit made after consensus therefore leaves `from` unchanged and is still applied using stale classifications and cascades. The current stale-plan test only changes the manifest version ([test](https://github.com/microsoft/oxidizer/blob/9000d1812bb4e9e707948b1e877e7c10be3096ba/scripts/tests/Pester/integration/ReleaseSkillApply.Tests.ps1#L154-L171)), so it does not exercise this case.

**Direction:** include a canonical snapshot identity (HEAD plus the relevant dirty-tree/facts digest) in resolver output and require an exact match before the first write.

**Done when:** changing any classified source, manifest, fixture, or request input after resolution makes apply fail before mutation, while the unchanged snapshot still applies.

.github/skills/release-packages/scripts/release-facts.ps1
```diff
@@ -0,0 +1,995 @@
...
+        exposureUnknown   = $uncheckedTarget
+        baselineSha       = $baselineSha
+        hasBaseline       = ($null -ne $baselineSha)
+        # Whether the crate has ever been published, determined from its release
+        # tags. A crate's introducing commit counts as a version bump, so
+        # hasBaseline alone cannot distinguish a first-ever release from a real one;
+        # cargo-semver-checks against an unpublished baseline would classify normal
+        # pre-publication churn as breaking. The release skill's Step 3 branches on
+        # this fact.
+        everReleased      = [bool](Invoke-Git -Arguments @('tag', '--list', "$($package.Name)-v*") -RepoRoot $RepoRoot)
```
geeknoid: **A tagless checkout turns every published crate into a first release** — *Correctness · High · High confidence*

`git tag --list` returning no local match is treated as authoritative `everReleased = false`. The repository already documents the relevant environment: its live integration test rewrites this field because PR merge checkouts have no tags ([lines 34–39](https://github.com/microsoft/oxidizer/blob/9000d1812bb4e9e707948b1e877e7c10be3096ba/scripts/tests/Pester/integration/ExposureCascade-RealWorkspace.Tests.ps1#L34-L39)). In such a checkout the resolver skips SemVer classification and `Get-EntryTargetVersion` keeps the declared version, so a previously published crate follows the first-release path instead of receiving a bump.

**Direction:** fail closed when release-tag completeness cannot be established, or derive publication state from an authoritative source that is available in supported agent checkouts.

**Done when:** a published crate in a tagless/shallow checkout cannot resolve as `everReleased = false`, with a regression fixture covering that checkout shape.

.github/skills/release-packages/scripts/release-facts.ps1
```diff
@@ -0,0 +1,995 @@
...
+                (
+                    $content.StartsWith('///', [StringComparison]::Ordinal) -and
+                    -not $content.StartsWith('////', [StringComparison]::Ordinal)
+                ) -or
+                $content.StartsWith('//!', [StringComparison]::Ordinal)
+            ) {
+                if ($docEligible) { $docComment = $true }
+                continue
+            }
+            if ($content.StartsWith('//', [StringComparison]::Ordinal)) { continue }
```
geeknoid: **Line-prefix filtering can hide executable string-content changes** — *Correctness · Medium · High confidence*

This diff classifier has no Rust lexical state: after trimming, any changed line beginning with `//`, `///`, or `//!` is classified as a comment. Those prefixes are also valid content lines inside multiline/raw string literals, including generated source, templates, and protocol text. If those are the only changed lines, `rustImplementationChanged` remains false and the resolver can reject a real nonbreaking/breaking classification as unsupported.

**Direction:** derive implementation-vs-doc changes with a Rust-aware parser/tokenizer, or conservatively treat ambiguous lines inside changed source as implementation.

**Done when:** fixtures changing `//`, `///`, and `//!` lines inside multiline/raw strings set `rustImplementationChanged`, while actual line-comment-only edits remain non-implementation changes.

.github/skills/release-packages/SKILL.md
```diff
@@ -0,0 +1,115 @@
...
+   - Require PowerShell 7, Cargo, `cargo semver-checks`, and a clean baseline.
+   - Record `git rev-parse HEAD` and `git status --porcelain`.
+
+2. **Gather facts**
+   - Run `.github/skills/release-packages/scripts/release-facts.ps1` and save its
+     JSON.
+   - Determine targeted, changed, or all mode.
+
+3. **Classify**
+   - For every ordinary, previously released library that may enter the plan,
```
geeknoid: **Repository-controlled prose is not separated from release instructions** — *Security · Medium · Medium confidence (Hardening)*

A repository contributor controls source comments, docs, fixtures, changelogs, and commit messages that the release models inspect. The workflow does not tell models to treat that material strictly as untrusted evidence, so embedded instructions can compete with the skill while the same agent can write request JSON and invoke the mutating apply helper. Resolver checks limit some outcomes, but they do not validate the semantic truth of model classifications; consensus also exposes multiple models to the same text.

**Direction:** define an explicit instruction hierarchy: repository content/tool output is data only, embedded directives are ignored and reported, and only the checked-in skill/reference rules plus the invoking user authorize release actions.

**Done when:** adversarial instructions in each reviewed content channel cannot alter the requested mode, evidence rules, plan, or apply decision; platform-level prompt isolation would also disprove the need for a local guard.

geeknoid (commented):
Early static feedback at `9000d1812bb4e9e707948b1e877e7c10be3096ba`: I found two release-integrity defects that should be addressed before relying on this workflow. Plans are not bound to the source/facts snapshot reviewed by consensus, and missing local tags silently route previously published crates through first-release semantics. I also found one source-classification parser defect and one AI trust-boundary hardening gap; details and acceptance criteria are inline.

The deterministic resolver structure, explicit ambiguity blocking, dependency-first ordering, rollback on ordinary command failure, and CI-gated cross-platform Pester suite are strong controls. I specifically traced the version/cascade/apply path, proc-macro evidence gates, filesystem/process boundaries, and the replacement tests.

All five taxonomies were assessed: security 150, correctness 162, testing 182, performance 183, and conformance 143. This was static-only: no project/test/build/benchmark/profile/Miri/coverage/mutation/fuzz/live execution occurred.

I deduplicated against all 51 existing inline comments. Already-raised issues—including the leaked baseline SHA, missing required plan envelope, unbound behavior evidence, exact/tilde requirement handling, changelog links, and documentation drift—are intentionally not repeated.

PR #664, opened 2026-08-13
fix(ci): let the nightly Miri job honour strict-provenance ignores

martintmk:

## Summary

Approach changed after review. The `[patch.crates-io]` git pin of `parking_lot_core` and the `deny.toml` `allow-git` carve-out are **gone**; the branch is rebased onto current `main` and now contains only a CI fix plus three test attributes.

## Root cause (revised)

The workspace *already* has the exclusion mechanism this needs. `crates/internity/tests/basic.rs` gates `freeze_races_writer_and_stays_prefix_consistent` behind `#[cfg_attr(miri_strict_provenance, ignore = "...")]`, `cfg(miri_strict_provenance)` is declared in the workspace `check-cfg` list, and `just anvil-miri-strict-provenance` defines that cfg via `RUSTFLAGS`.

The scheduled `miri-strict-provenance` job in `.github/workflows/nightly.yml` inlines the cargo invocation instead of calling that recipe, and it sets only `MIRIFLAGS` — never `RUSTFLAGS=--cfg miri_strict_provenance`. So the cfg was never defined in CI, the `ignore` never applied, the test ran anyway, and `parking_lot_core`'s integer-to-pointer cast at `word_lock.rs:320` aborted the whole `internity` test binary. That is why nightly kept failing even after the suppression was added.

## Changes

- `.github/workflows/nightly.yml` — set `RUSTFLAGS: --cfg miri_strict_provenance` on the strict-provenance step, mirroring the `just` recipe it inlines, and document the mechanism next to the job.
- `crates/internity/tests/basic.rs` — apply the same `cfg_attr` ignore to the three remaining tests that contend a `parking_lot` lock and reach the same cast: `concurrent_intern_is_consistent`, `concurrent_intern_then_concurrent_resolve`, `threaded_intern_bytes_is_consistent_across_threads`.

`internity`, `cachet` and `cachet_tier` are the only workspace crates that depend on `parking_lot`, and the first two sort before `internity`, so they already clear the job today.

## Why this shape

This is what was asked for on this PR: exclude the affected tests from strict-provenance Miri until a fixed `parking_lot_core` is released, rather than pull a third-party git revision into the build. `unknown-git = "deny"` stays intact and `deny.toml` is untouched. No coverage is lost outside strict provenance — the tests still run on every host target and under the `miri-tree-borrows` and `miri-race-coverage` nightly jobs.

crates.io still has no fixed release: `parking_lot_core` 0.9.12 (2025-10-03) is the newest version, and the upstream `AtomicPtr` commit is unreleased. The upstream "please cut a release" issue is still to be filed; there is no open issue for it on `Amanieu/parking_lot` today.

Note: setting `RUSTFLAGS` in the job env overrides `.cargo/config.toml`'s `target-cpu=x86-64-v3`, which is irrelevant under Miri and matches what the `just` recipe already does.

## Validation

- `cargo +1.96.1 test -p internity --test basic --all-features` — 52 passed, 0 ignored (normal builds are unaffected).
- Same command with `--cfg miri_strict_provenance` injected — 48 passed, **4 ignored**, no `unexpected_cfgs` warning. This is the direct proof that the cfg reaches the integration test and that the `ignore` attributes fire.
- `cargo +1.96.1 clippy -p internity --all-features --all-targets -D warnings` — clean.
- `rustfmt +nightly-2026-05-30 --check --config-path unstable-rustfmt.toml` on the changed file — clean.
- `nightly.yml` parses as valid YAML.

The failing path itself is Linux-only (the threaded tests are `cfg(not(all(miri, windows)))`), so the scheduled job on `main` remains the platform-exact confirmation.

Bug: https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7734570


Cargo.toml
```diff
@@ -345,6 +345,11 @@ clippy.wildcard_imports = "allow"
 clippy.empty_structs_with_brackets = "warn"
 clippy.panic = "warn"
 
+[patch.crates-io]
+# parking_lot_core 0.9.12 loses pointer provenance in WordLock. Use the
+# upstream AtomicPtr fix until the next crates.io release includes it.
+parking_lot_core = { git = "https://github.com/Amanieu/parking_lot.git", rev = "fbfe854e8b5407f7a898c64fb6d296c5e965c877" }
```
martintmk: we don't want to use unknown repos, it exposes us to attacks

martintmk: investigate alternatives to this fix

sandersaares:
Hmmm.

image

Do we even need parking_lot here? It can be a bit cargo-culted by AI sometimes. I suggest also trying with `std` synchronization primitives. They are better maintained and not necessarily slower.

Cargo.toml
martin-kolinek: 🤖: The pinned revision is provenance-correct **only under Miri**, so the strict-provenance job and every ordinary build compile different code.

At this rev, `core/src/word_lock.rs:17-32` selects `atomic_ptr_fetch_byte_sub`/`atomic_ptr_fetch_and` under `#[cfg(not(miri))]`; both cast `&AtomicPtr<T>` to `*const AtomicUsize` and perform an integer read-modify-write. Upstream's own comment on those lines says it: "we are loading the pointer stored in `p` at integer type, which discards provenance, and then storing that back, thus erasing the provenance in memory". The provenance-aware `AtomicPtr::fetch_byte_sub`/`fetch_and` versions (lines 34-42) are `#[cfg(miri)]` only.

Concretely: `WordLock::unlock` calls the non-Miri `fetch_byte_sub`, which erases the queue-head provenance that `with_queue_head` was careful to preserve via `with_addr`; `unlock_slow` then reloads that value and dereferences it. Contended `parking_lot::Mutex`/`RwLock` reach that bucket lock from entirely safe code in this workspace — e.g. concurrent `internity::ThreadedLexicon::intern` → `crates/internity/src/shard.rs:68-78`. And 0.9.12 at least got there through `thread_data as *const _ as usize`, an explicitly *exposing* pointer-to-integer cast that a later int-to-ptr cast can recover from; the polyfill drops that exposure.

The upstream FIXME above the `#[cfg(miri)]` versions reads "use this everywhere once we can depend on Rust 1.91" — this workspace already requires 1.93.1 (`Cargo.toml:10`), so the polyfill's reason to exist doesn't apply to us. Upstream keeps it because `parking_lot_core` declares `rust-version = "1.71.0"`, and that MSRV won't move on our timeline.

**There is no better revision to pin.** `fbfe854e` is still `HEAD` for `core/src/word_lock.rs` on upstream `master` — the commits master is ahead by don't touch that file — and master's copy still has the polyfill `cfg`-gated exactly as above. crates.io's newest release, 0.9.12, *predates* this commit, so no published version contains even the polyfilled form. So the realistic options are:

1. **Carry the one-line local patch** — drop the `cfg` gate and call `AtomicPtr::fetch_byte_sub`/`fetch_and` unconditionally, which is sound at our 1.93.1 MSRV. That's the `.patches/parking_lot_core/` shape an earlier revision of this PR had; re-adopting it means owning the CodeQL "Access of invalid pointer" alerts again (and it removes the need for the `deny.toml` `allow-git` carve-out entirely).
2. **Accept Miri-only correctness** and say so — the patch then buys a green scheduled job, not a provenance-correct shipped build. If that's the intent, it's worth stating in the `[patch.crates-io]` comment so the next reader doesn't over-read the PR title.

Upstreaming the ungating (an MSRV bump or a `rustversion`-style dispatch in `parking_lot`) is the durable fix, but it isn't available on this PR's timeline. Which of the two do you want here?

deny.toml
```diff
@@ -64,3 +64,5 @@ wildcards = "warn"
 unknown-registry = "deny"
 unknown-git = "deny"
 # <<< anvil-managed: anvil-deny-sources
+# Temporary, commit-pinned upstream fix for parking_lot_core strict provenance.
+allow-git = ["https://github.com/Amanieu/parking_lot.git"]
```
martin-kolinek: 🤖: This is the first hole in the repo's `unknown-git = "deny"` posture, and as written it's wider than the change needs.

**`allow-git` is URL-granular.** This entry doesn't permit only the pinned commit; it permits *any* revision of `Amanieu/parking_lot` to enter the build from here on, including an unpinned branch dependency added later. The `rev` that makes today's build reproducible lives only in `Cargo.toml`/`Cargo.lock` — i.e. in exactly the thing the sources policy exists to constrain. Trusting a third-party repository at build time, rather than one immutable commit, is the substance of the concern, and it's closable with one more line in `[sources]` rather than by reversing the approach:

```toml
required-git-spec = "rev"
```

cargo-deny then rejects any allowed git source not pinned to a commit SHA, so the carve-out reads "this repo, at an immutable commit" instead of "this repo". (Worth noting the option in the blocking comment on `Cargo.toml:351` — carrying the ungating as a local patch instead of a git `[patch]` — removes the need for this exemption altogether.)

**Placement and cleanup trigger.** This is the first adopter edit in the file to sit *outside* an anvil-managed block — `[advisories].ignore` (lines 10-29) and `[licenses].allow` (lines 34-49) both put adopter entries inside the block behind an `Adopter-owned:` comment, and those neighbours also carry an explicit removal trigger ("Tracked for removal once the upstream migration lands: <link>"). As written the exemption is detached from the `unknown-git = "deny"` at line 65 it exists to relax, and "Temporary" has no link or exit criterion, so nothing here points at when it can go. Either the in-block pattern survives regeneration (and this should follow it) or it doesn't (and those two lists are already at risk) — worth settling one way.

(Cross-reference: this is the concrete narrowing for the concern raised on the `Cargo.toml:351` thread — "we don't want to use unknown repos, it exposes us to attacks".)

Cargo.toml
martin-kolinek: 🤖: Non-blocking — on "investigate alternatives to this fix" and the `std` suggestion, one data point so the follow-up isn't closed on a false premise: dropping `parking_lot` would **not** retire this patch unless `internity` moves too.

- `cachet/src/refresh.rs:18` and `cachet_tier/src/testing.rs:13` use only `parking_lot::Mutex` — a drop-in for `std::sync::Mutex`.
- `internity/src/shard.rs:24,69,77` is built on `RwLockUpgradableReadGuard`, and `std::sync::RwLock` has no upgradable read at all. That fast path is a deliberate, documented concurrency tradeoff (`shard.rs:13-22`), so this is a redesign of the intern path, not a type swap.

And since `anvil-miri-strict-provenance` runs `--all-features`, `internity`'s `std` feature keeps pulling `parking_lot_core` in — so a partial migration leaves the Miri failure exactly where it is.

sandersaares (dismissed):
If we need to use random Git patch then I would rather just exclude this from the Miri checking until a proper patched version is available. Plus file an upstream "hey release a fix please" issue.

Vaiz:

@sandersaares @martintmk — taken as directed. The git patch and the `deny.toml` carve-out are both dropped; `unknown-git = "deny"` is untouched and `deny.toml` no longer appears in the diff at all. The branch is rebased onto current `main` and the PR is now two files.

While implementing the exclusion I found the actual reason nightly kept failing, which changes the fix from "add a suppression" to "make the existing suppression work":

- `crates/internity/tests/basic.rs` already gates `freeze_races_writer_and_stays_prefix_consistent` behind `#[cfg_attr(miri_strict_provenance, ignore = "...")]` (added in #706), and `cfg(miri_strict_provenance)` is already declared in the workspace `check-cfg` list.
- That cfg is defined by `just anvil-miri-strict-provenance` through `RUSTFLAGS`. The scheduled job inlines the cargo command instead of calling the recipe and sets only `MIRIFLAGS` — so the cfg was never defined in CI and the `ignore` never fired. The test ran, and `parking_lot_core` aborted the whole `internity` binary.

So the fix sets `RUSTFLAGS: --cfg miri_strict_provenance` on that job and extends the same `cfg_attr` to the three other tests that contend a `parking_lot` lock. Coverage is untouched everywhere else — those tests still run on every host target and under the tree-borrows and race-coverage Miri jobs.

Verified locally that the mechanism actually works rather than assuming it: `cargo test -p internity --test basic --all-features` reports 52 passed / 0 ignored normally, and 48 passed / **4 ignored** with `--cfg miri_strict_provenance` injected, with no `unexpected_cfgs` warning.

On the upstream ask: crates.io still shows 0.9.12 (2025-10-03) as newest, the `AtomicPtr` fix is unreleased, and there is no open "please cut a release" issue on `Amanieu/parking_lot`. I have **not** filed one — opening an issue on a third-party repo under a maintainer's identity is outside what I'll do unprompted. Say the word and I'll draft it, or file it yourself; either way it is the exit criterion for removing these four `cfg_attr`s.

Cargo.toml
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

This analysis is what settled the direction — thank you. The "Miri-only correctness" objection is decisive: pinning that revision would have bought a green scheduled job while shipped builds still ran the provenance-erasing polyfill, so the PR title would have overstated what landed.

Neither of the two options is taken. The `[patch.crates-io]` block is removed entirely, and instead the tests are excluded from strict-provenance Miri only — which is what the workspace already tried to do via `#[cfg_attr(miri_strict_provenance, ignore)]`, except the nightly job never defined the cfg. Resolving.

Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Agreed, and dropped. The `[patch.crates-io]` block is gone — no third-party git source enters the build and `unknown-git = "deny"` stays as-is.

On "investigate alternatives": the alternative turned out to be already in the tree. `basic.rs` gates the affected test behind `#[cfg_attr(miri_strict_provenance, ignore)]`, but the nightly job never defined that cfg, so it never took effect. The fix now sets `RUSTFLAGS: --cfg miri_strict_provenance` on the job and extends the gate to the other three lock-contending tests. Resolving.

Cargo.toml
```diff
@@ -248,6 +248,11 @@ wiremock = { version = "0.6.5", default-features = false }
 xxhash-rust = { version = "0.8.15", default-features = false }
 zerocopy = { version = "0.8.26", default-features = false }
 
+[patch.crates-io]
+# parking_lot_core 0.9.12 loses pointer provenance in WordLock. Use the
+# upstream AtomicPtr fix until the next crates.io release includes it.
```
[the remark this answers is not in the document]
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Obsolete — the `[patch.crates-io]` block this refers to no longer exists. Resolving.

deny.toml
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Resolved by removal rather than by narrowing: there is no `allow-git` entry any more, and `deny.toml` is not in the diff at all. The `unknown-git = "deny"` posture is unbroken, so the `required-git-spec = "rev"` question and the in-block placement question both fall away here — though the placement point about adopter edits inside anvil-managed blocks is worth raising separately if it still applies to the two existing lists. Resolving.

martin-kolinek (approved):

martin-kolinek:
@sasaares, just to double check, your random patch applies to a previous version of this PR, right? I don't see any random patches, just exclusion of a few tests.

.github/workflows/nightly.yml
```diff
@@ -130,6 +138,7 @@ jobs:
       - name: Miri (strict provenance)
         env:
           MIRIFLAGS: -Zmiri-strict-provenance
+          RUSTFLAGS: --cfg miri_strict_provenance
         run: cargo +${{ env.RUST_NIGHTLY }} miri test --all-features --workspace --lib --tests
```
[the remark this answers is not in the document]
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Valid premise, latent rather than live — fixed anyway in `f9f7a58d`, and applied to `MIRIFLAGS` as well as `RUSTFLAGS`.

The asymmetry is real: `justfiles/anvil/checks/miri-strict-provenance.just:14-15` prepends both variables (`$env:RUSTFLAGS = "--cfg miri_strict_provenance $($env:RUSTFLAGS)".Trim()`), while a step-level `env:` map replaces. So "mirrors the recipe" was only approximately true.

Nothing is being dropped today, though: grepping all of `.github/` for `RUSTFLAGS` returns exactly two hits, and both are this PR's own lines — there is no workflow-level, job-level or `$GITHUB_ENV` assignment to clobber. What makes it worth closing rather than waving through is that the sibling `miri-race-coverage` job at `nightly.yml:160-171` already writes `MIRIFLAGS` through `$GITHUB_ENV`, so the same pattern is live in this very file and the trap is one edit away.

Both lines now read `<flag> ${{ env.<VAR> }}`. When the inherited value is empty this expands to a trailing space, which is harmless — cargo whitespace-splits these and discards empty entries.

One thing appending does *not* recover, in either form: `.cargo/config.toml`'s `-C target-cpu=x86-64-v3` is a config value, not an environment variable, so setting `RUSTFLAGS` at all still overrides it. The `just` recipe has the same behaviour, and it is irrelevant under Miri, so the two stay consistent.

YAML re-validated after the change.

martin-kolinek (approved):

Vaiz (approved):

PR #671, opened 2026-08-14
feat: introduce observed, observed_macros and observed_testing crates

Vaiz:
Move the `observed` telemetry crate family from the internal ox-sdk monorepo into this repository, so the crates live alongside the rest of the open-source Oxidizer surface and can be published to crates.io. This is the publication step for the telemetry-enrichment work; until now every change had to be transferred between the two repositories by hand.

## The crates

- **`observed`** (0.24.0) — structured, typed telemetry events. Each event is a struct with a data classification on every field. Enrichment entries are scoped, stackable and propagate across threads and async tasks. One event struct can produce a log and a metric with independent field subsets per signal.
- **`observed_macros`** (0.24.0) — the `#[event(...)]` attribute and `#[derive(Enrichment)]`. Re-exported through `observed`; do not depend on it directly.
- **`observed_testing`** (`publish = false`) — the `MockProcessor` harness and the integration-test suite for both crates.

## Changes made during the move

- `observed` no longer depends on `opentelemetry`. `Value` is an owned enum, so an `opentelemetry` major bump is no longer a breaking change for a crate that only emits typed events. `opentelemetry` stays as a dev-dependency, because the examples and benchmarks still export through an OpenTelemetry sink.
- The `metric_number_of` helper moved into `observed_testing`, which is its only consumer in this repository.
- `observed_testing` drives futures by hand with `testing_aids::FutureTestExt` and `testing_aids::poll_once` instead of the ox-sdk-only helper crate.
- Tests and examples use `expect` with a message instead of `unwrap`, to satisfy this workspace's `clippy::unwrap_used` policy.
- The `#![doc(html_logo_url)]` and `#![cfg_attr(coverage_nightly, ...)]` attributes are enabled, matching the convention of the other crates here.
- `observed_testing` is registered in `automation::INTERNAL_CRATES`.
- `all_the_time` and `thread_local` are added to `[workspace.dependencies]`.

## Effects

- Telemetry changes no longer need a manual transfer between two repositories; they also no longer reach internal consumers until a release is published from here.
- Leaf crates that emit typed events but never export to OpenTelemetry stop compiling `opentelemetry` and its transitive tree; a crate that does export must now convert values itself.
- The public API is governed by this repository's gates — `unwrap_used`, external-type approval, coverage and mutation testing — which are stricter than the ones the crates were developed under.
- The `Value` variant set becomes public vocabulary rather than an implementation detail. `#[non_exhaustive]` keeps adding a variant source-compatible for matchers, but an exporter outside this repository needs a catch-all arm.

## Validation

`just check`, `just anvil-clippy`, `just anvil-fmt`, `just test` (6628 passed), `just test-docs`, `just readme-check`, `just anvil-license-headers`, `just anvil-external-types`, `just anvil-doc-build` and `cargo machete` all pass locally. `cargo spellcheck` cannot be installed on this machine and mutation testing has not been run locally; both are left to CI.


crates/observed_macros/Cargo.toml
```diff
@@ -0,0 +1,48 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+[package]
+name = "observed_macros"
```
martintmk, line 5: in this repo, I have seen the pattern of also having observed_macros_impl crate where the real implementation lives. 

Just curious, why is it not used here?

crates/observed_macros/src/event.rs
```diff
@@ -0,0 +1,2463 @@
...
+
+    Ok(event_metric)
+}
+
+/// Enforces requirement that `kind = counter` value fields are unsigned integers
+/// and `kind = updown_counter` value fields are signed integers. `gauge`/`histogram`
+/// place no signedness constraint on the value type. `Option<T>` never satisfies
+/// any of these, but it is rejected earlier with a dedicated diagnostic.
+fn enforce_value_type(kind: InstrumentKindValue, field: &FieldDef, attr: &syn::Attribute) -> Result<()> {
+    let required = match kind {
```
martintmk: [AI Agent]

The accepted counter widths do not match the generated `Value` conversion. `enforce_value_type` advertises `u8..u128`/`usize` and `i8..i128`/`isize`, but `Value` only converts `u32`, `i32`, and `i64`; a normal `u64` byte counter therefore fails inside the expansion.

            Verified with:

            ```rust
            use observed::{SinkId, emit, event};
            use observed_testing::test_emitter;

            const TEST_ID: SinkId = SinkId::new("counter-width");

            #[test]
            fn counter_records_u64_field_value() {
                #[event("bytes.written")]
                #[info]
                #[counter(bytes)]
                struct BytesWritten {
                    #[unredacted]
                    bytes: u64,
                }

                let (sink, processor) = test_emitter(TEST_ID);
                emit!(sink, BytesWritten { bytes: 7 });

                let captured = processor.single_event();
                let field_metrics = captured.field_metrics();
                assert_eq!(field_metrics.len(), 1);
                assert!((field_metrics[0].value() - 7.0).abs() < f64::EPSILON);
            }
            ```

            `cargo test -p observed_testing --test zz_review_probe` fails with `error[E0277]: the trait bound Value: From<u64> is not satisfied` at `#[event("bytes.written")]`. Narrow the accepted/documented widths to the representation that `Value` can carry, or add explicit conversions with a documented loss policy. Please add this test permanently to `crates/observed_testing/tests/metric_routing.rs`.

crates/observed_macros/src/event.rs
```diff
@@ -0,0 +1,2463 @@
...
+}
+
+fn generate_visit_fields_body(fields: &[FieldDef], has_log: bool) -> TokenStream {
+    let visits: Vec<TokenStream> = fields.iter().map(|f| generate_field_visit(f, has_log)).collect();
+    quote! { #(#visits)* }
+}
+
+fn generate_field_visit(field: &FieldDef, has_log: bool) -> TokenStream {
+    let field_ident = &field.ident;
+    let default_key = field.ident.to_string();
```
martintmk: [AI Agent]

            Raw Rust identifiers leak their `r#` escape into telemetry keys because `Ident::to_string()` preserves it. Thus `r#type` is exported as `"r#type"` rather than the domain key `"type"` in both events and enrichments.

            Verified with:

            ```rust
            use observed::{Enrichment as _, SinkId, Value, emit, event};
            use observed_testing::test_emitter;

            const TEST_ID: SinkId = SinkId::new("raw-ident");

            #[test]
            fn raw_identifier_field_logs_under_its_plain_name() {
                #[event("http.request")]
                #[info]
                struct HttpRequest {
                    #[unredacted]
                    r#type: i64,
                }

                let (sink, processor) = test_emitter(TEST_ID);
                emit!(sink, HttpRequest { r#type: 1 });
                assert_eq!(
                    processor.single_event().dimensions(),
                    vec![("type".to_owned(), Value::from(1_i64))]
                );
            }

            #[test]
            fn raw_identifier_enrichment_field_uses_its_plain_name() {
                #[derive(observed::Enrichment)]
                struct Ctx {
                    #[unredacted]
                    r#type: i64,
                }

                let entries = Ctx { r#type: 1 }.into_entries();
                assert_eq!(entries[0].key().to_string(), "type");
            }
            ```

            `cargo test -p observed_testing --test zz_review_probe3` fails both tests with observed values `"r#type"` and `[("r#type", I64(1))]`. Use `syn::ext::IdentExt::unraw()` at every default-key derivation site while retaining the original identifier for field access. Please add these tests permanently to `derive_event.rs` and `derive_enrichment.rs`.

crates/observed_macros/src/event.rs
```diff
@@ -0,0 +1,2463 @@
...
+                        #metric_expr,
+                        #has_field_metrics,
+                        #disabled,
+                    );
+
+                fn visit_fields(
+                    &self,
+                    visitor: &mut ::observed::processing::FieldVisitorFn<'_>,
+                ) -> ::core::ops::ControlFlow<()> {
+                    use ::observed::Value;
```
martintmk: [AI Agent]

The imports injected into `visit_fields` are in scope for the consumer's `#[data_class(...)]` expression. A consumer type named `Value` is silently shadowed by `observed::Value`, so otherwise valid macro input fails to compile; `enrichment.rs` already avoids this by fully qualifying generated paths.

            Verified with:

            ```rust
            use observed::event;
            use observed_testing::MicrosoftEnterpriseDataTaxonomy;

            struct Value;

            impl Value {
                const CLASS: MicrosoftEnterpriseDataTaxonomy =
                    MicrosoftEnterpriseDataTaxonomy::Eupi;
            }

            #[event("hygiene.probe")]
            #[info]
            struct HygieneProbe {
                #[data_class(Value::CLASS)]
                user: String,
            }

            #[test]
            fn consumer_type_named_value_is_not_shadowed() {}
            ```

            `cargo test -p observed_testing --test zz_review_probe5` fails with `error[E0599]: no variant, associated function, or constant named CLASS found for enum observed::Value`. Deleting these imports and fully qualifying the generated paths makes this probe and the full suite pass. Please add this test permanently to `crates/observed_testing/tests/derive_event.rs`.

crates/observed_macros/src/event.rs
```diff
@@ -0,0 +1,2463 @@
...
+
+/// Enforces requirement that `kind = counter` value fields are unsigned integers
+/// and `kind = updown_counter` value fields are signed integers. `gauge`/`histogram`
+/// place no signedness constraint on the value type. `Option<T>` never satisfies
+/// any of these, but it is rejected earlier with a dedicated diagnostic.
+fn enforce_value_type(kind: InstrumentKindValue, field: &FieldDef, attr: &syn::Attribute) -> Result<()> {
+    let required = match kind {
+        InstrumentKindValue::Counter => IntegerSignedness::Unsigned,
+        InstrumentKindValue::UpDownCounter => IntegerSignedness::Signed,
+        InstrumentKindValue::Gauge | InstrumentKindValue::Histogram => return Ok(()),
```
martintmk, line 674: `[AI Agent]`

            `Gauge | Histogram => return Ok(())` accepts non-numeric value fields, but the metric pipeline cannot turn such a `Value` into a measurement and silently records nothing. The width-agnostic intent is fair, but it still needs a numeric-type check.

            Verified with:

            ```rust
            use observed::{SinkId, emit, event};
            use observed_testing::{PublicString, test_emitter};

            const TEST_ID: SinkId = SinkId::new("numeric-gauge");

            #[test]
            fn gauge_on_non_numeric_field_records_a_measurement() {
                #[event("pool.state")]
                #[info]
                #[gauge(label, name = "pool.state.gauge")]
                struct PoolState {
                    label: PublicString,
                }

                let (sink, processor) = test_emitter(TEST_ID);
                emit!(sink, PoolState { label: PublicString("idle".into()) });
                assert_eq!(
                    processor.single_event().field_metrics().len(),
                    1,
                    "gauge produced no measurement"
                );
            }
            ```

            `cargo test -p observed_testing --test zz_review_probe2` fails with `left: 0, right: 1` (`gauge produced no measurement`). Require a recognized numeric primitive for gauge/histogram fields (including `f32`/`f64`) and issue a macro diagnostic otherwise. Please add this test permanently to `crates/observed_testing/tests/metric_routing.rs`.

crates/observed/src/sink/core.rs
```diff
@@ -0,0 +1,653 @@
...
+    /// Dispatches an event through the sink.
+    ///
+    /// It's automatically called by the `emit!` macro expansion, and can be called directly for
+    /// [`DynEvent`s](DynEvent)
+    pub(crate) fn emit_impl<'a, T: Event, F: FnOnce() -> T + 'a>(&self, state: IntermediateEvent<'a, F>) {
+        if self.is_noop() {
+            return;
+        }
+
+        let Some(_guard) = super::try_acquire_reentrancy_guard() else {
```
martintmk: [AI Agent]

            The reentrancy guard is acquired before `state.evaluate()`, so telemetry emitted by an ordinary helper used to initialize an event field is silently dropped, even when it targets an unrelated sink. The documented reentrancy hazard begins when processors run, not while the event value is being built.

            Verified with:

            ```rust
            use std::sync::{Arc, Mutex};
            use observed::metadata::EventDescription;
            use observed::processing::{EventProcessor, EventView};
            use observed::{FlushError, Sink, emit, event};
            use tick::SimpleClock;

            #[event("probe.event")]
            #[info]
            struct ProbeEvent {
                #[unredacted]
                n: i64,
            }

            #[derive(Default)]
            struct Recorder(Mutex<usize>);

            impl EventProcessor for Recorder {
                fn is_interested(&self, _: &EventDescription) -> bool { true }
                fn process(&self, _: &EventView<'_>) { *self.0.lock().unwrap() += 1; }
                fn flush(&self) -> Result<(), FlushError> { Ok(()) }
            }

            fn recording(id: &'static str, clock: &SimpleClock) -> (Sink, Arc<Recorder>) {
                let recorder = Arc::new(Recorder::default());
                let processor = Arc::clone(&recorder) as Arc<dyn EventProcessor>;
                (Sink::new(id, vec![processor], clock), recorder)
            }

            fn compute_n(helper: &Sink) -> i64 {
                emit!(helper, ProbeEvent { n: 0 });
                7
            }

            #[test]
            fn helper_telemetry_during_field_initialization_is_not_dropped() {
                let clock = SimpleClock::new_frozen();
                let (outer, outer_recorder) = recording("outer", &clock);
                let (helper, helper_recorder) = recording("helper", &clock);

                emit!(outer, ProbeEvent { n: compute_n(&helper) });

                assert_eq!(*outer_recorder.0.lock().unwrap(), 1);
                assert_eq!(*helper_recorder.0.lock().unwrap(), 1);
            }
            ```

            `cargo test -p observed --features test-util --test zz_findings` fails the second assertion with `left: 0, right: 1`. Move guard acquisition to after `state.evaluate()`, immediately before processor dispatch. Please add this test permanently to the sink integration tests.

crates/observed/src/context/transfer.rs
```diff
@@ -0,0 +1,106 @@
...
+    ///
+    /// Because this mutates a thread-local, the guard **must not be held across
+    /// an `.await`**: while the task is suspended the enrichment would remain
+    /// active and leak into unrelated tasks the runtime schedules on the same
+    /// thread. For async code use
+    /// [`EnrichFutureExt::attach`](crate::enrichment::EnrichFutureExt::attach),
+    /// which re-applies the transfer on each poll and drops the guard before the
+    /// future yields.
+    #[must_use]
+    pub fn apply_current_thread(&self) -> impl Drop {
```
martintmk: [AI Agent]

            The guard returned by `apply_current_thread` is structurally `Send`, although dropping it restores thread-local state by blindly replacing the current thread's slot. Moving it to another thread wipes that thread's live enrichment and leaves the origin thread's transferred enrichment stranded; the warning about `.await` does not prevent this safe-code misuse.

            Verified with:

            ```rust
            use observed::enrichment::EnrichFnExt;
            use observed::{Enrichment, Sink};

            #[derive(Enrichment)]
            struct Ctx {
                #[unredacted]
                value: i64,
            }

            #[test]
            fn transfer_guard_cannot_corrupt_a_foreign_thread() {
                let sink = Sink::noop();
                let transfer =
                    (|| sink.transfer_context()).enrich(&sink, Ctx { value: 7 })();

                let guard = transfer.apply_current_thread();
                assert_eq!(sink.current_enrichments().len(), 1);

                let worker_sink = sink.clone();
                std::thread::scope(|scope| {
                    scope.spawn(move || {
                        (|| {
                            assert_eq!(worker_sink.current_enrichments().len(), 1);
                            drop(guard);
                            assert_eq!(worker_sink.current_enrichments().len(), 1);
                        })
                        .enrich(&worker_sink, Ctx { value: 9 })();
                    })
                    .join()
                    .unwrap();
                });

                assert_eq!(sink.current_enrichments().len(), 0);
            }
            ```

            `cargo test -p observed --features test-util --test zz_findings` fails in the worker with `left: 0, right: 1` (`worker's own enrichment scope was wiped by a foreign guard`). Make this guard `!Send` so the thread-local lifetime requirement is enforced by the type system. Please add this test permanently to the context-transfer suite.

crates/observed/src/value.rs
```diff
@@ -0,0 +1,396 @@
...
+            // A `RedactedDisplay` impl may emit an event of its own, re-entering
+            // this function while the buffer is already borrowed. That is too
+            // rare to optimize for, but it must not panic.
+            let Ok(mut buffer) = cell.try_borrow_mut() else {
+                return Self::String(Text::from(rendered.to_string()));
+            };
+
+            buffer.clear();
+            // Writing to a `String` cannot fail and redaction is infallible.
+            _ = write!(&mut *buffer, "{rendered}");
```
martintmk: [AI Agent]

            A `RedactedDisplay` implementation can return `fmt::Error` after writing a prefix, but this discards that failure and exports the partial text as if redaction completed. A failed redaction should erase the value rather than potentially widening what reaches telemetry.

            Verified with:

            ```rust
            use std::fmt;
            use data_privacy::simple_redactor::{SimpleRedactor, SimpleRedactorMode};
            use data_privacy::{RedactedDisplay, RedactionEngine, Redactor};
            use observed::Value;

            struct FailsMidway;

            impl RedactedDisplay for FailsMidway {
                fn fmt(
                    &self,
                    _: &dyn Redactor,
                    f: &mut fmt::Formatter<'_>,
                ) -> fmt::Result {
                    f.write_str("prefix-")?;
                    Err(fmt::Error)
                }
            }

            #[test]
            fn failed_redaction_does_not_export_partial_output() {
                let engine = RedactionEngine::builder()
                    .set_fallback_redactor(SimpleRedactor::with_mode(
                        SimpleRedactorMode::Passthrough,
                    ))
                    .build();

                assert_ne!(
                    Value::from_redacted(&FailsMidway, &engine),
                    Value::from("prefix-")
                );
            }
            ```

            `cargo test -p observed --features test-util --test zz_findings` fails with both sides equal to the string `"prefix-"`. On `write!` failure, clear the scratch buffer and return the erased representation. Please add this test permanently to `crates/observed/src/value.rs`.

martintmk (changes requested):
[AI Agent]

_Automated review by an AI agent (GitHub Copilot CLI), requested by @martintmk._

I validated the three added crates locally: 172 targeted unit/doc tests passed (1 ignored), and the minimal-feature `observed` plus `observed_macros` builds pass. I found seven reproducible defects: four in the proc-macro contract and three in runtime behavior. **Verdict: changes requested.**

martintmk (approved):

martintmk:
Approved, feel free to ignore my agent.

crates/observed_macros/Cargo.toml
Vaiz: I don't like splitting macros crates into two just for tests. It's not necessary.

martintmk: If it's not necessary, let's not do it then. I thought there was strong reason to do so. This deserves some consistency across our crates.

cc @ralfbiedert

crates/observed_macros/src/event.rs
```diff
@@ -0,0 +1,2486 @@
...
+                    "`#[{}({field_ident})]` requires field `{field_ident}` to hold a value on \
+                     every emission, but it is an `Option<T>`; a metric value cannot be optional. \
+                     Use a non-optional field, or record it as a metric dimension instead \
+                     (`#[dimension(metric = ...)]`)",
+                    kind.attr_name(),
+                ),
+            ));
+        }
+
+        enforce_value_type(kind, field, &attr)?;
```
martintmk, line 603: [AI Agent]

**Metric fields with classification silently produce no measurement.** This validation accepts numeric metric-value fields regardless of redaction, but `value_expr` later routes default and `#[data_class(...)]` values through `Value::from_redacted`, which always yields a string. Metric processors can only extract numeric `Value`s, so a valid-looking counter/gauge/histogram/up-down-counter declaration is silently dropped at runtime. Please either reject metric-value fields unless they are `#[unredacted]`, or add a privacy-safe conversion that preserves a numeric measurement; add a routing test covering a classified numeric metric field.

martintmk (changes requested):
[AI Agent]

_Automated review by an AI agent (GitHub Copilot CLI), requested by @martintmk._

I found one additional correctness issue in the metric-field contract. **Verdict: changes requested.**

martintmk (approved):

crates/observed/examples/three_enrichment_styles.rs
```diff
@@ -0,0 +1,144 @@
...
+struct HttpRequest {
+    #[data_class(EXAMPLE_DC)]
+    status: i64,
+}
+
+async fn handle_request(sink: &Sink, route: HttpRoute) {
+    async {
+        tokio::task::yield_now().await;
+        emit!(sink, HttpRequest { status: 1 });
+        (|| {
```
martintmk, line 87: nit: I cannot wrap my head around the `(|| { ... }).enrich(...)` syntax.

WDYT about this alternative:

```rust
let enriched = enriched(sink, MethodContext { method: HttpMethod("GET") );

enriched.invoke(|| {
   // some code
   emit!(sink, HttpRequest { status: 2 });
})
```

Async syntax is already pretty neat, but the tuple like syntax  for lambda execution doesn't sit well with me.

(non blocking)

Vaiz: definitely not in this PR

crates/observed_macros/src/event.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Confirmed and fixed in c89a02c3.

`Ident::to_string()` preserves the `r#` escape, so the default key derivation leaked it into telemetry. All three default-key sites now go through `syn::ext::IdentExt::unraw()` — `FieldDef::log_key`, `generate_field_visit`'s `default_key` (which feeds the field name, the log key and the metric dimension key), and `generate_entry_stmt` in `enrichment.rs`. The identifier used to *address* the field keeps its raw form.

Tests added, both of which I ran against the unfixed code and saw fail:

- `derive_event.rs::raw_identifier_field_uses_its_plain_name`
- `derive_enrichment.rs::raw_identifier_entry_uses_its_plain_name` — failed with `left: Key("r#type")`, `right: "type"`.

crates/observed_macros/src/event.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Confirmed and fixed in c89a02c3.

The two `use` statements are gone and every path generated into `visit_fields` is now fully qualified: `::observed::Value::{from, from_redacted}`, `::observed::metadata::{FieldDescriptor, LogFieldEntry, MetricFieldEntry}` — including the `FIELD_DESC` type annotation and the `Option` placeholder arm, which had the same exposure. `enrichment.rs` already did this, so the two macros are now consistent.

Test added: `derive_event.rs::consumer_type_named_value_is_not_shadowed`. Run against the unfixed code it fails with exactly the error you quoted:

```
error[E0599]: no variant or associated item named `class` found for enum `observed::Value`
   --> crates\observed_testing\tests\derive_event.rs:993:29
```

The 30 expansion snapshots in `observed_macros` are regenerated for the qualified paths.

crates/observed/src/sink/core.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Agreed, and fixed in c89a02c3. Your reading of the contract is the right one: the hazard is a *processor* re-entering the pipeline, not the construction of the event value, which is ordinary user code.

`emit_impl` now orders the steps: `is_noop` → `description` → `is_interested_in` → `state.evaluate()` → **acquire guard** → `dispatch_to_processors`. Recursion is still bounded, because a processor that emits reaches the acquisition point and is turned away there; it just costs one extra `evaluate` on the dropped path.

I also corrected the docs that described the old scope, since they now state something different:

- `recursion_guard.rs` module docs — "emitted *during* an emission" → "emitted *while processors are running*", plus an explicit note that construction is outside the guard.
- `dispatch_to_processors` — the stale `Sink::emit` reference.
- `EventProcessor::process`'s "nested telemetry is not supported" note still holds as written, since `process()` runs under the guard.

Test added: `basic_emission.rs::helper_telemetry_during_field_initialization_is_not_dropped`, a two-sink version of your reproducer. Against the unfixed code it fails with `left: 0, right: 1` on the helper sink.

crates/observed/src/context/transfer.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Agreed, and fixed in c89a02c3. The `.await` warning was documentation where the type system could carry the constraint.

`apply_current_thread` now returns a private `ThreadBoundGuard` holding the enrichment `Guard` plus a `PhantomData<Rc<()>>`, so the returned `impl Drop` is `!Send` (and `!Sync`). Moving it to another thread is now a compile error rather than silent corruption.

Two things I checked before making it `!Send`, since narrowing an opaque return type can break callers:

- `Transferred::poll` (`context/transferred.rs`) creates and drops the guard entirely *within* `poll`, so it never enters the generator state and attached futures stay `Send`.
- `EnrichFutureExt::attach` uses `EnrichmentTransfer::apply` directly, not this method, so the async path is untouched.

No runtime test for this one — the property is the absence of a `Send` impl, so the reproducer you wrote no longer compiles, which is the intended outcome. If you want it pinned against regression I can add a `trybuild` compile-fail case; the workspace already uses `trybuild` in `internity`, `fundle`, `ohno` and `templated_uri`. Say the word and I'll add it.

crates/observed/src/value.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Confirmed and fixed in c89a02c3. The comment claiming "redaction is infallible" was simply wrong — it is only the *write to a `String`* that cannot fail, and a `RedactedDisplay` impl can return `Err` after emitting a prefix.

`from_redacted` now clears the scratch buffer on a write error, so a failed redaction yields the erased value rather than an unapproved partial string.

While in there I fixed the same class of bug on the re-entrant path a few lines up: it used `rendered.to_string()`, and `ToString` **panics** when the `Display` impl returns an error. It now writes into a local `String` and applies the same erase-on-error rule, so a failing redaction inside a nested `from_redacted` no longer takes the process down.

Test added: `value.rs::from_redacted_erases_a_value_whose_redaction_fails`. Against the unfixed code it fails with `left: String(Shared("prefix-"))`, `right: String(Static(""))` — i.e. the partial text really was reaching telemetry.

crates/observed_macros/Cargo.toml
ralfbiedert: This only "works" because you now cram everything into the `src` folder; like tests and snapshots, which is ugly. If you ever wanted to run integration tests, macro benchmarks, or anything else that needs to use "macros as library" to work this here will give you a headache. Let's standardize on `macros_impl` please, because it leads to a cleaner separation of concerns (and code).

crates/observed_macros/src/event.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Confirmed, and I've verified it against `Value`'s actual `From` impls. This finding and your two others on the metric-value contract (#discussion_r3782414471 on gauge/histogram, #discussion_r3783511368 on classified metric fields) are one underlying problem, so I'm answering all three here and linking from those.

**What `Value` can actually carry.** `crates/observed/src/value.rs` has `From` for exactly `bool`, `i32`, `u32`, `i64`, `f32`, `f64`, `String`, `&'static str` and `Arc<str>`. There is no `From<u64>`, `u8`, `u16`, `u128`, `usize`, `i8`, `i16`, `i128` or `isize`.

So `enforce_value_type` currently accepts three shapes that cannot work:

1. **Widths `Value` cannot carry** — your case. `#[counter(bytes)] bytes: u64` passes macro validation and then fails deep inside the expansion with `Value: From<u64> is not satisfied`. The diagnostic advertising `u8, u16, u32, u64, u128, usize` is actively misleading: of that list only `u32` works.
2. **Non-numeric gauge/histogram values** — `Gauge | Histogram => return Ok(())` performs no type check at all, so a `PublicString` gauge compiles and records nothing.
3. **Classified metric values** — `value_expr` routes anything that is not `#[unredacted]` through `Value::from_redacted`, which always yields `Value::String`. A metric processor cannot extract a measurement from a string, so it is dropped silently.

Case 3 is not hypothetical in this repo. `crates/observed/tests/otel_log_record.rs:55` is a *metric-only* event whose gauge value is classified:

```rust
#[event("system.memory.usage")]
#[gauge(bytes, name = "system.memory.usage")]
struct MemoryUsage {
    #[data_class(TEST_DC)]
    bytes: i64,
}
```

That event exists to emit a metric and emits none.

**Where I'd like your call.** The diagnosis is settled; the remedy forks, and one branch touches `Value`'s public surface, which was reviewed separately as a deliberate breaking change. I don't want to make that call unilaterally inside a PR whose scope is "move the crates in".

- **(A) Macro-only.** Require metric value fields to be `#[unredacted]` *and* one of the five numeric types `Value` carries; `counter` → `u32`, `updown_counter` → `i32`/`i64`, `gauge`/`histogram` → any of the five. All three failures become macro diagnostics. Cost: `u64` counters stay impossible, which is the natural type for a byte or request counter, and existing call sites in this repo have to change — there are 70 metric attributes across 18 files.
- **(B) Widen `Value`, then tighten the macro.** Add `From` for the remaining integer widths with a documented conversion/saturation policy, then apply (A)'s numeric + `#[unredacted]` rule. Keeps `u64` counters working and makes the current diagnostic honest. Cost: new public API on `Value` and a loss policy to agree.

I lean **(B)**, because (A) makes the macro correct by making a common, legitimate counter shape unexpressible. Either way the `#[unredacted]` requirement in case 3 applies unchanged.

Tell me which you want and I'll implement it with routing tests for all three cases in `metric_routing.rs`. I've flagged the same question to @Vaiz, since it reaches the `Value` design.

The other five findings in this review are fixed and pushed in c89a02c3.

crates/observed_macros/src/event.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Confirmed — `Gauge | Histogram => return Ok(())` performs no type check at all, so a non-numeric value field compiles and records nothing.

This is one of three symptoms of the same metric-value contract gap, so I've answered all three together on the counter-width thread rather than splitting the analysis: **#discussion_r3784031101**. It sets out what `Value` can actually carry, and the two remedies — macro-only tightening vs. widening `Value` first — for you to pick between.

crates/observed_macros/src/event.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Confirmed, and this is the sharpest of the three. `value_expr` routes anything not `#[unredacted]` through `Value::from_redacted`, which always returns `Value::String`, so a classified numeric metric field is silently dropped downstream.

It already bites in this repo: `crates/observed/tests/otel_log_record.rs:55` defines `MemoryUsage`, a *metric-only* event whose `#[gauge(bytes)]` value carries `#[data_class(TEST_DC)]` — an event that exists to emit a metric and emits none.

I've answered this together with the counter-width and gauge/histogram findings on **#discussion_r3784031101**, since all three are the same contract and fixing one without the others just moves the hole. Your "reject unless `#[unredacted]`" option is part of both remedies I laid out there; the open question is whether we also widen `Value`'s integer `From` impls first, which is why I'd like a decision before implementing.

crates/observed_macros/Cargo.toml
Vaiz: not in this PR, it's a substantial change

martintmk: I think we need to document the benefits of the split. Even for me, I don't see anything wrong with having tests and snapshots under `src`:

image

I use something similar pretty regularly. Never seen benchmarks nor samples for macros either.

Vaiz: I've started a topic in teams

crates/observed/examples/three_enrichment_styles.rs
Vaiz: tbh, I haven't seen much usage for this in sync code. I'll think if we can adopt this syntax

crates/observed_macros/src/event.rs
Vaiz: unredacted values should preserve their original type

Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Thanks — I want to make sure I implement exactly that, because the phrase decides a couple of things and I would rather ask once than guess.

Reading it back into this thread, I take "unredacted values should preserve their original type" to mean:

1. A metric **value** field must be `#[unredacted]`. Anything classified goes through `Value::from_redacted`, which always produces a `String`, and no metric processor can take a measurement from a string — so a classified metric value can never work and should be a macro error, not a silent drop.
2. An unredacted numeric must reach `Value` as a **numeric**, not be widened into something lossy or stringified.

If that is right, point 2 is where the remaining work is, because `Value` cannot currently carry most integer widths at all. The existing precedent is lossless widening: `i32` and `u32` both become `Value::I64` via `i64::from`.

- `i8`, `i16`, `u8`, `u16` — same lossless `i64::from` widening. I'll just add these; no decision needed.
- `u64`, `usize`, `u128`, `i128` — there is no infallible conversion into `i64`, so these need a call:
  - **(i)** add a `U64` variant (and use it for `usize`). Truest to "preserve their original type", and `Value` is `#[non_exhaustive]` precisely so a variant can be added source-compatibly — but every exporter gains an arm.
  - **(ii)** saturate into `I64`. No new variant, silently wrong above `i64::MAX`.
  - **(iii)** leave them unsupported, but as a clear macro diagnostic instead of the current `Value: From<u64> is not satisfied` from inside the expansion.

My suggestion: **(i) for `u64` and `usize`** — a byte or request counter is the motivating case and it should not need a cast — and **(iii) for `u128`/`i128`**, which no instrument realistically needs.

Confirm or correct that and I'll implement it in one change: the `#[unredacted]` requirement, a numeric-type check for `gauge`/`histogram` (which today check nothing), the widened `From` impls, and routing tests in `metric_routing.rs` covering all three of @martintmk's cases. It will also touch `crates/observed/tests/otel_log_record.rs`, whose `MemoryUsage` gauge is classified today and therefore emits no measurement.

crates/observed_macros/src/event.rs
Vaiz: Decision: **(B)** — widen `Value`, then tighten the macro.

- `u64` must be a valid `Value`. It's the natural type for byte and request counters, and making it unexpressible would be the wrong trade. We'll add the missing integer `From` impls with a documented conversion policy.
- `u128` and `i128` are **rejected as unsupported**. They don't map onto any telemetry backend's numeric representation, so the macro should emit a diagnostic rather than silently truncating.

With that settled, the macro-side rule from (A) applies as well: metric value fields must be `#[unredacted]` and one of the supported numeric types, so the classified-metric case becomes a compile-time diagnostic instead of a silent drop. Routing tests for all three cases go into `metric_routing.rs`.

crates/observed_macros/src/event.rs
Vaiz: Agreed — instruments should reject any unsupported type at **compile time**, not accept it and silently record nothing.

`Gauge | Histogram => return Ok(())` gets replaced with a real check: the value field must be `#[unredacted]` and one of the supported numeric types (including `f32`/`f64` for gauge/histogram). Anything else — non-numeric, classified, or an unsupported width such as `u128`/`i128` — becomes a macro diagnostic pointing at the field.

Type support itself is settled on #discussion_r3784613672: `u64` will be a valid `Value`, `u128`/`i128` are rejected as unsupported. A routing test for the non-numeric gauge case goes into `metric_routing.rs`.

crates/observed_macros/src/event.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Implemented in 0a1f36e9.

**`Value` widened.** New `U64` variant, plus `From` for `i8`, `i16`, `u8`, `u16`, `u64`, `usize` and `isize`. `u64` gets its own variant rather than folding into `I64` — more than half its range does not fit there, and byte and request counters live exactly in that half, so widening would have wrapped `u64::MAX` to `-1`. `Value` is `#[non_exhaustive]`, so the new variant stays source-compatible for matchers outside the crate. The conversion table is documented on `Value` itself:

| Source | Variant | Conversion |
|--------|---------|------------|
| `i8`, `i16`, `i32`, `i64` | `I64` | widening, lossless |
| `u8`, `u16`, `u32` | `I64` | widening, lossless |
| `u64` | `U64` | exact |
| `isize` / `usize` | `I64` / `U64` | exact on every supported target |
| `f32`, `f64` | `F64` | widening, lossless |

`u128`/`i128` have no conversion, as you decided.

**Macro tightened.** `enforce_value_type` now requires the value field to be `#[unredacted]` and a supported numeric primitive, with a dedicated diagnostic for the 128-bit widths rather than the generic one. `counter` still requires unsigned, `updown_counter` signed.

**Two call sites in this repo were silently broken** and are fixed rather than exempted:

- `crates/observed/tests/otel_log_record.rs` — `MemoryUsage`, the metric-only event whose classified gauge emitted no measurement.
- `crates/observed/examples/optional_fields.rs` — its `#[histogram(status)]` targeted a classified newtype over `u16`, so it recorded nothing. The histogram now targets an unredacted `f64` duration, which is what you would actually histogram.

I also added a `U64` arm to `observed_testing::metric_number_of` — without it a `u64` counter would have passed the new macro check and still recorded nothing — and mapped `U64` explicitly in the OTel example instead of letting it fall through to the debug-string arm.

Tests: `metric_routing.rs` gains `u64_counter_records_its_value`, `u64_counter_preserves_values_above_i64_max` (asserts the measurement stays above `1.8e19`, which an `I64` fold would have turned negative), `gauge_records_each_supported_width` and `gauge_records_a_float_value`. The rejection cases are compile-time, so they live in the macro's own tests, and `value.rs` covers the conversions directly.

Docs updated to match: the `#[event(...)]` attribute reference, `DESIGN.md`, `FEATURES.md`, `CHANGELOG.md`. A stale `TODO` in `metric_routing.rs` that promised classified metric values "once the codegen supports" them is deleted — that is now expressly rejected.

crates/observed_macros/src/event.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Implemented in 0a1f36e9. `Gauge | Histogram => return Ok(())` is gone; both now go through the same check as the counters.

The value field must be `#[unredacted]` and one of `i8`, `i16`, `i32`, `i64`, `isize`, `u8`, `u16`, `u32`, `u64`, `usize`, `f32`, `f64`. Gauge and histogram stay signedness- and width-agnostic within that set, floats included; only `counter` (unsigned) and `updown_counter` (signed) narrow it further. Non-numeric, classified, and the 128-bit widths each get their own diagnostic pointing at the attribute.

The example that proved the gap is fixed rather than exempted: `crates/observed/examples/optional_fields.rs` had `#[histogram(status)]` on a `#[classified]` newtype over `u16`, so it recorded nothing on every emission. It now histograms an unredacted `f64` duration.

Routing tests in `metric_routing.rs` cover the accepted shapes end-to-end (`gauge_records_each_supported_width`, `gauge_records_a_float_value`); the rejections are compile-time, so they are asserted in the macro's own tests — `non_numeric_metric_value_field_is_rejected` checks both `gauge` and `histogram`.

Full detail on #discussion_r3784850824.

crates/observed_macros/src/event.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Implemented in 0a1f36e9. A classified metric value field is now a compile-time error naming the fix:

```
`#[gauge(bytes)]` requires field `bytes` to be `#[unredacted]`; a classified value is
rendered through the redaction engine as a string, which carries no measurement for the
instrument to record
```

That diagnostic fired on the two real cases in this repo the moment it was added — `otel_log_record.rs::MemoryUsage` and the `optional_fields` example — which is the evidence that this was not hypothetical. Both are fixed, not exempted.

Full detail, including the widened `Value` conversions, on #discussion_r3784850824.

crates/observed/src/value.rs
```diff
@@ -0,0 +1,599 @@
...
+    ///
+    /// This is the only way to turn a classified value into a `Value`, and it
+    /// is what the derive macros generate for a classified field.
+    ///
+    /// A [`RedactedDisplay`](data_privacy::RedactedDisplay) implementation that
+    /// fails part-way through has already written an unknown prefix of its
+    /// output. That prefix was never approved by the redactor, so exporting it
+    /// could widen what reaches telemetry; a failed redaction therefore yields
+    /// the erased (empty) value rather than the partial text.
+    pub fn from_redacted(value: &(impl data_privacy::RedactedDisplay + ?Sized), engine: &data_privacy::RedactionEngine) -> Self {
```
Vaiz: this should be `redactor: &dyn [Redactor](https://docs.rs/data_privacy_core/latest/data_privacy_core/trait.Redactor.html)`

also check if we can limit dependency to data_privacy_core

Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Agreed on both counts. Done in 2ba5fcd1.

**`&dyn Redactor`** — you're right that the parameter was over-constrained. `RedactedDisplay::fmt` and `RedactedToString::to_redacted_string` both take `&dyn Redactor`, so demanding a concrete `RedactionEngine` was narrower than the work actually being done. Changed on:

- `Value::from_redacted`
- `EnrichmentEntry::redacted_value`
- `FieldValueFn`, and therefore every field/enrichment getter closure

No caller had to change: `RedactionEngine` implements `Redactor`, so `&engine` still coerces at the call site, and the macro-generated getter closures are unannotated (`&|engine| …`) — so no `observed_macros` change and no snapshot regeneration. A processor can now redact with a bare strategy instead of being forced to build an engine, which `from_redacted_accepts_a_bare_redactor` pins by passing a plain `SimpleRedactor` — that did not compile before.

**Limiting the dependency to `data_privacy_core`** — partly, and the change above gets the valuable half of it. `observed`'s **public API** is now core-only:

```diff
-    "data_privacy::redaction_engine::RedactionEngine",
     "data_privacy_core::redacted::RedactedDisplay",
+    "data_privacy_core::redactor::Redactor",
```

`anvil-external-types` reports 0 errors with `RedactionEngine` removed from the allow list, which proves it is genuinely gone from the surface rather than merely unlisted.

It cannot go core-*only* at the crate level, because of exactly one thing: **`Sensitive<T>`**. `data_class = <expr>` generates `::observed::__private::Sensitive::new(value, expr)`, and `data_privacy_core` has no equivalent — it ships the `Classified` trait and `DataClass`, but not a wrapper type that pairs a value with a class and implements `RedactedDisplay`. This is the same position `templated_uri` is in: it lists seven `data_privacy_core::*` types plus `data_privacy::sensitive::Sensitive`, and depends on `data_privacy`. So `observed` now matches the convention already in the repo.

I also dropped the `__private::RedactionEngine` re-export, which nothing referenced.

If you want the dependency gone entirely, the options are (a) move `Sensitive` into `data_privacy_core`, or (b) reimplement it inside `observed::__private`. Both are `data_privacy` decisions rather than part of moving these crates, so I have not done either — happy to file a follow-up if you'd like (a).

Gates run before pushing: `just anvil-clippy`, `anvil-external-types`, `anvil-doc-build`, `readme-check`, both rustfmt configs, and the full test suite for all three crates.

martintmk (approved):
I played with the branch a bit and I think this is really nice. 

Two nits that we can address in follow-up:

- Support for histogram boundaries. This is something that is encoded in defaults for each histogram and we should preserve this. See [MetricsHandler](https://github.com/microsoft/oxidizer/blob/9877935cc807673ab88ffcf32f0842d70c0564c5/crates/fetch/src/handlers/metrics.rs#L267) for example.
- Have you thought about using `CompactStr` in `Text`? Could help with some small allocations. 
- I could see the following pattern being repeated when dimensions are shared for both logs and metrics:

`#[dimension(logs = "server.url.port", metrics = "server.url.port")]`

Maybe also allow this syntax?

`#[dimension("server.url.port", metrics = true)]` // logs included by default

crates/observed/examples/three_enrichment_styles.rs
Vaiz: filed a work item for this

evgenyfedorov2 (approved):

PR #646, opened 2026-08-06
fix(release): cascade exposed dependency breaks

kate-shine:
## Summary

Release planning could ship a **breaking API change inside a semver-compatible version**.

`cargo-semver-checks` analyses each crate in isolation. But a crate's public API is not only what it declares — it is also every foreign type it re-exports. When `bytesbuf` breaks, `bytesbuf_io` breaks too, because `bytesbuf_io`'s API *is* `bytesbuf` types:

```rust
// crates/bytesbuf_io/src/testing/fake_write.rs
pub fn contents(&self) -> &BytesBuf
pub fn reserve(&self, min_bytes: usize) -> BytesBuf
```

`bytesbuf_io` never changed, so semver-checks saw nothing, and the planner gave it the mechanical patch floor:

| | before | after |
|---|---|---|
| `bytesbuf` | 0.7.0 → **0.8.0** | 0.7.0 → **0.8.0** |
| `bytesbuf_io` | 0.7.0 → **0.7.1** ❌ | 0.7.0 → **0.8.0** ✅ |

That `0.7.1` is the bug: a patch release that breaks every downstream build, delivered automatically by `cargo update`.

This PR teaches the planner to combine semver-checks with the `cargo_check_external_types` allowlists each crate already declares, and to propagate incompatibility across public-API exposure edges to a fixpoint.

Work item: https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7705926

---

## What the cascade actually does

Running `release-packages bytesbuf@breaking` against today's workspace pulls in **9 crates, all breaking**. Solid arrows are exposure edges; dashed arrows are dependencies that do *not* surface the type in their public API:

```mermaid
graph LR
    bytesbuf["bytesbuf 0.7.0 → 0.8.0"]
    bytesbuf_io["bytesbuf_io 0.7.0 → 0.8.0"]
    cachet["cachet 0.10.0 → 0.11.0"]
    multitude["multitude 0.7.1 → 0.8.0"]
    http_extensions["http_extensions 0.8.0 → 0.9.0"]
    seatbelt_http["seatbelt_http 0.6.0 → 0.7.0"]
    fetch["fetch 0.14.0 → 0.15.0"]
    fetch_hyper["fetch_hyper 0.5.0 → 0.6.0"]
    fetch_azure["fetch_azure 0.4.0 → 0.5.0"]

    bytesbuf --> bytesbuf_io
    bytesbuf --> cachet
    bytesbuf --> multitude
    bytesbuf --> http_extensions
[...]

PR #672, opened 2026-08-14
fix(release): close two fail-open gaps in the exposure cascade

### LINKED ISSUE #665: release: two fail-open gaps in exposure analysis (follow-up to #646)
Two exposure-analysis gaps found by automated review on #646. Both are **fail-open**: they let a breaking dependency ship inside a semver-compatible version, which is exactly the class of bug #646 exists to prevent. Neither is reachable against today's workspace, so they did not block that PR.

Both were reproduced before filing; neither is theoretical.

## 1. A mixed direct/indirect topology is only ever tested on the direct edge

`Get-PublishedDependentsExposingTarget` (`scripts/lib/release-flow.ps1`) picks the exposure predicate with an `if`/`else` on whether a direct edge is declared:

```powershell
if ($_.Deps -contains $targetCargoName) {
    Test-PackageExposesTarget ...          # direct
} else {
    $reachable.Contains($_.Folder) -and
    (Test-PackageAllowlistNamesTarget ...) # indirect
}
```

The two are mutually exclusive, so a crate holding **both** kinds of path is judged solely on the direct one. That matters because the two predicates accept different roots by design: the direct predicate builds roots from `DepAliases` (edge-local, rename-aware) and deliberately rejects the target's global crate root, since a `package = "..."` rename shadows it on that edge. The indirect predicate builds roots from the target's own record.

So if `facade` depends on `defining` under a rename *and* also reaches `defining`'s types re-exported through `relay`, its allowlist legitimately carries `defining`'s `[lib]` root — earned on the indirect path — which the direct predicate correctly refuses and the `else` never lets the indirect predicate see.

Reproduced with `defining` = package `defining`, `[lib] name = "defining_core"`, renamed to `aliased_dep` on the direct edge:

| predicate | result |
|---|---|
| `Test-PackageExposesTarget` (the one actually run) | `False` |
| `Test-PackageAllowlistNamesTarget` (never reached) | `True` |
| accepted roots on the direct edge | `aliased_dep`, `defining` |

`facade` stays on its patch floor while its public API is `defining`'s types.

**Not reachable today:** no workspace crate declares a `[lib]` name differing from its package name, so both root sets collapse to the same value and the direct predicate matches anyway.

**Fix:** determine the transitive path independently of the direct edge and accept *either* qualifying relationship, rather than branching exclusively.

## 2. A scalar `allowed_external_types` fails open

`Get-WorkspacePackages` (`scripts/lib/releasing.ps1`) normalises the allowlist with:

```powershell
$allowedTypes = @($allowed.Value)
```

`allowed_external_types` must be an array. When a manifest gives a scalar — `allowed_external_types = "std::*"` — `@(...)` promotes it into a *valid-looking* one-entry allowlist. That entry then matches nothing and the crate reads as "not exposed".

This is the only malformed shape that fails open:

| `allowed_external_types` | `Test-PackageExposesTarget` | direction |
|---|---|---|
| scalar string `"std::*"` | `False` | **fails open** |
| non-string container | `True` | fails closed |
| absent (`$null`) | `True` | fails closed |

It contradicts the fail-closed policy established in d42c6cb1, which fixed the same class of bug for malformed *entries* while the malformed *container* slipped through.

**Not reachable today:** all 44 `allowed_external_types` declarations in the workspace are arrays.

**Fix:** leave `$allowedTypes` as `$null` unless the value is genuinely an array. The existing absent-metadata branch then fails closed on its own, with no new branch. Worth a unit test in `PureFunctions.Tests.ps1` alongside the existing malformed-entry cases.

## Why both are worth fixing despite being unreachable

Each becomes live the moment someone adds an ordinary manifest construct — a `[lib]` rename, or a typo'd allowlist. The failure is silent, produces a green release, and surfaces as broken downstream builds after `cargo update`. The cost of closing them now is a few lines each.

kate-shine:
Closes #665. Follow-up to #646, which the two findings below came out of. Both are
fail-opens in the exposure cascade: cases where the planner reports "not exposed"
and leaves a crate on the patch floor while its public API really does carry the
bumped types.

Neither is reachable in the workspace as it stands today, which is why #646 was
merged ahead of them — see the issue for the reachability analysis.

## 1. Judge both exposure paths when a crate holds each

`Get-PublishedDependentsExposingTarget` picked its exposure predicate with an
`if`/`else` on whether a direct dependency edge existed, making the direct and
indirect tests mutually exclusive. A crate holding both paths was judged only on
the direct one.

That is a fail-open, because **the two predicates accept different allowlist roots
by design**:

| | roots built from | treats the target's own crate root |
|---|---|---|
| direct | the edge's `DepAliases` (rename-aware) | **rejects** it — a `package = "..."` rename shadows it |
| indirect | the target's own record | accepts it |

So a crate that imports the target under a rename *and* also reaches it through a
conduit re-exporting its types loses the root its indirect path legitimately
earned. Both edges are now evaluated independently, and the crate is raised if
either claims exposure.

The load-bearing subtlety: the indirect test asks whether a path reaches the
target **through some other package**, not whether the crate is merely reachable
from it. Plain reachability would count the direct edge itself and readmit exactly
the renamed root the direct predicate just rejected.

## 2. Fail closed on a malformed `allowed_external_types` container

`@($allowed.Value)` promoted a malformed scalar into a valid-looking one-entry
allowlist that matches nothing, yielding "not exposed". `[package.metadata]` is
arbitrary TOML that cargo passes through unvalidated, so this is reachable by a
typo.

This was the **only** malformed shape that failed open — a non-string container
already returned `True`, as did absent metadata. The value is now left `$null`,
reusing the existing absent-metadata branch, which fails closed.

Two shapes to be careful with:
- a string is itself `IEnumerable` (over its chars), so it must be excluded explicitly;
- `@()` must survive as a positive claim of "no external types" — collapsing it to
  `$null` would be a different fail-open.

## Verification

Every change was mutation-verified in both directions — a guard test that does not
exercise the condition it guards is worthless:

- reverting the `if`/`else` fails **exactly** the new mixed-path test;
- swapping path-independence for plain reachability fails **exactly** the control
  test, whose only path to the target is the renamed direct edge;
- restoring `@($allowed.Value)` fails **exactly** the scalar and table tests, while
  the array and empty-array tests keep passing, proving no over-rejection.

Suites: `unit\release-packages` **87/87**, `unit\releasing` **67/67**,
`integration\ExposureCascade-RealWorkspace` **14/14** (re-run after the rebase,
since it reads live manifests and `main` has moved).

Also included: contract corrections on `Test-PackageAllowlistNamesTarget`,
`Get-AcceptedExposureRoots` and the indirect-edge section of `docs/releasing.md`,
all of which described the old mutually-exclusive behaviour. The `releasing.ps1`
part of the first commit is comment-only.

### DIFF
--- a/scripts/lib/release-flow.ps1
+++ b/scripts/lib/release-flow.ps1
@@ -211,6 +211,58 @@ function Parse-ReleaseTokens {
     return $results.ToArray()
 }
 
+# BFS over a workspace baseline collecting every transitive dependent of a
+# cargo package (identified by its underscore-normalized cargo name), in the
+# two shapes the cascade needs:
+#
+#   PublishedFolders    - folders of the published dependents only. What joins
+#                         a release set.
+#   DependentCargoNames - cargo names of ALL dependents, published or not.
+#                         Used to decide whether a crate reaches the target
+#                         through some *other* package, which requires seeing
+#                         unpublished conduits too.
+#
+# The target itself appears in neither: it is seeded into the visited set so
+# the walk terminates, but it is not its own dependent.
+#
+# Operates against an in-memory baseline snapshot rather than disk, so it
+# produces deterministic answers even after disk state changes.
+function Get-TransitiveDependentClosureFromBaseline {
+    param(
+        [Parameter(Mandatory = $true)][AllowEmptyCollection()][object[]]$Baseline,
+        [Parameter(Mandatory = $true)][string]$TargetCargoName
+    )
+
+    $toVisit = [System.Collections.Generic.Queue[string]]::new()
+    $toVisit.Enqueue($TargetCargoName)
+    $visited = [System.Collections.Generic.HashSet[string]]::new()
+    [void]$visited.Add($TargetCargoName)
+
+    $dependents = New-Object 'System.Collections.Generic.List[string]'
+    $dependentNames = [System.Collections.Generic.HashSet[string]]::new(
+        [System.StringComparer]::Ordinal)
+    while ($toVisit.Count -gt 0) {
+        $current = $toVisit.Dequeue()
+        foreach ($candidate in $Baseline) {
+            $candidateNorm = $candidate.Name.Replace('-', '_')
+            if ($visited.Contains($candidateNorm)) { continue }
+            if ($candidate.Deps -contains $current) {
+                [void]$visited.Add($candidateNorm)
+                [void]$dependentNames.Add($candidateNorm)
+                $toVisit.Enqueue($candidateNorm)
+                if ($candidate.Published) {
+                    $dependents.Add($candidate.Folder)
+                }
+            }
+        }
+    }
+
+    return [pscustomobject]@{
+        PublishedFolders    = @($dependents)
+        DependentCargoNames = $dependentNames
+    }
+}
+
 # BFS over a workspace baseline to find all published transitive dependents of
 # a cargo package (identified by its underscore-normalized cargo name). Mirrors
 # Get-AllTransitiveDependents but operates against an in-memory baseline
@@ -225,28 +277,8 @@ function Get-TransitivePublishedDependentsFromBaseline {
         [Parameter(Mandatory = $true)][string]$TargetCargoName
     )
 
-    $toVisit = [System.Collections.Generic.Queue[string]]::new()
-    $toVisit.Enqueue($TargetCargoName)
-    $visited = [System.Collections.Generic.HashSet[string]]::new()
-    [void]$visited.Add($TargetCargoName)
-
-    $dependents = New-Object 'System.Collections.Generic.List[string]'
-    while ($toVisit.Count -gt 0) {
-        $current = $toVisit.Dequeue()
-        foreach ($candidate in $Baseline) {
-            $candidateNorm = $candidate.Name.Replace('-', '_')
-            if ($visited.Contains($candidateNorm)) { continue }
-            if ($candidate.Deps -contains $current) {
-                [void]$visited.Add($candidateNorm)
-                $toVisit.Enqueue($candidateNorm)
-                if ($candidate.Published) {
-                    $dependents.Add($candidate.Folder)
-                }
-            }
-        }
-    }
-
-    return @($dependents)
+    return @((Get-TransitiveDependentClosureFromBaseline `
+                -Baseline $Baseline -TargetCargoName $TargetCargoName).PublishedFolders)
 }
 
 # Raises a resolved release-set entry's EffectiveChangeType to at least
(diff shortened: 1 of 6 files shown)


scripts/lib/release-flow.ps1
```diff
@@ -394,26 +426,79 @@ function Test-EntryPlansBreakingRelease {
...
+    )
+
+    foreach ($dep in @($Dependent.Deps)) {
+        # Belt and braces: the target seeds the BFS visited set, so it is never
+        # added to its own dependent closure and the Contains test below would
+        # reject it anyway. Skipping it explicitly keeps "some package other
+        # than the target" true by construction rather than by a property of a
+        # different function.
+        if ($dep -eq $TargetCargoName) { continue }
+        if ($DependentCargoNames.Contains($dep)) { return $true }
```
sandersaares, line 461: [Copilot speaking]

# Resolve indirect paths by workspace package identity

`DependentCargoNames` is the set of workspace packages in the target's reverse dependency closure, and an indirect path means that the candidate reaches the target through one of those packages. `Test-PackageReachesTargetIndirectly` currently treats a matching dependency name as proof of that path. However, `Get-WorkspacePackages` stores every non-development dependency in `Deps` using only its normalized Cargo package name, discarding the dependency's `path` and `source`. A registry or non-member path dependency can therefore be mistaken for a same-named workspace package.

**Reproducible reasoning**: The failure composes across three layers. First, `Get-WorkspacePackages` converts each normal or build dependency to `dep.name` and appends it to `Deps`, regardless of whether the dependency resolves to a workspace member. Second, `Get-TransitiveDependentClosureFromBaseline` builds `DependentCargoNames` from the names of baseline workspace packages that depend on the target. Third, `Test-PackageReachesTargetIndirectly` returns true when any entry in the candidate's name-only `Deps` appears in that set. Consider a workspace package `relay` that depends on the breaking target, plus a candidate that directly imports the target under a rename and separately depends on a registry or non-member path package also named `relay`. The closure contains the workspace `relay`; the candidate's unrelated dependency has the same text name, so the indirect-path predicate returns true even though the candidate has no edge to the workspace `relay`. If the candidate allowlists the target's own crate root, the renamed direct check rejects that root but the fabricated indirect check accepts it and raises the candidate to breaking. Cargo metadata exposes dependency `path` and `source`, so the loader has enough information to avoid conflating these package identities.

**Consequence**: The release planner can assign and propagate an unnecessary breaking release when an external dependency happens to share a package name with a workspace conduit, even though no workspace path carries the target's types to the candidate.

**Recommended action**: Record workspace-member dependency edges separately from name-only dependency declarations. While loading metadata, resolve each dependency's `path` to a workspace member (or use resolved Cargo package IDs), and use only those workspace edges in the reverse closure, direct-edge gate, and indirect-path predicate. Registry and non-member path dependencies should not participate in workspace cascade reachability; real workspace edges should retain the existing direct and indirect exposure behavior. Add a regression with a non-member dependency whose package name matches a workspace conduit and verify that it does not establish an indirect path.

**References**:

- scripts/lib/releasing.ps1 — Get-WorkspacePackages
- [cargo_metadata::Dependency exposes name, source, and path](https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.Dependency.html)
- [Cargo metadata JSON format](https://doc.rust-lang.org/cargo/commands/cargo-metadata.html)

**Impacted locations**:

- `scripts/lib/release-flow.ps1:246-256`
- `scripts/lib/release-flow.ps1:454-461`
- `scripts/lib/releasing.ps1:624-630`

sandersaares (commented):
[Copilot speaking]

| Diagnostic | Value |
|------------|-------|
| Cache | Miss |

</details>

sandersaares (approved):

scripts/lib/release-flow.ps1
kate-shine: GHCP: Verified and fixed in e337b1e9.

Confirmed the conflation is real: `Get-WorkspacePackages` recorded `dep.name` for every non-dev dependency and discarded the resolved path, so nothing downstream could tell a member from a same-named registry or out-of-workspace path crate.

Two things the investigation added to your account of it:

1. **It predates the exposure cascade.** `git blame` puts the name-only match at #436 (2026-06-10), so it already affected the reverse closure and the direct-edge gate, not just the new indirect predicate.
2. **Every consumer wanted identity already.** All eight `Deps` consumers filter by workspace membership afterwards -- two with a literal `# external package` skip -- so they were all asking the right question against the wrong key.

So the fix went to the loader rather than the three call sites you listed: a dependency now joins the graph only when its `path` resolves to a member manifest directory. That closes the pre-existing paths at the same time. `DepAliases` is filtered by the same guard, since leaving it would keep the conflation one level down and let an external crate's `package = "..."` rename supply an allowlist root for a same-named member.

One deliberate behaviour change: a member depended on through the registry rather than by path is now excluded. That is correct -- bumping the local copy does not affect a consumer resolving to the published one -- and it is not hypothetical-only in the other direction: I probed the live workspace and all 58 members are depended on by path, so nothing changes today.

Regression added as requested (`GitFs.Tests.ps1`), using a new non-member fixture under `external/`: a member `relay` depending on the target, plus a candidate depending on a *different* `relay` outside the workspace. Mutation-verified -- removing the path guard fails exactly the two identity assertions (the non-member becomes an edge; the candidate enters the target's dependent closure) while both controls keep passing, so genuine member edges are not over-rejected.

Suites: `unit\release-packages` 87/87, `integration\DependencyRename` 16/16, `ExposureCascade-RealWorkspace` 14/14, `Releasing-Integration` 44/44.

martintmk (approved):

PR #673, opened 2026-08-14
perf(multitude): improve steady-state allocation throughput

geeknoid:
## Summary

- make Criterion and Callgrind comparisons measure identical, warmed, allocation-free Multitude and Bumpalo hot paths
- optimize arena reset, reference marking, fixed copies, clone initialization, and known-length iterator allocation
- add repeatable same-process reporting and reset regressions, update the performance report, and bump `multitude` to 0.8.1

The updated report has Multitude ahead on scalar allocation, slice copy/clone, iterator fill, and growable string/vector scenarios; the two remaining scenarios are within 4% of Bumpalo.

## Validation

- `cargo clippy -p multitude --profile dev --all-targets --all-features --locked`
- `cargo test -p multitude --locked --all-features --test arena reset::reset_preserves_smart_owner_from_retired_mixed_chunk`
- package formatting, README generation, and spellcheck

### DIFF
--- a/crates/multitude/docs/DESIGN.md
+++ b/crates/multitude/docs/DESIGN.md
@@ -225,15 +225,32 @@ after the arena is gone.
 smart pointer is pinned until reset even after its `Arc`s drop. This is the
 deliberate, acknowledged cost of letting one current chunk serve both
 styles; the arena tracks a single "did this chunk hand out a reference?"
-flag so that only genuinely mixed chunks pay it.
+flag so that only genuinely mixed chunks pay it. Local reservations sample
+that flag after bounds checks but before committing the bump cursor, and write
+it only for the false-to-true transition. This keeps the steady-state path free
+of repeated marker stores and avoids a marker load after the cursor store.
+The marker transition and chunk-exhaustion edges are laid out as cold blocks,
+leaving the marked, in-capacity path as straight-line fall-through code.
 
 **Reset is a cursor rewind.** `Arena::reset` takes `&mut self`, which
 statically guarantees no `Alloc` (which borrows `&self`) is live. It runs
 **no** destructors — every `Alloc` already ran its own on drop, and
-smart-pointer values remain owned by their still-live handles. It simply
-reconciles the current chunk's refcount surplus (below) and returns chunk
-bytes to the cache (or leaves chunks alive if escaped handles still hold
-them).
+smart-pointer values remain owned by their still-live handles. A current chunk
+that handed out no smart pointer is retained with its pre-credited refcount and
+rewound in place. Otherwise reset reconciles the current chunk's refcount
+surplus (below) and detaches it, returning its bytes to the cache when the last
+escaped handle releases it. Reset similarly releases only the list-owned
+reference to each older retired chunk; an outstanding smart owner keeps that
+chunk alive until the owner releases it.
+
+The marker is cleared even when reset retains the current chunk: the next
+generation may use that chunk exclusively for smart pointers and must still
+qualify for early reclamation. Leaving a stale mark would silently pin it.
+Encoding an additional "rewound but previously local" state would remove a
+small part of the first-allocation boundary, but the measured path already
+executes fewer instructions than Bumpalo and the extra lifecycle state is not
+justified by the remaining wall-clock delta in
+[`PERF.md`](PERF.md#reset-plus-the-next-allocation).
 
 `ArenaStats` keeps its lifetime counters and live gauges across this boundary,
 while its explicitly named `*_since_reset` counters start a new generation
--- a/crates/multitude/src/arena/alloc_slice_ref.rs
+++ b/crates/multitude/src/arena/alloc_slice_ref.rs
@@ -62,6 +62,20 @@ fn empty_slice<'a, T>() -> &'a mut [T] {
     unsafe { core::slice::from_raw_parts_mut(NonNull::<T>::dangling().as_ptr(), 0) }
 }
 
+/// Adopts a fully initialized slice while preserving a length already known by
+/// the caller.
+///
+/// # Safety
+///
+/// `slot` must contain exactly `len` initialized elements that the returned
+/// [`Alloc`] may exclusively own.
+unsafe fn adopt_slice_with_len<T>(slot: &mut [T], len: usize) -> Alloc<'_, [T]> {
+    debug_assert_eq!(slot.len(), len);
+    let ptr = slot.as_mut_ptr();
+    // SAFETY: guaranteed by the caller.
+    unsafe { Alloc::from_mut(core::slice::from_raw_parts_mut(ptr, len)) }
+}
+
 /// Records the caller invariant that `len` is non-zero.
 #[inline(always)]
 #[cfg_attr(test, mutants::skip)]
@@ -281,11 +295,11 @@ impl<A: Allocator + Clone> Arena<A> {
     /// into a fresh arena slot and takes ownership of it in an [`Alloc`].
     #[inline(always)]
     fn impl_alloc_slice_copy<T: Copy>(&self, src: &[T]) -> Result<Alloc<'_, [T]>, AllocError> {
+        let len = src.len();
         let slot = self.alloc_slice_copy_raw::<T>(src)?;
-        // SAFETY: `alloc_slice_copy_raw` returns the unique `&mut [T]` for a
-        // freshly-written arena slice that the arena hands out exactly once and
-        // never drops itself, so `Alloc` may adopt it and own its destructor.
-        Ok(unsafe { Alloc::from_mut(slot) })
+        // SAFETY: `alloc_slice_copy_raw` initialized exactly `src.len()`
+        // elements in the unique arena slot returned.
+        Ok(unsafe { adopt_slice_with_len(slot, len) })
     }
 
     /// Raw allocation path shared by `alloc_slice_copy` and
@@ -348,11 +362,11 @@ impl<A: Allocator + Clone> Arena<A> {
     /// into a fresh arena slot and takes ownership of it in an [`Alloc`].
     #[inline(always)]
     fn impl_alloc_slice_clone<T: Clone>(&self, src: &[T]) -> Result<Alloc<'_, [T]>, AllocError> {
+        let len = src.len();
         let slot = self.alloc_slice_clone_raw::<T>(src)?;
-        // SAFETY: `alloc_slice_clone_raw` returns the unique `&mut [T]` for a
-        // freshly-written arena slice that the arena hands out exactly once and
-        // never drops itself, so `Alloc` may adopt it and own its destructor.
-        Ok(unsafe { Alloc::from_mut(slot) })
+        // SAFETY: `alloc_slice_clone_raw` initialized exactly `src.len()`
+        // elements in the unique arena slot returned.
+        Ok(unsafe { adopt_slice_with_len(slot, len) })
     }
 
     /// Closure-free fast path for `alloc_slice_clone` /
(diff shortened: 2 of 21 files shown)


martintmk (approved):

ralfbiedert (approved):

crates/multitude/docs/DESIGN.md
```diff
@@ -225,15 +225,32 @@ after the arena is gone.
...
 styles; the arena tracks a single "did this chunk hand out a reference?"
-flag so that only genuinely mixed chunks pay it.
+flag so that only genuinely mixed chunks pay it. Local reservations sample
+that flag after bounds checks but before committing the bump cursor, and write
+it only for the false-to-true transition. This keeps the steady-state path free
+of repeated marker stores and avoids a marker load after the cursor store.
+The marker transition and chunk-exhaustion edges are laid out as cold blocks,
+leaving the marked, in-capacity path as straight-line fall-through code.
 
 **Reset is a cursor rewind.** `Arena::reset` takes `&mut self`, which
```
Vaiz, line 235: 🤖 **Clawpilot here!** Automated review comment from an AI agent -- not a human reviewer. Please verify before acting.
AI reviewer: GPT-5.6 Sol

The lifecycle diagram immediately above this section is now stale: it combines "fills up / reset" into one edge that removes `CURRENT` and eventually sends an `Alloc`-serving chunk to cache/free. The new `reset` path keeps a local-only current chunk installed and rewinds it in place. Could you split the refill and reset edges and show the retained/rewound-current branch so the diagram agrees with the updated prose and implementation?

crates/multitude/src/arena/alloc_slice_ref.rs
```diff
@@ -473,23 +485,21 @@ impl<A: Allocator + Clone> Arena<A> {
...
     /// `try_alloc_slice_fill_iter`. The iterator length is sampled once
-    /// via [`ExactSizeIterator::len`] before reservation. Keeping the refill
+    /// via [`ExactSizeIterator::len`] before this helper. Keeping the refill
     /// continuation out of line avoids materializing iterator state on the
     /// common path. The iterator is consumed only on success arms that return.
     #[inline(always)]
-    fn alloc_slice_fill_iter_raw<T, I: ExactSizeIterator<Item = T>>(&self, iter: I) -> Result<&mut [T], AllocError> {
-        reject_over_aligned::<T>()?;
-        let len = iter.len();
+    fn alloc_slice_fill_iter_raw<T, I: Iterator<Item = T>>(&self, iter: I, len: usize) -> Result<&mut [T], AllocError> {
```
Vaiz, line 502: 🤖 **Clawpilot here!** Automated review comment from an AI agent -- not a human reviewer. Please verify before acting.
cargo-evaluate guideline: [M-PARAMETER-CONSISTENCY](https://microsoft.github.io/rust-guidelines/guidelines/libs/ux/#M-PARAMETER-CONSISTENCY)

Reason:
The iterator-taking helpers place `iter` inconsistently relative to their closure-analog counterparts: `alloc_slice_fill_iter_raw(&self, iter: I, len: usize)` (line 502) puts `iter` first and `len` second, while its cold continuation `alloc_slice_fill_iter_refill(&self, len: usize, refill_hint: usize, iter: I)` (line 529) puts `len` first and `iter` last. This mirrors the fill_with pair (alloc_slice_fill_with_raw(len, f) / alloc_slice_fill_with_refill(len, refill_hint, f), both f-last), so the iterator variant's raw/refill split breaks the crate's own established ordering convention of placing the closure/iterator-like parameter last.

Recommendation:
Reorder alloc_slice_fill_iter_raw's parameters to `(&self, len: usize, iter: I)` to match the len-first, iterator-last convention used by alloc_slice_fill_iter_refill and the analogous fill_with_raw/fill_with_refill pair, and update its call site in impl_alloc_slice_fill_iter (line 490) accordingly.

Note: the `len` parameter is new in this PR, so this is a choice being made here rather than existing style being inherited. The unit test at line 563 also passes `(iter, len)` and would need the same update.

crates/multitude/src/arena/mod.rs
```diff
@@ -365,7 +365,12 @@ impl<A: Allocator + Clone> Arena<A> {
         // outstanding handles). Fold in the currently-active chunk's free
         // tail so the reported value also reflects the slack that would
         // become wasted if the next alloc forced a refill right now.
-        let current_free = u64::from(self.current.borrow().wasted_tail_for_stats());
+        let current_is_active = self.current_has_reference.get() || self.local_shared_count.get() != 0;
```
Vaiz, line 368: 🤖 **Clawpilot here!** Automated review comment from an AI agent -- not a human reviewer. Please verify before acting.
AI reviewer: Claude Opus 5

The comment directly above this line still says "Fold in the currently-active chunk's free tail so the reported value also reflects the slack that would become wasted if the next alloc forced a refill right now" — and this new line now conditionally *doesn't* fold it in.

The reason isn't obvious from the code: the gate exists so `wasted_tail_bytes` still reports zero after a reset, now that the chunk stays installed rather than being detached. Without a sentence saying that, the gate reads like a bug. Something like:

```rust
// A retained-but-rewound chunk has no live slack yet: `reset` leaves the
// chunk installed, so gate on generation activity to keep the gauge at
// zero across a reset.
```

Worth noting that this also quietly widens `current_has_reference`: it now doubles as a "this chunk is active in the current generation" signal for `stats()`, not just "handed out an arena-lifetime reference". If that dual role is intended, the field doc is the natural place to say so.

crates/multitude/src/internal/chunk_mutator.rs
```diff
@@ -150,6 +175,25 @@ impl<A: Allocator + Clone> ChunkMutator<A> {
     // Always rejecting requests would make caller refill loops infinite.
     #[cfg_attr(test, mutants::skip)]
     pub(crate) fn try_alloc(&self, size: usize, align: usize) -> Option<InChunk<u8>> {
+        self.try_alloc_before_commit(size, align, || {})
+    }
+
+    /// Local-reference form of [`Self::try_alloc`].
```
Vaiz, line 181: 🤖 **Clawpilot here!** Automated review comment from an AI agent -- not a human reviewer. Please verify before acting.
AI reviewer: GPT-5.6 Sol

This private comment mostly restates the helper name/link ("Local-reference form of `try_alloc`"). The same pattern appears on the newly added `try_alloc_uninit_local`, `try_alloc_bytes_local`, `try_alloc_uninit_slice_local`, and `try_alloc_freezable_slice_local`.

Could you replace these with the non-obvious contract instead — why the marker must transition before the cursor commit, and what that guarantees for chunk retirement? That's the thing a reader can't recover from the signature. (If the intent is specifically to disambiguate "local" as *local-reference* rather than thread-local, that's worth keeping — just say it once, with the invariant.)

crates/multitude/src/internal/chunk_mutator.rs
```diff
@@ -150,6 +175,25 @@ impl<A: Allocator + Clone> ChunkMutator<A> {
...
+    #[inline]
+    #[cfg_attr(test, mutants::skip)] // see `try_alloc`
+    fn try_alloc_local(&self, size: usize, align: usize, marker: &Cell<bool>) -> Option<InChunk<u8>> {
```
Vaiz, line 184: 🤖 **Clawpilot here!** Automated review comment from an AI agent -- not a human reviewer. Please verify before acting.
AI reviewer: Claude Opus 5

The marker guard appears verbatim three times — here, in `try_alloc_bytes_local`, and in `try_alloc_smart_prefixed_local`:

```rust
|| {
    if !marker.get() {
        set_reference_marker(marker);
    }
}
```

The load-then-cold-store pattern is the actual optimization and is load-bearing for the invariant documented in `DESIGN.md` ("write it only for the false-to-true transition"), so it's worth having exactly one copy. You already extracted the `#[cold]` store; consider extracting the guard too:

```rust
/// Records the first local reservation of a generation. The load guard keeps
/// the steady state store-free; the transition itself is cold.
#[inline]
fn mark_local_reference(marker: &Cell<bool>) {
    if !marker.get() {
        set_reference_marker(marker);
    }
}
```

Each wrapper then becomes `self.try_alloc_before_commit(size, align, || mark_local_reference(marker))`.

crates/multitude/src/internal/uninit.rs
```diff
@@ -13,6 +13,26 @@ use core::{mem, str};
...
+const FIXED_COPY_BYTES: usize = 64;
+
+/// Copies one cache-line-sized region without a dynamic `memcpy` dispatch.
+///
+/// # Safety
+///
+/// `dst` and `src` must be valid for [`FIXED_COPY_BYTES`] non-overlapping
+/// bytes.
+//
+// Keeping this out of line is deliberate: when LLVM inlined the fixed copy
```
Vaiz, line 25: 🤖 **Clawpilot here!** Automated review comment from an AI agent -- not a human reviewer. Please verify before acting.
cargo-evaluate guideline: [M-NO-META-DESIGN-DOCUMENTATION](https://microsoft.github.io/rust-guidelines/guidelines/ai/#M-NO-META-DESIGN-DOCUMENTATION)

Reason:
Two comments narrate implementation-history / 'why we picked X over Y' engineering journeys rather than documenting end-state behavior: lines 25-27 explain that keeping a function 'out of line is deliberate' because 'when LLVM inlined the fixed copy beside the dynamic fallback, it merged both paths back into libc memcpy', and lines 205-208 explain 'A conventional iterator loop made it coalesce each short trivial slice into an out-of-line memcpy; this shape instead produced inline vector copies in disassembly'. These are design-journal-style rationales about a past alternative implementation and compiler behavior observed during development, not documentation of current behavior for users.

Recommendation:
Replace the historical/comparative narrative with a concise forward-looking comment, e.g. 'Kept out-of-line and manually unrolled to ensure LLVM emits inline vector copies instead of a memcpy call; do not simplify to an iterator or #[inline] without re-checking codegen.' Avoid describing what a previous version did or why an alternative was rejected.

Note: please keep the constraint itself — that `#[inline(never)]` is deliberate and that this must not be simplified to an iterator loop without re-checking codegen. Only the "here's what we tried and what LLVM did" framing is at issue.

crates/multitude/src/internal/uninit.rs
```diff
@@ -177,12 +199,52 @@ impl<'a, T> Uninit<'a, [T]> {
...
+        // Keep fixed clone sites visible to LLVM. A conventional iterator
+        // loop made it coalesce each short trivial slice into an out-of-line
+        // `memcpy`; this shape instead produced inline vector copies in
+        // disassembly while the guard preserves panic cleanup.
+        // SAFETY: `dst` has `len` uninitialized `T` slots, `src` has the same
+        // length, and the guard drops exactly the initialized prefix on panic.
+        unsafe {
+            let dst = slice_ptr.as_ptr().cast::<T>();
+            let mut guard = InitGuard { dst, initialized: 0 };
+            while len - guard.initialized >= 8 {
```
Vaiz, line 214: 🤖 **Clawpilot here!** Automated review comment from an AI agent -- not a human reviewer. Please verify before acting.
AI reviewer: Claude Opus 5
cargo-evaluate guideline: [M-DOCUMENTED-MAGIC](https://microsoft.github.io/rust-guidelines/guidelines/universal/#M-DOCUMENTED-MAGIC)

Reason:
The clone-unrolling loop uses a hardcoded unroll factor of `8` (line 214, `while len - guard.initialized >= 8`) with no comment explaining why 8 was chosen specifically (as opposed to 4 or 16) or what tradeoff informed it; the surrounding comment (lines 205-208) explains why manual unrolling exists at all, but not the specific magic constant `8`.

Recommendation:
Add a brief comment near the unroll loop (or promote `8` to a named constant like `CLONE_UNROLL_FACTOR`) documenting why this specific factor was chosen (e.g., based on benchmarking or register/vector-width considerations).

Adding to that: the 8-wide step is eight byte-identical copies of the same three lines, and the literal `8` is both the loop bound *and*, implicitly, the body's repetition count — with nothing tying them together, so a future edit that lands in seven of eight blocks is invisible in review. Naming the constant fixes half of that; expressing the unroll as a constant-trip-count inner loop would fix both:

```rust
while len - guard.initialized >= 8 {
    for _ in 0..8 {
        let index = guard.initialized;
        dst.add(index).write((*src.as_ptr().add(index)).clone());
        guard.initialized += 1;
    }
}
```

I haven't built this, so please re-check against the disassembly you used to justify the current shape. If the nested form *does* regress, that's worth putting in the comment — "a constant-trip-count inner loop coalesced too" is what stops the next reader from simplifying it back.

PR #676, opened 2026-08-18
fix(rest_over_grpc_examples): rename the duplicate `tower_service` example

Vaiz:
Two packages declared an example target named `tower_service`: `rest_over_grpc` (autoexample, `required-features = ["tower"]`) and `rest_over_grpc_examples` (explicit `[[example]]`). Cargo warned about an output filename collision on both `tower_service.exe` and its `.pdb`, then built the two targets concurrently into that one path.

On Windows the resulting `link.exe` race is fatal — `anvil-examples` failed with `LNK1104: cannot open file ...\examples\tower_service.exe` on an unrelated PR. On Linux and macOS the same race silently overwrote the binary, so only the Windows leg surfaced it. Cargo also warns the collision "may become a hard error in the future" (rust-lang/cargo#6313).

Rename the target in `rest_over_grpc_examples` (`publish = false`) to `generated_tower_service`, which also states what distinguishes it: it serves the generated tonic-bridged `Transcoder`, whereas the `rest_over_grpc` example wraps a hand-written one. The published crate's example name is unchanged.

`cargo build --workspace --examples --all-features --locked` now completes with no filename-collision warning anywhere in the workspace.

---

## Also: unblock CI for test-less affected sets

The first commit made this PR the first one whose affected crate set is a single crate with no test targets, which exposed a pre-existing CI bug. `delta` reported `Modified 1 / Affected 1`, so the `testing` job's `Tests` step ran effectively as `-p rest_over_grpc_examples`. That crate has no test targets, so:

```
Starting 0 tests across 1 binary
Summary [0.000s] 0 tests run: 0 passed, 0 skipped
error: no tests to run
(hint: use `--no-tests` to customize)
```

`cargo nextest run` exits 4 and all four `testing` legs fail, even though nothing is wrong. The condition predates the rename — `rest_over_grpc_examples` had no test targets on `main` either, so any docs- or examples-only PR would hit this.

The second commit passes `--no-tests=pass` on the two nextest runs whose crate selection is delta-narrowed (`testing` → `Tests`, `extended-analysis` → `Careful`), so an empty selection is a pass rather than an error.

The `Serde negative contract` run is deliberately left alone: it pins `-p internity -E 'binary(compile_fail)'` with no delta narrowing, so zero tests there really would mean the filter had stopped matching, and should still fail.

Verified on the pinned nextest 0.9.137: the command above exits 4, and exits 0 with `--no-tests=pass`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -125,7 +125,11 @@ jobs:
         run: cargo hack build --each-feature --workspace ${{ needs.delta.outputs.exclude_not_affected }} --verbose --locked --color always
       - name: Tests
         if: success() || failure()
-        run: cargo nextest run --verbose --workspace ${{ needs.delta.outputs.exclude_not_affected }} --all-features
+        # --no-tests=pass: the crate selection here is narrowed by `delta` to the
+        # affected crates, so a PR that only touches crates without test targets
+        # (docs, examples) legitimately selects zero tests. Without this, nextest
+        # treats that as an error and exits 4.
+        run: cargo nextest run --verbose --workspace ${{ needs.delta.outputs.exclude_not_affected }} --all-features --no-tests=pass
       - name: Docs
         if: success() || failure()
         env:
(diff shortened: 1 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -125,7 +125,11 @@ jobs:
         run: cargo hack build --each-feature --workspace ${{ needs.delta.outputs.exclude_not_affected }} --verbose --locked --color always
       - name: Tests
         if: success() || failure()
-        run: cargo nextest run --verbose --workspace ${{ needs.delta.outputs.exclude_not_affected }} --all-features
+        # --no-tests=pass: the crate selection here is narrowed by `delta` to the
```
sgalkin, line 128: this change is not mentioned in description

sgalkin (approved):

.github/workflows/main.yml
Vaiz: it showed up later :(

martintmk (approved):

PR #677, opened 2026-08-18
feat: introduce the observed_utils crate

Vaiz:
Move the OpenTelemetry helper crate from the internal repo into this repository, following `observed`, `observed_macros` and `observed_testing`. It is the optional companion for consumers that export typed events to OpenTelemetry.

The crate is named `observed_utils` and starts at version 0.1.0. It ships internally as `observed_helpers` at 0.24.0; both were changed on review here, while nothing is published and the change is therefore free.

`observed` itself no longer depends on `opentelemetry`, so an exporter has to convert values on its own. That conversion lives here once rather than being re-derived in every exporter: `any_value_of`, `otel_value_of` and `otel_severity_of`. The crate also carries `format_any_value`, which renders an `AnyValue` in human-readable form instead of its `Debug` shape, and `SensitiveSlice`, a bounded, type-erased collection of `RedactedDisplay` references that renders without allocating.

Changes made during the move:

- `any_value_of` and `otel_value_of` handle `Value::U64`. The variant was added in 0.24 after this crate was written, so a `u64` dimension used to fall through to the `#[non_exhaustive]` guard arm and be exported as a debug string. Values past `i64::MAX` still export as a decimal string, because neither `AnyValue` nor `opentelemetry::Value` has an unsigned variant.
- `metric_number_of` returns to this crate. It was parked in `observed_testing` because that was its only consumer here; `observed_testing` now uses it from this crate instead of keeping a second copy.
- The example uses the `#[path = "support/taxonomy.rs"]` stand-in taxonomy the `observed` examples use, instead of the ox-sdk-only `microsoft_enterprise_data_taxonomy` crate, and `expect` with a message instead of `unwrap`, to satisfy this workspace's `clippy::unwrap_used` policy.
- The `#![doc(html_logo_url)]` and `#![cfg_attr(coverage_nightly, ...)]` attributes are enabled, matching the convention of the other crates here.
- `arrayvec` and `const-hex` are added to `[workspace.dependencies]`.

Effects:

- A crate that emits typed events and exports them no longer has to write its own `Value` mapping, and the mapping is governed by this repository's gates.
- `observed_testing` gains a dependency on `observed_utils`; the two copies of `metric_number_of` can no longer drift apart.
- Telemetry helper changes no longer need a manual transfer between two repositories.
- The crate name and version now differ from the internal copy until the follow-up PR removes it there.

---

### Rename and version (review request)

`observed_helpers` becomes `observed_utils`, and the version becomes 0.1.0 instead of inheriting `observed`'s 0.24.0. Renamed in place: the directory, the package name, the `repository` and doc-URL attributes, both `observed_testing` call sites and its dependency entry, the `[workspace.dependencies]` entry, two prose references under `crates/observed`, and the lockfile. Git records every file as a rename, so the move carries no content change. The README is regenerated by cargo-doc2readme rather than hand-edited.

### Correctness: an unreachable overflow probe removed

`SensitiveSlice::new` ended with a post-loop `iter.next()` used to detect overflow. It cannot be reached: a `for` loop exits normally only after `next()` returned `None`, so the extra call could report `Some` only for an iterator that resumes after `None`, which the `Iterator` contract leaves unspecified. The loop already detects overflow on its own, because `is_full()` is checked before the push, so the `N+1`-th item is pulled, observed and discarded.

The gates could not have caught this: the line executed on every non-overflowing call, so it counted as covered, and `is_some()` was simply always false.

### Bytes render without allocating (review request)

`format_any_value`'s `Bytes` arm used `const_hex::encode`, which returns an owned `String`. It now uses `const_hex::display`, a `Display` adapter that writes through const-hex's own encoder straight into the formatter, so every arm writes directly and rendering allocates nothing.

`display` is not gated behind const-hex's `alloc` feature and `encode` was the only user of it, so the feature is dropped. const-hex enables `proptest` through `alloc`, and nothing else in the workspace pulled it in, so `proptest`, `rand_xorshift` and `unarray` leave `Cargo.lock` too.

### Other review fixes

- **Spellcheck**: `implementors` becomes `implementers` and the British `signalled` becomes `signaled`. Two genuinely new technical words introduced by the module docs, `recursing` and `redactable`, were added to `.spelling`.
- **External types**: the approved-type list names `opentelemetry::logs::record::Severity`, the canonical path `cargo-check-external-types` reports, instead of the re-export path.
- **Docs**: the example's module doc said `SensitiveCollection`, a type that does not exist; `examples/support/taxonomy.rs` claimed to be shared by the `observed` crate's examples, and now describes this crate's copy and why the copy exists; `format_any_value.rs` and `sensitive_slice.rs` gained `//!` headers, matching `otel.rs`.
- **Tests**: added for the paths nothing reached - `AnyValue::Map` rendering, `metric_number_of` on `Value::U64`, `otel_value_of` on `Bool` / `F64` / `BoolArray` / `F64Array`, `Text::Shared`, and the `Debug` and `Clone` impls of `SensitiveSlice`.

### Deliberate decision: the `#[non_exhaustive]` guard arms stay

The only lines no test reaches are the wildcard arms guarding the foreign `#[non_exhaustive]` enums `observed::Value`, `observed::Text` and `opentelemetry::logs::AnyValue`. All variants that exist today are matched explicitly, so those arms are unreachable and cannot be covered by a test. They are kept on purpose: they are what makes a variant added upstream degrade to a readable debug form instead of failing to compile or silently dropping data.

Rather than delete them for the 100 % line-coverage gate, the four dispatch functions carry `#[cfg_attr(coverage_nightly, coverage(off))]` with a comment stating exactly what the exclusion covers - including that the inline scalar and `Bytes` arms are excluded too, and are pinned by tests and mutation testing rather than by the gate. The work they delegate to (`unsigned_any_value`, `unsigned_otel_value`, `list_of`, `write_list` / `write_map`) stays measured.

### `cargo-deny`: h2 advisory fixed, stale ignores removed

`anvil-deny` failed on the advisory "h2 unbounded empty DATA frames" against h2 0.4.15, which this repository carries transitively. The advisory predates this branch, but the fix is a one-line lockfile bump, so it is included: h2 moves to the patched 0.4.16. Only the h2 entry is touched - a full `cargo update h2` also re-resolved unrelated `windows-sys` edges downwards.

Two ignores that no longer match any crate were removed from `deny.toml`: `RUSTSEC-2025-0141` (bincode, no longer in the lockfile at all) and `RUSTSEC-2026-0173` (proc-macro-error2, still present but no longer matched). The quick-xml ignores stay, because quick-xml 0.39.4 is still pulled in through the Azure SDK. `cargo deny check all` now reports advisories, bans, licenses and sources ok with no `advisory-not-detected` warnings.

### Verification

On the renamed crate: tests, clippy `-D warnings`, doc, both rustfmt configs, `doc2readme --check`, spellcheck, cargo-sort, cargo-machete and `cargo deny check all` are clean; coverage is 100 % with no uncovered lines; `cargo mutants` over the package reports no surviving mutants.

### Second review round: public documentation corrected

Three claims in the public documentation were not supported by the implementation as written. All three are fixed as documentation; no behaviour changed.

- **The `u64` threshold is now public.** `any_value_of` and `otel_value_of` described their result as the value's OpenTelemetry counterpart, but a `Value::U64` has no single output type: up to `i64::MAX` it becomes the integer variant, past it a decimal string. That boundary was documented only on the private helpers, which never reach rustdoc. It is now on both public functions and in the crate overview. The policy itself is unchanged - neither destination enum has an unsigned variant, so the alternatives are wrapping into a negative number or dropping the value.
- **`SensitiveSlice`'s capacity has a lower bound.** `N` was documented only as the maximum number of stored items, while `new` asserts `N > 0` at compile time. The type parameter and the constructor now state it.
- **`SensitiveSlice`'s guarantees are scoped to what the wrapper controls.** The module said items are "never rendered in the clear by this module" and the crate said rendering happens "without allocating". Rendering is delegated to each item's own `RedactedDisplay` implementation under the caller's `Redactor`, which is permitted to pass values through, and an item may allocate. The wording now promises inline storage of the collection and names the delegation instead.

Declined in this PR: changing `metric_number_of` to preserve the numeric kind, or returning it to `observed_testing`. The precision analysis is correct and is already acknowledged by the `cast_precision_loss` expectations at both call sites, but a new numeric contract is a redesign of a helper this PR only moves, and returning it to `observed_testing` reverses the de-duplication that is part of the move. Follow-up material, together with the adjacent-`u64` boundary test suggested in review.

### Third review round: the `Sync` bound is documented

`SensitiveSlice` was documented as storing `&dyn RedactedDisplay` fat pointers, and `new` as accepting implementers of `RedactedDisplay`. The stored type is actually `&(dyn RedactedDisplay + Sync)` and the constructor requires `T: RedactedDisplay + Sync`, so a caller with a non-`Sync` item type satisfied the documented bound and still failed to compile.

The bound is documented rather than removed. `&(dyn Trait + Sync)` is itself `Send` and `Sync` while a bare `&dyn Trait` is neither, so the bound is what lets an event struct holding a `SensitiveSlice` cross threads; dropping it would narrow where the type can be used. Documentation only, and the generated README is unaffected because these docs live on the type rather than in `lib.rs`.

---

**Rebase note (2026-08-19):** rebased onto `main` after it advanced; the branch had gone into a conflicting state. The only conflict was in `Cargo.lock`, where upstream's new `arty_executor` package entry landed in the same alphabetical slot as this branch's `arrayvec` entry. Both were kept, in alphabetical order; `cargo metadata --locked` accepts the merged lockfile. No source change.

---

**Conversion policy change (2026-08-24, owner ruling):** `any_value_of` and `otel_value_of` now convert a `Value::U64` unconditionally to an `i64`, **saturating at `i64::MAX`**. Previously a value past `i64::MAX` was exported as its decimal string, which preserved the digits but changed the exported *type* for large values — the same dimension could reach the backend sometimes as a number and sometimes as a string. Saturating keeps the exported type stable and ordered; the cost is that everything past `i64::MAX` collapses onto one reading. Wrapping was never on the table, as it reports a large count as a negative one. `unsigned_any_value` / `unsigned_otel_value` collapse into a single `saturating_i64` helper, and tests pin the exact `i64::MAX` boundary as well as values past it.


crates/observed_helpers/examples/sensitive_collection.rs
```diff
@@ -0,0 +1,155 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Demonstrates using `SensitiveCollection` inside an emit `Event`, processed
```
Vaiz: 🤖 **Clawpilot here!** Automated review comment from an AI agent -- not a human reviewer. Please verify before acting.
AI reviewers: Claude Opus 5, GPT-5.6 Sol, Gemini 3.7 Flash, Grok 4.5

This module doc names `SensitiveCollection`, but that type does not exist anywhere in the workspace — a
grep across `crates/` returns only this line. The type the example actually imports and uses is
`SensitiveSlice` (line 18, and the `NotificationBatch` fields at lines 64-66).

It looks like a leftover from the rename during the move from the internal repo, and it is the first line
a reader of the crate's only example sees.

```suggestion
//! Demonstrates using `SensitiveSlice` inside an emit `Event`, processed
```

The file name `sensitive_collection.rs` carries the same old name; renaming it to `sensitive_slice.rs`
would finish the cleanup, though that is optional and noisier.

crates/observed_helpers/examples/support/taxonomy.rs
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! A small stand-in data taxonomy shared by the `observed` crate's examples.
```
Vaiz: 🤖 **Clawpilot here!** Automated review comment from an AI agent -- not a human reviewer. Please verify before acting.
AI reviewers: Claude Opus 5, Grok 4.5

This file is byte-identical to `crates/observed/examples/support/taxonomy.rs` (`git diff --no-index`
reports no differences), so its module doc still describes its original home: this line says the taxonomy
is "shared by the `observed` crate's examples", and lines 11-13 give a rationale about not depending on
`observed_testing` that was written for that crate.

Duplicating the file is reasonable — a published crate cannot `#[path]` outside its own package directory —
but could the doc be reworded to describe *this* crate's examples? As it stands, a reader who opens the copy
under `observed_helpers` is told it belongs somewhere else, which is exactly the kind of text that stops
being true during a repo move. If drift between the two copies becomes a concern later, a small shared
`examples_support` dev-dependency crate would remove the duplicate entirely.

crates/observed_utils/src/format_any_value.rs
```diff
@@ -0,0 +1,185 @@
+// Copyright (c) Microsoft Corporation.
```
Vaiz, line 1: 🤖 **Clawpilot here!** Automated review comment from an AI agent -- not a human reviewer. Please verify before acting.
cargo-evaluate guideline: experimental rule; no published guideline

Reason: this file exposes a public API but has no `//!` module documentation

Recommendation: add a `//!` inner-doc header at the top of the file covering what the module contains, when to use it, and any observable side effects. See M-MODULE-DOCS

The same finding is reported for `src/sensitive_slice.rs:1`.

One clarification on the linter's wording, so this is not over-sold: `lib.rs` declares these modules
privately (`mod format_any_value;`, `mod sensitive_slice;`) and re-exports their items at the crate root, so
a `//!` header here would not appear in published rustdoc output. The concrete point that does hold is
internal consistency — `src/otel.rs:4` carries a `//!` header and these two modules do not. Adding one to
each is cheap and helps the next maintainer; treat it as a consistency nit rather than a docs.rs gap.

crates/observed_utils/src/format_any_value.rs
```diff
@@ -0,0 +1,185 @@
...
+struct DisplayAnyValue<'a>(&'a AnyValue);
+
+impl DisplayAnyValue<'_> {
+    // The last arm guards the `#[non_exhaustive]` `AnyValue`, so it cannot be
+    // reached by any variant that exists today, and coverage instrumentation
+    // counts an arm that is never taken as an uncovered line. The dispatch is
+    // therefore excluded from the coverage gate instead of the guard being
+    // deleted; the rendering it delegates to stays measured, and mutation
+    // testing still applies here.
+    #[cfg_attr(coverage_nightly, coverage(off))]
```
Vaiz, line 67: 🤖 **Clawpilot here!** Automated review comment from an AI agent -- not a human reviewer. Please verify before acting.
AI reviewers: Claude Opus 5, Grok 4.5

This justification ends with "the rendering it delegates to stays measured", but that holds for only two of
the seven arms. `Int`, `Double`, `String`, `Boolean` and `Bytes` (including the `const_hex::encode` call)
are rendered **inline inside the excluded function** at lines 55-59, so they are not measured either — the
exclusion is wider than "dispatch only". And while `write_list` / `write_map` do sit outside the exclusion,
they only emit the delimiters and then call back into `fmt_value` for each element (lines 80 and 93), so
nested element rendering is excluded as well.

`otel.rs:12-13` has the same shape: `string_value_of` is excluded in full and points at the `any_value_of`
rationale about delegated conversions staying measured, but `string_value_of` delegates nothing — its whole
`Static` / `Shared` conversion is inside the exclusion.

To be clear about the impact: nothing is untested today — `format_bytes`, `format_string` and the
`Text::Shared` test do exercise these arms. What is given up is the coverage *gate* that would catch a
future regression in them, and the comment currently understates that.

Two ways to square it: extract the scalar/`Bytes` rendering into a measured helper, the same move already
made for `write_list` / `write_map`, so the excluded function becomes pure dispatch + guard and the comment
becomes literally true; or keep the code and reword the comment to say the scalar rendering is excluded
along with the dispatch and is guarded by tests rather than by the gate. Applying the same wording fix at
`otel.rs:12` would keep the two consistent.

crates/observed_helpers/src/sensitive_slice.rs
```diff
@@ -0,0 +1,267 @@
...
+
+        for item in iter.by_ref() {
+            if items.is_full() {
+                overflowed = true;
+                break;
+            }
+            items.push(item as &(dyn RedactedDisplay + Sync));
+        }
+
+        // If we haven't detected overflow yet, check if the iterator has more.
```
Vaiz: 🤖 **Clawpilot here!** Automated review comment from an AI agent -- not a human reviewer. Please verify before acting.
AI reviewers: Claude Opus 5, Gemini 3.7 Flash, GPT-5.6 Sol

I think this post-loop probe is unreachable. The `if !overflowed` branch is only taken when the `for` loop
exited normally, and a `for` loop exits normally only after `iter.next()` returned `None` — so this extra
`next()` can return `Some` only for an iterator that resumes after `None`, which the `Iterator` contract
leaves unspecified.

The loop above already detects overflow correctly on its own: `items.is_full()` is checked *before* the
push, so the `N+1`-th item is pulled, observed and discarded, which is exactly what makes `overflowed`
accurate.

Suggest deleting the `if !overflowed { ... }` block along with the comment above it, which narrates the call
while asserting something that cannot be true at that point. If a comment is useful here, the non-obvious
part is in the loop itself — that the item pulled when `items.is_full()` is intentionally dropped, and that
pulling it is what makes `overflowed` correct.

crates/observed_helpers/examples/sensitive_collection.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Correct, and fixed in 722f308f — the doc now says `SensitiveSlice`. It was indeed a leftover from the rename during the move.

Declining the file rename: `sensitive_collection.rs` still describes what the example demonstrates (a collection of classified items inside an event), the type name appears correctly in the doc and the code, and a rename is churn that this move-only PR does not need.

crates/observed_helpers/examples/support/taxonomy.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Agreed on the doc, fixed in 722f308f. It now describes this crate's examples, and states why the copy exists — an example cannot include a file outside its own package directory — instead of pointing at `observed_testing`, which this crate does not depend on either.

Declining the shared `examples_support` crate: that is a workspace-wide change to how every crate's examples are structured, which is outside the scope of moving this one crate in. Worth raising separately if the copies ever diverge.

crates/observed_utils/src/format_any_value.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Taken as the consistency nit it is — 722f308f adds a `//!` header to both `format_any_value.rs` and `sensitive_slice.rs`, so all four modules now carry one.

Your clarification is right and worth keeping on the record: both modules are declared privately in `lib.rs` and their items re-exported at the crate root, so these headers do not reach published rustdoc. `just readme-check` confirms the generated README is unchanged.

crates/observed_utils/src/format_any_value.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

You are right, and the comment was mine from an earlier pass — it overstated what survives the exclusion. Fixed in 722f308f by taking your second option, rewording rather than extracting.

The comment on `fmt_value` now says plainly that the exclusion covers the inline scalar and `Bytes` arms too, that only `write_list` / `write_map` stay measured, and that each nested element re-enters the excluded dispatch. It also states what is actually given up: those arms are pinned by `format_int` / `format_bool` / `format_string` / `format_bytes` and by mutation testing, not by the coverage gate. `otel.rs` gets the matching treatment — `string_value_of` no longer points at a delegation rationale that does not apply to it, since both of its real conversions are inline, and `any_value_of`'s comment now names what genuinely stays measured (`unsigned_any_value`, `list_of`, the element closures).

I did not extract the scalar rendering: a helper would need its own match over the same `#[non_exhaustive]` enum, so it would carry the same unreachable guard arm and the exclusion would simply move rather than shrink. An accurate comment is the honest fix here.

`cargo mutants` over `otel.rs`, `format_any_value.rs` and `sensitive_slice.rs` reports no surviving mutants, and the crate still measures 100 % lines.

crates/observed_helpers/src/sensitive_slice.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Confirmed and removed in 722f308f. Your reasoning holds: the `for` loop exits normally only after `next()` returned `None`, so the probe could report `Some` only for an iterator that resumes after `None` — unspecified by the `Iterator` contract. The `by_ref()` on the loop is gone with it.

Worth noting how invisible this was to the gates: the line was *executed* on every non-overflowing call, so it counted as covered, and `is_some()` simply always returned `false`. Coverage could never have flagged it.

I took the comment suggestion too — it now explains the part that is actually non-obvious: `is_full()` is checked before the push, so the `N+1`-th item is pulled and deliberately dropped, and pulling it is what makes `overflowed` mean "more than `N`" rather than "at least `N`". All 32 tests still pass, `cargo mutants` over the file reports no survivors, and coverage stays at 100 %.

crates/observed_helpers/Cargo.toml
```diff
@@ -0,0 +1,48 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+[package]
+name = "observed_helpers"
```
martintmk:
```suggestion
name = "observed_utils"
```

feels more idiomatic

Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Flagging this to the PR owner rather than acting on it — a published crate name is his call, not mine, and I do not want an agent to rename a crate on a naming preference.

Two facts that seem relevant to the decision, offered neutrally:

- The crate ships today as `observed_helpers` in the internal ox-sdk repo, and this PR is a move of it. Renaming here makes the two repos disagree until the follow-up PR that removes it internally lands, and any internal consumer of `observed_helpers` has to be updated in step.
- Mechanically the rename is small: 14 files, ~29 references, the directory, and the `[workspace.dependencies]` entry. Nothing is published yet, so there is no crates.io compatibility cost — this is the cheapest moment it will ever be.

Held for the owner. No code change on this thread meanwhile.

Vaiz: rename, and change version to 0.1.0

Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Both done in 9a61d4bf: the crate is `observed_utils` and the version is 0.1.0.

Renamed in place — the directory, the package name, the `repository` and `html_logo_url` / `html_favicon_url` attributes, both `observed_testing` call sites plus its dependency entry, the `[workspace.dependencies]` entry, two prose references under `crates/observed` (`FEATURES.md` and the `observed` OTel example), and the lockfile. Git records all 12 files as renames, so the move itself carries no content change. The README is regenerated by cargo-doc2readme rather than hand-edited, and a repo-wide grep for the old name returns nothing.

One incidental formatting change: the shorter crate name lets the `html_favicon_url` attribute fit back on one line, so rustfmt collapsed it.

The PR title and description are updated to match. Gates on the renamed crate: tests, clippy `-D warnings`, doc, both rustfmt configs, `doc2readme --check`, spellcheck and `cargo deny check all` clean; coverage 100 % with no uncovered lines; `cargo mutants` over the package reports no survivors. Rebased onto current main (dbbef151) and the gate re-run after the rebase.

crates/observed_utils/src/format_any_value.rs
```diff
@@ -0,0 +1,200 @@
...
+    // `write_list` / `write_map` below, stays measured - and each nested element
+    // re-enters this excluded dispatch.
+    #[cfg_attr(coverage_nightly, coverage(off))]
+    fn fmt_value(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self.0 {
+            AnyValue::Int(v) => write!(f, "{v}"),
+            AnyValue::Double(v) => write!(f, "{v}"),
+            AnyValue::String(v) => write!(f, "{v}"),
+            AnyValue::Boolean(v) => write!(f, "{v}"),
+            AnyValue::Bytes(v) => f.write_str(&const_hex::encode(v.as_slice())),
```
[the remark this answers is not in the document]
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Correct — `const_hex::encode` returns an owned `String` (the crate enables const-hex's `alloc` feature for exactly this call), so the sentence was wrong for the `Bytes` arm. That sentence was mine, added a couple of pushes ago; fixed in 8c4ed0b9. The doc now says every arm writes straight to the formatter except `Bytes`, which builds one temporary hex string.

I fixed the claim rather than the allocation. Writing hex directly into the formatter would change the behaviour of code this PR only moves from another repository, so it belongs in a follow-up if the allocation ever matters — this crate formats telemetry values on an export path, not in a hot loop.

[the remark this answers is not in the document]
Vaiz: actually, it's a small change, do it in this PR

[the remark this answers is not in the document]
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Done in 74417e94. The `Bytes` arm now uses `const_hex::display`, which returns a `Display` adapter that writes through const-hex's own encoder straight into the formatter — so the fast encoding is kept and the temporary `String` is gone. Every arm now writes directly, and the module doc says plainly that rendering allocates nothing.

It turned out slightly better than a like-for-like swap: `display` is not gated behind const-hex's `alloc` feature and `encode` was the only thing using it, so the feature is dropped. const-hex enables `proptest` through `alloc`, and nothing else in the workspace pulled it in, so `proptest`, `rand_xorshift` and `unarray` drop out of `Cargo.lock` as well.

Gates: tests, clippy `-D warnings`, doc, both rustfmt configs, `doc2readme --check`, spellcheck, cargo-sort, cargo-machete and `cargo deny check all` clean; coverage still 100 % with no uncovered lines; `cargo mutants` over the file reports no survivors. Rebased onto current main (2f653cc5) and the gate re-run after the rebase.

crates/observed_utils/src/lib.rs
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#![cfg_attr(coverage_nightly, feature(coverage_attribute))]
+#![doc(html_logo_url = "https://media.githubusercontent.com/media/microsoft/oxidizer/refs/heads/main/crates/observed_utils/logo.png")]
+#![doc(html_favicon_url = "https://media.githubusercontent.com/media/microsoft/oxidizer/refs/heads/main/crates/observed_utils/favicon.ico")]
+// TODO(doc-coverage): remove once `missing_docs` is promoted to [workspace.lints.rust].
+#![deny(missing_docs)]
+
+//! Helper types and utilities that expand `observed` functionality.
```
martintmk, line 10: Are these general-purpose helpers or opentelemetry related ones?

If latter, it would make sense to rename this (again) to observed_opentelemetry

crates/observed_utils/src/lib.rs
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Microsoft Corporation.
```
martin-kolinek, line 1: Is there a document or diagram somewhere that explains the intended end state of crates and dependencies in the observed space? Who is supposed to depend on observed_utils? I would really like to prevent having applications have to depend on something called "observed_utils" - for me, a utils crate is something that should generally be buried in the dependency tree.

martin-kolinek (approved):

crates/observed_utils/src/lib.rs
Vaiz: generally speaking, I put here all common functionality that often is required by observed users, but doesn't have place in `observed` crate itself due to increased risk of breaking changes

crates/observed_utils/src/lib.rs
Vaiz: it's not limited to observed-otel adapters

crates/observed_utils/src/lib.rs
Vaiz: I expected that certain percent of crates using observed, also use observed_util instead of reimplementing similar functionality

crates/observed_utils/src/otel.rs
```diff
@@ -0,0 +1,222 @@
...
+/// either.
+fn unsigned_otel_value(value: u64) -> opentelemetry::Value {
+    i64::try_from(value).map_or_else(
+        |_| opentelemetry::Value::String(value.to_string().into()),
+        opentelemetry::Value::I64,
+    )
+}
+
+/// Converts a [`Value`] into an `OTel` [`AnyValue`], as used for log record
+/// attributes and bodies.
```
sandersaares, line 45: [Copilot speaking]

# Reconcile unsigned conversion documentation with the value-dependent output type

`any_value_of` converts to OpenTelemetry's log-value type (`AnyValue`), and `otel_value_of` converts to its attribute-value type (`opentelemetry::Value`). Their public documentation calls the result the source value's OpenTelemetry counterpart, but a `Value::U64` does not have one stable output type: values through `i64::MAX` become signed integers, while larger values become decimal strings. That boundary policy is documented only on private helpers.

**Reproducible reasoning**: `observed::Value::U64` preserves the complete unsigned value. Each public conversion matches that variant and delegates to `unsigned_any_value` or `unsigned_otel_value`; those helpers use `i64::try_from`, select the signed-integer variant on success, and select the string variant on failure because neither destination enum has an unsigned-integer variant. The public function pages and crate overview omit that chain, so the documented 'counterpart' description does not tell callers that the emitted schema changes when the same field crosses the signed-integer boundary.

**Consequence**: The same field changes from an integer-valued OpenTelemetry attribute to a string-valued attribute after crossing `i64::MAX`. Callers matching destination variants must handle two schemas, and numeric processing no longer applies to the larger observations.

**Recommended action**: Reconcile the public contract and implementation without assuming which direction is intended. Either describe the threshold and decimal-string fallback on both public conversion functions and in the crate overview, or change the conversion policy so one documented representation applies across the supported `u64` range. Keep the generated README and boundary tests consistent with the chosen contract.

**References**:

- [OpenTelemetry AnyValue variants](https://docs.rs/opentelemetry/0.32.0/opentelemetry/logs/enum.AnyValue.html)
- [OpenTelemetry attribute Value variants](https://docs.rs/opentelemetry/0.32.0/opentelemetry/enum.Value.html)

**Impacted locations**:

- `crates/observed_utils/src/otel.rs:52-53`
- `crates/observed_utils/src/otel.rs:83-84`
- `crates/observed_utils/src/lib.rs:18-20`
- `crates/observed_utils/README.md:24-26`

crates/observed_utils/src/sensitive_slice.rs
```diff
@@ -0,0 +1,275 @@
...
+/// - **Zero heap allocations** — items are stored inline in a fixed-size array.
+/// - Only the first `N` items are stored; whether the source had more items
+///   is tracked to detect overflow.
+/// - Virtual dispatch (`dyn RedactedDisplay`) per item during formatting,
+///   bounded by `N`.
+///
+/// # Type parameters
+///
+/// - `N` — maximum number of items to render (default: `5`).
+/// - `D` — single-character delimiter between items (default: `','`).
```
sandersaares, line 46: [Copilot speaking]

# Reconcile the documented SensitiveSlice capacity domain with the zero-capacity rejection

The public type-parameter documentation presents `N` only as the maximum number of stored items, but `SensitiveSlice::new` contains a compile-time assertion that rejects `N == 0`; the formatter repeats the same assertion. A zero-capacity specialization therefore falls inside the documented wording but outside the implemented domain.

**Reproducible reasoning**: `ArrayVec<_, 0>` is a valid storage type, so the lower bound does not follow from the field declaration. Construction reaches `const { assert!(N > 0, ...) }` before collecting the iterator, which makes `SensitiveSlice::<0>::new(...)` fail during compilation. Because the fields are private, `new` is the supported construction path; the repeated formatting assertion does not make zero capacity usable through another public path. The API documentation currently states the upper-bound role and default but no lower bound.

**Consequence**: Generic code can select zero to represent a disabled or empty collection and encounter a compile-time failure at the constructor rather than learning from the API contract that the specialization is unsupported.

**Recommended action**: Reconcile the documented and implemented capacity domains without prescribing which is authoritative. Either state the nonzero precondition in the type-parameter and constructor documentation, or make the implementation support zero capacity with defined empty/overflow behavior. Add a boundary test or compile-time check that pins the chosen contract.

**Impacted locations**:

- `crates/observed_utils/src/sensitive_slice.rs:35-38`
- `crates/observed_utils/src/sensitive_slice.rs:75-86`

crates/observed_utils/src/sensitive_slice.rs
```diff
@@ -0,0 +1,275 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! A bounded, type-erased view over a collection of classified items.
+//!
+//! [`SensitiveSlice`] borrows at most `N` items from any iterator of
+//! [`RedactedDisplay`] references and stores them inline, so a struct field can
+//! hold a redactable collection without naming the iterator's type and without
+//! allocating. Formatting goes through the caller's [`Redactor`], so the items
+//! themselves are never rendered in the clear by this module.
```
sandersaares: [Copilot speaking]

# Reconcile SensitiveSlice's confidentiality and allocation guarantees with delegated formatting

`SensitiveSlice` stores its collection of references inline, but it does not control how each referenced item is rendered. `RedactedDisplay` is the item-formatting trait, and `Redactor` is the caller-supplied strategy passed to it. The crate and module documentation nevertheless describe rendering as allocation-free and say items are never rendered in the clear, while the implementation accepts arbitrary `RedactedDisplay` implementations and redactors that can pass values through.

**Reproducible reasoning**: Construction stores trait-object references in `ArrayVec`, so the wrapper performs no collection allocation. During `RedactedDisplay::fmt`, however, every item is handed to its own `RedactedDisplay::fmt` together with the caller's `Redactor`. The redactor contract explicitly permits pass-through or suppressed redaction, and the included example configures suppression and prints clear values. An item implementation can also allocate before writing to the formatter. The inline-storage property therefore composes with behavior outside the wrapper to produce output and costs that the absolute public wording does not capture.

**Consequence**: A consumer can treat `SensitiveSlice` as a confidentiality boundary or an end-to-end no-allocation formatter even though clear output and item-local allocations are both possible under supported inputs. That can invalidate privacy reviews or hot-path allocation assumptions.

**Recommended action**: Reconcile the public guarantees and supported behavior without choosing the direction for the author. Either scope the promises to wrapper-controlled inline storage and delegated rendering, or constrain the accepted item/redactor behavior so the absolute confidentiality and allocation guarantees hold. Make the custom-delimiter example identify the redactor assumption behind any literal clear-text output, and keep the crate, module, and type documentation consistent.

**References**:

- crates/data_privacy_core/src/redacted.rs defines delegated RedactedDisplay formatting
- crates/data_privacy_core/src/redactor.rs permits pass-through and suppressed redaction
- crates/observed_utils/examples/sensitive_collection.rs demonstrates suppressed redaction

**Impacted locations**:

- `crates/observed_utils/src/sensitive_slice.rs:6-10`
- `crates/observed_utils/src/sensitive_slice.rs:27-33`
- `crates/observed_utils/src/sensitive_slice.rs:55-67`
- `crates/observed_utils/src/lib.rs:24-25`

crates/observed_utils/src/metric_number.rs
sandersaares, line 1: [Copilot speaking]

# Preserve metric numeric kinds until the destination selects an instrument

`metric_number_of` maps `Value::I64`, `Value::U64`, and `Value::F64` to `Option<f64>`. This erases signedness and can lose integer precision before a metric processor considers the field's `InstrumentKind`, even though the source model and OpenTelemetry both provide typed integer and floating-point paths.

**Reproducible reasoning**: `observed::Value` stores `I64`, `U64`, and `F64` separately and documents the integer conversions as exact. The event macro reinforces that distinction by accepting unsigned integers for counters and signed integers for up-down counters, while gauges and histograms can use any supported numeric kind. The repository's destination sketch then places `metric_number_of` before instrument recording, and the in-repository test processor likewise converts into `CapturedFieldMetric { value: f64 }`. OpenTelemetry 0.32 exposes `u64`, `i64`, and `f64` instruments. Because an `f64` has 53 bits of integer precision, `9_007_199_254_740_992_u64` and the next integer produce the same extracted value; no later instrument dispatch can recover the original number or signedness.

**Consequence**: A large counter or gauge can silently collapse distinct observations, and a destination that supports exact integer instruments cannot recover the source value. The same normalization also prevents tests using `CapturedFieldMetric` from detecting one-count changes above the exact `f64` range.

**Recommended action**: Avoid exposing `f64` normalization as the general exporter boundary. The simplest option is to keep that conversion local to `observed_testing`, where the capture type explicitly stores `f64`. If exporters need a shared helper, return a small `I64`/`U64`/`F64` representation (or equivalent accessors), then match it together with `InstrumentKind` and convert only when the selected destination instrument requires `f64`. Add boundary coverage with adjacent `u64` values above the exact `f64` range.

**References**:

- crates/observed/src/value.rs documents exact integer storage and exporter-visible signedness
- crates/observed_macros/src/event.rs enforces metric-kind signedness
- crates/observed/FEATURES.md shows metric_number_of before destination recording
- [OpenTelemetry 0.32 Meter exposes typed u64, i64, and f64 instruments](https://docs.rs/opentelemetry/0.32.0/opentelemetry/metrics/struct.Meter.html)

**Impacted locations**:

- `crates/observed_utils/src/metric_number.rs:15-27`
- `crates/observed/src/value.rs:32-38`
- `crates/observed/src/value.rs:585-593`
- `crates/observed_testing/src/mock_processor.rs:18-21`

sandersaares (commented):
[Copilot speaking]

Published 4 findings. No finding follows up on an existing discussion thread.

| Diagnostic | Value |
|------------|-------|
| Cache | Miss |

</details>

sandersaares (approved):

crates/observed_utils/src/lib.rs
sandersaares: If this is "ordinary" functionality, it sounds like we should stabilize it, so it could go in main crate. For "ultra-stable" things, maybe the "re-export critical types from observed_core that you never touch" pattern helps avoid accidental breakage? Just speaking in generalizations, I guess it is not something that has a universal pattern we have worked out so far but in general if we expect most users to use a `_utils` crate it stops having value on its own IMO and might as well get merged into the top-level crate. But probably any detailed analysis needs to be case by case - I know little about observed structure and what is needed by which users. Maybe opportunity for design discussion at future arch meeting?

martintmk (approved):

crates/observed_utils/src/lib.rs
Vaiz: I would call it `experimental`. Let's see what people actually use, and where to put it. Right now it solves the problem of reinventing the wheel, and it's good enough for me.

for example, I also plan to add `info!` shortcut here that defines event struct inline, so you don't need to declare a separate struct for simple events.

crates/observed_utils/src/otel.rs
Vaiz: Accepted, as a documentation fix — in `ec533c65`.

You are right that the boundary was documented only on `unsigned_any_value` / `unsigned_otel_value`, which are private and therefore never reach rustdoc. Both public functions and the crate overview now state it: a `Value::U64` up to `i64::MAX` becomes the integer variant, past it a decimal string.

The policy itself stays as it is. `AnyValue` and `opentelemetry::Value` have no unsigned variant, so the alternatives are wrapping into a negative number, which is silently wrong, or dropping the value. A decimal string is the only lossless option, and it is the reason the variant is handled explicitly at all: before this move a `u64` fell through the `#[non_exhaustive]` guard arm and was exported as a `Debug` string.

crates/observed_utils/src/sensitive_slice.rs
Vaiz: Accepted, as a documentation fix — in `ec533c65`.

The `const { assert!(N > 0, ...) }` was the only statement of the lower bound, and it is not visible from the API docs. `N` is now documented as "must be greater than zero: `SensitiveSlice::<0>` is rejected at compile time", and `new` says the same.

The implementation keeps rejecting zero rather than growing an empty specialization: a zero-capacity redacted collection renders nothing while looking like it renders something, which is the failure mode this type exists to avoid.

crates/observed_utils/src/sensitive_slice.rs
Vaiz: Accepted, as a documentation fix — in `ec533c65`.

Both claims were absolute where the behaviour is delegated. Items are rendered by their own `RedactedDisplay` implementation, driven by the caller's `Redactor`, and that redactor is permitted to pass values through — the example in this crate configures exactly that. An item implementation may also allocate before writing.

The wording now scopes both promises to what the wrapper actually controls: the collection is stored inline and costs no allocation of its own, and rendering delegates, so what reaches the output — redacted, suppressed, or in the clear — is decided by the redactor and the items, not by this module.

The accepted item and redactor types are unchanged. Constraining them would make this a different type, and the pass-through redactor is a supported configuration of `data_privacy`, not an accident.

crates/observed_utils/src/metric_number.rs
Vaiz: Declined for this PR.

The analysis of the numerics is right — `f64` carries 53 bits, so adjacent `u64` values above 2^53 do collapse, and the existing `#[expect(clippy::cast_precision_loss, reason = "metric recording precision loss is acceptable")]` says as much at both call sites. What is being asked for, though, is an API change to a function this PR only moves.

Two specific reasons:

- Changing the return type to an `I64`/`U64`/`F64` representation, matched against `InstrumentKind` at the destination, is a redesign of the helper API. This PR's scope is the move plus what the move itself forces: workspace conventions, this repository's stricter lints, the `Value::U64` gap in the OpenTelemetry conversions, and the de-duplication. New numeric contracts are follow-up material.
- Keeping it in `observed_testing` is the arrangement this PR deliberately undoes. It sat there only because the mock processor was its single consumer at the time; leaving it there while an exporter helper crate exists means the next exporter writes a second copy that can drift. `observed_testing` now depends on it rather than owning it.

Worth filing as a follow-up against `metric_number_of` itself, and it would want the adjacent-`u64` boundary test suggested here. It should not land as a side effect of relocating the crate.

Vaiz:

Re the two suppressed comments in [review #4971552344](https://github.com/microsoft/oxidizer/pull/677#pullrequestreview-4971552344), both about `SensitiveSlice`'s undocumented `Sync` bound — accepted, fixed in `8db85fb3`.

Both are correct and they are the same gap seen from two angles. The type documentation said it stores `&dyn RedactedDisplay` fat pointers, and `new` said it accepts implementers of `RedactedDisplay`, but the stored type is `&(dyn RedactedDisplay + Sync)` and the constructor requires `T: RedactedDisplay + Sync`. A caller with a non-`Sync` item type therefore satisfied the documented bound and still failed to compile.

Documented on both the type and the constructor rather than removing the bound. The bound is load-bearing: `&(dyn Trait + Sync)` is itself `Send` and `Sync`, while a bare `&dyn Trait` is neither, so it is what allows an event struct holding a `SensitiveSlice` to cross threads. Dropping it would narrow where the type can be used, which is a behaviour change to code this PR only relocates.

Documentation only — no behaviour change, and the generated README is unaffected since these docs live on the type rather than in `lib.rs`.

martin-kolinek (approved):

crates/observed_utils/Cargo.toml
```diff
@@ -0,0 +1,48 @@
...
+[package.metadata.cargo_check_external_types]
+allowed_external_types = [
+    "data_privacy_core::redacted::RedactedDisplay",
+    "data_privacy_core::redactor::Redactor",
+    "observed::severity::Severity",
+    "observed::value::Value",
+    "opentelemetry::common::Value",
+    "opentelemetry::logs::record::AnyValue",
+    "opentelemetry::logs::record::Severity",
+]
```
ralfbiedert, line 29: 🤖 This crate combines independently usable OpenTelemetry conversion, metric extraction, formatting, and privacy collection APIs. A `SensitiveSlice` consumer therefore compiles the OpenTelemetry dependencies, while conversion users compile the privacy stack. Please split the independently usable surfaces (for example, a focused OpenTelemetry crate and a privacy collection crate) or otherwise avoid forcing unrelated dependency sets.

ralfbiedert (changes requested):
🤖 Requesting changes for the lossy integer-to-float metric conversion. The crate-boundary concern is secondary.

crates/observed_utils/src/metric_number.rs
```diff
@@ -29,13 +29,15 @@ pub fn metric_number_of(value: &Value) -> Option<f64> {
...
 
+#[cfg_attr(coverage_nightly, coverage(off))]
 #[cfg(test)]
 mod tests {
     use super::*;
 
     #[test]
     fn numeric_values_convert() {
         assert_eq!(metric_number_of(&Value::from(42_i64)), Some(42.0));
+        assert_eq!(metric_number_of(&Value::from(7_u64)), Some(7.0));
```
ralfbiedert, line 40: 🤖 The implementation still converts `i64` and `u64` to `f64`; this small-value test misses the boundary where adjacent counters above 2^53 collapse to the same observation. Please preserve the numeric kind until instrument selection and add boundary tests around 2^53.

Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

This point was already raised and answered on this PR — see https://github.com/microsoft/oxidizer/pull/677#discussion_r3812035654 for the reasoning (the numerics are correct and acknowledged by the `#[expect(clippy::cast_precision_loss, ...)]` at both arms; changing the return type is a redesign of a helper this PR only relocates, so it is follow-up material along with the 2^53 boundary test). Flagging it to the PR owner rather than re-deciding it here.

martin-kolinek (approved):

crates/observed_utils/src/metric_number.rs
Vaiz: I'll file it as a follow up. Current goal is to have this conversion in one place instead of tens of different implementations of the same conversions by our users

crates/observed_utils/Cargo.toml
Vaiz: we can do the split later, and dependency size is not a concern for this specific crate

crates/observed/examples/support/otel.rs
```diff
@@ -37,10 +37,10 @@ const CODE_NAMESPACE: &str = "code.namespace";
 /// Converts a [`Text`] into an `OTel` string, preserving the borrowed-versus-
 /// shared distinction so neither representation copies.
 ///
-/// `observed_helpers` offers this conversion (and the two below) ready-made, but
+/// `observed_utils` offers this conversion (and the two below) ready-made, but
 /// it depends on `observed`, so using it here would make `observed`'s dependency
 /// graph cyclic. A consumer outside this crate should call
-/// `observed_helpers::any_value_of` rather than copy this.
+/// `observed_utils::any_value_of` rather than copy this.
```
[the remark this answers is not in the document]
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Accepted — fixed in `daf1b59d`. You are right that this PR created the divergence: the doc immediately above this helper is edited by this PR to point readers at `observed_utils::any_value_of`, so the two have to agree, and after the saturation change they did not. The copy now does `AnyValue::Int(i64::try_from(v).unwrap_or(i64::MAX))` with a comment naming the utility it has to match.

To be explicit about why this file is touched at all, given the PR is otherwise scoped away from `observed`: it is an examples-support helper this PR already edits, not the `observed` library, and the fix is keeping the PR's own change self-consistent rather than new work.

Vaiz:

Answering the two suppressed comments from review #5006128223, since a review body is not a thread.

**`sensitive_slice.rs:130` — redaction path never exercised: ACCEPTED, fixed in `daf1b59d`.**
This was a real gap and a well-aimed one. Every assertion used `passthrough_engine()`, so the behaviour the type exists for — forwarding the caller's `Redactor` to each item — was pinned by nothing, and `cargo-mutants` would not have caught it either, since it mutates bodies and return values rather than the argument being passed through. Added `items_render_through_the_supplied_redactor`, which renders the same slice under a rewriting engine (`SimpleRedactorMode::Insert`) and under the pass-through engine and asserts the two differ (`"X, X"` vs `"alpha, beta"`). The delimiter still comes from the slice, so the test separates the slice's own formatting from the delegated rendering.

**`format_any_value.rs:181` — `.expect(...)` should be `.unwrap()`: DECLINED, the premise does not hold.**
The stated reason is that "the repository convention is to use `.unwrap()` rather than `.expect(...)`" in test code. Measured across `crates/*/src/**`, this repository has **3430** `.unwrap(` and **1809** `.expect(` occurrences, the latter across 1422 distinct lines — `.expect` is not against convention here, it is used pervasively. There is also no lint enforcing it: `cargo clippy --all-targets -D warnings` is clean on this crate either way.

On the merits, this particular call is one where the message earns its keep. The assertion is `strip_prefix('{').and_then(strip_suffix('}'))`, and `"map renders inside braces"` records *which* of the two strips failed in intent terms. Replacing it with `.unwrap()` would trade that for a bare `called Option::unwrap() on a None value`, which is strictly less information at the moment the test fails. Happy to revisit if a maintainer points at an actual written convention I have missed.

martin-kolinek (approved):

Vaiz:

Answering the two suppressed comments from review #5006279993.

**`examples/sensitive_collection.rs:93` — "from a `Vec`" is wrong: ACCEPTED, fixed in `697e4815`.**
Good catch. `recipients` is built from `let emails = [...]`, an array. It matters more than a stray word here, because the example exists to show `SensitiveSlice` accepting several different iterator sources — the neighbouring fields say `HashMap`'s keys and `VecDeque` — so mislabelling this one blunted the point the example is making. Now reads "from an array".

**`format_any_value.rs:181` — `.expect(...)` → `.unwrap()`: this is a repeat**, and your own note flags it as "previously missed... in code that hasn't changed since the last review". It was answered ~15 minutes earlier in https://github.com/microsoft/oxidizer/pull/677#issuecomment-5393067314, with the measurements: this repository's `crates/*/src/**` holds 3430 `.unwrap(` against 1809 `.expect(`, and clippy `--all-targets -D warnings` is clean either way, so the stated convention does not exist here. Declined again, on the same evidence and without re-arguing it further.

martin-kolinek (approved):

Vaiz:

On the `sensitive_slice.rs:97` suppressed comment from review #5006491374 — relaxing `T` to `?Sized`.

The technical observation is correct. `new` is declared `pub fn new<I, T>(iter: I) where I: IntoIterator<Item = &'a T>, T: RedactedDisplay + Sync + 'a`, and the implicit `Sized` on `T` does mean an iterator whose items are already `&(dyn RedactedDisplay + Sync)` will not bind, even though that is exactly what the collection stores.

Declining it here anyway, on scope rather than on correctness. This PR relocates the crate from an internal repository essentially unchanged, and the one API-shaped change in it was an explicit owner ruling. Widening a bound on a public constructor is a deliberate API decision for the crate's owner to make, not something to slip in from a review of a move — particularly now, with the PR otherwise complete and carrying two approvals.

It is also cheap to defer: relaxing `Sized` is backwards compatible, so it can land later without breaking any existing caller, and it would want a test covering the heterogeneous trait-object case you describe rather than just the bound change. Good follow-up material, and I have noted it as such.

ralfbiedert (approved):

PR #679, opened 2026-08-18
fix(multitude): compile the test suite under the UTC codegen backend

wukchung:
&#x1F916; This description was written by an AI agent.

## Problem

`cargo test -p multitude` fails on the internal `ms-prod` toolchains, whose UTC
codegen backend caps type alignment at 8192 bytes:

```text
error: invalid alignment for type `audit_repro::...::Wide`
       (32768 bytes), UTC supports at most 8192B power-of-two alignment
error: could not compile `multitude` (test "audit_repro") due to 1 previous error
```

The mitigation for this was already in the tree — and inert. Five test files
carry 41 `#[cfg(not(utc_backend))]` attributes, and the workspace declares
`'cfg(utc_backend)'` in its `check-cfg` list. But nothing anywhere set the cfg:

```text
$ git grep -n utc_backend -- . ":(exclude)crates/multitude/tests"
Cargo.toml:287:    'cfg(utc_backend)',
```

`multitude` had no build script, and no `.cargo/config.toml`, justfile recipe or
CI variable passes `--cfg utc_backend`. Every gate was therefore permanently
false and skipped nothing.

The `check-cfg` entry is what let this survive: without it, `unexpected_cfgs`
would have warned on every gate and made the dangling cfg obvious.

This is invisible to this repository's own CI, which never builds with a
`ms-prod` toolchain — so the failure only reaches developers using one locally.

## Change

Adds a build script to `multitude` that sets `utc_backend` when the codegen
backend cannot compile the alignment the gated tests need.

There is no built-in cfg identifying the codegen backend
([upstream issue](https://developercommunity.visualstudio.com/t/Conditional-compilation-lacks-built-in-c/11107823)),
so the script probes the capability rather than inferring it from a toolchain
name. Probing the limit that actually matters keeps the gate correct if a
backend raises its cap, or if another backend adopts the same restriction.

Three properties are load-bearing:

* **The probe must materialize a place of the over-aligned type.** UTC rejects
  the alignment when laying out a value, not when the type merely appears in a
  reference parameter — a slice-taking function compiles cleanly under UTC even
  at 128 KiB. The probe therefore returns the value by value
  (`pub fn probe() -> Probe { Probe }`). `Probe` stays a ZST so no over-aligned
  stack frame is required, and an over-aligned `static` is deliberately avoided
  because it trips the COFF section-alignment limit and crashes stock `rustc`,
  which would set the cfg on every Windows host.

* **A control probe separates "backend rejected it" from "could not run here".**
  A non-zero `rustc` exit is not by itself evidence of an alignment cap: a
  target whose `core` is not installed, a custom target spec, or an unusable
  sysroot all fail the same way, and misreading one as UTC would silently
  disable the gated tests on a healthy toolchain. A control probe at `align(8)`
  runs first, and the cfg is set only when the control compiles and the
  over-aligned probe does not. Failure to spawn `rustc` at all also leaves the
  tests enabled.

* **`PROBE_ALIGN` is the largest gated alignment (131072).** Probing a smaller
  value would leave the cfg unset on a backend that accepts the probe but still
  rejects a larger gated type, such as `HugeAlign` or `HugeAlignBox`.

`/build.rs` is already in the `[workspace.package]` `include` allowlist, so the
script ships with the published crate.

## Validation

Against an internal toolchain whose codegen backend is UTC:

| | without this PR | with this PR |
| --- | --- | --- |
| `cargo +ms-prod test -p multitude --no-run` | exit 101, `invalid alignment` | **exit 0** |
| build script output under `ms-prod` | no cfg emitted | `cargo::rustc-cfg=utc_backend` |

Fail-open behaviour, exercised by running the build script directly against each
environment:

| environment | cfg emitted |
| --- | --- |
| stock `rustc` | no |
| `ms-prod` (UTC) | **yes** |
| `RUSTC` pointing at a missing binary | no |
| `TARGET` whose `core` is not installed | no |

Stock toolchains are unaffected:

```text
cargo test -p multitude          973 tests, all suites pass
just package=multitude clippy    clean, no warnings
just package=multitude format    no changes
```

The cfg gates 19 tests (973 without it, 954 with it). The remaining
`cfg(not(utc_backend))` attributes sit on the over-aligned type declarations
those tests use.

## Reviewer notes

`PROBE_ALIGN` must stay in sync with the largest `#[repr(align(...))]` guarded by
`cfg(utc_backend)`; it is documented as such in the build script.

AB#7707893


geeknoid (approved):

crates/multitude/build.rs
```diff
@@ -0,0 +1,86 @@
...
+    // hits the COFF section-alignment limit and crashes stock `rustc`, which
+    // would make the probe report a false positive on every Windows host.
+    // No attributes beyond `repr`/`inline` are used, so the source stays valid
+    // regardless of which edition `rustc` defaults to.
+    let source = format!(
+        "#![no_std]\n\
+         #[repr(align({PROBE_ALIGN}))]\n\
+         pub struct Probe;\n\
+         #[inline(never)]\n\
+         pub fn probe(values: &[Probe]) -> usize {{ values.len() }}\n"
```
wukchung: &#x1F916; **Critical — the probe never fires under UTC, so this fix is inert on the toolchain it targets.**

An `ms-prod` toolchain was available to me (`rustc 1.95.0-ms-20260618.5`, `codegen-utc-hash` / `x86_64-utc-builder-path: amd64\r2c2.dll`), so I ran the repro the PR notes say could not be executed.

Compiling this `build.rs` verbatim and running it with `RUSTC` pointed at `ms-prod` and `TARGET=x86_64-pc-windows-msvc` emits:

```
cargo::rerun-if-changed=build.rs
cargo::rerun-if-env-changed=RUSTC
cargo::rerun-if-env-changed=RUSTUP_TOOLCHAIN
```

No `cargo::rustc-cfg=utc_backend`. The gate never sets, so `cargo test -p multitude` still fails exactly as before.

**Root cause:** UTC rejects the alignment when a *place* of the over-aligned type is materialized, not when the type merely appears in a reference parameter. Measured against `ms-prod` at `align(32768)` (exit codes):

| probe shape | stock rustc | ms-prod (UTC) |
| --- | --- | --- |
| `fn f(v: &[P]) -> usize { v.len() }` — **current probe** | 0 | **0** |
| same, but `P(u8)` sized rather than a ZST | 0 | **0** |
| `fn f() -> P { P }` | 0 | **1** |
| `let a: [P; 1] = [P]; (&a[..]).len()` | 0 | **1** |

Making `Probe` sized does not help — materialization is what matters, not size.

The suggestion below returns the value instead of taking a slice. I verified it is accepted by stock rustc and rejected by `ms-prod` at both `32768` and `131072`, and it keeps `Probe` a ZST so no 32 KiB stack frame is required and the COFF `static` trap is still avoided.

Note the comment above at lines 50-55 explains the slice-taking choice, and would need rewording — the COFF concern it describes is real, but returning a ZST by value avoids it too.

Also worth correcting in the PR description: the `probe rejected (1 GiB, stock rustc)` validation row trips rustc's front-end `repr(align)` parser limit (2^29), which is a different code path from the backend rejection being detected. It does not demonstrate that the probe detects a codegen cap.

```suggestion
         pub fn probe() -> Probe {{ Probe }}\n"
```

crates/multitude/build.rs
```diff
@@ -0,0 +1,86 @@
...
+        .stdout(Stdio::null())
+        .stderr(Stdio::null());
+
+    // Build scripts run on the host, but the cfg applies to the target being
+    // compiled, so the probe has to target it too.
+    if let Some(target) = env::var_os("TARGET") {
+        command.arg("--target").arg(target);
+    }
+
+    command.status().is_ok_and(|status| status.success())
```
wukchung: &#x1F916; **The probe fails closed, which is the opposite of the documented contract, and silently deletes 64 tests.**

The doc comment at lines 39-40 states:

> Returns `true` when the probe cannot be run at all, so an unexpected environment leaves the tests enabled rather than silently skipping them.

`is_ok_and` does not do that. Two distinct problems:

1. **Spawn failure returns `false`.** `Result::is_ok_and` yields `false` on `Err`, and `Command::status()` returns `Err` precisely when the binary cannot be run at all — the case the doc singles out. Confirmed with a repro: `status() = Err(Error { kind: NotFound })` → `is_ok_and(|s| s.success()) = false` → `utc_backend` is set. The other two fallbacks (lines 43 and 65) correctly return `true`, so this one is inconsistent with them.

2. **Any non-zero exit is attributed to the alignment cap.** `#![no_std]` still needs `core` for the requested target, so a target that is not installed, a custom JSON target spec (`RUST_TARGET_PATH` is not forwarded), or a `-Zbuild-std` flow all exit 1 with `error[E0463]: can't find crate for core` — verified locally. With stderr at `Stdio::null()` that is indistinguishable from a UTC rejection, and the result is a green build with 64 tests silently gone and no diagnostic.

The suggestion below fixes (1) only, to match the stated contract. For (2), the robust shape is a control probe: compile the identical crate at a benign alignment (e.g. `align(8)`) with the same flags first, and only trust a negative result when the control succeeds. Emitting `cargo::warning=` on the skip path would also make a misfire visible rather than silent.

```suggestion
    match command.status() {
        Ok(status) => status.success(),
        // The probe could not be run at all, so leave the tests enabled.
        Err(_) => true,
    }
```

crates/multitude/build.rs
```diff
@@ -0,0 +1,86 @@
...
+
+use std::path::Path;
+use std::process::{Command, Stdio};
+use std::{env, fs};
+
+/// Alignment the probe requests, in bytes.
+///
+/// This is the largest alignment any gated test relies on. It must stay in sync
+/// with the `#[repr(align(...))]` values guarded by `cfg(utc_backend)`.
+const PROBE_ALIGN: u32 = 32768;
```
wukchung: &#x1F916; **`PROBE_ALIGN` is not the largest gated alignment, so the documented invariant is already false.**

The doc says "This is the largest alignment any gated test relies on", but larger values sit directly under `#[cfg(not(utc_backend))]`:

- `crates/multitude/tests/arena.rs:2253-2254` — `#[repr(align(65536))] struct ChunkAlign;`
- `crates/multitude/tests/arena.rs:6068-6069` — `#[repr(align(131072))] struct HugeAlign(u8);`
- `crates/multitude/tests/arena.rs:6322-6323` — `#[repr(align(131072))] struct HugeAlignBox(u8);`
- `crates/multitude/tests/arena.rs:6775` and `6798` — further `#[repr(align(131072))]` under the same gate

This does not bite today only because UTC's cap (8192) is below 32768. But the module doc at lines 13-15 claims probing "keeps the gate correct if a backend raises its cap" — it does not. Any backend capping in `[32768, 131072)`, including a future UTC raised to 64 KiB, would pass the probe, leave the cfg unset, and fail to compile `ChunkAlign` / `HugeAlign` with the very error this PR exists to prevent.

I verified stock rustc accepts the probe pattern up to at least `262144`, and that `131072` is rejected by `ms-prod`, so raising the constant costs nothing in either direction.

```suggestion
const PROBE_ALIGN: u32 = 131072;
```

wukchung (commented):
&#x1F916; This review was produced by an AI agent.

Reviewed by fanning out to five independent models, then verifying every claim against the code in a worktree. Nine raw findings came back; three survived verification and are posted below. Refuted claims (the probe needing a sized rather than zero-sized type, a missing `cargo::rustc-check-cfg`, `/build.rs` missing from the `include` allowlist, and `RUSTFLAGS` needing to be forwarded) are not included.

One thing worth flagging up front: an `ms-prod` toolchain **is** installed on this machine, so the repro the PR notes describe as unverifiable was executed. Finding 1 is the result, and it blocks the PR — the build script does not set the cfg under a real UTC toolchain, so `cargo test -p multitude` still fails there.

The internal toolchain has UTC baked into that `rustc` binary rather than selected via `RUSTFLAGS`. That confirms the build script's toolchain-based assumption is sound, and refutes the suggestion that `CARGO_ENCODED_RUSTFLAGS` needs forwarding.

crates/multitude/build.rs
wukchung: &#x1F916; Follow-up: I have now proven this end-to-end at the `cargo test` level, and there is good news alongside the bug.

**This PR as-is, under `ms-prod`** (build script left to decide):

```
cargo +ms-prod test -p multitude --no-run
error: invalid alignment for type
  `core::mem::maybe_uninit::MaybeUninit<pin_support::try_alloc_uninit_box_pin_rejects_over_alignment::HalfChunk>`
  (32768 bytes), UTC supports at most 8192B power-of-two alignment
error: could not compile `multitude` (test "pin_support") due to 1 previous error
exit 101
```

**The same tree with the cfg forced on** (`RUSTFLAGS="--cfg utc_backend"`), simulating a probe that fires correctly:

```
cargo +ms-prod test -p multitude --no-run
    Finished `test` profile [unoptimized + debuginfo] target(s)
exit 0
```

So the 64 existing `#[cfg(not(utc_backend))]` gates are **complete and correct** — nothing else in the crate needs gating, and the premise of the PR holds. The only defect is that the probe never sets the cfg. Fixing the probe shape is sufficient to make the whole thing work.

One claim I checked and am explicitly **not** raising, in case it comes up: `non_freezable_overaligned_vec_grows_via_oversized_path` at `tests/arena.rs:3893` declares an ungated `#[repr(align(32768))] struct Over(u8)`, which looks like it should break under UTC. It does not — `try_reserve` never materializes a place of `Over`, and the suite compiles cleanly with the cfg set, as above. No gate is needed there.

wukchung: &#x1F916; Fixed in 3fe43da0. The probe now returns the value rather than taking a slice:

```rust
pub fn probe() -> Probe {{ Probe }}
```

Verified against the real `ms-prod` toolchain. `cargo +ms-prod test -p multitude --no-run` went from `exit 101` with `invalid alignment for type MaybeUninit<HalfChunk>` to `Finished ... exit 0`, and the build script now emits `cargo::rustc-cfg=utc_backend` under `ms-prod` while emitting nothing under stock rustc. Stock `cargo test -p multitude` still passes with the gated tests present.

crates/multitude/build.rs
wukchung: &#x1F916; Fixed in 3fe43da0. A control probe at `align(8)` now runs first, and the cfg is set only when the control compiles and the over-aligned probe does not. `compile_probe` returns `Option<bool>` so "could not run" is distinct from "rejected".

Both failure modes verified to fail open — neither now emits the cfg:

| scenario | before | after |
| --- | --- | --- |
| `RUSTC` pointing at a missing binary | cfg set | no cfg |
| `TARGET` whose `core` is not installed | cfg set | no cfg |
| stock rustc | no cfg | no cfg |
| `ms-prod` (UTC) | no cfg | **cfg set** |

crates/multitude/build.rs
wukchung: &#x1F916; Fixed in 3fe43da0 — raised to `131_072`, matching `HugeAlign` / `HugeAlignBox`. Verified stock rustc still accepts the probe at that alignment (so no false positive) and `ms-prod` rejects it.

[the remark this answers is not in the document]
wukchung: &#x1F916; Already fixed in 3fe43da0 — `PROBE_ALIGN` raised to `131_072` to match `HugeAlign` / `HugeAlignBox`. Verified stock rustc still accepts the probe at that alignment, so there is no false positive.

crates/multitude/build.rs
```diff
@@ -0,0 +1,86 @@
...
+        .stderr(Stdio::null());
+
+    // Build scripts run on the host, but the cfg applies to the target being
+    // compiled, so the probe has to target it too.
+    if let Some(target) = env::var_os("TARGET") {
+        command.arg("--target").arg(target);
+    }
+
+    command.status().is_ok_and(|status| status.success())
+}
```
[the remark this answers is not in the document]
wukchung: &#x1F916; Already fixed in 3fe43da0 — `compile_probe` now returns `Option<bool>`, so a spawn failure yields `None` and leaves the tests enabled. A control probe at `align(8)` additionally distinguishes "backend rejected the alignment" from "rustc could not run here", which covers the non-zero-exit case too (e.g. a target whose `core` is not installed).

wukchung:
&#x1F916; Closing this in favour of a simpler approach.

The gated tests reach the arena's alignment guards by declaring a Rust type at `#[repr(align(N))]`, which forces the codegen backend to lay out an over-aligned type. That type declaration is the only reason a backend alignment cap is ever involved.

The guards themselves take a `Layout`. Driving them through the layout-taking entry points reaches the same code with no over-aligned Rust type declared, so no backend detection is needed at all. That removes the build script, the `cfg`, and the gates rather than making the detection work.

Superseded by that change; no reviewer action needed here.

PR #683, opened 2026-08-20
refactor(ohno_macros): split the implementation into ohno_macros_impl

Vaiz:

## What this changes

`ohno_macros` becomes a thin proc-macro shim. A new crate, `ohno_macros_impl`, holds the implementation. This adopts the Microsoft Rust guideline M-MACRO-MAIN-CRATE. It matches the pattern already used by `fundle`, `data_privacy`, `internity`, `multitude`, `templated_uri` and `thread_aware`.

Replaces #682, which mixed this change with the `observed_macros` split.

## Details

- Add `ohno_macros_impl` 0.5.0. It holds every implementation module. It is wired through `[workspace.dependencies]` with a matching path and version. The version is deliberately 0.5.0 rather than 0.4.0, so the crate does not publish a version that nothing consumes; `ohno` and `ohno_macros` stay at 0.4.0. The root `Cargo.toml` requirement and the `Cargo.lock` entry both read 0.5.0 to match.
- Reduce `ohno_macros` to three entry points. Each converts `proc_macro::TokenStream` to `proc_macro2::TokenStream`, delegates, and converts back. Macro names, `attributes(...)` lists and rustdoc are unchanged.
- Keep every module in `ohno_macros_impl` private. The public API is exactly `derive_error`, `enrich_err` and `error`. No item is `pub` or `#[doc(hidden)]` for a test.
- Move the tests to `tests/public_api.rs`, which calls only those three entry points. Each test snapshots the expansion, pretty-printed with `prettyplease`, so a snapshot records the generated Rust. Faults are snapshotted the same way, which pins the diagnostic text.
- Group related inputs into one snapshot under `// === label ===` headers. 20 snapshot files replace the 50 the crate carried before.
- Remove the out-of-range guard in `derive_error::model::Shape::new`. `validate::select_core` only yields indices from `ast.fields`. `validate` maps that same list one-to-one into `fields`. `Shape::new` now returns `Self` and documents the invariant.
- Remove `derive_error::display::argument::Root::field_name`. `display::scope_to_self` matches the `Root` variants directly and never called it.
- Register the crate in the root CHANGELOG index, `.anvil.lock`, `scripts/mutants.rs` and the CI miri exclusion list.
- Give the shim the `coverage-gate` opt-out that the other proc-macro shims carry.

## Effects

- `ohno_macros_impl` is a new crate in the workspace. `ohno_macros` depends on it, which adds one crate boundary to the build.
- Macro behaviour does not change. The `ohno` crate, its 12 `trybuild` UI tests and its doctests are untouched and pass.
- `Shape::new` now panics when `core` does not index `fields`. It returned `None` before. No caller can supply such an index, so the panic is unreachable today.
- The suite is 20 tests, down from 144 unit tests. Line and function coverage of `ohno_macros_impl` is 100%.
- A grouped snapshot fails as a whole. A reviewer reads the insta diff to see which labelled case changed.
- A test can no longer reach a private item. A future test for an internal helper must go through a macro entry point, or the helper must be proved unreachable and removed.
- `tests/public_api.rs` is `#![cfg(not(miri))]`, because insta reads the workspace from disk. CI already excludes this crate from the miri job.
- The workspace carries two version numbers for the `ohno` family: `ohno_macros_impl` at 0.5.0, everything else at 0.4.0.
- `observed_macros` is not touched here. That half of the original change lands on `u/eshutov/split-observed-macros-impl`.

## Validation

- Coverage was measured with the internal tests disabled, to find code unreachable from the public API. The entry points alone cover 58.10% of lines. The full suite reaches 100.00% of lines and functions. The two removed items are what stayed uncovered after that suite stopped growing.


Vaiz:

Thanks — answering the two suppressed comments in the review body (`derive_error/ast.rs:28` and `derive_error/model.rs:25`), which make the same point: internal structs use `pub` fields, so the public surface could accidentally widen later.

Declining both, because the premise does not hold for this crate.

- `Ast`, `AstField`, `Model` and `Shape` are themselves `pub(crate)`, and every module they live in is `pub(crate)` or private — `lib.rs` declares `mod derive_error;` (private), and the crate contains no `pub mod` and no `pub use` at all. Only three items are `pub`: `derive_error`, `enrich_err` and `error`.
- In Rust a field's effective visibility is capped by the visibility of the type that owns it and of the modules enclosing it. `pub` on a field of a `pub(crate)` struct in a private module is not reachable from outside the crate; it means "visible everywhere this struct is", i.e. crate-internal. Rewriting it to `pub(crate)` would not change the public API by one item — it is the same surface, spelled longer.
- The "if modules are later re-exported" risk is the same risk either way: adding a `pub use` for a `pub(crate)` type does not compile, and making the type `pub` is a deliberate act that a reviewer would see regardless of how the fields are spelled. The guard here is the type and module visibility, which is already in place.

So this does not contradict the stated goal of a three-function public API — that goal is already enforced by the type and module visibility, and CI is green on the crate's external-types gate.

For the record, the reverse direction is a firm decision on this PR: nothing in `ohno_macros_impl` is made `pub` or `#[doc(hidden)]` in order to reach it from a test. An earlier revision of this work widened 89 items and 7 modules for that reason and it was deliberately reverted; the tests drive only the three public entry points.

martintmk (approved):

martin-kolinek (approved):

Vaiz:

Answering the two version findings from the review bodies at 06:12Z and 06:35Z. Both are about `ohno_macros_impl`'s version; they describe two different moments, and the first is already fixed.

**06:12Z — "crate version 0.5.0 but workspace and lockfile say 0.4.0": correct, and fixed.** That was a real break, not a style point: `[workspace.dependencies]` still required `^0.4.0`, so resolution failed before anything compiled —

```
error: failed to select a version for the requirement `ohno_macros_impl = "^0.4.0"`
candidate versions found which didn't match: 0.5.0
required by package `ohno_macros v0.4.0`
```

That is what turned 28 checks red at once. Fixed in b1022b51: the workspace requirement now reads 0.5.0 and the `Cargo.lock` entry records 0.5.0. Two lines, no other dependency touched.

**06:35Z — "PR description says 0.4.0 but the crate is 0.5.0": correct.** The description was stale, and I have updated it rather than changing the version.

To be explicit about which way that alignment goes, since the finding offered both: **0.5.0 is the intended version and stays.** It was chosen deliberately so the crate does not publish a version that nothing consumes. `ohno` and `ohno_macros` remain at 0.4.0, so the `ohno` family carries two version numbers on purpose. The PR description now states this, so it should not read as an accident.

Nothing else changed in response to these.

PR #684, opened 2026-08-21
feat: use opaque memory in HTTP APIs

martintmk:
## Summary

- change `HttpBodyBuilder` to store `OpaqueMemory` while accepting any `MemoryShared` provider
- make `OpaqueMemory::new` preserve an existing `OpaqueMemory` instead of adding another wrapper
- change fetch dependencies to expose `OpaqueMemory`, allowing customers to choose their memory provider
- keep `GlobalPool` as the default provider and retain the existing `bytesbuf` type names

## Validation

- full workspace clippy with all targets and features
- full workspace tests with all features
- affected-package formatting, no-default-features build, and external-type checks
- repository spelling and generated README checks

### DIFF
--- a/crates/bytesbuf/src/mem/opaque_memory.rs
+++ b/crates/bytesbuf/src/mem/opaque_memory.rs
@@ -83,6 +83,14 @@ mod tests {
     assert_impl_all!(OpaqueMemory: MemoryShared);
 
     #[test]
+    #[ignore = "stub"]
+    fn opaque_pool_wraps_custom_provider() {
+        // Arrange a custom MemoryShared provider.
+        // Wrap it in OpaquePool and reserve memory.
+        // Assert the custom provider handled the reservation.
+    }
+
+    #[test]
     fn wraps_inner() {
         let provider = GlobalPool::new();
         let memory = OpaqueMemory::new(provider);
--- a/crates/bytesbuf/src/mem/opaque_memory.rs
+++ b/crates/bytesbuf/src/mem/opaque_memory.rs
@@ -28,8 +28,8 @@ impl OpaqueMemory {
     ///
     /// # Panics
     ///
-    /// Panics only if runtime type identification reports [`OpaqueMemory`] but downcasting the
-    /// same value to [`OpaqueMemory`] fails, which would indicate a standard library defect.
+    /// Panics only if runtime type identification reports [`OpaqueMemory`] but the downcast of
+    /// the same value to [`OpaqueMemory`] fails, which would indicate a standard library defect.
     #[must_use]
     pub fn new<M: MemoryShared>(inner: M) -> Self {
         if TypeId::of::<M>() == TypeId::of::<Self>() {
--- a/crates/bytesbuf/src/mem/opaque_memory.rs
+++ b/crates/bytesbuf/src/mem/opaque_memory.rs
@@ -53,6 +53,13 @@ impl OpaquePool {
     }
 }
 
+#[cfg(feature = "std")]
+impl From<crate::mem::GlobalPool> for OpaquePool {
+    fn from(pool: crate::mem::GlobalPool) -> Self {
+        Self::new(pool)
+    }
+}
+
 impl Clone for OpaquePool {
     fn clone(&self) -> Self {
         Self {
--- a/crates/bytesbuf/src/mem/opaque_memory.rs
+++ b/crates/bytesbuf/src/mem/opaque_memory.rs
@@ -1,10 +1,9 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-use core::any::{Any, TypeId};
-
 #[cfg(not(test))]
 use alloc::boxed::Box;
+use core::any::{Any, TypeId};
 
 use thread_aware::ThreadAware;
 
--- a/crates/bytesbuf/src/mem/opaque_memory.rs
+++ b/crates/bytesbuf/src/mem/opaque_memory.rs
@@ -82,15 +82,17 @@ mod tests {
 
     use super::*;
     use crate::mem::GlobalPool;
+    use crate::mem::testing::TransparentMemory;
 
     assert_impl_all!(OpaquePool: MemoryShared);
 
     #[test]
-    #[ignore = "stub"]
     fn opaque_pool_wraps_custom_provider() {
-        // Arrange a custom MemoryShared provider.
-        // Wrap it in OpaquePool and reserve memory.
-        // Assert the custom provider handled the reservation.
+        let pool = OpaquePool::new(TransparentMemory::new());
+
+        let buffer = pool.reserve(1024);
+
+        assert_eq!(buffer.capacity(), 1024);
     }
 
     #[test]
--- a/crates/bytesbuf/src/mem/opaque_memory.rs
+++ b/crates/bytesbuf/src/mem/opaque_memory.rs
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+use core::any::{Any, TypeId};
+
 #[cfg(not(test))]
 use alloc::boxed::Box;
 
@@ -25,8 +27,15 @@ pub struct OpaqueMemory {
 impl OpaqueMemory {
     /// Creates a new instance of the adapter.
     #[must_use]
-    pub fn new(inner: impl MemoryShared) -> Self {
-        Self { inner: Box::new(inner) }
+    pub fn new<M: MemoryShared>(inner: M) -> Self {
+        if TypeId::of::<M>() == TypeId::of::<Self>() {
+            let inner: Box<dyn Any> = Box::new(inner);
+            *inner
+                .downcast::<Self>()
+                .expect("the concrete type was verified as OpaqueMemory above")
+        } else {
+            Self { inner: Box::new(inner) }
+        }
     }
 
     /// Reserves at least `min_bytes` bytes of memory capacity.
@@ -88,6 +97,16 @@ mod tests {
         let memory = OpaqueMemory::new(provider);
 
         let builder = memory.reserve(1024);
+        assert!(builder.capacity() >= 1024);
+    }
+
+    #[test]
+    fn accepts_existing_opaque_memory() {
+        let memory = OpaqueMemory::new(GlobalPool::new());
+        let memory = OpaqueMemory::new(memory);
+
+        let builder = memory.reserve(1024);
+
         assert!(builder.capacity() >= 1024);
     }
 
--- a/crates/bytesbuf/src/mem/opaque_memory.rs
+++ b/crates/bytesbuf/src/mem/opaque_memory.rs
@@ -8,7 +8,7 @@ use thread_aware::ThreadAware;
 
 use crate::mem::{Memory, MemoryShared};
 
-/// Type-erased memory pool backed by any [`MemoryShared`] implementation.
+/// Adapter to erase the type of a [`MemoryShared`] implementation.
 ///
 /// This adapter adds some inefficiency due to additional indirection overhead for
 /// every memory reservation, so avoid this adapter if you can tolerate alternatives (generics).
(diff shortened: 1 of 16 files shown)


crates/bytesbuf/src/mem/opaque_memory.rs
```diff
@@ -18,12 +18,12 @@ use crate::mem::{Memory, MemoryShared};
 /// provider. Cloning the adapter clones the wrapped provider; whether the clones then share any
 /// state is up to that provider.
 #[derive(Debug, ThreadAware)]
-pub struct OpaqueMemory {
+pub struct OpaquePool {
```
martintmk: rever this rename and all changes in bytesbuf, this PR is only about using OpaqueMemory not renaming the types in this crate

martintmk: &#x1F916; Agreed - I will revert the `bytesbuf` rename and use the existing `OpaqueMemory` type throughout the HTTP APIs.

---

crates/http_extensions/src/body/builder.rs
```diff
@@ -94,11 +98,7 @@ impl HttpBodyBuilder {
     /// relocated along with it.
     #[must_use]
     pub fn with_custom_memory(memory: impl MemoryShared, clock: &Clock) -> Self {
```
martintmk: this constructor is no longer needed

crates/fetch/src/custom.rs
```diff
@@ -53,7 +53,7 @@ where
     /// Clock for timing operations and timeouts.
     pub clock: Clock,
     /// Memory pool for usage-neutral memory allocations.
-    pub global_pool: GlobalPool,
+    pub global_pool: OpaquePool,
```
martintmk:
```suggestion
    pub memory: OpaquePool,
```

crates/http_extensions/src/body/builder.rs
martintmk: &#x1F916; Agreed - I will remove the redundant `with_custom_memory` constructor now.

---

crates/bytesbuf/src/mem/opaque_memory.rs
martintmk: &#x1F916; Fixed: `bytesbuf` is restored exactly to `main`, and the downstream HTTP APIs now use the existing `OpaqueMemory` adapter.

---

crates/fetch/src/custom.rs
martintmk: &#x1F916; Agreed - I will rename the dependency field to `memory`, using `OpaqueMemory` now that the `bytesbuf` rename has been reverted.

---

martintmk: &#x1F916; Fixed: `CustomDeps` now exposes `memory: OpaqueMemory`, and all custom transport call sites use the renamed field.

---

crates/http_extensions/src/body/builder.rs
martintmk: &#x1F916; Fixed: removed `with_custom_memory`; callers now construct `OpaqueMemory` explicitly and pass it to `HttpBodyBuilder::new`.

---

crates/http_extensions/src/body/builder.rs
```diff
@@ -71,31 +73,16 @@ impl HttpBodyBuilder {
...
+        Self::new(OpaqueMemory::new(GlobalPool::new()), &Clock::new_frozen())
     }
 
     /// Creates a new instance of [`HttpBodyBuilder`].
     ///
-    /// This method uses a per-thread memory pool from [`GlobalPool`].
+    /// The provided type-erased memory can wrap any [`MemoryShared`] implementation.
     #[must_use]
-    pub fn new(memory: GlobalPool, clock: &Clock) -> Self {
+    pub fn new(memory: OpaqueMemory, clock: &Clock) -> Self {
```
martintmk:
```suggestion
    pub fn new(memory: impl Into<Memory>, clock: &Clock) -> Self {
```

and transparently create OpaqueMemory?

martintmk: is it possible to no double crate OpaqueMemory if user already passed it?

martintmk: &#x1F916; Not without either double-wrapping an existing `OpaqueMemory` or adding conversion implementations in `bytesbuf`. A generic `impl MemoryShared` constructor would wrap every input, including `OpaqueMemory`; a blanket conversion for all `MemoryShared` types overlaps with Rust's identity conversion for `OpaqueMemory`. Keeping `new` typed as `OpaqueMemory` avoids the extra indirection and lets callers explicitly erase custom providers once.

---

martintmk: and how about accepting impl Memory then checking it if it's Opaque or wrapping it with OpaqueMemory

martintmk: or could OpaqueMemory::new handle this transparently?

martintmk: &#x1F916; Yes - because `MemoryShared` requires `'static`, the constructor can compare the concrete `TypeId`, safely downcast an existing `OpaqueMemory`, and only wrap other providers. I will implement that without changing `bytesbuf`.

---

martintmk: &#x1F916; Fixed: `HttpBodyBuilder::new` now accepts any `MemoryShared`. It uses the concrete `TypeId` to preserve an existing `OpaqueMemory` without another wrapper and type-erases all other providers.

---

crates/http_extensions/src/body/builder.rs
```diff
@@ -44,11 +48,22 @@ use crate::{HttpError, Result};
...
 }
 
+fn into_opaque_memory<M: MemoryShared>(memory: M) -> OpaqueMemory {
+    if TypeId::of::<M>() == TypeId::of::<OpaqueMemory>() {
+        let memory: Box<dyn Any> = Box::new(memory);
+        *memory
+            .downcast::<OpaqueMemory>()
+            .expect("the concrete type was verified as OpaqueMemory above")
+    } else {
+        OpaqueMemory::new(memory)
```
martintmk: could we do something like this in OpaqueMemory?

martintmk: &#x1F916; Yes - this detection belongs in `OpaqueMemory::new` so every caller gets the no-double-wrap behavior and `HttpBodyBuilder` stays simple. I will move it there without changing the public type name.

---

martintmk: &#x1F916; Fixed: the no-double-wrap logic now lives in `OpaqueMemory::new`, with a dedicated regression test. `HttpBodyBuilder` delegates directly to it.

---

crates/bytesbuf/src/mem/opaque_memory.rs
```diff
@@ -24,9 +25,21 @@ pub struct OpaqueMemory {
...
+    /// # Panics
+    ///
+    /// Panics only if runtime type identification reports [`OpaqueMemory`] but the downcast of
+    /// the same value to [`OpaqueMemory`] fails, which would indicate a standard library defect.
     #[must_use]
-    pub fn new(inner: impl MemoryShared) -> Self {
-        Self { inner: Box::new(inner) }
+    pub fn new<M: MemoryShared>(inner: M) -> Self {
+        if TypeId::of::<M>() == TypeId::of::<Self>() {
+            let inner: Box<dyn Any> = Box::new(inner);
```
martintmk, line 36: if the type is the same why do I need to go through the `let inner: Box<dyn Any> = Box::new(inner);`

martintmk: &#x1F916; `TypeId` only proves at runtime that `M` is `OpaqueMemory`; it does not let Rust cast an owned generic `M` to `Self`. The `Box<dyn Any>` is needed because `Any::downcast` is the safe standard-library operation that converts an owned erased value back to `Box<Self>`, after which we can move `Self` out. Avoiding that temporary allocation would require an unsafe cast or specialization, neither of which is justified here.

---

crates/bytesbuf/src/mem/opaque_memory.rs
```diff
@@ -91,6 +104,16 @@ mod tests {
         assert!(builder.capacity() >= 1024);
     }
 
+    #[test]
+    fn accepts_existing_opaque_memory() {
```
martintmk, line 108: **Posted by an AI agent**

**The no-double-wrap regression test cannot distinguish nested wrappers**

**Problem**
`accepts_existing_opaque_memory` only checks that `reserve` still works after wrapping twice. The previous implementation that always nested an `OpaqueMemory` produces the same capacity, so this test stays green if the new `TypeId`/downcast branch is removed.

**Why this matters**
The no-double-wrap behavior requested in this PR can regress without this test detecting it.

**Suggested fix**
Assert an internal invariant that distinguishes one erased layer from two, such as a test-only depth or type probe.

crates/fetch/src/custom.rs
```diff
@@ -311,6 +313,36 @@ mod tests {
...
+
+    impl Memory for CustomMemory {
+        fn reserve(&self, min_bytes: usize) -> BytesBuf {
+            self.inner.reserve(min_bytes)
+        }
+    }
+
+    #[cfg_attr(miri, ignore)]
+    #[tokio::test]
+    async fn custom_deps_accept_custom_opaque_memory() {
```
martintmk, line 329: **Posted by an AI agent · Non-blocking**

**The custom-provider test does not prove provider wiring**

**Problem**
This custom provider behaves exactly like `GlobalPool`, and the test only checks the HTTP status. It would still pass if `CustomDeps::memory` were ignored and the default pool were substituted; the corresponding `HttpBodyBuilder` test has the same gap.

**Why this matters**
The new provider-plumbing contract can break without either test detecting that body allocation bypasses the supplied provider.

**Suggested fix**
Use a counting or callback provider and assert that body allocation reaches it.

crates/fetch/src/tokio.rs
```diff
@@ -33,7 +33,7 @@ pub struct TokioDeps {
     /// Clock for timing operations and timeouts.
     pub clock: Clock,
     /// Memory pool for usage-neutral memory allocations.
-    pub global_pool: bytesbuf::mem::GlobalPool,
+    pub global_pool: bytesbuf::mem::OpaqueMemory,
```
martintmk, line 36: **Posted by an AI agent · Non-blocking**

**The Tokio dependency field understates the provider contract**

**Problem**
`global_pool` now accepts an `OpaqueMemory`, so it can hold any `MemoryShared` provider rather than specifically a `GlobalPool`. `CustomDeps::memory` already uses the broader name.

**Why this matters**
The two public dependency structs describe the same contract inconsistently, which can mislead callers about what providers the Tokio path accepts.

**Suggested fix**
Rename this field and its generated builder and example call sites to `memory`.

martintmk (commented):
**Posted by an AI agent**

I found one test gap that leaves the new no-double-wrap invariant unprotected, plus two non-blocking coverage/naming improvements. The intentional public API breaks were not raised: the existing PR discussion explicitly requested the constructor removal and CustomDeps rename, and release tooling handles versioning. Correctness, resilience, telemetry, consistency, and measured performance checks found no runtime regression. cargo check -p bytesbuf -p fetch -p http_extensions --all-targets --all-features passed, and current required CI is green.

geeknoid (commented):
Early feedback on this draft: I found no additional security, correctness, performance, or conformance issue beyond what is already in the discussion, so there are no new inline comments.

I specifically traced the `TypeId`/`Any` no-double-wrap path, `MemoryShared` cloning and thread-aware relocation, custom and Tokio provider propagation, `GlobalPool` defaults, every affected `HttpBodyBuilder::new` caller, and the request-builder benchmark. The runtime claims in the summary hold on static inspection. The semver breaks are already surfaced by CI and were explicitly discussed.

The two concrete test gaps are already covered by prior comments: the no-double-wrap test does not kill deletion of the special case, and the custom-provider tests do not prove allocations reach the supplied provider. I did not duplicate those threads or the existing `TokioDeps::global_pool` naming comment.

All 820 taxonomy classes were assessed (150 security, 162 correctness, 182 testing, 183 performance, 143 conformance). Nothing was executed: no project code, compilation, tests, benchmarks, profiles, Miri, coverage, mutation, fuzzing, or live probes.

PR #685, opened 2026-08-21
docs(ohno): document how error text is rendered

Vaiz:

## What this changes

Fixes AB#7675161. The `ohno` crate documentation describes only `#[display("...")]`. It does not say what an error renders when that attribute is absent, which is the common case. One sentence implies the type name is always the default message. That is true only for an error with no source.

## Details

- Replace the `Display Error Override` section with two sections. `How error text is rendered` covers the default. `Overriding error text` covers `#[display("...")]`. The default is documented first.
- Document the with-source default. The source message is printed as it stands, with no type name and no `caused by:` line. `source()` still returns the concrete error.
- Document the sourceless default. The type name is printed.
- Document that a cause given as a string renders like an error cause, but does not join the chain. `source()` returns `None` for it.
- Add an `Enrichment and backtraces` subsection. It shows an `#[enrich_err]` entry rendering under the message, then a captured backtrace.
- State that a chain of wrappers prints the message once and one backtrace block per level. Every error owns an `OhnoCore`, and every core renders its own backtrace.
- Correct the override text. Under a template the source is never printed alone, and the type name is never used.
- Add `crates/ohno/docs/COMPARISON.md`, a `thiserror` translation guide. It maps the attributes and names the places a port does not translate one for one.
- Remove the two `thiserror` references from the crate documentation, so the rendered documentation names no other crate.

## Effects

- Documentation only. No public API change, no behavior change, no new dependency.
- Three doctests are added. They assert the documented output on every build.
- `crates/ohno/README.md` is regenerated from `lib.rs` by `cargo doc2readme`. Its heading order changes.
- The `Display Error Override` heading is replaced by two headings. An external link to `#display-error-override` breaks. No link inside this repository uses that anchor.
- `docs/**/*.md` is in the workspace include allowlist. `COMPARISON.md` therefore ships in the published crate.
- `COMPARISON.md` states the behavior of `thiserror`, which this repository does not build. It can go stale without CI reporting it.
- The backtrace and enrichment output blocks are illustrative text. No test asserts their exact layout.
- The bug also asked for a `#[no_constructors]` recommendation for transparent wrappers. It is not added. That attribute is constructor guidance, and the `Automatic Constructors` section already documents it.

## Validation

- The rendered output in the documentation was copied from real runs, not written from memory. Scratch binaries confirmed the enrichment line format `> <message> (at <file>:<line>)`, and confirmed that a two-level wrapper prints two `Backtrace:` blocks under `RUST_BACKTRACE=1`.
- The string-cause claim was verified by running both forms. An error cause returns `Some` from `source()`. A string cause returns `None`. Both print the same text.

### DIFF
--- a/crates/ohno/README.md
+++ b/crates/ohno/README.md
@@ -46,8 +46,7 @@ fn open_file(path: impl AsRef<Path>) -> Result<String, ConfigError> {
 
 Derive macro for automatically implementing error traits.
 
-When applied to a struct or enum containing an [`OhnoCore`][__link5] field,
-this macro automatically implements [`std::error::Error`][__link6], [`std::fmt::Display`][__link7], [`std::fmt::Debug`][__link8], and [`From`][__link9] conversions.
+When applied to a struct containing an [`OhnoCore`][__link5] field, this macro automatically implements [`std::error::Error`][__link6], [`std::fmt::Display`][__link7], [`std::fmt::Debug`][__link8], and [`From`][__link9] conversions.
 
  > 
  > **Note**: `From<std::convert::Infallible>` is implemented by default and calls via [`unreachable!`][__link10] macro.
@@ -143,8 +142,9 @@ that it needs either a cause or a template.
 
 ### Enrichment and backtraces
 
-The message is only the first line of what `Display` writes. Each enrichment entry follows it
-on its own line, marked with `>` and tagged with the place it was added:
+The message is only the first part of what `Display` writes — with a template and a cause it is
+already two lines. Each enrichment entry follows it on its own line, marked with `>` and tagged
+with the place it was added:
 
 ```rust
 use std::io;
@@ -167,7 +167,7 @@ println!("{error}");
 //         > failed to load the service configuration (at src/config.rs:6)
 ```
 
-A captured backtrace comes last, after any enrichment:
+A captured backtrace comes last for that level, after that level’s enrichment:
 
 ```text
 no such file: /etc/app.toml
@@ -182,12 +182,15 @@ Backtrace:
    ...
 ```
 
-A backtrace is captured only when `RUST_BACKTRACE` asks for one. Use
-[`ErrorExt::message()`][__link12] to read the message on its own.
+Capture is the standard library’s decision: a backtrace is taken only when `RUST_LIB_BACKTRACE`
+asks for one, or when `RUST_BACKTRACE` does and `RUST_LIB_BACKTRACE` is unset. Use
+[`ErrorExt::message()`][__link12] to read the message without the backtrace.
 
 Every error owns its [`OhnoCore`][__link13], and every core renders its own backtrace, so a chain of
 wrappers that all use the default rendering prints the message once and one backtrace block per
-level:
+level. The levels are written in turn, innermost first — a wrapper’s own enrichment and
+backtrace follow the complete rendering of the level it wraps, so an outer enrichment entry
+appears *after* the inner level’s `Backtrace:` block, not alongside the other enrichment:
 
 ```text
 no such file: /etc/app.toml
@@ -457,7 +460,7 @@ uniformly via [`Labeled::label`][__link25].
 This crate was developed as part of <a href="https://github.com/microsoft/oxidizer">The Oxidizer Project</a>. Browse this crate's <a href="https://github.com/microsoft/oxidizer/tree/main/crates/ohno">source code</a>.
 </sub>
 
- [__cargo_doc2readme_dependencies_info]: ggGmYW0CYXZlMC43LjJhdIQb11VxC_uAPOQbtUn4Wx2-BfAbid3Nt1Y27Pobprn8Z6FjFy9hYvRhcoQbytrpabsE8xobjqG9axpJ2FMbPtKXWLOg3e8bzFvTXkcI6oFhZIKCZG9obm9lMC40LjCCa29obm9fbWFjcm9zZTAuNC4w
+ [__cargo_doc2readme_dependencies_info]: ggGmYW0CYXZlMC43LjJhdIQb11VxC_uAPOQbtUn4Wx2-BfAbid3Nt1Y27Pobprn8Z6FjFy9hYvRhcoQbNR5mLV9Kcl8bZv6w7esxE5UbZsOYl4G8UygbmNzvN7_GkWxhZIKCZG9obm9lMC40LjCCa29obm9fbWFjcm9zZTAuNC4w
  [__link0]: https://doc.rust-lang.org/stable/std/?search=fmt::Display
  [__link1]: https://doc.rust-lang.org/stable/std/?search=fmt::Debug
  [__link10]: https://doc.rust-lang.org/stable/std/macro.unreachable.html
--- a/crates/ohno/README.md
+++ b/crates/ohno/README.md
@@ -92,13 +92,13 @@ pub struct NetworkError {
 ## How error text is rendered
 
 Without a `#[display("...")]` attribute, the text an error renders depends on whether it has a
-source.
+cause — an error or a string handed to `caused_by`.
 
-**With a source**, the source’s message is printed as it stands: no type name, and no
+**With a cause**, the cause’s message is printed as it stands: no type name, and no
 `caused by:` line, so the wrapper leaves no trace in the message line. Enrichment and a
 backtrace, described below, are still written after it.
-[`source()`][__link11] still returns the concrete error, so a caller that
-walks the chain still finds it.
+A cause that is an error also stays in the [`source()`][__link11] chain, so a
+caller that walks the chain still finds it.
 
 ```rust
 use std::io;
@@ -126,7 +126,7 @@ Wrapping therefore adds nothing to the text. A wrapper that should say what it w
 “failed to load the configuration”, say — has to be given a template; see
 [Overriding error text](#overriding-error-text).
 
-**Without a source**, there is no message to pass through, so the type’s own name is printed:
+**Without a cause**, there is no message to pass through, so the type’s own name is printed:
 
 ```rust
 #[ohno::error]
@@ -139,7 +139,7 @@ println!("{error}");
 ```
 
 That is a symbol, not an explanation, so an error that renders as its own bare name is a sign
-that it needs either a source or a template.
+that it needs either a cause or a template.
 
 ### Enrichment and backtraces
 
@@ -186,7 +186,8 @@ A backtrace is captured only when `RUST_BACKTRACE` asks for one. Use
 [`ErrorExt::message()`][__link12] to read the message on its own.
 
 Every error owns its [`OhnoCore`][__link13], and every core renders its own backtrace, so a chain of
-wrappers prints the message once and one backtrace block per level:
+wrappers that all use the default rendering prints the message once and one backtrace block per
+level:
 
 ```text
 no such file: /etc/app.toml
@@ -456,7 +457,7 @@ uniformly via [`Labeled::label`][__link25].
 This crate was developed as part of <a href="https://github.com/microsoft/oxidizer">The Oxidizer Project</a>. Browse this crate's <a href="https://github.com/microsoft/oxidizer/tree/main/crates/ohno">source code</a>.
 </sub>
 
- [__cargo_doc2readme_dependencies_info]: ggGmYW0CYXZlMC43LjJhdIQb11VxC_uAPOQbtUn4Wx2-BfAbid3Nt1Y27Pobprn8Z6FjFy9hYvRhcoQbF7t022SddTAbMru5KrO1GBsbj6WlvWEqJMgbKR_mg0kL5sxhZIKCZG9obm9lMC40LjCCa29obm9fbWFjcm9zZTAuNC4w
+ [__cargo_doc2readme_dependencies_info]: ggGmYW0CYXZlMC43LjJhdIQb11VxC_uAPOQbtUn4Wx2-BfAbid3Nt1Y27Pobprn8Z6FjFy9hYvRhcoQbytrpabsE8xobjqG9axpJ2FMbPtKXWLOg3e8bzFvTXkcI6oFhZIKCZG9obm9lMC40LjCCa29obm9fbWFjcm9zZTAuNC4w
  [__link0]: https://doc.rust-lang.org/stable/std/?search=fmt::Display
  [__link1]: https://doc.rust-lang.org/stable/std/?search=fmt::Debug
  [__link10]: https://doc.rust-lang.org/stable/std/macro.unreachable.html
--- a/crates/ohno/README.md
+++ b/crates/ohno/README.md
@@ -182,11 +182,12 @@ Backtrace:
    ...
 ```
 
-Capture is the standard library’s decision: a backtrace is taken only when `RUST_LIB_BACKTRACE`
-asks for one, or when `RUST_BACKTRACE` does and `RUST_LIB_BACKTRACE` is unset. Use
-[`ErrorExt::message()`][__link12] to read the message without the backtrace.
+Whether a backtrace is captured at all is the standard library’s decision — see its
+[environment variables][__link12].
+Use [`ErrorExt::message()`][__link13] to read the message without this level’s own
+backtrace.
 
-Every error owns its [`OhnoCore`][__link13], and every core renders its own backtrace, so a chain of
+Every error owns its [`OhnoCore`][__link14], and every core renders its own backtrace, so a chain of
 wrappers that all use the default rendering prints the message once and one backtrace block per
 level. The levels are written in turn, innermost first — a wrapper’s own enrichment and
 backtrace follow the complete rendering of the level it wraps, so an outer enrichment entry
@@ -209,7 +210,7 @@ several layers deep.
 
 The `#[display("...")]` attribute replaces the rendered message with a template of its own,
 while still printing the cause after it. A cause that is an error also stays in the
-[`source()`][__link14] chain; a cause given as a string is printed the same way
+[`source()`][__link15] chain; a cause given as a string is printed the same way
 but does not join the chain, exactly as under the default rendering.
 
 ```rust
@@ -340,15 +341,15 @@ let my_err: MyError = io_err.into(); // Works automatically
 
 ## Error Enrichment
 
-The [`#[enrich_err("message")]`][__link15] attribute macro adds error enrichment with file and line info to function errors.
+The [`#[enrich_err("message")]`][__link16] attribute macro adds error enrichment with file and line info to function errors.
 
-Functions annotated with [`#[enrich_err("message")]`][__link16] automatically wrap any returned `Result`. If
+Functions annotated with [`#[enrich_err("message")]`][__link17] automatically wrap any returned `Result`. If
 the function returns an error, the macro injects a message, including file and line information, into the error chain.
 
 **Requirements:**
 
 * The function must return a type that implements the `map_err` method (such as `Result` or `Poll`)
-* The error type must implement the [`Enrichable`][__link17] trait (automatically implemented for all ohno error types)
+* The error type must implement the [`Enrichable`][__link18] trait (automatically implemented for all ohno error types)
 
 **Supported syntax patterns:**
 
@@ -404,10 +405,10 @@ fn open_file(path: &str) -> Result<String, MyError> {
 
 ## AppError
 
-For applications that need a simple, catch-all error type, use [`AppError`][__link18]. It
+For applications that need a simple, catch-all error type, use [`AppError`][__link19]. It
 automatically captures backtraces and can wrap any error type.
 
-To avoid accidental usage in libraries, [`AppError`][__link19] is only available when the `app-err`
+To avoid accidental usage in libraries, [`AppError`][__link20] is only available when the `app-err`
 feature is enabled.
 
 Example usage:
@@ -423,7 +424,7 @@ fn process() -> Result<(), AppError> {
 
 ## Error Labeling
 
-[`ErrorLabel`][__link20] is a low-cardinality string label for errors, intended for use as a metric
+[`ErrorLabel`][__link21] is a low-cardinality string label for errors, intended for use as a metric
 tag or structured log field. Labels must be chosen from a small, bounded set known at
 development time to avoid high-cardinality metric series.
 
@@ -437,7 +438,7 @@ let label = ErrorLabel::from_parts(["http", "client", "timeout"]);
 assert_eq!(label, "http.client.timeout");
 ```
 
-Use [`ErrorLabel::from_error_chain`][__link21] to walk an error’s [`source`][__link22]
+Use [`ErrorLabel::from_error_chain`][__link22] to walk an error’s [`source`][__link23]
 chain and build a dotted label from recognized errors:
 
 ```rust
@@ -451,8 +452,8 @@ let label = ErrorLabel::from_error_chain(&io_err, |e| {
 assert_eq!(label, "connection_refused");
 ```
 
-Types that carry an [`ErrorLabel`][__link23] can implement the [`Labeled`][__link24] trait to expose it
-uniformly via [`Labeled::label`][__link25].
+Types that carry an [`ErrorLabel`][__link24] can implement the [`Labeled`][__link25] trait to expose it
+uniformly via [`Labeled::label`][__link26].
 
 
 <hr/>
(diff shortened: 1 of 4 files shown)


crates/ohno/src/lib.rs
```diff
@@ -84,9 +84,123 @@
...
-//! # Display Error Override
+//! # How error text is rendered
 //!
-//! The `#[display("...")]` attribute customizes the main error message
+//! Without a `#[display("...")]` attribute, the text an error renders depends on whether it has a
+//! source.
+//!
+//! **With a source**, the source's message is printed as it stands: no type name, and no
+//! `caused by:` line, so the wrapper leaves no trace in the output.
+//! [`source()`](std::error::Error::source) still returns the concrete error, so a caller that
```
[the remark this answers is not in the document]
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Agreed — accepted. A wrapper with a source is transparent in the *message line*, but its own `OhnoCore` still contributes enrichment entries and a per-level backtrace to the full `Display` output, so "leaves no trace in the output" overstated it. Reworded to "leaves no trace in the message line", with a pointer to the enrichment/backtrace section that follows. Fixed in 7392146f.

crates/ohno/src/lib.rs
```diff
@@ -104,9 +218,11 @@
 //! // Output: "Failed to read config with path: /etc/config.toml\nCaused by:\n\tfile not found"
 //! ```
 //!
-//! The template string supports field interpolation using `{field_name}` syntax. The underlying
-//! error (if any) is automatically shown as "Caused by:" in the error chain. If the inner error
-//! has no source, only the custom message is displayed.
+//! The template string supports field interpolation using `{field_name}` syntax. Unlike the
+//! default rendering, the source is never printed on its own: the custom message always leads, and
+//! the underlying error (if any) follows as "Caused by:" in the error chain. If the error has no
+//! source, only the custom message is displayed — the type name is never used once a template is
```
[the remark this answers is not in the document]
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Correct, and a good catch — the docs were describing output the crate never produces. `MessageFormatter` in `crates/ohno/src/core.rs` writes `write!(f, "{msg}\n{CAUSED_BY} {source}")` with `const CAUSED_BY: &str = "caused by:"`, i.e. lowercase, on a single line, with no tab indent. That matches the assertions in `tests/display_error_override.rs`.

The example output is now `"Failed to read config with path: /etc/config.toml\ncaused by: file not found"`, and the prose says the cause "follows on the next line, after a `caused by:` label". Fixed in 7392146f; the README was regenerated from the crate docs to match.

kate-shine (approved):

Vaiz:

Both suppressed comments in [review #4992974183](https://github.com/microsoft/oxidizer/pull/685#pullrequestreview-4992974183) are correct, and they are the same defect seen twice. Accepted and fixed in 188004f1.

The "Overriding error text" section claimed the template "preserv[es] the underlying error as a cause in the error chain", while the example immediately below passes a *string* cause. Verified against `crates/ohno/src/core.rs`: `OhnoCore::source()` returns `Some` only for `Source::Error`; a string is classified by `is_string_error` into `Source::Transparent` over a `String`-backed error whose own `source()` is `None`, so it is rendered after `caused by:` but never joins the `source()` chain. That is exactly the split the default-rendering section already spells out, so the override section was contradicting it.

The section now reads:

> The `#[display("...")]` attribute replaces the rendered message with a template of its own, while still printing the cause after it. A cause that is an error also stays in the `source()` chain; a cause given as a string is printed the same way but does not join the chain, exactly as under the default rendering.

and the following paragraph now says "cause" throughout rather than "the underlying error" / "the source". README regenerated from the crate docs to match.

Vaiz:

Both suppressed comments in [review #4993095736](https://github.com/microsoft/oxidizer/pull/685#pullrequestreview-4993095736) are correct. Accepted and fixed in 00c39b9a.

**"has a source" (lib.rs:90).** Right — a string cause renders identically to an error cause, so keying the section on `source()` was the wrong axis. The section now opens on whether the error has a *cause* — an error or a string handed to `caused_by` — and the following paragraph says which kind stays in the `source()` chain. "Without a source" became "Without a cause" for the same reason. This also lines up the section with the override section, which moved to "cause" in 188004f1.

**"prints the message once" (lib.rs:187).** Also right, and it was the broader of the two. That sentence sits under a heading about the *default* rendering but read as a general claim, while `MessageFormatter` gives an overridden level `{msg}\ncaused by: {source}` — so a chain with any `#[display("...")]` level does produce several message lines. Scoped to "a chain of wrappers that all use the default rendering".

README regenerated from the crate docs. Local gates clean for `ohno`: format-check, readme-check, spellcheck, test-docs (39 passed), clippy.

crates/ohno/src/lib.rs
```diff
@@ -101,12 +219,14 @@
 //! // Usage
 //! let error = ConfigError::caused_by("/etc/config.toml", "file not found");
 //!
-//! // Output: "Failed to read config with path: /etc/config.toml\nCaused by:\n\tfile not found"
+//! // Output: "Failed to read config with path: /etc/config.toml\ncaused by: file not found"
```
martin-kolinek, line 228: 🤖: Non-blocking — the same correction is still owed one file over, in a file this PR doesn't touch: `crates/ohno/examples/display_error_complex.rs:4-7` documents its output as

```
Operation 'update' failed with code 404
Caused by:
        Not found
```

which is exactly the capitalized, tab-indented form this line fixes. `MessageFormatter` writes `write!(f, "{msg}\n{CAUSED_BY} {source}")` with `const CAUSED_BY: &str = "caused by:"` (`crates/ohno/src/core.rs:212,222`), so that example really prints `Operation 'update' failed with code 404\ncaused by: Not found`. `examples/chained.rs:52-57` already shows the lowercase form, so after this PR the crate documents its cause label two contradictory ways — and the example header is what a reader who runs `cargo run --example display_error_complex` compares the output against. Sweeping it in here looks like two lines.

crates/ohno/src/lib.rs
```diff
@@ -84,10 +84,128 @@
...
+//!
+//! let error = read_config().unwrap_err();
+//!
+//! println!("{error}");
+//! // Output: no such file: /etc/app.toml
+//! //         > failed to load the service configuration (at src/config.rs:6)
+//! # assert!(error.to_string().contains("> failed to load the service configuration (at "));
+//! ```
+//!
+//! A captured backtrace comes last, after any enrichment:
```
martin-kolinek: 🤖: Non-blocking — "A captured backtrace comes last, after any enrichment" holds per level, but not for a wrapped chain — which this section goes on to discuss at `:186-198`. The outer error's `format_error` writes the inner error's *entire* rendering (message + inner enrichment + inner backtrace) through `write!(f, "{source}")` (`core.rs:225`) before it writes its own enrichment entries (`core.rs:144-146`) and its own backtrace (`core.rs:148-150`). So a wrapper that adds enrichment prints that enrichment *after* the inner level's `Backtrace:` block, not after all enrichment; the multi-level example at `:190-198` only avoids showing this because it carries no enrichment — the new `:187` qualifier ("that all use the default rendering") addresses the `caused by:` case, not this one. Worth a clause here (e.g. "…last for a given level; on a wrapped chain each level's entries and backtrace are written in turn, innermost first"), since ordering is what this section exists to state.

Two companions to the same wording, if you touch it:
- `:143` says the message is "only the first line" of `Display`. With a display override and a cause it is the template plus `\ncaused by: …` (see `tests/display_error_override.rs:27-41`), i.e. already multiline — "the first *part*" would be exact.
- `crates/ohno/docs/COMPARISON.md:57-60` states the same message → enrichment → backtrace ordering as a global promise and needs the same per-core phrasing, so the standalone comparison guide doesn't promise more than the implementation provides.

crates/ohno/src/lib.rs
```diff
@@ -84,10 +84,128 @@
...
+//! Backtrace:
+//!    0: std::backtrace::Backtrace::capture
+//!    1: ohno::backtrace::Backtrace::capture
+//!    2: ohno::core::OhnoCore::from_source
+//!    3: my_app::config::ConfigError::caused_by
+//!    4: my_app::config::read_config
+//!    ...
+//! ```
+//!
+//! A backtrace is captured only when `RUST_BACKTRACE` asks for one. Use
```
martin-kolinek: 🤖: Non-blocking — `OhnoCore` delegates to `std::backtrace::Backtrace::capture`, which also honors `RUST_LIB_BACKTRACE` and only falls back to `RUST_BACKTRACE` when that variable is unset. Saying capture happens only when `RUST_BACKTRACE` asks for it excludes a supported control; naming both standard-library variables here (and regenerating the README) would keep this new rendering contract from implying `RUST_BACKTRACE` is the only way a backtrace can appear.

crates/ohno/src/lib.rs
```diff
@@ -84,10 +84,128 @@
...
+//!         io::ErrorKind::NotFound,
+//!         "no such file: /etc/app.toml",
+//!     )))
+//! }
+//!
+//! let error = read_config().unwrap_err();
+//!
+//! println!("{error}");
+//! // Output: no such file: /etc/app.toml
+//! # assert!(error.to_string().starts_with("no such file: /etc/app.toml"));
```
martin-kolinek: 🤖: Non-blocking — this `starts_with` assertion doesn't pin the headline negative guarantees the prose just made at `:92-94`: it still passes if rendering later appends a wrapper name or a `caused by:` line. Since backtrace capture makes exact `Display` equality environment-dependent, consider asserting the exact `ErrorExt::message()` output instead (the generated impl uses the same `MessageFormatter`), here and for the cause-less example at `:135`.

Related, in the same section: the override example at `:220-222` is the only one that still pins its output in a comment rather than an assertion — and it is the one that drifted, since the `Caused by:` / tab-indented text this PR corrects lived there. Its output is a single deterministic string, so it is the best candidate for a hidden exact assertion:

```rust
# assert_eq!(error.to_string(), "Failed to read config with path: /etc/config.toml\ncaused by: file not found");
```

crates/ohno/src/lib.rs
```diff
@@ -84,10 +84,128 @@
...
+//!    3: my_app::config::ConfigError::caused_by
+//!    4: my_app::config::read_config
+//!    ...
+//! ```
+//!
+//! A backtrace is captured only when `RUST_BACKTRACE` asks for one. Use
+//! [`ErrorExt::message()`](ErrorExt::message) to read the message on its own.
+//!
+//! Every error owns its [`OhnoCore`], and every core renders its own backtrace, so a chain of
+//! wrappers that all use the default rendering prints the message once and one backtrace block per
```
martin-kolinek, line 191: 🤖: Non-blocking — this multi-level backtrace contract is currently represented only by the `text` block: the existing forced-backtrace tests format a single `OhnoCore`, so nothing would catch a regression that omitted or duplicated a wrapper's block, or that printed the message more than once. An integration test that re-executes the test binary with `RUST_BACKTRACE=1`, builds two default-rendered `#[ohno::error]` wrappers, and asserts the source message occurs once and exactly two `\n\nBacktrace:\n` blocks appear in per-level order would pin it. A child process avoids races from mutating `RUST_BACKTRACE` in-process, and other tests in the repo already use the `current_exe()` child-test pattern.

crates/ohno/docs/COMPARISON.md
```diff
@@ -0,0 +1,84 @@
...
+Code that logs `{e}` and expects a single line therefore needs review when
+porting. `ErrorExt::message()` returns the message on its own.
+
+Because every `ohno` error owns its own `OhnoCore`, and every core renders its
+own backtrace, a chain of wrappers prints the message once and one backtrace
+block per level.
+
+## Structs only
+
+`ohno` derives on structs; an enum is rejected. A `thiserror` enum ports to one
```
martin-kolinek, line 92: 🤖: Non-blocking — this is correct (`derive_error/parse.rs` rejects `Data::Enum` with "`#[derive(ohno::Error)]` supports structs only"), but it now contradicts the crate docs being edited in this same PR: `crates/ohno/src/lib.rs:45` — and the generated `README.md:49` — still say the derive applies "to a struct or enum containing an `OhnoCore` field". Worth correcting `lib.rs:45` in this PR, since that file is already open and this new section is what puts the two statements side by side.

martin-kolinek (approved):
🤖: Approved.

crates/ohno/docs/COMPARISON.md
Vaiz: yep, it doesn't work with enums, and it's not supposed to

crates/ohno/docs/COMPARISON.md
sandersaares, line 1: [Copilot speaking]

# Surface the comparison guide from public crate entry points

The new comparison guide is written for readers encountering `ohno` from `thiserror`, but neither the crate-level rustdoc nor the generated crate README links to it. Packaging the Markdown file makes it available in the source archive, but does not expose it through the documentation entry points its intended audience normally reads.

**Reproducible reasoning**: `crates/ohno/src/lib.rs` is the source for both the public crate documentation and, through `cargo-doc2readme`, `crates/ohno/README.md`. The reviewed versions contain no reference to `COMPARISON.md`, and rustdoc does not automatically add standalone packaged Markdown files to the crate documentation. The guide also refers back to "How error text is rendered" as plain text rather than a link. The repository's `internity` crate demonstrates the complete navigation pattern by linking its comparison guide from crate-level documentation, which carries that link into its generated README.

**Consequence**: A developer evaluating or performing a migration from `thiserror` can read the primary `ohno` documentation without learning that the purpose-built guide exists.

**Recommended action**: Link the comparison guide from the crate-level documentation and regenerate the README so both public entry points expose it; make the guide's back-reference to the canonical rendering section directly navigable as well.

**References**:

- crates/internity/src/lib.rs
- Repository README, "Documenting Crates"

**Impacted locations**:

- `crates/ohno/docs/COMPARISON.md:4-8`
- `crates/ohno/src/lib.rs:17-25`
- `crates/ohno/README.md:17-34`

crates/ohno/src/lib.rs
sandersaares: [Copilot speaking]

# Use one rendering model throughout the migration guide

The comparison guide uses the related terms cause, source, message, `Display` output, and caller-visible output as though they described the same boundary. They do not: a cause is an error or string supplied to `caused_by`; only an error cause participates in the `source()` chain; `ErrorExt::message()` returns the message portion but may include a newline and `caused by:`; `Display` adds enrichment and backtrace data; and any `Error: ` prefix belongs to a caller outside `ohno`. The guide's source-based default categories, prefixed sourceless example, and presentation of `message()` beside a single-line warning each cross one of these boundaries.

**Reproducible reasoning**: `OhnoCore::from` stores a recognized string cause as `Source::Transparent` and a typed error cause as `Source::Error`. `MessageFormatter` renders both variants through the same cause-present branches, while `OhnoCore::source()` exposes only `Source::Error`; rendering behavior therefore cannot be classified solely by whether `source()` returns a value. The generated `ErrorExt::message()` calls `format_message`, which runs that formatter and can produce `<custom message>\ncaused by: <cause>`. The generated `Display` implementation instead calls `format_error`, which writes the same formatter output and then appends enrichment entries and a captured backtrace. None of those layers adds `Error: `. These are successive parts of one rendering model, so explaining them with one set of boundaries resolves the repeated contradictions rather than adding isolated caveats.

**Consequence**: A reader porting an error can select the sourceless default for a string cause, expect `ohno` itself to emit an `Error: ` prefix, or replace `Display` with `message()` while still unintentionally producing a multiline log field.

**Recommended action**: Establish one explicit rendering model at the start of the comparison's rendering guidance that distinguishes cause from source and message from full `Display` and caller presentation. Reconcile the default-case classification, the sourceless example, and the single-line migration advice with that model so each passage names the exact input or output surface it describes.

**References**:

- crates/ohno/src/core.rs
- crates/ohno_macros_impl/src/derive_error/generate/traits.rs
- crates/ohno/tests/display_error_override.rs
- crates/ohno/src/lib.rs, "How error text is rendered"

**Impacted locations**:

- `crates/ohno/docs/COMPARISON.md:29-46`
- `crates/ohno/docs/COMPARISON.md:51-56`
- `crates/ohno/docs/COMPARISON.md:39-41`
- `crates/ohno/docs/COMPARISON.md:73-74`

sandersaares (approved):
[Copilot speaking]

Published 2 findings. One finding follows up on an existing discussion thread.

| Diagnostic | Value |
|------------|-------|
| Cache | Hit |

</details>

crates/ohno/src/lib.rs
martin-kolinek: 🤖: Non-blocking — agreed on the rendering model, and I think the two points are compatible rather than competing: establishing one vocabulary settles *what each surface contains*, not *in what order a wrapped chain emits them*, which is the gap this thread is about. `format_error` writes the inner error's entire rendering — message + inner enrichment + inner `Backtrace:` block — via `write!(f, "{source}")` (`core.rs:225`) before appending its own enrichment (`core.rs:144-146`) and its own backtrace (`core.rs:148-150`). So "a captured backtrace comes last, after any enrichment" is true per level and false for the chain: a wrapper that adds enrichment prints it *after* the inner level's backtrace. One clause fixes it ("…last for a given level; on a wrapped chain each level's entries and backtrace are written in turn, innermost first"), and the same per-level phrasing is owed to `docs/COMPARISON.md:57-60`, which states the ordering as a global promise. Whatever vocabulary the guide settles on, this ordering fact still needs stating.

crates/ohno/src/lib.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Accepted and swept in — you are right that leaving it would make the crate document its cause label two contradictory ways. Fixed in 34571e3d.

I ran the example rather than reasoning about it: `cargo run -p ohno --example display_error_complex` prints

```
Operation 'update' failed with code 404
caused by: Not found
```

so the header now says exactly that. `examples/chained.rs` already matched, so the two are consistent again.

crates/ohno/src/lib.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Accepted — this was the most substantive finding on the PR, and both of you are right that it survives the earlier `caused by:` qualifier. Fixed in 34571e3d.

Confirmed in `core.rs`: `format_error` runs the `MessageFormatter` first, and for a wrapper with a cause that formatter writes `{source}` — the inner error's *entire* rendering, message plus its enrichment plus its `Backtrace:` block — before the outer level appends its own enrichment (`core.rs:144-146`) and its own backtrace (`core.rs:148-150`). So the ordering is per level, and an outer enrichment entry really does land after the inner level's backtrace.

Changes:
- The backtrace sentence now reads "comes last **for that level**, after that level's enrichment".
- The multi-level paragraph now states the chain rule: levels are written innermost first, so a wrapper's enrichment appears after the inner level's `Backtrace:` block rather than grouped with the inner entries.
- `docs/COMPARISON.md` carried the same ordering as a global promise; it now states it per level and points at the chain behaviour.

Both companions taken too: "the message is only the first **part**" (with a template and a cause it is already two lines), and `COMPARISON.md` no longer offers `ErrorExt::message()` as the single-line answer — it drops enrichment and the backtrace, but returns `<message>\ncaused by: <cause>` when there is a template and a cause.

On the rendering-model finding in the same thread: the specific contradictions it lists are fixed — the guide's defaults are now classified by *cause* rather than by `source()`, with a sentence saying why (`source()` does not predict what is rendered), the `Error: ` prefix is attributed to the caller rather than to `ohno`, and the `message()` advice is corrected as above. I stopped short of restructuring the guide around a formal vocabulary section: the crate docs are the canonical account and the guide is a mapping onto it, so a second normative model in `docs/` is a place for the two to drift apart. Happy to reconsider if you'd rather have the explicit section.

crates/ohno/src/lib.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Accepted — correct. `Backtrace::capture` in `crates/ohno/src/backtrace.rs:37-40` delegates straight to `std::backtrace::Backtrace::capture`, so the standard library's precedence applies and naming only `RUST_BACKTRACE` understated it. Fixed in 34571e3d:

> Capture is the standard library's decision: a backtrace is taken only when `RUST_LIB_BACKTRACE` asks for one, or when `RUST_BACKTRACE` does and `RUST_LIB_BACKTRACE` is unset.

Phrased as the standard library's rule rather than `ohno`'s, since that is where the decision is made. README regenerated.

crates/ohno/src/lib.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Accepted in full — including the observation about *which* example drifted, which is the argument for the change. Fixed in 34571e3d.

All three examples in the section now assert their exact `ErrorExt::message()` output:

```rust
# assert_eq!(error.message(), "no such file: /etc/app.toml");
# assert_eq!(error.message(), "ConfigError");
# assert_eq!(error.message(), "Failed to read config with path: /etc/config.toml\ncaused by: file not found");
```

`message()` is the right surface for this: it runs the same `MessageFormatter` but excludes the backtrace, so the assertions are exact without being environment-dependent — which is what forced `starts_with` before. The override example was indeed the only one pinned by a comment alone, and it is the one whose text this PR had to correct; it is now the one with the strictest assertion.

Doctests pass on the new head (39 passed, 0 failed).

crates/ohno/src/lib.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Declining this one for this PR — it is a real gap, but it is not a documentation change.

The gap is genuine: the multi-level contract is carried only by a `text` block, and the existing forced-backtrace tests format a single `OhnoCore`, so nothing would catch a regression that dropped or duplicated a level's block. Your sketch is also the right shape — a child process via `current_exe()` avoids the in-process `RUST_BACKTRACE` mutation race.

But this PR is documentation-only: it changes no runtime behaviour and no public API, and everything in it is verifiable by reading the rendering code or running a doctest. Adding an integration test that re-executes the test binary under a modified environment is new test infrastructure exercising runtime behaviour, and it deserves its own PR where a flake or a platform difference in backtrace capture can be dealt with on its own merits rather than blocking a docs correction. You marked it non-blocking, which I read the same way.

I did take the part of it that fits here: the section now states the chain ordering explicitly rather than leaving it implicit in the `text` block (see the sibling thread), so a future test has a written contract to pin. Happy for this to be filed as a follow-up issue — say the word and I will open one with your sketch attached.

crates/ohno/docs/COMPARISON.md
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Done in 34571e3d, following @Vaiz's confirmation above that enums are not supported and not meant to be. `lib.rs` line 45 now reads "When applied to a struct containing an [`OhnoCore`] field", and the README was regenerated, so the crate docs and the new comparison section no longer say different things about the same restriction.

crates/ohno/docs/COMPARISON.md
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Declining the rustdoc half of this, accepting the navigation half.

**Not linking the guide from `lib.rs` / the README.** That is a deliberate rule for this crate, not an oversight: `thiserror` is not named in `ohno`'s public rustdoc or its published API documentation. References to it live in the repo's `docs/` folder only. The guide is titled "Coming from `thiserror`", so a crate-level link necessarily surfaces a competing crate's name in the published docs — which is the exact thing the rule exists to prevent. The `internity` precedent does not transfer, because its comparison guide is not about a third-party crate.

The consequence you describe is real but bounded: `docs/COMPARISON.md` is discoverable where its audience actually looks — the repository, and the source archive, which is why it is packaged. Someone migrating from `thiserror` is reading the repo, not only docs.rs.

**Taken:** the back-reference is now navigable. `COMPARISON.md` line 48 was plain text pointing at "How error text is rendered"; it is now a link to that section in the crate README, which resolves on GitHub and in the packaged sources. Fixed in 34571e3d.

If the no-`thiserror`-in-rustdoc rule should be revisited, that is a call for @Vaiz rather than something I will decide inside this PR.

crates/ohno/src/lib.rs
Vaiz: FYI, capturing multiple backtraces was never a goal but a limitation of current Rust language that doesn't give a way to verify if source error already contains a backtrace

crates/ohno/README.md
```diff
@@ -89,10 +88,129 @@ pub struct NetworkError {
...
+Backtrace:
+   0: std::backtrace::Backtrace::capture
+   1: ohno::backtrace::Backtrace::capture
+   2: ohno::core::OhnoCore::from_source
+   3: my_app::config::ConfigError::caused_by
+   4: my_app::config::read_config
+   ...
+```
+
+Capture is the standard library’s decision: a backtrace is taken only when `RUST_LIB_BACKTRACE`
```
Vaiz: don't go too deep into details, just share this link - https://doc.rust-lang.org/std/backtrace/index.html#environment-variables

crates/ohno/src/lib.rs
```diff
@@ -84,10 +83,134 @@
...
+//!    1: ohno::backtrace::Backtrace::capture
+//!    2: ohno::core::OhnoCore::from_source
+//!    3: my_app::config::ConfigError::caused_by
+//!    4: my_app::config::read_config
+//!    ...
+//! ```
+//!
+//! Capture is the standard library's decision: a backtrace is taken only when `RUST_LIB_BACKTRACE`
+//! asks for one, or when `RUST_BACKTRACE` does and `RUST_LIB_BACKTRACE` is unset. Use
+//! [`ErrorExt::message()`](ErrorExt::message) to read the message without the backtrace.
```
martin-kolinek: 🤖: Non-blocking — this push sharpens "read the message on its own" into the specific negative claim "without the backtrace", and that claim has one exception. `message()` runs `MessageFormatter` only, so it does drop *this* level's enrichment and backtrace — but with no `#[display]` template and a cause, `MessageFormatter` renders `write!(f, "{source}")` (`core.rs:225`), i.e. the cause's entire `Display`, which for an `ohno` cause includes that level's enrichment and its `Backtrace:` block. So for a transparent wrapper `message()` can still contain a backtrace. Two words settle it — "without its own backtrace" — no extra prose needed. `crates/ohno/docs/COMPARISON.md:77` carries the same claim ("drops the enrichment and the backtrace") and wants the same qualifier.

crates/ohno/README.md
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Done — 7b1f734c. Cut the precedence prose and linked std instead:

> Whether a backtrace is captured at all is the standard library's decision — see its [environment variables](https://doc.rust-lang.org/std/backtrace/index.html#environment-variables).

Better anyway: the rule belongs to `std` and can change there, so restating it here was a copy that could go stale. README regenerated.

crates/ohno/src/lib.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Accepted — and you are right that my own push is what sharpened this into a false claim. Fixed in 7b1f734c.

Confirmed: with no template and a cause, `MessageFormatter` takes the `(None, Source::Transparent | Source::Error)` branch and writes `{source}` — the cause's entire `Display`. If that cause is itself an `ohno` error, its `Display` is `format_error`, so its enrichment entries and its `Backtrace:` block come back with it. `message()` drops *this* level's own enrichment and backtrace, not everything beneath it.

`lib.rs` now reads "without this level's own backtrace". `COMPARISON.md` needed more than the two words, since it was making the claim as migration advice — someone swapping `{e}` for `message()` to get a single log line is exactly who this bites — so it now says `message()` drops this level's own enrichment and backtrace, and spells out that on a transparent wrapper it renders the cause's full `Display` and brings that level's enrichment and backtrace back.

Gates clean on the new head; README regenerated.

wukchung (approved):

Vaiz:

The suppressed comment in [review #5005736773](https://github.com/microsoft/oxidizer/pull/685#pullrequestreview-5005736773) is correct. Accepted and fixed in 2563f777.

`message()` calls `format_message`, which runs `MessageFormatter` alone. Both the enrichment entries and the backtrace are appended by `format_error` (`core.rs:144-150`), so neither survives — naming only the backtrace implied enrichment did. `lib.rs` now reads:

> Use [`ErrorExt::message()`](ErrorExt::message) to read the message without this level's own enrichment or backtrace.

`docs/COMPARISON.md` already said "this level's own enrichment and backtrace", so this is the crate docs catching up rather than a new claim. README regenerated; gates clean for `ohno`.

martin-kolinek (approved):

PR #686, opened 2026-08-21
refactor(observed_macros): split the implementation into observed_macros_impl

Vaiz:

## What this changes

`observed_macros` becomes a thin proc-macro shim. A new crate, `observed_macros_impl`, holds the implementation. This follows the Microsoft Rust guideline M-MACRO-MAIN-CRATE. Seven other crate families in this workspace already use the pattern.

The split created a sixth copy of a macro-expansion test helper. This change moves that helper to `testing_aids` and deletes the copies.

### The split

- Add `observed_macros_impl` 0.24.0. It holds the `enrichment`, `event` and `field_attrs` modules. The version matches `observed_macros`, as every other `*_macros_impl` crate does.
- Reduce `observed_macros` to two entry points. Each converts `proc_macro::TokenStream` to `proc_macro2::TokenStream`, delegates, then converts back. Macro names, the `attributes(...)` list and the rustdoc do not change.
- Keep every module in `observed_macros_impl` private. The public API is `event` and `derive_enrichment` only. No item is `pub` or `#[doc(hidden)]` for a test.
- Move the tests into two integration tests. `tests/expansion.rs` snapshots each expansion. `tests/entry_points.rs` holds the rejections and the classification checks. It uses no insta, so it runs under Miri.
- Add an `observed` test group to `scripts/mutants.rs`. The family had no group before, so the `observed_macros` shim was mutated alone and no test could reach it.
- Register the crate in the root `CHANGELOG.md` and the CI Miri exclusion list. Give the shim the `coverage-gate` opt-out that the other proc-macro shims carry.

### The shared test helper — fixes AB#7776892

- Add `testing_aids::render_expansion`. It parses an expansion as a `syn::File`, then formats it with `prettyplease`.
- Use it in `observed_macros_impl`, `ohno_macros_impl`, `data_privacy_macros_impl`, `fundle_macros_impl` and `thread_aware_macros_impl`. No snapshot file changes, so the output is byte-identical.
- `render_expansion` panics when the tokens are not a Rust file, and reports the parse error and the tokens. Two of the replaced helpers returned raw token text instead. That records a broken macro as a snapshot which only looks reformatted.
- Add `testing_aids::render_tokens_lossy` for macro input. Some `ohno_macros_impl` cases feed a macro tokens that are not an item, and those inputs still belong in the snapshot. The helper is documented as input-only.
- Add a `macros` feature to `testing_aids` with optional `syn`, `prettyplease` and `proc-macro2`. Move `futures`, `opentelemetry`, `opentelemetry_sdk`, `tracing` and `tracing-subscriber` behind a `runtime` feature. `runtime` stays on by default.
- Remove the `prettyplease` dev-dependency from five macro crates.

## Effects

- `observed_macros_impl` is a new published crate. `observed_macros` depends on it. The build gains one crate boundary.
- Macro behaviour does not change. `observed` and `observed_testing` are untouched and pass.
- A test can no longer reach a private item in `observed_macros_impl`. A future test for an internal helper must go through an entry point, or the helper must be removed as unreachable.
- `render_expansion` panics where two helpers previously produced a snapshot. A suite that relies on that fallback now fails instead. `thread_aware_macros_impl` passes without it, so nothing relies on it today.
- Consumers that take `testing_aids` through the workspace dependency must now request `runtime`. `arty_executor` and `observed_testing` are updated. The seven crates that take it by path keep default features and do not change.
- A proc-macro crate that takes `testing_aids` with `features = ["macros"]` does not build `opentelemetry` or `tracing`. Impact-scoped CI stops marking those crates affected by changes to those stacks.
- Mutation testing now runs the whole `observed` family's tests against each mutant of any crate in the family. Per-mutant runtime grows.
- The `observed_macros_impl` suite is 32 tests over 15 snapshots. `observed_macros` carried 125 unit tests over 44 snapshots.
- `tests/expansion.rs` is `#![cfg(not(miri))]`. insta reads the workspace from disk, which Miri isolation rejects.
- Region coverage of `observed_macros_impl` is 98.46%, against 100% of lines and functions. The gate measures lines.

## Validation

- `cargo hack --package testing_aids --feature-powerset check` passes for all six feature combinations. PR CI does not run `cargo-hack`; it belongs to the scheduled exhaustive group.
- Not verified under Miri. CI excludes `observed_macros_impl` from the Miri job, and `tests/expansion.rs` is compiled out there.

## Review decisions (2026-08-21)

- **`testing_aids` no longer has features.** The `runtime` / `macros` split was reverted at the maintainer's request: every extra feature combination makes cargo rebuild the same crate again, which costs more than the trimmed dependency graph saves. All dependencies are unconditional, the `#[cfg(feature = ...)]` gates are gone, and consumers take `testing_aids = { workspace = true }`. This also removed the `[features]` table that was breaking `anvil-cargo-sort`.
- **The old `observed_macros` implementation sources are deleted.** `src/{event,enrichment,field_attrs}.rs` and `src/snapshots/` were unreferenced but would still have shipped via the `/src/**` packaging allowlist.
- **`scripts/mutants.rs` keeps the `observed` / `observed_macros` / `observed_macros_impl` grouping.** Every other group has the same shape and none includes a `*_testing` sibling; the mutation gate is diff-scoped, so the extra crate would cost time without widening what the gate catches.
- **`observed_macros_impl` stays on the Miri exclusion list.** That list covers every `*_macros_impl` crate for the same cost reason, and any `unsafe` these macros emit is checked in the consuming crate's tests. `tests/entry_points.rs` exists so the non-snapshot assertions are not trapped in the `#![cfg(not(miri))]` snapshot file, and it remains runnable under `just miri` locally.
- **The strict/lenient render split stays in `testing_aids`, not in `ohno_macros_impl/tests/public_api.rs`.** `thread_aware_macros_impl/tests/derive.rs` carried the same raw-token fallback, so a crate-local fix would have left that copy broken and let the next `*_macros_impl` crate reintroduce it.

**Correction (2026-08-21):** removing the features initially also dropped `default-features = false` from the root `[workspace.dependencies]` entry for `testing_aids`, which failed `anvil-ensure-no-default-features` on all four `pr-fast` platforms and on `static-analysis`. That gate requires the flag on every workspace dependency regardless of whether the crate defines features, so it is restored in `f44ee691`; with no `[features]` table it is a no-op.

## Test strategy (updated 2026-08-24)

Every test in `observed_macros_impl` asserts an **insta snapshot** — of the whole expansion, or of the whole diagnostic. No substring assertions. A `contains` marks a branch covered while leaving the rest of the output free to be wrong, so it buys the coverage number without the proof this PR is after; a snapshot pins all of it, so a branch whose behaviour changes shows in the diff whether or not anyone thought to assert on that part.

A table-driven case joins its rows into one labelled snapshot, so the test count stays at 32 and one test still owns one snapshot. `tests/entry_points.rs` existed only to hold the assertions that carried no snapshot and is therefore deleted; everything lives in `tests/expansion.rs`.

Consequence: `expansion.rs` is `#![cfg(not(miri))]`, so the crate has no Miri-runnable tests. That costs no CI coverage — the crate is on the Miri exclusion list in `main.yml` — but a local `just miri` runs nothing for it.

### Snapshot format (updated 2026-08-24, `49048c77`)

Review feedback on the snapshot conversion: a table row was labelled with a hand-written
summary, so the snapshot never showed which source produced it, and the assertion was an
inline iterator chain, so insta recorded that whole chain as the `expression:` header.

Both are fixed:

- **Every row records the whole input.** A row is now the full source the macro was given,
  spelled the way a user would write it (`#[event("e")]` / `#[derive(Enrichment)]` followed
  by the item), then `---- yields ----`, then the diagnostic or expansion. A snapshot reads
  on its own, and the source names its own entry point, so the two rows of one input in a
  case that drives both macros are told apart by the input rather than by a label.
- **Reports are bound to `output` before being asserted**, matching the convention already
  documented in `ohno_macros_impl/tests/public_api.rs`. The header becomes `expression: output`.
  Side effect: the header no longer contains the table, so reformatting a test can no longer
  invalidate its snapshot.
- **A single-line diagnostic is asserted inline** rather than in a file of its own, so the
  input and the expected text sit on adjacent lines. Two snapshot files were deleted for this.
  Multi-row reports stay in files: now that each row carries its full input they are too large
  to read inside the test body.

Test count is unchanged — 32 in the crate, 6588 in the workspace.


crates/testing_aids/Cargo.toml
```diff
@@ -29,11 +29,27 @@ allowed_external_types = ["*"]
...
+opentelemetry = { workspace = true, features = ["metrics"], optional = true }
+opentelemetry_sdk = { workspace = true, features = ["metrics", "testing"], optional = true }
+prettyplease = { workspace = true, optional = true }
+proc-macro2 = { workspace = true, optional = true }
+syn = { workspace = true, features = ["full", "parsing", "printing", "clone-impls"], optional = true }
+tracing = { workspace = true, features = ["std"], optional = true }
+tracing-subscriber = { workspace = true, features = ["fmt", "registry"], optional = true }
+
+[features]
+default = ["runtime"]
```
Vaiz: remove features and opt deps. By increasing the number of possible combinations, the compile time only increases by recompiling the same crate over and over

Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Done in d827822e. `[features]` is gone from `testing_aids`: `futures`, `opentelemetry`, `opentelemetry_sdk`, `prettyplease`, `proc-macro2`, `syn`, `tracing` and `tracing-subscriber` are all unconditional again, the `#[cfg(feature = "runtime")]` / `#[cfg(feature = "macros")]` gates in `src/lib.rs`, `src/macros.rs` and `tests/log_buffer.rs` are removed, the workspace dependency no longer carries `default-features = false`, and all seven consumers are back to a plain `testing_aids = { workspace = true }`.

Side effect worth noting: this also fixed the `anvil-cargo-sort` failure, which was caused by the `[features]` table's placement and formatting.

crates/observed_macros/src/lib.rs
```diff
@@ -33,10 +29,9 @@ use proc_macro::TokenStream;
...
 #[cfg_attr(test, mutants::skip)] // a proc-macro entry point cannot be invoked from this crate's own tests
 pub fn event(attr: TokenStream, item: TokenStream) -> TokenStream {
-    match crate::event::event_attr(attr.into(), item.into()) {
-        Ok(tokens) => tokens.into(),
-        Err(err) => err.to_compile_error().into(),
-    }
+    observed_macros_impl::event(attr.into(), item.into())
+        .unwrap_or_else(|err| err.to_compile_error())
+        .into()
 }
```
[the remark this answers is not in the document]
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Correct, and fixed in 77ee0a33. `crates/observed_macros/src/{event,enrichment,field_attrs}.rs` and `src/snapshots/` were dead — `lib.rs` declares no `mod` for them — but the `/src/**` packaging allowlist would still have shipped them. All four paths are deleted; the shim is now `lib.rs` only.

scripts/mutants.rs
```diff
@@ -42,6 +42,7 @@ const TEST_GROUPS: &[&[&str]] = &[
     &["bytesbuf"],
     &["data_privacy", "data_privacy_core", "data_privacy_macros", "data_privacy_macros_impl"],
     &["fundle", "fundle_macros", "fundle_macros_impl"],
+    &["observed", "observed_macros", "observed_macros_impl"],
     &["ohno", "ohno_macros", "ohno_macros_impl"],
```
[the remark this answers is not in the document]
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Declining. Every group in `TEST_GROUPS` is `<crate>`, `<crate>_macros`, `<crate>_macros_impl`; none of them pulls in a sibling `*_testing` crate, and `observed` follows that shape exactly. Adding `observed_testing` here would make `observed` the one exception, and the mutation gate is scoped to the diff, so the cost would land on every run for no change in what the gate can catch. Changing the grouping convention is a separate call from this PR.

.github/workflows/main.yml
```diff
@@ -383,7 +383,7 @@ jobs:
         # already `#[cfg_attr(miri, ignore)]`, and the tests that do run cover the
         # runtime transcoding path over `bytes`/`hyper`, which is exactly the kind
         # of buffer handling Miri needs to check.
-        run: cargo +${{ env.RUST_NIGHTLY }} miri test --all-features --workspace ${{ needs.delta.outputs.exclude_not_affected }} --exclude data_privacy_macros_impl --exclude fundle_macros_impl --exclude internity_macros_impl --exclude multitude_macros_impl --exclude ohno_macros_impl --exclude routerama_build --exclude templated_uri_macros_impl --exclude thread_aware_macros_impl --lib --tests
+        run: cargo +${{ env.RUST_NIGHTLY }} miri test --all-features --workspace ${{ needs.delta.outputs.exclude_not_affected }} --exclude data_privacy_macros_impl --exclude fundle_macros_impl --exclude internity_macros_impl --exclude multitude_macros_impl --exclude observed_macros_impl --exclude ohno_macros_impl --exclude routerama_build --exclude templated_uri_macros_impl --exclude thread_aware_macros_impl --lib --tests
```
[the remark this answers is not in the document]
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Declining. The exclusion list is a cost policy that already covers every `*_macros_impl` crate in the workspace — `data_privacy`, `fundle`, `internity`, `multitude`, `ohno`, `templated_uri`, `thread_aware` — because interpreting `syn` parsing under Miri costs about 75 minutes of the step for behaviour that is pure computation over token trees, and any `unsafe` these macros *emit* is executed and checked in the consuming crate's tests, which stay in the run. `observed_macros_impl` follows that rule rather than becoming its exception.

What `tests/entry_points.rs` buys is separate: it keeps the non-snapshot assertions out of a file that must be `#![cfg(not(miri))]` because insta reads the workspace from disk. So the crate stays Miri-clean and `just miri` runs it locally on demand, without pinning ~75 minutes onto every CI run.

crates/observed_macros_impl/tests/entry_points.rs
```diff
@@ -0,0 +1,377 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Behaviour of the crate's two entry points, `event` and `derive_enrichment`.
+//!
+//! Everything the macros reject, plus the classification decisions that are easier to state
+//! as an assertion than to read out of a snapshot. The expansions themselves are snapshotted
+//! in `expansion.rs`; nothing here uses insta, so this file also runs under miri.
+
+#![cfg_attr(coverage_nightly, feature(coverage_attribute))]
```
Vaiz: `#![cfg_attr(coverage_nightly, feature(coverage_attribute))]` is not needed in `tests` folder, it's automatically excluded from coverage

crates/observed_macros_impl/tests/entry_points.rs
```diff
@@ -0,0 +1,377 @@
...
+
+use observed_macros_impl::{derive_enrichment, event};
+use proc_macro2::{Delimiter, Group, TokenStream, TokenTree};
+use quote::quote;
+use testing_aids::tokenize;
+
+#[cfg_attr(coverage_nightly, coverage(off))]
+#[cfg(test)]
+mod tests {
```
Vaiz: there is no point in extra submodule here

crates/observed_macros_impl/tests/entry_points.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Agreed, removed in f5076967 — both the crate-level `#![cfg_attr(coverage_nightly, feature(coverage_attribute))]` and the `#[cfg_attr(coverage_nightly, coverage(off))]` it existed to enable. Applied to `expansion.rs` as well, which carried the same pair.

crates/observed_macros_impl/tests/entry_points.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Agreed, flattened in f5076967 — the `mod tests` wrapper and its `use super::*;` are gone from both `entry_points.rs` and `expansion.rs`, along with the now-redundant `#[cfg(test)]` (an integration test file is already its own crate).

One knock-on: dropping the module changes insta's snapshot naming, so the fourteen `expansion__tests__*.snap` files are renamed to `expansion__*.snap` in the same commit. Content is byte-identical (git records all fourteen as 100% renames) and the 32 tests in the crate still pass.

crates/observed_macros_impl/tests/entry_points.rs
```diff
@@ -0,0 +1,369 @@
...
+
+    // `hits` satisfied the counter's unsigned-integer contract, so the numeric check saw
+    // through the group to the `u64`.
+    assert!(expanded.contains("InstrumentKind :: Counter"), "{expanded}");
+    // `maybe` was recognized as an `Option`, so `#[if_none(drop)]` was accepted and the
+    // drop arm generated instead of an unconditional visit.
+    assert!(expanded.contains("if let :: core :: option :: Option :: Some"), "{expanded}");
+    // `label` was recognized as a borrowed `str`, so the copy into an `Arc<str>` is
+    // spelled out rather than a plain conversion being emitted.
+    assert!(expanded.contains("Arc :: < str >"), "{expanded}");
```
Vaiz: assert that uses `contains` has smell of a bad tests. Consider using insta snapshots instead

Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Holding off on this one until you confirm the scope, because it cuts against the two-file split rather than just this line.

`entry_points.rs` exists specifically to hold the assertions that carry **no** snapshot; `expansion.rs` holds the insta ones and is `#![cfg(not(miri))]` because insta reads the workspace from disk. Turning these three into snapshots means moving the test into `expansion.rs`, so the split shrinks to whatever is left.

One correction to my own earlier reasoning, since it matters here: the crate is on the Miri exclusion list in `main.yml`, so `entry_points.rs` gets **no Miri coverage in CI** today — it is only Miri-clean for a local `just miri`. That makes the "keep it snapshot-free so Miri can run it" argument weaker than I stated on the thread above.

The case for keeping `contains` *in this particular test* is that each assertion pins one named consequence — the numeric check saw through the invisible group, the `Option` was still recognized, the borrowed `str` still produced the `Arc` copy — and a whole-expansion snapshot would bury those three facts in ~200 lines where a future reader cannot tell which part is load-bearing.

Which do you want?
1. This test only → snapshot, moved to `expansion.rs`.
2. Every `contains` in `entry_points.rs` → snapshots (the diagnostics tests assert on error text, so those become snapshots too).
3. Leave as is.

I have made no change and will do whichever you pick.

crates/testing_aids/src/macro_expansion.rs
```diff
@@ -0,0 +1,96 @@
...
+
+/// Parses Rust source text into a token stream, for feeding a macro entry point.
+///
+/// # Panics
+///
+/// Panics if `source` is not valid Rust tokens.
+#[must_use]
+pub fn tokenize(source: &str) -> TokenStream {
+    source.parse().expect("the source tokenizes")
+}
```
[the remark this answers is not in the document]
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Agreed, fixed in `5af7e769`. It now reports the lex error and the offending source:

```
the source tokenizes as Rust: {err}
--- source ---
{source}
```

which matches what `render_expansion` already did for tokens.

crates/testing_aids/src/macro_expansion.rs
```diff
@@ -0,0 +1,96 @@
...
+/// # Panics
+///
+/// Panics if `tokens` do not parse as a Rust file. A macro that emits something else is
+/// broken, so this is deliberately not a recoverable case.
+#[must_use]
+pub fn render_expansion(tokens: TokenStream) -> String {
+    let raw = tokens.to_string();
+    let file = syn::parse2(tokens).unwrap_or_else(|err| panic!("the expansion parses as a Rust file: {err}\n--- tokens ---\n{raw}"));
+    prettyplease::unparse(&file)
+}
```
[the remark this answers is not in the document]
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Correct — the eager `to_string()` ran on every call and only the panic arm read it. Fixed across two commits, with one correction to your suggestion.

`5af7e769` did exactly what you proposed (clone, stringify only in the panic arm), and that turned out to break `just clippy`: with the value no longer consumed, `clippy::needless_pass_by_value` fires (it is on via `pedantic` in the workspace lints). So `f07e3d1a` takes `&TokenStream` instead, which settles the lint and matches `render_tokens_lossy`, already by reference. `syn::parse2` needs an owned stream, so the parse still clones — but nothing is formatted on the successful path, which was your point.

Call sites in `observed_macros_impl`, `ohno_macros_impl`, `thread_aware_macros_impl`, `data_privacy_macros_impl` and `fundle_macros_impl` are updated. Full workspace suite is 6588/6588, and clippy, spellcheck, cargo-sort, ensure-no-default-features and nightly rustfmt are clean.

crates/testing_aids/src/macro_expansion.rs
```diff
@@ -0,0 +1,101 @@
...
+//! Helpers for snapshot-testing what a procedural macro expands to.
+//!
+//! For a procedural macro the generated code *is* the behaviour, so the natural test is to
+//! snapshot the whole expansion rather than probe it with substring assertions. Raw
+//! `TokenStream` text is unreadable in a diff, so an expansion is parsed back into a
+//! [`syn::File`] and pretty-printed before it reaches the snapshot.
+//!
+//! Both helpers panic rather than degrade. An expansion that no longer parses is a defect in
+//! the macro, and rendering it as raw tokens instead would hide that defect behind a snapshot
+//! that merely looks reformatted.
```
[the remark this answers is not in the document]
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Right, and thank you — that sentence was written when the module had two helpers and went stale when `render_tokens_lossy` arrived. Fixed in `ee3b78d7`: it now names `tokenize` and `render_expansion` as the strict pair and calls out `render_tokens_lossy` as the deliberate exception, for macro *input* only.

Vaiz:

Answering the suppressed finding in the latest Copilot review, on `crates/observed_macros_impl/src/event.rs:918` — duplicate `where` predicates from `event_predicates` when several fields share a generic type.

**The observation is correct.** In `expansion__a_generic_event_spells_out_the_bounds_its_body_relies_on.snap` the impl's `where` clause repeats `T: Send + Sync` three times, and `T: Display` and `T: RedactedDisplay` twice each. `event_predicates` pushes per field and never de-duplicates.

**Declining it in this PR, on scope.** This is pre-existing behaviour that the PR moved, not behaviour it introduced. `event_predicates` is byte-identical to `origin/main:crates/observed_macros/src/event.rs:894` — same body, same order, only the file path changed — and the reviewer itself files this under "previously missed, in code that hasn't changed since the last review".

This PR splits `observed_macros` into a shim plus `observed_macros_impl` and moves the shared snapshot helper into `testing_aids`. Changing what the macro emits — even to emit less of it — is a behaviour change to `observed` that belongs in its own PR, where the snapshot churn is the point of the diff rather than noise inside a refactor. Duplicate predicates are accepted by rustc and semantically harmless; the cost is expansion size and a little compile time.

Worth doing separately: de-duplicate while preserving first-seen order, and let the snapshots record the smaller output.

crates/observed_macros_impl/tests/entry_points.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

You were right and my reasoning was wrong. Done in `38cc0279`.

For the record, since I argued against this above: what I offered was that a `contains` line documents *which* consequence matters. That is a readability claim, and it is not one of this PR's goals. It also trades against the goal it looks like it serves — `contains` marks the branch covered while leaving the rest of the output free to be wrong, so it buys the coverage number without the proof. The snapshot is the stronger instrument. My second argument, that converting would shrink the two-file split, was defending file structure, which is not a goal either.

**What changed**

- Every `contains` / `assert_eq` probe in `entry_points.rs` is now a snapshot, and the file is deleted — it existed only to hold the assertions that carried no snapshot.
- A table-driven case joins its rows into one labelled snapshot (`==== #[gauge] ====` and so on), so **the test count is unchanged at 32** and one test still owns one snapshot. 17 new snapshot files, not ~90.
- Diagnostics now pin the whole message instead of a fragment.

**What that immediately caught:** `each_entry_point_rejects_input_it_cannot_parse` used to throw the error away entirely (`_ = …expect_err(item)`). The snapshot shows the two entry points do not agree — for `enum NotAStruct { A }`, `#[derive(Enrichment)]` says *"Enrichment can only be derived for structs, not enums"* while `#[event]` only manages *"expected `struct`"*. That asymmetry was invisible before and is now pinned; worth a look if you want `event` to say something better.

Two consequences you should know about:

1. `expansion.rs` is `#![cfg(not(miri))]`, so the crate now has no Miri-runnable tests at all. No CI coverage is lost — the crate is on the exclusion list in `main.yml` — but a local `just miri` now runs nothing for it.
2. The two biggest snapshots are the severity table (8.5 KB, six full expansions) and the `Option`-lookalike table (11.5 KB, seven). Trimming them to the interesting line would just be `contains` again, so I left them whole. Largest pre-existing snapshot was 7.3 KB, for scale.

Full workspace suite 6588/6588; clippy, spellcheck, cargo-sort, ensure-no-default-features and nightly rustfmt all clean.

crates/observed_macros_impl/tests/expansion.rs
```diff
@@ -0,0 +1,689 @@
...
+#![cfg(not(miri))]
+
+use observed_macros_impl::{derive_enrichment, event};
+use proc_macro2::{Delimiter, Group, TokenStream, TokenTree};
+use quote::quote;
+use testing_aids::{render_expansion, tokenize};
+
+fn expand_event(attr: &str, item: &str) -> String {
+    render_expansion(&event(tokenize(attr), tokenize(item)).expect("the event attribute expands"))
+}
```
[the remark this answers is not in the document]
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Declining, on three grounds.

**1. The stated premise is not this repo's convention.** `.expect(...)` / `.expect_err(...)` appear in more than sixty integration-test files across this workspace — `multitude`, `routerama`, `internity`, `rest_over_grpc`, `observed_testing`, `http_path_template`, `cachet`, `ohno` and others. And `clippy.toml` sets `allow-unwrap-in-tests = true` *specifically so that `unwrap` is also legal in tests*, with the comment "Test code asserts via unwrap()/panic!()". Both forms are permitted here. There is no "expect is production-code-only" rule to conform to, so this is a preference, not a convention.

**2. It reverses a change you asked for on this same PR.** On 2026-08-21 you flagged `testing_aids::tokenize`'s bare `.expect("the source tokenizes")` for not explaining the invariant or what the caller did wrong, and I agreed — `5af7e769` made it report the lex error and the offending source. These four helpers are the same shape: `expect("the event attribute expands")` says which of the two entry points failed and whether it failed in expansion or in rejection. A bare `unwrap` backtrace lands inside `syn` and says neither. I am not going to move in both directions on one file across three days.

**3. It serves none of this PR's goals** — splitting the crates, 100% coverage through the public API, the minimal number of tests that reaches it, and proof of no dead branches. The panic form changes nothing about what is covered or what is proven.

No change made. This also answers the three duplicates of the same point under "Suppressed comments" in the review body (lines 33, 43 and 454).

crates/observed_macros_impl/tests/expansion.rs
```diff
@@ -0,0 +1,689 @@
...
+
+fn expand_enrichment(item: &str) -> String {
+    render_expansion(&derive_enrichment(tokenize(item)).expect("the derive expands"))
+}
+
+fn event_error(attr: &str, item: &str) -> String {
+    event(tokenize(attr), tokenize(item))
+        .expect_err("the event attribute is rejected")
+        .to_string()
+}
```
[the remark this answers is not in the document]
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Declining, same reasons as [the `.expect` thread](https://github.com/microsoft/oxidizer/pull/686#discussion_r3841366811) — in short: `.expect_err` is used in test files throughout this workspace, `clippy.toml` deliberately permits *both* forms in tests (`allow-unwrap-in-tests = true`), and the request reverses a finding from this same reviewer that I accepted in `5af7e769`, which asked for panic messages to name the invariant rather than leave it to the backtrace.

Here the message is doing real work: `event_error` and `enrichment_error` both call `expect_err`, and `"the event attribute is rejected"` versus `"the derive is rejected"` is what tells you which entry point unexpectedly *succeeded* when a negative case regresses.

No change made.

crates/observed_macros_impl/tests/snapshots/expansion__a_log_message_may_only_reference_an_attribute_that_exists.snap
```diff
@@ -0,0 +1,122 @@
+---
+source: crates/observed_macros_impl/tests/expansion.rs
+expression: report(rejected.into_iter().chain(accepted))
+---
+==== rejects {absent} ====
```
Vaiz: from this snapshot it's not clear what is tested here

crates/observed_macros_impl/tests/snapshots/expansion__an_event_requires_named_fields.snap
Vaiz, line 1: it's better to keep such small snapshots as inline

crates/observed_macros_impl/tests/snapshots/expansion__both_macros_validate_the_shared_field_attributes_identically.snap
```diff
@@ -0,0 +1,111 @@
+---
+source: crates/observed_macros_impl/tests/expansion.rs
+expression: "report([r#\"#[dimension(\"positional\")] f: T\"#, r#\"#[dimension = \"x\"] f: T\"#,\n\"#[dimension()] f: T\", \"#[dimension(bogus)] f: T\",\n\"#[dimension(log = bogus)] f: T\",\nr#\"#[dimension(log = \"a\", log = \"b\")] f: T\"#,\n\"#[dimension(log = exclude, log = exclude)] f: T\",\nr#\"#[dimension(metric = \"a\", metric = \"b\")] f: T\"#,\n\"#[dimension(metric, metric)] f: T\", \"#[dimension] #[dimension] f: T\",\n\"#[unredacted(foo)] f: T\", r#\"#[unredacted = \"no\"] f: T\"#,\n\"#[unredacted] #[data_class(Euii)] f: T\",\n\"#[data_class(Euii)] #[unredacted] f: T\", \"#[if_none] f: Option<T>\",\n\"#[if_none(bogus)] f: Option<T>\",\n\"#[if_none(drop)] #[if_none(drop)] f: Option<T>\",\n\"#[if_none(drop)] f: T\",].into_iter().flat_map(|field|\n[(format!(\"#[event] on `{field}`\"),\nevent_error(r#\"\"e\"\"#, &format!(\"#[info] struct Subject {{ {field} }}\")),),\n(format!(\"Enrichment on `{field}`\"),\nenrichment_error(&format!(\"struct Subject {{ {field} }}\")),),]))"
```
Vaiz: expression is non-readable

crates/observed_macros_impl/tests/snapshots/expansion__both_macros_validate_the_shared_field_attributes_identically.snap
```diff
@@ -0,0 +1,111 @@
...
+---
+==== #[event] on `#[dimension("positional")] f: T` ====
```
Vaiz: can we have full input code here?

crates/ohno_macros_impl/tests/public_api.rs
```diff
@@ -22,24 +22,21 @@
...
+///
+/// The source side is rendered leniently: several cases deliberately feed a macro something that
+/// is not an item at all, and that input is still worth showing. The expansion side is rendered
+/// strictly, because a macro that emits unparsable tokens is broken.
 fn case(source: &TokenStream, expanded: &TokenStream) -> String {
     format!(
         "{}\n// ---- expands to ----\n\n{}",
-        pretty(source).trim_end(),
-        pretty(expanded).trim_end()
+        render_tokens_lossy(source).trim_end(),
```
Vaiz, line 38: why `lossy` is acceptable? is there a good reason for that?

crates/observed_macros_impl/tests/snapshots/expansion__a_log_message_may_only_reference_an_attribute_that_exists.snap
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Agreed — fixed in `49048c77`. The row label was a hand-written summary (`rejects {absent}`), which named the *placeholder* but never showed the struct that decides which attributes exist, so the message's "available attributes: [status, renamed]" had nothing to check it against.

Every row now records the whole source the macro was given, then what it produced:

```
---- input ----
#[event("e")]
#[info("{absent}")]
struct Subject {
    #[unredacted]
    status: i64,
    #[dimension(log = "renamed")]
    original: ClassifiedString,
    #[dimension(log = exclude)]
    request_id: ClassifiedString,
}

---- yields ----
log message references `{absent}` but no log attribute with that name exists; available attributes: [status, renamed]
```

So the rename and the exclusion that make `{original}` and `{request_id}` invalid are now visible in the snapshot itself, which is what the case is actually about.

crates/observed_macros_impl/tests/snapshots/expansion__an_event_requires_named_fields.snap
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Agreed — done in `49048c77`. This file and `expansion__an_instrument_belongs_on_the_struct_not_on_the_field.snap` were each a single line of diagnostic in a file of their own; both are deleted and asserted inline:

```rust
insta::assert_snapshot!(
    event_error(r#""e""#, "#[info] struct Tuple(#[unredacted] i64);"),
    @"#[event] can only be applied to structs with named fields"
);
```

Input and expected text now sit on adjacent lines, which is strictly better than a two-file round trip for one string.

I stopped at those two rather than inlining every small snapshot, because of your other comment on this push: now that each row carries its full input, the multi-row reports grew (the smallest, `an_enrichment_requires_a_struct_with_named_fields`, went from 578 B to ~1.4 KB). Inlining those would put more expected text in the test body than test. Say the word if you want the line drawn somewhere else.

crates/observed_macros_impl/tests/snapshots/expansion__both_macros_validate_the_shared_field_attributes_identically.snap
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Agreed — fixed in `49048c77`. The assertion was `assert_snapshot!(report(<the whole table literal and its iterator chain>))`, so insta recorded that entire expression as the header, escaped onto one logical line. Unreadable, and it churned on every reformat of the test.

The report is now bound before it is asserted, so the header is:

```
expression: output
```

That also matches `ohno_macros_impl/tests/public_api.rs`, which already did `let output = cases(...)`. Useful side effect: because the expression no longer contains the table, rustfmt can no longer invalidate the header — the regenerate/format/regenerate dance these snapshots needed is gone.

crates/observed_macros_impl/tests/snapshots/expansion__both_macros_validate_the_shared_field_attributes_identically.snap
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Yes — done in `49048c77`. The row showed only the field fragment (``#[event] on `#[dimension("positional")] f: T` ``), so the struct around it, and the fact that the `#[event]` row carries an extra `#[info]`, were both invisible. Now:

```
---- input ----
#[event("e")]
#[info] struct Subject { #[dimension("positional")] f: T }

---- yields ----
a positional string name is not supported; use `#[dimension(log = "...")]` to rename the log key

---- input ----
#[derive(Enrichment)]
struct Subject { #[dimension("positional")] f: T }

---- yields ----
a positional string name is not supported; use `#[dimension(log = "...")]` to rename the log key
```

The `#[event(...)]` / `#[derive(Enrichment)]` line is now part of the recorded input rather than a label I wrote by hand, so the two rows for one field are still adjacent and a divergence between the entry points is still a one-row diff — but nothing about which entry point ran is taken on trust from the label any more.

crates/ohno_macros_impl/tests/public_api.rs
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Good question, and the asymmetry is the whole point of the helper rather than a shortcut.

`case()` renders two different things. The **expansion** goes through `render_expansion`, which *panics* if the output does not parse as a Rust file — that is the bug this PR fixes. Both sides used to fall back to raw token text, so a macro that emitted unparsable garbage produced a snapshot of that garbage on one line and the test still passed. Nothing caught it.

The **source** side cannot be strict, because two cases deliberately hand the macro something that is not an item at all:

- `("an unparsable item", attributed(quote!(1 + 1)))` in `the_error_attribute_reports_what_it_cannot_rewrite` — source `#[ohno::error] 1 + 1`
- `("an unparsable item", enriched(TokenStream::new(), quote!(1 + 1)))` in `the_enrich_attribute_reports_what_it_cannot_enrich` — source `#[enrich_err] 1 + 1`

The behaviour under test is precisely that the macro *reports* that input instead of panicking on it. `syn::parse2::<File>` cannot parse `1 + 1`, so a strict renderer would panic on the **input** and kill the test before the macro's diagnostic was ever recorded — the tool would break the one case it exists to document.

So the leniency is scoped to the side where malformed input is the subject, and it is the only side where malformed input is legitimate:

- `render_tokens_lossy`'s rustdoc says input-only, and `ohno_macros_impl` is the only crate that calls it.
- `observed_macros_impl` does not use it anywhere — its inputs are all parsable items.
- Every expansion in the workspace goes through the strict renderer.

The alternative would be to drop the source line from those two rows, which costs the reader the input on exactly the cases where the input is the interesting part. I would rather keep it, but say so if you disagree and I will change it.

Vaiz:

Answering the "Suppressed comments" finding in [review 5005561631](https://github.com/microsoft/oxidizer/pull/686#pullrequestreview-5005561631) — a review body cannot take a threaded reply.

> `crates/testing_aids/src/macro_expansion.rs:90` — `catch_unwind` panic payloads are not guaranteed to be a `String` (they can be `&str` as well), so `downcast_ref::<String>()` makes this test brittle.

Declining. The general rule is right, but it does not apply to this call site.

**The payload type here is determined, not incidental.** `catch_unwind` returns `Box<dyn Any>` whose concrete type depends on how the panic was raised: `panic!("a literal")` boxes a `&'static str`, while `panic!` with formatting arguments boxes the formatted `String`. The panic this test catches is five lines above it in the same file:

```rust
syn::parse2(tokens.clone())
    .unwrap_or_else(|err| panic!("the expansion parses as a Rust file: {err}\n--- tokens ---\n{tokens}"));
```

That interpolates `{err}` and `{tokens}`, so the payload is always a `String`. There is no input to `render_expansion` that produces a `&str` payload, because there is only one panic site and it is a formatted one.

**And the failure mode is loud, not silent.** If someone later flattened that message to a literal, `.expect("the panic payload is a formatted string")` fails and names the invariant that broke. Accepting both types would instead let the test keep passing while quietly losing the thing it exists to check — that the offending tokens reach the message. Broadening the downcast would make this test *less* diagnostic, not more robust.

**`uniflight`'s `extract_panic_message` is not the same situation.** It handles panics from arbitrary caller-supplied closures, where the payload genuinely can be either and the code has no way to know. Here the panic site is in the same file as the test.

No change made.

martin-kolinek (approved):

sgalkin (approved):

PR #691, opened 2026-08-24
docs(ohno): add an example replacing a thiserror error enum

Vaiz:
## What

`#[derive(Error)]` rejects enums, which is the first question anyone migrating from `thiserror` asks, since its headline pattern is an enum with one variant per failure condition. The `Structs only` section of `docs/COMPARISON.md` answers it in one sentence, with nothing to look at.

Adds `crates/ohno/examples/error_enum_replacement.rs`, which ports one **real** six-variant enum — [`cargo_metadata::Error`](https://docs.rs/cargo_metadata/0.23.1/src/cargo_metadata/errors.rs.html#25-52), reproduced verbatim in the header — to one error type per failure condition behind a wrapper that lists them in `#[from(..)]`. The wrapper declares no `#[display]`, so it renders its cause verbatim and adds no text of its own.

The condition types stay **private**; the wrapper exposes `is_start_failure()` / `is_bad_output()` instead, per [M-ERRORS-CANONICAL-STRUCTS](https://microsoft.github.io/rust-guidelines/guidelines/libs/ux/index.html#M-ERRORS-CANONICAL-STRUCTS). Callers therefore neither name the condition types nor import `ErrorExt`, and conditions can be added or split without breaking them.

`docs/COMPARISON.md` points its `Structs only` section at the example.

## Notes

- No new dependency: `serde_json::Error` is replaced by a small local stand-in, labelled as such.
- The example header calls out the one porting detail that does not translate: `thiserror` interpolates the cause inline with `: {0}`, ohno renders it on its own `caused by:` line, so that suffix should be dropped when moving a message over.
- The `#[from]` list is written one per line as **repeated attributes**, not a multi-line list. rustfmt collapses `#[from(A, B, C)]` onto one line because it parses as a meta-list; separate `#[from(..)]` attributes accumulate (`parse_from_attribute` extends rather than replaces) and survive formatting.
- Actual program output is recorded at the bottom of the example.

## Validation

- `cargo clippy -p ohno --all-targets --all-features` with `-D warnings` — clean
- `cargo run -p ohno --example error_enum_replacement` — output matches the recorded block
- `just spellcheck` — clean
- rustfmt with `unstable-rustfmt.toml` on `nightly-2026-05-30` — clean

Relates to [AB#7675160](https://o365exchange.visualstudio.com/O365%20Core/_queries/edit/7675160). See the review thread on `lib.rs` — that bug also asked for crate-level docs, which this PR deliberately no longer adds.

### DIFF
--- a/crates/ohno/examples/error_enum_replacement.rs
+++ b/crates/ohno/examples/error_enum_replacement.rs
@@ -1,0 +1,169 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Replacing a `thiserror` error enum.
+//!
+//! Ohno derives on structs, not enums, so one variant per failure condition becomes one error type
+//! per condition, behind a wrapper that lists them in `#[from(..)]` so `?` converts. The wrapper
+//! declares no `#[display(..)]`, so it renders its cause verbatim and adds no text of its own.
+//!
+//! The condition types stay private and the wrapper answers `is_*()` questions about them, per
+//! [M-ERRORS-CANONICAL-STRUCTS][guideline]. Callers never name a condition, so conditions can be
+//! added or split without breaking them.
+//!
+//! Ported from [`cargo_metadata::Error`][upstream]. Note that `thiserror` interpolates the cause
+//! into the message with `{0}` while ohno renders it on its own `caused by:` line, so drop the
+//! trailing `: {0}` when moving a message over.
+//!
+//! [guideline]: https://microsoft.github.io/rust-guidelines/guidelines/libs/ux/index.html#M-ERRORS-CANONICAL-STRUCTS
+//! [upstream]: https://docs.rs/cargo_metadata/0.23.1/src/cargo_metadata/errors.rs.html#25-52
+
+use std::error::Error as StdError;
+use std::fmt;
+
+fn main() {
+    for error in [
+        metadata::run(),
+        metadata::parse(b"\xff"),
+        metadata::parse(b"no json here"),
+        metadata::rejected("error: no such command: `metadata`"),
+    ]
+    .map(|result| result.expect_err("every stub path fails"))
+    {
+        // The wrapper contributes no message of its own, so this prints the condition's text.
+        println!("{error}");
+
+        // Callers ask the error what happened, instead of matching on an exposed enum.
+        if error.is_start_failure() {
+            println!("  -> could not start the process, retriable");
+        } else if error.is_bad_output() {
+            println!("  -> the process ran but its output was unusable");
+        }
+    }
+}
+
+/// Stands in for `serde_json::Error`, so this example needs no third-party dependency.
+///
+/// Any foreign error type behaves the same way: name it in `#[from(..)]` and it becomes a source.
+#[derive(Debug)]
+struct JsonSyntaxError;
+
+impl fmt::Display for JsonSyntaxError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("expected value at line 1 column 1")
+    }
+}
+
+impl StdError for JsonSyntaxError {}
+
+mod metadata {
+    use std::error::Error as StdError;
+    use std::str::{Utf8Error, from_utf8};
+    use std::string::FromUtf8Error;
+
+    use super::JsonSyntaxError;
+
+    // One type per failure condition, each owning the message its enum variant used to carry.
+    // They stay private: nothing outside this module names them.
+
+    #[ohno::error]
+    #[display("`cargo metadata` exited with an error: {stderr}")]
+    struct CargoMetadataError {
+        stderr: String,
+    }
+
+    #[ohno::error]
+    #[from(std::io::Error)]
+    #[display("failed to start `cargo metadata`")]
+    struct StartError;
+
+    #[ohno::error]
+    #[from(Utf8Error)]
+    #[display("cannot convert the stdout of `cargo metadata`")]
+    struct StdoutError;
+
+    #[ohno::error]
+    #[from(FromUtf8Error)]
+    #[display("cannot convert the stderr of `cargo metadata`")]
+    struct StderrError;
+
+    #[ohno::error]
+    #[from(JsonSyntaxError)]
+    #[display("failed to interpret `cargo metadata`'s json")]
+    struct JsonError;
+
+    #[ohno::error]
+    #[display("could not find any json in the output of `cargo metadata`")]
+    struct NoJsonError;
+
+    /// The error this module returns.
+    ///
+    /// No `#[display(..)]`, so it renders its cause verbatim and stays invisible in the output.
+    #[ohno::error]
+    #[from(CargoMetadataError)]
+    #[from(StartError)]
+    #[from(StdoutError)]
+    #[from(StderrError)]
+    #[from(JsonError)]
+    #[from(NoJsonError)]
+    pub(crate) struct Error;
+
+    impl Error {
+        /// Returns `true` if `cargo metadata` could not be started at all.
+        pub(crate) fn is_start_failure(&self) -> bool {
+            // A failure condition is always the wrapper's immediate source.
+            self.source().is_some_and(<dyn StdError>::is::<StartError>)
+        }
+
+        /// Returns `true` if the process ran but its output could not be used.
+        pub(crate) fn is_bad_output(&self) -> bool {
+            self.source().is_some_and(|source| {
+                source.is::<StdoutError>() || source.is::<StderrError>() || source.is::<JsonError>() || source.is::<NoJsonError>()
+            })
+        }
+    }
+
+    /// Runs `cargo metadata` and returns its JSON output.
+    pub(crate) fn run() -> Result<String, Error> {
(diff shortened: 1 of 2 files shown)


crates/ohno/examples/error_enum_replacement.rs
```diff
@@ -0,0 +1,282 @@
...
+    #[display("could not find any json in the output of `cargo metadata`")]
+    pub(crate) struct NoJsonError;
+
+    /// The error this module returns.
+    ///
+    /// There is deliberately no `#[display(..)]`: an ohno error with a source and no message
+    /// override displays that source verbatim, so the wrapper is invisible in the output and
+    /// the reader sees the failure condition's own message.
+    #[ohno::error]
+    #[from(CargoMetadataError, StartError, StdoutError, StderrError, JsonError, NoJsonError)]
```
Vaiz: more canonical form is to use one line per error

crates/ohno/examples/error_enum_replacement.rs
```diff
@@ -0,0 +1,282 @@
...
+    }
+}
+
+impl StdError for JsonSyntaxError {}
+
+/// The default shape: one error type per failure condition, behind a wrapper that adds no text.
+///
+/// Each condition owns its message, so the messages stay exactly where the variants had them.
+/// The wrapper exists only to give the crate one error type to return.
+mod wrapper {
```
Vaiz: - keep only `wrapper` example, remove `kind`
- move main on top of the file, that's more convenient for human readers
- according to [M-ERRORS-CANONICAL-STRUCTS](https://microsoft.github.io/rust-guidelines/guidelines/libs/ux/index.html) errors should have a methods to verify their source
    impl HttpError {
        pub fn is_io(&self) -> bool { matches!(self.kind, ErrorKind::Io(_)) }
        pub fn is_protocol(&self) -> bool { matches!(self.kind, ErrorKind::Protocol) }
    }

crates/ohno/src/lib.rs
```diff
@@ -341,6 +341,122 @@
 //!
 //! **Note:** Error's fields must implement `Default` when using `#[from]` to ensure they can be properly initialized.
 //!
+//! # Multiple failure conditions
```
Vaiz: rollback lib.rs, it's enough to have an example which can be discovered by anyone who needs it by simply browsing the repo

crates/ohno/examples/error_enum_replacement.rs
Vaiz: Done, but not the way I first tried — worth recording why.

Writing the list one per line inside a single attribute does not survive formatting:

```rust
#[from(
    CargoMetadataError,
    StartError,
    ...
)]
```

rustfmt collapses that straight back onto one line, because `#[from(A, B, C)]` parses as a meta-list and gets normalized. (The `#[from(Type(field: expr))]` form in the other examples only survives because it *isn't* valid meta-item syntax, so rustfmt leaves it alone.)

Repeated attributes do survive, and `parse_from_attribute` supports them — it does `ast.conversions.extend(..)` per attribute with no "only one" guard, unlike `display`:

```rust
#[ohno::error]
#[from(CargoMetadataError)]
#[from(StartError)]
#[from(StdoutError)]
#[from(StderrError)]
#[from(JsonError)]
#[from(NoJsonError)]
pub(crate) struct Error;
```

Verified: builds, runs, and `rustfmt --check` is clean. Let me know if you meant the single-attribute form instead and would rather take the reflow — in that case it can't be enforced by the formatter.

crates/ohno/examples/error_enum_replacement.rs
Vaiz: All three done.

**Removed the `kind` module.** You were right beyond the redundancy argument: M-ERRORS-CANONICAL-STRUCTS says an inner `ErrorKind` "should not be exposed directly for future-proofing reasons", and my module did exactly that — a public `ErrorKind` plus a `kind()` accessor returning `&ErrorKind`. It was demonstrating the shape the guideline warns about, in an example whose job is to show people what to write. Gone.

**Moved `main` to the top**, right after the header doc.

**Added `is_*()` predicates.** Your snippet uses `matches!(self.kind, ..)`, which belongs to the kind shape; with that shape gone I implemented the same API over the source chain:

```rust
impl Error {
    pub(crate) fn is_start_failure(&self) -> bool {
        self.find_source::<StartError>().is_some()
    }

    pub(crate) fn is_bad_output(&self) -> bool {
        self.find_source::<StdoutError>().is_some()
            || self.find_source::<StderrError>().is_some()
            || ...
    }
}
```

This let the six condition types become **private** — nothing outside the module names them any more, so a condition can be added or split without breaking callers, which is the future-proofing the guideline is after. It also relocated the `ErrorExt` import note: instead of callers needing `use ohno::ErrorExt`, the error answers for itself and the import stays inside the module. Two predicates rather than six, grouped by what a caller would actually do about it (retry vs. don't).

Note this drops one thing the originating bug asked for — it explicitly wanted both shapes documented with "does anything actually branch on the category?" as the deciding question. The example now only shows the recommended one. I think that's the right call for an example, and the guideline link carries the reasoning, but flagging it so AB#7675160 isn't closed on the assumption it's fully covered.

crates/ohno/src/lib.rs
Vaiz: Rolled back. `crates/ohno/src/lib.rs` and the regenerated `crates/ohno/README.md` are both back to `origin/main` byte-for-byte — `git diff origin/main --name-status` now reports only:

```
M   crates/ohno/docs/COMPARISON.md
A   crates/ohno/examples/error_enum_replacement.rs
```

I kept the `COMPARISON.md` change, since that file's `Structs only` section is where a `thiserror` migrant actually lands and it previously answered this in one sentence with nowhere to go; it now links the example and M-ERRORS-CANONICAL-STRUCTS. Say the word if you'd rather that came out too and the example stood entirely on its own.

One consequence worth naming: AB#7675160 was filed as "neither the README nor the rustdoc answers it", so with `lib.rs` reverted the PR no longer addresses the bug as written — the answer now lives only in an example and a doc file that ships with the crate but isn't rendered on docs.rs. Happy to leave it there; just don't want the bug auto-closed on this PR without that being a deliberate choice.

crates/ohno/examples/error_enum_replacement.rs
```diff
@@ -0,0 +1,203 @@
...
+    #[from(JsonError)]
+    #[from(NoJsonError)]
+    pub(crate) struct Error;
+
+    impl Error {
+        /// Returns `true` if `cargo metadata` could not be started at all.
+        pub(crate) fn is_start_failure(&self) -> bool {
+            // `find_source` is a trait method, so `ohno::ErrorExt` has to be in scope. Answering
+            // here is what spares callers both that import and the private condition types.
+            self.find_source::<StartError>().is_some()
```
Vaiz: don't use `find_source` when u supposed to inspect only one error deep

Vaiz: Fixed. A condition is always the wrapper's *immediate* source, so walking the chain was both unnecessary work and wrong in principle — `find_source` would happily match a `StartError` buried further down someone else's chain and report it as this error's condition.

Now one level, one downcast:

```rust
impl Error {
    pub(crate) fn is_start_failure(&self) -> bool {
        self.condition().is_some_and(<dyn StdError>::is::<StartError>)
    }

    pub(crate) fn is_bad_output(&self) -> bool {
        self.condition().is_some_and(|condition| {
            condition.is::<StdoutError>()
                || condition.is::<StderrError>()
                || condition.is::<JsonError>()
                || condition.is::<NoJsonError>()
        })
    }

    /// The failure condition, which is always the wrapper's immediate source.
    fn condition(&self) -> Option<&(dyn StdError + 'static)> {
        self.source()
    }
}
```

Output is byte-identical before and after, which confirms the chain walk was finding nothing the single-level check doesn't. `use ohno::ErrorExt` is gone from the example entirely as a result — `source()` and `is::<T>()` are both std.

Also trimmed the header doc per your other note: 44 lines down to 16. The inline copy of the original `thiserror` enum is gone (the link to it remains), along with the `OhnoCore`-lives-in-one-place explanation, which belongs in the crate docs rather than repeated here.

Vaiz: No good reason — removed, now calling `self.source()` directly.

My intent was to name the concept: the header doc calls these "conditions", so `condition()` matched that vocabulary and gave me one place to hang the "always the immediate source" comment. That doesn't survive scrutiny. It was a private one-line method that forwarded to `source()` and nothing else, so it bought a layer of indirection and cost a reader the most recognisable name in the whole file — `source()` is std vocabulary every Rust reader already knows, and renaming it locally makes the example *less* readable, not more. The invariant it was carrying is one line of comment, which is where it now lives.

It was also straightforwardly at odds with the "keep the example doc concise" note — I added a method in the same change where I was trimming.

```rust
pub(crate) fn is_start_failure(&self) -> bool {
    // A failure condition is always the wrapper's immediate source.
    self.source().is_some_and(<dyn StdError>::is::<StartError>)
}
```

Output unchanged; clippy, rustfmt and spellcheck clean.

wukchung (approved):

martin-kolinek (approved):

PR #692, opened 2026-08-24
fix(multitude): compile the test suite on alignment-capped codegen backends

wukchung:
&#x1F916; This description was written by an AI agent.

## Problem

`cargo test -p multitude` fails to compile on codegen backends whose maximum
supported type alignment is lower than the alignments the tests deliberately
construct. Several tests declare types aligned to 32 KiB, 64 KiB and 128 KiB in
order to reach the arena's over-alignment guards, and a backend that cannot lay
out those types rejects the test crate outright.

The mitigation for this was already in the tree, and inert. Five test files gate
the affected tests on a cfg, and the workspace pre-declares that cfg in its
`check-cfg` list, but nothing anywhere set it: `multitude` had no build script,
and no `.cargo/config.toml`, justfile recipe or CI variable passed it. Every
gate was therefore permanently false and skipped nothing.

The `check-cfg` entry is what let this survive: without it, `unexpected_cfgs`
would have warned on every gate and made the dangling cfg obvious.

This is invisible to this repository's CI, which only builds with toolchains
that have no such cap, so the failure reaches developers locally instead.

## Change

**1. A build script sets the cfg by probing the backend's capability.**

There is no built-in cfg identifying the codegen backend
([upstream issue](https://developercommunity.visualstudio.com/t/Conditional-compilation-lacks-built-in-c/11107823)),
so the script probes the limit that actually matters rather than inferring it
from a toolchain name. That keeps the gate correct if a backend raises its cap,
or if another backend adopts the same restriction.

Four properties are load-bearing:

* **The probe must materialize a place of the over-aligned type.** A backend
  that caps alignment rejects it when laying out a value, not when the type
  merely appears in a reference parameter, so a slice-taking function compiles
  cleanly even where the gated tests do not. The probe returns the value by
  value, and is `pub` so the standalone library has to generate it. `Probe`
  stays a ZST so no over-aligned stack frame is required, and an over-aligned
  `static` is deliberately avoided because it trips the COFF section-alignment
  limit and crashes stock `rustc`, which would set the cfg on every Windows
  host.

* **A control probe separates "backend rejected it" from "could not run here".**
  A non-zero `rustc` exit is not by itself evidence of an alignment cap: a
  target whose `core` is not installed, a custom target spec, or an unusable
  sysroot all fail the same way, and misreading one would silently disable the
  gated tests on a healthy toolchain. A control probe at `align(8)` runs first,
  and the cfg is set only when the control compiles and the over-aligned probe
  does not. Failure to spawn `rustc` at all also leaves the tests enabled.

* **The probe is compiled the way the crate is.** Cargo's own rustc flags are
  forwarded to it, because `-Zcodegen-backend` selects the very thing being
  detected, and the script reruns when `RUSTC`, `RUSTUP_TOOLCHAIN` or those
  flags change so a cached result cannot outlive the compiler it measured.

* **`PROBE_ALIGN` is the largest gated alignment (131072).** Probing a smaller
  value would leave the cfg unset on a backend that accepts the probe but still
  rejects a larger gated type. The value is asserted to be an expressible
  `repr(align(...))` before `rustc` runs, so a mistyped constant fails the build
  instead of being misread as a backend cap.

`/build.rs` is already in the `[workspace.package]` `include` allowlist, so the
script ships with the published crate.

**2. The cfg is renamed to `align_capped_backend`,** which describes the
constraint being detected rather than any one backend. That is what the gates
and the probe actually test for.

**3. The cfg is declared where it is emitted.** The build script issues
`cargo::rustc-check-cfg`, which covers every unit of the crate, so the
crate-local entry in the workspace `check-cfg` allowlist is removed and that
anvil-managed region returns to its catalog default.

**4. The over-aligned doctest in `error.rs` is gated too.** The example for
`is_alignment_too_large` constructs a 32 KiB-aligned type, so `cargo test
-p multitude` still failed to compile on a capped backend with only the
integration tests gated. It is now rendered but not compiled there.

## Validation

Run against the current tree.

| check | result |
| --- | --- |
| `cargo test -p multitude` (stock toolchain) | 1274 passed, 0 failed |
| `cargo test -p multitude` (alignment-capped backend) | 1254 passed, 0 failed, 1 ignored |
| same, with `--cfg align_capped_backend` forced on | 1255 passed, 0 failed, compiles clean |
| `cargo clippy -p multitude --all-targets --all-features` | no warnings |
| `cargo fmt -p multitude -- --check` | no changes |

The capped-backend row is the one that matters: the crate now compiles and its
whole suite runs there, where before it failed to build. The forced-on run shows
the gates are complete on a stock toolchain as well.

## Reviewer notes

`PROBE_ALIGN` must stay in sync with the largest `#[repr(align(...))]` guarded
by the cfg; it is documented as such in the build script.

The gate is a single boolean driven by the largest gated alignment, so a backend
that rejects it also disables the 32 KiB and 64 KiB tests that might still have
compiled. Reviewer feedback proposing a per-threshold cfg is open.

WI: https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7707893


crates/multitude/build.rs
```diff
@@ -0,0 +1,119 @@
...
+/// probe but still rejects a larger gated type.
+const PROBE_ALIGN: u32 = 131_072;
+
+/// Alignment used by the control probe, in bytes.
+///
+/// Any backend can satisfy this, so a control failure means the probe could
+/// not be compiled here for reasons unrelated to alignment.
+const CONTROL_ALIGN: u32 = 8;
+
+fn main() {
```
Vaiz, line 55: is it possible to limit the execution of this build script only to test runs? that way we won't do any awkward diagnostic when this crate is used as a dependency

another option is to move those tests into a separate crate

crates/multitude/build.rs
```diff
@@ -0,0 +1,119 @@
...
+    println!("cargo::rerun-if-changed=build.rs");
+    println!("cargo::rerun-if-env-changed=RUSTC");
+    println!("cargo::rerun-if-env-changed=RUSTUP_TOOLCHAIN");
+
+    if !supports_probe_alignment() {
```
Vaiz, line 72: have you tried to replace it with `for cap in [32kib, 64kib, 128kib]` ?

that way we can set `cargo::rustc-cfg=align_capped_backend={cap}` and then enable some of the tests. [link](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-cfg)

crates/multitude/build.rs
```diff
@@ -0,0 +1,119 @@
...
+const PROBE_ALIGN: u32 = 131_072;
+
+/// Alignment used by the control probe, in bytes.
+///
+/// Any backend can satisfy this, so a control failure means the probe could
+/// not be compiled here for reasons unrelated to alignment.
+const CONTROL_ALIGN: u32 = 8;
+
+fn main() {
+    println!("cargo::rerun-if-changed=build.rs");
```
martin-kolinek, line 60: 🤖: Non-blocking — now that this crate has a build script, the cfg can be *declared* where it is emitted:

```rust
println!("cargo::rustc-check-cfg=cfg(align_capped_backend)");
```

That covers every unit the cfg can reach (lib, tests, benches, examples of `multitude` — and all the gate sites are under `crates/multitude/tests/**`), which would let the workspace allowlist entry at the repo-root `Cargo.toml:290` go away. That entry is a crate-local name sitting inside the anvil-managed `[workspace.lints]` region, whose own comment notes that a customized line there is owned by the adopter and gets a `.anvil-proposed` sibling on every future catalog bump. Retiring the customization in the same PR that makes it unnecessary keeps the shared region at its catalog default and puts the declaration next to the `cargo::rustc-cfg` that produces it. The current form works; this is purely a tidy-up.

crates/multitude/tests/arena.rs
```diff
@@ -6063,13 +6063,13 @@ mod public_surface_behavior {
     // including Windows, whose default 1 MiB stack can't accommodate the
     // 128 KiB-aligned frame the guarded body would otherwise require.
     //
-    // Skipped under the UTC codegen backend (`--cfg utc_backend`): UTC caps
-    // type alignment at 8192 bytes, well below the 128 KiB this test needs.
-    #[cfg(not(utc_backend))]
+    // Skipped on codegen backends that cap type alignment below the value
+    // this test requires (`--cfg align_capped_backend`).
```
martin-kolinek: 🤖: Non-blocking — the second commit falsifies this parenthetical. It still describes the gate as something the invoker supplies (`--cfg align_capped_backend`), but after `build.rs` lands the cfg is auto-detected by the build script and is never passed on the command line. Since this is now the only prose in the repo that explains where the cfg comes from, it's worth saying so — e.g. "set automatically by `build.rs`, which probes whether the active codegen backend accepts a 128 KiB-aligned type".

martin-kolinek (approved):
🤖: Approved.

crates/multitude/build.rs
```diff
@@ -0,0 +1,119 @@
...
+/// `#[repr(align(...))]` values guarded by `cfg(align_capped_backend)`; probing a
+/// smaller value would leave the gate unset on a backend that accepts the
+/// probe but still rejects a larger gated type.
+const PROBE_ALIGN: u32 = 131_072;
+
+/// Alignment used by the control probe, in bytes.
+///
+/// Any backend can satisfy this, so a control failure means the probe could
+/// not be compiled here for reasons unrelated to alignment.
+const CONTROL_ALIGN: u32 = 8;
```
sandersaares, line 48: [Copilot speaking]

# Document alignment-capability test gating in the implementation guide

The new build script adds a crate-wide verification mechanism: it first compiles a control type for the target, then compiles a type aligned to the largest alignment used by the gated tests, and emits `cfg(align_capped_backend)` only when the control succeeds and the over-aligned probe does not. That cfg conditionally removes the guarded over-alignment tests from their test binaries. The published, nontrivial `multitude` crate has an extensive design document describing alignment as a fragile invariant and integration tests as part of its verification model, but it has no implementation guide describing this mechanism or its cross-file maintenance constraints.

**Reproducible reasoning**: The control probe, target capability probe, conservative fallback, emitted cfg, and synchronization with the largest guarded `repr(align(...))` value are high-level implementation decisions rather than local command details. The mechanism spans `build.rs`, the custom-cfg declaration in the workspace manifest, and tests covering 32 KiB, 64 KiB, and 128 KiB alignments. The build script documents its local decisions, but the design document's testing section presents the crate-wide verification model without explaining that backend capability determines whether this coverage is compiled. The architecture documentation therefore does not provide one place where a maintainer can discover the full maintenance relationship among the probe, cfg declaration, and gated tests.

**Consequence**: A maintainer can add a larger guarded alignment, change the probe or fallback policy, or alter the gated test set without coordinating all participating files. The probe can then accept a backend that cannot compile the largest test type, or suppress more verification than intended, while the documented verification model overstates the alignment coverage available on that backend.

**Recommended action**: Add `crates/multitude/docs/implementation.md`, scoped at minimum to build-time alignment-capability detection, and link it from the testing section of `crates/multitude/docs/DESIGN.md`. Document the control and capability probes, the conservative classification when the control succeeds but the over-aligned compilation fails, where `align_capped_backend` is declared and emitted, which over-alignment coverage it gates, and the invariant that the probe alignment matches the largest alignment used by every guarded test. Keep command-level details in code unless they are necessary to preserve these constraints.

**References**:

- crates/multitude/docs/DESIGN.md, Fragile invariants and Testing and verification
- crates/multitude/build.rs, PROBE_ALIGN and capability-probe documentation
- [Cargo build scripts](https://doc.rust-lang.org/cargo/reference/build-scripts.html)

**Impacted locations**:

- `Cargo.toml:290`
- `crates/multitude/build.rs:1-119`
- `crates/multitude/docs/DESIGN.md:781-873`
- `crates/multitude/tests/arena.rs:2244-2253`
- `crates/multitude/tests/arena.rs:6066-6072`
- `crates/multitude/tests/arena.rs:6901-6983`

crates/multitude/build.rs
```diff
@@ -0,0 +1,119 @@
...
+
+fn main() {
+    println!("cargo::rerun-if-changed=build.rs");
+    println!("cargo::rerun-if-env-changed=RUSTC");
+    println!("cargo::rerun-if-env-changed=RUSTUP_TOOLCHAIN");
```
sandersaares, line 68: [Copilot speaking]

# Explain why toolchain-selection changes invalidate the probe

The build script asks Cargo to rerun when `RUSTUP_TOOLCHAIN` changes, but the surrounding code does not explain why this variable matters in addition to `RUSTC`. This is a non-obvious cache-invalidation decision because `RUSTUP_TOOLCHAIN` is not otherwise read by the script.

**Reproducible reasoning**: The probe result is cached through Cargo's build-script change tracking. A rustup proxy can keep the same `RUSTC` command while `RUSTUP_TOOLCHAIN` selects a different compiler, and the selected compiler can have different alignment support. Therefore the extra rerun trigger protects the capability result from becoming stale, but a maintainer cannot reproduce that rationale from the current code because only the directive is present.

**Consequence**: A future maintainer may remove the apparently unused trigger or add related triggers inconsistently, leaving `align_capped_backend` based on a compiler selected before the toolchain change.

**Recommended action**: Add a brief comment above the rerun directives explaining that the probe must be invalidated when the compiler command or rustup-selected toolchain changes, including that the rustup proxy path may remain unchanged while its selected compiler changes.

**References**:

- [Cargo build-script change detection](https://doc.rust-lang.org/cargo/reference/build-scripts.html#change-detection)

**Impacted locations**:

- `crates/multitude/build.rs:46-47`

crates/multitude/build.rs
```diff
@@ -0,0 +1,119 @@
...
+    let rustc = env::var_os("RUSTC").unwrap_or_else(|| "rustc".into());
+    let mut command = Command::new(rustc);
+    command
+        .arg("--crate-type=lib")
+        .arg("--emit=obj")
+        .arg("-o")
+        .arg(&object_path)
+        .arg(&source_path)
+        .stdout(Stdio::null())
+        .stderr(Stdio::null());
```
sandersaares, line 176: [Copilot speaking]

# Explain why probe diagnostics are discarded

The compiler probe redirects both standard output and standard error to the null device without explaining the decision. Suppression is consequential here because failure of the over-aligned compilation is an expected capability result, while a failed control probe causes the script to leave the tests enabled.

**Reproducible reasoning**: Silencing the child compiler prevents an expected over-alignment rejection from appearing as an alarming compiler error during an otherwise successful Cargo build. The same redirection also hides the compiler's reason for any nonzero exit. After the control succeeds, every nonzero exit from the over-aligned probe is classified as an alignment limitation, so the suppression is part of the probe's user-visible diagnostic policy rather than incidental process setup. The surrounding comments explain how exit results are classified but not why both output streams are intentionally hidden.

**Consequence**: A maintainer can remove the redirection and expose expected probe errors during ordinary builds, or preserve it without recognizing that the emitted cfg is the only normal indication of the classified result.

**Recommended action**: Add a brief comment before the redirections explaining that compiler output is suppressed because rejection of the capability probe is an expected result and should not pollute normal build output.

**Impacted locations**:

- `crates/multitude/build.rs:109-110`

crates/multitude/build.rs
```diff
@@ -0,0 +1,119 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Detects codegen backends that cap type alignment below what the tests need.
+//!
+//! Not every codegen backend accepts the maximum type alignment the reference
+//! implementation does. Several tests deliberately construct types aligned to
+//! half a chunk to exercise the arena's alignment handling, well above the cap
+//! such a backend imposes, so those tests cannot be compiled there at all. They
```
sandersaares: [Copilot speaking]

# Reconcile the capability gate with its module documentation

The module documentation says that several tests construct half-chunk-aligned types "well above" the cap imposed by an alignment-capped backend. The build script does not establish that relationship: it sets one boolean cfg when the backend rejects the 128 KiB probe, while that cfg also guards tests using 32 KiB and 64 KiB alignments. The same paragraph compares backends with an undefined "reference implementation," although the gate is based on a fixed test requirement rather than an identified implementation.

**Reproducible reasoning**: The prose at lines 6-9 characterizes every gated test in terms of half-chunk alignment and states that this alignment exceeds the detected backend cap. `PROBE_ALIGN` at lines 28-35 instead records only whether the largest gated alignment, 128 KiB, compiles. The guarded tests include 32 KiB, 64 KiB, and 128 KiB types, so a backend can trigger the cfg while still accepting some of the smaller guarded alignments. The probe establishes that the complete gated set cannot compile, not that every individual gated alignment is unsupported. In addition, neither this file nor the crate documentation defines whether "reference implementation" means rustc, its default LLVM backend, a stock toolchain, or something else.

**Consequence**: On a backend that accepts 32 KiB or 64 KiB but not 128 KiB, readers can incorrectly conclude that every skipped test is unrepresentable there. The undefined comparison baseline can also make the cfg appear to depend on a particular implementation instead of the fixed maximum alignment represented by `PROBE_ALIGN`.

**Recommended action**: Reconcile the opening documentation and the gate semantics so they make the same claim. If the boolean gate remains intentionally coarse, describe it as suppressing a set that cannot all compile when the maximum alignment is rejected; if each suppressed test is intended to be known unsupported, refine the capability detection and gates by alignment threshold. In either direction, replace "reference implementation" with direct wording that identifies the fixed test requirement and the active codegen backend.

**References**:

- crates/multitude/docs/DESIGN.md, Thin smart pointers: the alignment/masking trick

**Impacted locations**:

- `crates/multitude/build.rs:5-9`
- `crates/multitude/build.rs:28-35`
- `crates/multitude/tests/arena.rs:2244-2253`
- `crates/multitude/tests/arena.rs:6901-6983`

crates/multitude/build.rs
```diff
@@ -0,0 +1,119 @@
...
+/// Returns `true` whenever the answer cannot be established, so an unexpected
+/// environment leaves the tests enabled rather than silently skipping them.
+fn supports_probe_alignment() -> bool {
+    let Some(out_dir) = env::var_os("OUT_DIR") else {
+        return true;
+    };
+    let out_dir = Path::new(&out_dir);
+
+    // Only a control that compiles makes a failure of the real probe
+    // attributable to the alignment cap rather than to the environment.
```
sandersaares: [Copilot speaking]

# Use the established over-aligned probe term consistently

The build-script documentation distinguishes the control probe from the alignment-capability probe, which it otherwise calls the "over-aligned probe." The implementation comment instead calls the latter the "real probe," introducing an informal third label for the same concept.

**Reproducible reasoning**: The module documentation first defines a control probe at a benign alignment and contrasts it with an over-aligned probe. At the branch that interprets their results, the phrase "real probe" refers to that same over-aligned probe. Reusing the established term makes the two probe roles immediately recognizable and avoids implying that the control is somehow not a genuine probe.

**Consequence**: A reader tracing the control logic may briefly have to determine whether "real probe" names another probe or is an informal synonym for the documented over-aligned probe.

**Recommended action**: Replace "real probe" with "over-aligned probe" in the comment so the terminology remains precise and consistent.

**Impacted locations**:

- `crates/multitude/build.rs:64-65`

crates/multitude/build.rs
```diff
@@ -0,0 +1,119 @@
...
+    // No attributes beyond `repr`/`inline` are used, so the source stays valid
+    // regardless of which edition `rustc` defaults to.
+    let source = format!(
+        "#![no_std]\n\
+         #[repr(align({align}))]\n\
+         pub struct Probe;\n\
+         #[inline(never)]\n\
+         pub fn probe() -> Probe {{ Probe }}\n"
+    );
+
```
sandersaares, line 157: [Copilot speaking]

# Remove the redundant no-inline attribute from the probe

The generated probe source applies `#[inline(never)]` to a public function that has no callers. The public function is compiled as an externally reachable library symbol, which is what makes rustc generate the by-value return ABI for the over-aligned `Probe`; with no call site, preventing inlining does not contribute to that mechanism.

**Reproducible reasoning**: The probe compiles a standalone library directly to an object. `pub fn probe() -> Probe` must remain externally reachable because an unused private function can be discarded before code generation, but no code in the generated crate calls `probe`, so there is no inlining opportunity for `#[inline(never)]` to control. The comments correctly explain why returning an aligned zero-sized value exercises backend layout without an over-aligned stack frame, but they do not identify public reachability as the retention mechanism and instead leave an unrelated code-generation attribute in place.

**Consequence**: A maintainer can mistake `#[inline(never)]` for the load-bearing part of the probe and remove or narrow the public visibility that actually keeps the function available for code generation. The redundant attribute also creates an additional compiler-facing detail that future changes have to reason about.

**Recommended action**: Remove `#[inline(never)]` and add an adjacent explanation that the function is public so the standalone library must expose and generate the by-value return ABI. Preserve the existing direct capped- and uncapped-backend probe validation when making this simplification.

**Impacted locations**:

- `crates/multitude/build.rs:80-98`

crates/multitude/build.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    compile_probe(out_dir, PROBE_ALIGN).unwrap_or(true)
+}
+
+/// Compiles a crate containing a type aligned to `align`.
+///
+/// Returns `Some(true)` when codegen accepted it, `Some(false)` when `rustc`
+/// ran and rejected it, and `None` when the probe could not be run at all.
+fn compile_probe(out_dir: &Path, align: u32) -> Option<bool> {
```
sandersaares, line 126: [Copilot speaking]

# Validate probe alignments before classifying compiler failure

`compile_probe` accepts its `align` requirement as an unconstrained `u32`, although the value is interpolated into `repr(align(...))` and therefore must be nonzero, a power of two, and within Rust's supported alignment range. The current constants satisfy that contract, but the function does not distinguish an invalid source-language alignment from the backend limitation it is intended to detect.

**Reproducible reasoning**: The control probe distinguishes an unusable compiler environment from an over-aligned compilation failure. Once the control succeeds, however, any successfully spawned over-aligned probe that returns a nonzero status causes `supports_probe_alignment` to emit `align_capped_backend`. If `PROBE_ALIGN` is later changed to zero, a non-power-of-two value, or a value above Rust's allowed `repr(align)` maximum, rustc rejects the generated attribute for a source-language reason and the script misclassifies that rejection as a backend capability limit. A small local invariant check can prevent that state without introducing a new domain type for two private constants.

**Consequence**: A mistyped probe constant can silently disable the over-alignment tests across otherwise capable backends, making the build script report a backend limitation that does not exist.

**Recommended action**: Validate the alignment before invoking rustc, using an unconditional assertion that it is a nonzero power of two and no greater than Rust's documented `repr(align)` maximum. Failing the build at the invalid declaration is preferable to emitting a misleading capability cfg and is simpler than adding a dedicated validated type.

**References**:

- [Rust Reference: alignment modifiers](https://doc.rust-lang.org/reference/type-layout.html#the-align-and-packed-modifiers)

**Impacted locations**:

- `crates/multitude/build.rs:36-42`
- `crates/multitude/build.rs:77`

sandersaares (approved):
[Copilot speaking]

Published 7 findings. No finding follows up on an existing discussion thread.

| Diagnostic | Value |
|------------|-------|
| Cache | Hit |

</details>

wukchung:
&#x1F916; Closing this.

Carrying it to a good conclusion needs judgement about the arena's alignment
guards and about the design trade-offs raised in review that I am not well
placed to make. It is better picked up by someone with deeper context on this
crate than left open against my name.

Thanks to everyone who reviewed; the feedback was substantive and none of it is
wasted. The tracking work item stays open, with the review threads summarised on
it for whoever takes it forward.

PR #693, opened 2026-08-24
Design portable fetch transport API

martin-kolinek:
🤖 Defines the proposed stable API boundary for `fetch` before implementation work begins.

## Summary

- designs one concrete, transport-erased `HttpClientBuilder` around portable library requirements
- stabilizes only the portable requirement surface and a generic typed transport-config registry
- defines `fetch_hyper_common` as a reusable TLS-neutral HTTP engine
- composes Rustls and native TLS through dependency-isolated `fetch_hyper_rustls` and `fetch_hyper_native_tls` crates
- keeps backend-typed verifier, signer, provider, and connector APIs with their composition crates
- introduces dependency-light config companions only after a demonstrated library need
- unifies mTLS selection through application-provisioned logical credential identifiers
- models independent network destination, exact TLS server name, and HTTP authority
- removes inherited socket and HTTP/2 tuning knobs from the library-facing surface
- gives portable HTTP/1.1/2 constraints precedence over WinHTTP''s optional HTTP/3 preference
- defines fallible request/response trailer semantics and fetch-owned response decompression
- requires full-duplex HTTP/2 and includes a native WinHTTP proof for known and unknown-length uploads

## Backend evidence

- adds a WinHTTP HTTP/2 probe demonstrating independent DNS routing, SNI/certificate validation, and `:authority`
- adds a calibrated delayed-ACK experiment showing tested WinHTTP HTTP/1.1 small writes behave like the `TCP_NODELAY` control
- records where WinHTTP behavior is measured rather than covered by a documented compatibility guarantee

## Tuning policy

- Nagle is not configurable; socket-owning transports enable `TCP_NODELAY`
- HTTP/2 receive-window sizing remains transport-owned
- kernel send/receive buffers remain under OS autotuning
- initial congestion behavior remains OS/network policy


crates/fetch/docs/design/README.md
```diff
@@ -0,0 +1,218 @@
...
+```rust,ignore
+let transport = fetch::transport::winhttp()
+    .proxy(proxy)
+    .integrated_authentication(true);
+
+let builder = fetch::HttpClient::builder(transport);
+let client = service_library::build_client(builder)?;
+```
+
+The library accepts any builder whose transport implements the semantic capabilities it needs. It
```
martin-kolinek: All transport need to implement all semantic capabilities, right?

crates/fetch/docs/design/README.md
```diff
@@ -0,0 +1,218 @@
...
+lowest-common-denominator options bag.
+
+For example, connection maximum lifetime means that a connection is not selected for a new request
+after the configured age. Hyper can enforce that contract by retiring or poisoning pooled
+connections. WinHTTP can enforce it with `WINHTTP_OPTION_EXPIRE_CONNECTION`. The mechanism differs;
+the guarantee does not.
+
+When no faithful common contract exists, the option remains transport-specific. Coarse or partial
+support is not silently treated as success. A weaker behavior requires a separately named portable
+contract or an explicit preference API with observable resolution results.
```
martin-kolinek: What does this mean? What is an explicit preference AP?

crates/fetch/docs/design/README.md
```diff
@@ -0,0 +1,218 @@
...
+
+## Transport-owned performance policy
+
+The stable API exposes service requirements, not copies of socket and protocol-stack knobs.
+Supported transports choose and validate defaults for HTTP flow control, kernel buffering, and
+congestion behavior.
+
+Small writes are the exception because avoiding Nagle/delayed-ACK stalls is a general HTTP client
```
martin-kolinek: How are they the exception? The configuration is not exposed to applications, the transports make the decision.

crates/fetch/docs/design/README.md
```diff
@@ -0,0 +1,218 @@
...
+uses adaptive flow control. Kernel send and receive buffers remain under operating-system
+autotuning. Initial congestion behavior remains operating-system policy. None is configurable
+through `HttpClientBuilder`.
+
+## TLS and credentials
+
+TLS backend selection and backend-native customization are transport-specific. Portable security
+requirements remain on `HttpClientBuilder` because libraries may own them.
+
+Client-certificate authentication uses a logical credential identifier. Libraries name the
```
martin-kolinek: This paragraph could use a rewrite and/or an example. This is hard to understand without prior knowledge about our previous discussion.

crates/fetch/docs/design/README.md
```diff
@@ -0,0 +1,218 @@
...
+when constructing the transport. A Windows application may bind the name to a certificate-store
+selector, while a Linux application may bind it to provisioned key material. The library does not
+observe either modality.
+
+A named binding may represent a set of rotating certificates. Rustls and WinHTTP can select a
+certificate using issuer hints received during the handshake. The current native-TLS adapter must
+resolve the binding to one identity when constructing the connector and therefore requires a client
+rebuild to pick up rotation.
+
+Portable server validation is split into platform chain trust and endpoint identity. Platform
```
martin-kolinek, line 268: Also could use an example

crates/fetch/docs/design/capability-matrix.md
```diff
@@ -0,0 +1,149 @@
...
+| Custom verifier callback | Supported | No equivalent current `fetch_tls` API | No userspace callback | Transport-specific |
+| Exact TLS server-name override while preserving request authority | Connector dials the request endpoint and supplies the override to rustls | Connector dials the request endpoint and supplies the override to native TLS | `WinHttpConnect` uses the TLS name, resolution override uses the endpoint, and replaced `Host` preserves authority | Baseline |
+| Custom SAN/subject server-identity policy | Enforced before application data by a verifier | No equivalent current adapter | Server certificate is queryable only after TLS negotiation; equivalent pre-disclosure enforcement is unproven | Add a semantic capability only if exact-name mapping cannot replace the TVS rules |
+| Certificate or public-key pins | Implementable in a verifier | No equivalent current adapter | Certificate context is queryable after negotiation | Separate transport-specific feature until a safe portable contract exists |
```
martin-kolinek: Such a contract can never exist. Either needs to be removed (libraries don't need it), or this whole design falls apart.

crates/fetch/docs/design/capability-matrix.md
```diff
@@ -0,0 +1,149 @@
...
+| Arbitrary external signing service | Supported through rustls signing traits | Unsupported | Unsupported unless it provides a compatible Windows key handle | Keep rustls-specific until a non-Windows library use exists |
+| Custom verifier callback | Supported | No equivalent current `fetch_tls` API | No userspace callback | Transport-specific |
+| Exact TLS server-name override while preserving request authority | Connector dials the request endpoint and supplies the override to rustls | Connector dials the request endpoint and supplies the override to native TLS | `WinHttpConnect` uses the TLS name, resolution override uses the endpoint, and replaced `Host` preserves authority | Baseline |
+| Custom SAN/subject server-identity policy | Enforced before application data by a verifier | No equivalent current adapter | Server certificate is queryable only after TLS negotiation; equivalent pre-disclosure enforcement is unproven | Add a semantic capability only if exact-name mapping cannot replace the TVS rules |
```
martin-kolinek: Is it implementable if needed? I'd assume not.

crates/fetch/docs/design/capability-matrix.md
```diff
@@ -0,0 +1,149 @@
...
+| Certificate or public-key pins | Implementable in a verifier | No equivalent current adapter | Certificate context is queryable after negotiation | Separate transport-specific feature until a safe portable contract exists |
+| Custom trust roots | Expressible through custom rustls configuration | Not exposed by the current adapter | Uses OS trust unless additional validation is implemented | Transport-specific until a portable requirement is demonstrated |
```
martin-kolinek: Again, it won't be possible to add this as some transports simply cannot support it. Are we reasonably confident that that's ok?

crates/fetch/docs/design/capability-matrix.md
```diff
@@ -0,0 +1,149 @@
...
+modality a transport-construction concern rather than a library-facing capability axis.
+
+## HTTP protocols
+
+| Capability | Hyper + either TLS backend | WinHTTP | Public treatment |
+| --- | --- | --- | --- |
+| HTTP/1.1 and HTTP/2 preference | Supported | Supported | Baseline |
+| Strictly require HTTP/2 | Supported by Hyper's HTTP/2-only mode | Supported by enabling HTTP/2 and setting `WINHTTP_OPTION_HTTP_PROTOCOL_REQUIRED` | Baseline; gRPC is a demonstrated consumer |
+| Initial HTTP/2 stream receive window | Fixed or adaptive policy | OS default; a fixed window option exists | Transport-owned default, not public configuration |
+| HTTP/3 | Unsupported | Supported on recent Windows | Transport-specific until a library requires HTTP/3 |
```
martin-kolinek: Also until hyper supports H3, right?

crates/fetch/docs/design/capability-matrix.md
```diff
@@ -0,0 +1,149 @@
...
+| Fine-grained HTTP/2 flow control | Supported | Different partial native controls | Internal transport policy |
+
+An ordered version preference and a protocol requirement are different APIs. A transport may honor
+an HTTP/2 preference by falling back to HTTP/1.1, but that is not sufficient for a gRPC library
+that requires HTTP/2. WinHTTP can prevent fallback by combining its HTTP/2 enable flag with
+`WINHTTP_OPTION_HTTP_PROTOCOL_REQUIRED`; no user-space emulation is needed. On systems that lack
```
martin-kolinek: Which systems lack that option?

crates/fetch/docs/design/capability-matrix.md
```diff
@@ -0,0 +1,149 @@
...
+## Connections
+
+| Capability | Hyper + either TLS backend | WinHTTP | Public treatment |
+| --- | --- | --- | --- |
+| Fixed maximum connection lifetime | Enforced by retiring aged connections | Planned through `WINHTTP_OPTION_EXPIRE_CONNECTION` | Baseline |
+| Per-connection lifetime callback | Supported by current Hyper options | No portable callback model | Transport-specific; fixed lifetime covers the demonstrated library need |
+| Maximum idle age | Arbitrary duration or unlimited | Shortening is supported; longer retention depends on protocol and native scavenging | Baseline with value/protocol validation |
+| Total connections per origin | Not provided by the current Hyper option | Native per-server cap | Baseline requirement, but Hyper needs a real total-concurrency implementation |
+| Maximum idle connections per host | Current Hyper `max_connections` behavior | No equivalent meaning; WinHTTP's cap is total connections | Rename and keep transport-specific |
+| Coarse idle HTTP/2 health check | Supported | Supported with a native minimum interval | Avoid a trait unless a library requires this outcome independently of idle-age policy |
```
martin-kolinek: What does avoid a trait mean?

crates/fetch/docs/design/README.md
```diff
@@ -0,0 +1,218 @@
...
+Configuration is classified by semantics and ownership:
+
+1. **Pipeline policy** is implemented above the transport and is available for every client.
+   Routing, resilience, telemetry, redaction, response policy, and custom middleware are in this
+   category.
+2. **Portable transport requirements** describe observable networking behavior that a library or
+   application may require. Examples include connection lifetime, connection limits, protocol
+   constraints, certificate authentication, and portable trust policy. Every transport must honor
+   an explicit requirement or reject client construction.
+3. **Transport-specific configuration** controls mechanisms that have no portable contract.
```
martin-kolinek, line 22: Just looking through the list of transport-specific configuration options, I think we'll need an escape hatch - a way for a library to check if a specific transport is used and access and change transport-specific configuration.

crates/fetch/docs/design/README.md
martin-kolinek: 🤖 Yes. I clarified that there are no partial portable capability profiles: every supported transport, external transport, and fake implements the complete portable baseline. Transport-specific extensions sit outside that baseline.

crates/fetch/docs/design/README.md
martin-kolinek: 🤖 The hypothetical preference API was unnecessary and unclear, so I removed it. A library requiring a nonportable mechanism now uses a typed transport extension and returns an unsupported-transport error when it is absent.

crates/fetch/docs/design/README.md
martin-kolinek: 🤖 Agreed—the exception wording conflated configurability with invariants. I rewrote this to say that avoiding Nagle/delayed-ACK stalls is an unconditional transport invariant; callers do not configure it.

crates/fetch/docs/design/README.md
martin-kolinek: 🤖 Rewritten with a concrete credential example: the library requests the logical role `tvs-client`, while Windows can bind it to a store selector and Linux to certificate/private-key material.

crates/fetch/docs/design/README.md
martin-kolinek: 🤖 Added the localhost/TVS example: the request and HTTP authority remain `https://localhost:50042`, while the transport sends `tvs.prod.example` as SNI and validates that exact certificate DNS name.

crates/fetch/docs/design/capability-matrix.md
martin-kolinek: 🤖 Agreed. Pinning is now an explicit portable non-goal rather than a potential future contract. A transport-bound library may configure it through a typed extension and must reject transports that cannot enforce it before request disclosure.

crates/fetch/docs/design/capability-matrix.md
martin-kolinek: 🤖 Correct: custom SAN/subject validation is not implementable across the supported transports. The matrix now marks it as a rustls-only typed extension, never a portable capability. TVS must use an exact certificate DNS identity to remain transport-independent.

crates/fetch/docs/design/capability-matrix.md
martin-kolinek: 🤖 Agreed. Per-client custom roots are now an explicit portable non-goal. The design assumes transport-independent libraries do not require them; a library that does must bind itself to a supporting transport through the typed extension escape hatch.

crates/fetch/docs/design/capability-matrix.md
martin-kolinek: 🤖 Yes. Clarified that HTTP/3 cannot enter the portable baseline until Hyper and every other supported transport implement it, in addition to a library demonstrating the requirement.

crates/fetch/docs/design/capability-matrix.md
martin-kolinek: 🤖 `WINHTTP_OPTION_HTTP_PROTOCOL_REQUIRED` requires Windows 10 version 1903 or later. The matrix now states that explicitly and describes strict HTTP/2 construction failing on older supported hosts.

crates/fetch/docs/design/capability-matrix.md
martin-kolinek: 🤖 Removed the stale trait language. Coarse HTTP/2 health checking is now described as transport-owned policy with no public option.

crates/fetch/docs/design/README.md
martin-kolinek: 🤖 Added a typed pre-build extension escape hatch. The concrete builder carries type-indexed transport configuration values; libraries can query/mutate a known extension and treat absence as optional or as a construction error. This intentionally introduces a runtime check and dependency on the concrete transport crate, without expanding the portable contract or exposing live handlers/sockets.

crates/fetch/docs/design/transport-configuration.md
```diff
@@ -0,0 +1,381 @@
...
+`HttpClientBuilder` owns an erased transport configuration and an accumulated set of portable
+requirements. Its setters merge constraints and retain enough provenance to diagnose conflicts.
+`build` returns the concrete, transport-erased `HttpClient`.
+
+`Transport` defines the complete library-facing baseline:
+
+```rust,ignore
+pub trait Transport: Send + Sync + 'static {
+    fn build(
+        self: Box<Self>,
```
martintmk, line 48:
```suggestion
      &self,
```

why not like this? The transport should still stay boxeable

crates/fetch/docs/design/transport-configuration.md
```diff
@@ -0,0 +1,381 @@
...
+## Composition and erasure
+
+Transport erasure occurs when the application creates the builder. Libraries then configure one
+stable type without understanding the underlying transport:
+
+```rust,ignore
+pub fn configure(builder: HttpClientBuilder) -> Result<HttpClient> {
+    builder
+        .connection_lifetime(LIFETIME)
+        .client_certificate(ClientCredentialId::new("service-client"))
```
martintmk, line 77: imho, all tls specific configuration should have `tls_` prefix to keep clean (reqwest did the same)

crates/fetch/docs/design/transport-configuration.md
martin-kolinek: 🤖 Agreed that the trait should not expose boxing. I changed `Transport` to `Clone + Send + Sync` with `validate(self, ...)`, and `HttpClient::builder<T: Transport>` now performs type erasure internally through a cloneable construction closure. Each builder build consumes a fresh transport-configuration clone; no boxed receiver is required from implementors.

crates/fetch/docs/design/transport-configuration.md
martin-kolinek: 🤖 Applied this to the broad portable builder: the methods are now `tls_client_credential(...)` and `tls_server_name(...)`. I kept names concise inside TLS-specific composition builders such as `fetch_hyper_rustls`, where the crate/type scope already establishes that the mechanisms are TLS-specific.

crates/fetch/docs/design/transport-configuration.md
martin-kolinek: 🤖 Correction after further discussion: `Transport` should remain dyn-compatible. The design now stores `Arc<dyn Transport>` internally and uses `validate(self: Arc<Self>, ...)`, which is an object-safe receiver. `HttpClient::builder<T>` performs the `Arc` erasure internally, while runtime-selected callers may pass an erased transport directly. This avoids both `Box<Self>` and `clone_box` while allowing the factory to retain owned configuration.

martin-kolinek: 🤖 Final correction: your original `&self` suggestion is the cleaner contract. `Transport::validate(&self, ...)` is dyn-compatible and treats the transport as reusable immutable composition configuration. It returns a `TransportFactory` owning independent validated client state. `HttpClient::builder<T>` boxes internally, and `builder_erased(Box<dyn Transport>)` is only the explicit dynamic-construction boundary; neither `Box` nor `Arc` appears in the trait receiver.

crates/fetch_winhttp/docs/design.md
```diff
@@ -116,37 +86,19 @@ each other's connections. This is a security boundary: a strict client and one b
...
-| rustls client-certificate resolver | Ignored; client certificates are out of scope (§4.1). |
-
-(The option mapping and the reasoning behind each floor are implementation.md §10.3.)
+### 2.1 Portable connection requirements
+
+The transport honors the complete portable connection contract:
+
+- total connect deadline;
+- maximum idle age, accepting values whose WinHTTP representation still implies the bound;
+- total concurrent connections per origin through `WINHTTP_OPTION_MAX_CONNS_PER_SERVER`;
```
geeknoid: **The advertised per-origin connection cap is multiplied by every isolated session** — *Correctness · Medium · High*

`WINHTTP_OPTION_MAX_CONNS_PER_SERVER` is session-scoped, while `implementation.md:454-492` specifies one session for every runtime thread × `multiple_pools` slot. Applying a requested cap of `N` to each session therefore permits up to `N × threads × slots` connections to the same origin, rather than the total cap promised here. This can exceed a caller's downstream connection budget even though validation succeeds.

**Direction:** enforce one client-wide permit budget per origin before dispatching into isolated sessions, or redefine the portable option explicitly as a per-partition limit and keep that semantic consistent across transports.

**Done when:** a multi-thread/multi-pool client cannot exceed the documented limit for one origin, or the public contract and compatibility tests intentionally specify the multiplied limit.

crates/fetch_winhttp/docs/design.md
```diff
@@ -251,43 +211,34 @@ bundle and configures only a small set of `WinHttpTlsConfig` knobs (§1.2):
...
-- **Request-response sequencing.** The request body is fully sent before response
-  reception begins.
-- **Trailers.** Response trailers exposed by WinHTTP are returned as `HttpBody` trailer
-  frames rather than discarded. HTTP/1.1 supports trailer fields after a chunked body,
-  but WinHTTP does not expose them; response trailers are therefore available only for
-  HTTP/2 and HTTP/3. Outgoing trailer frames are unsupported and fail the request rather
-  than being silently dropped. A trailer frame is reached only once the body yields it,
-  so that failure arrives after the headers and every preceding data frame have been
-  sent (§7).
+- **Full duplex.** For HTTP/2, response headers and body data may arrive while a known- or
```
geeknoid: **The response deadline cannot cover an upload after `execute` has returned** — *Correctness · Medium · High*

This permits response headers/body to be returned while upload continues, but §6.1 and `implementation.md:1712-1716` enforce `ResponseTimeout` only by wrapping the `HttpClient::execute` pipeline future and claim that it continues through the complete upload. Once that future resolves with the response, its wrapper is gone, so it cannot time out or cancel the detached upload represented by `UploadCompletion`. The two contracts cannot both hold.

**Direction:** propagate one absolute response deadline into the shared request/upload state and make `UploadCompletion` enforce/cancel it, or end the response timeout at response headers and define a separate upload-completion deadline.

**Done when:** the documented timeout owner remains alive until every operation it promises to bound completes, with a contract test where response headers precede a stalled upload.

crates/fetch_winhttp/docs/design.md
```diff
@@ -116,37 +86,19 @@ each other's connections. This is a security boundary: a strict client and one b
...
-| rustls crypto provider or certificate verifier | Ignored; Schannel owns cryptography and certificate verification. |
-| rustls client-certificate resolver | Ignored; client certificates are out of scope (§4.1). |
-
-(The option mapping and the reasoning behind each floor are implementation.md §10.3.)
+### 2.1 Portable connection requirements
+
+The transport honors the complete portable connection contract:
+
+- total connect deadline;
+- maximum idle age, accepting values whose WinHTTP representation still implies the bound;
```
geeknoid: **Unrepresentable idle policies are silently changed instead of rejected** — *Correctness · Medium · High*

The next paragraph requires validation to reject values WinHTTP cannot honor, but `implementation.md:1649-1655,1726-1730` rounds every finite value below five seconds up to five seconds and maps `Unlimited` to roughly 49 days. The first mapping weakens an upper-bound requirement (a requested one-second maximum may reuse for five seconds); the second is observably finite for long-running services. Both contradict the exact/monotonic requirement semantics proposed here.

**Direction:** reject finite bounds below WinHTTP's minimum and represent `Unlimited` without an idle-eviction option if possible; otherwise mark it unsupported rather than coercing it.

**Done when:** every accepted value preserves its stated bound/meaning, and unsupported values fail during validation with a specific error.

crates/fetch/docs/design/capability-matrix.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Timeouts, I/O, and platform services
+
+| Capability | Hyper + either TLS backend | WinHTTP | Public treatment |
+| --- | --- | --- | --- |
+| End-to-end and attempt deadlines | Pipeline-owned | Pipeline-owned | Pipeline policy |
+| Connect deadline | Wraps connector establishment | Native resolve/connect controls with different phase boundaries | Baseline after defining one observable deadline |
+| Separate resolve/send/receive timers | Not exposed by the supported Hyper path | Native controls | Transport-specific |
+| Streaming request and response bodies | Supported | Native upload/download paths implemented | Required `Transport` invariant |
+| Full-duplex HTTP/2 | Supported | Native behavior demonstrated on Windows 11 build 26100; directional implementation required | Required invariant; retain platform compatibility coverage |
```
geeknoid: **A required invariant is being promoted from a single-host manual probe** — *Testing · Medium · High*

The experiment records five manual runs on Windows 11 24H2 build 26100, while Microsoft only guarantees duplex behavior on “some versions.” `fetch_winhttp/docs/design.md:68-75` requires every supported client/server baseline to pass before release, and `implementation.md:1333-1341` describes future integration coverage, but this change adds only examples and no CI/release gate that executes them across the supported matrix. A later OS or runner change can therefore invalidate a required `Transport` invariant without failing the suite.

**Direction:** turn the decisive probe cases into gated integration tests and make supported-SKU qualification consume their results before advertising the capability.

**Done when:** every supported Windows baseline is automatically qualified (or rejected) and a duplex regression blocks release.

crates/fetch_winhttp/examples/resolution_hostname.rs
```diff
@@ -0,0 +1,580 @@
...
+        );
+        ensure!(
+            authority_override.sni.as_deref() == Some(LOGICAL_HOST),
+            "authority-override request changed the TLS SNI"
+        );
+        ensure!(
+            authority_override
+                .http_authority
+                .as_deref()
+                .is_some_and(|authority| authority.starts_with(RESOLUTION_HOST)),
```
geeknoid: **The authority probe accepts values that violate the exact-identity contract** — *Testing · Low · High*

Both this assertion and the positive case at line 80 use `starts_with`, so `resolution.test.invalid.attacker` or an authority with the wrong port would still pass. The design requires the HTTP authority to preserve the exact logical/override host and port; this probe is intended to prove that separation, but its oracle only checks a prefix.

**Direction:** construct the exact expected authority, including the effective port, and compare for equality in both cases.

**Done when:** changing either the hostname suffix or port makes the probe fail.

geeknoid (commented):
# Early feedback: the target contracts currently outrun their enforcement

This is a static review of draft #693 at `ba0bd5ecf38d1901f7ff5bec02aa62bf84f9f0a9` against merge base `36b9e57bd6c36779195b9989b06a36d94e78c960`.

The main blockers are contract consistency: the per-origin connection cap is multiplied by isolated WinHTTP sessions; `ResponseTimeout` cannot cover an upload after the pipeline future returns response headers; and idle policies that the design says must be exact/rejected are silently coerced. The native full-duplex claim also needs an automated supported-Windows qualification gate, and the authority probe's prefix oracle is weaker than the exact identity contract.

No exploitable security issue was established. I reviewed all 36 `unsafe` occurrences in the three new WinHTTP examples (35 FFI operations/blocks and the scoped `SendPtr` impl); handle ownership, pointer lifetimes, directional concurrency, and joins have local safety rationale, with no proven UB or unsound safe abstraction.

**Open performance decision (not inline):** per-thread/per-pool-slot sessions fragment connection reuse and can repeat DNS/TLS setup. That behavior predates this diff, while the new document both calls it preferable and defers the decision to profiling. Representative handshake, reuse, scaling, and tail-latency measurements should resolve shared versus isolated sessions before stabilization.

I deduplicated against the existing inline discussion. Generated `Cargo.lock` was excluded. No project code, tests, build, benchmark, profiler, Miri, coverage, mutation, fuzzing, or live executable was run, and nothing was posted.

PR #681, opened 2026-08-18
perf(thread_aware)!: partition relocation storage per affinity

sandersaares:
[Copilot speaking]

## Motivation

`thread_aware::Arc<T, S>::relocate` is called on every cross-affinity handoff for every thread-aware `Arc` reachable in the relocated object graph. It previously took an unconditional exclusive `RwLock` write lock, including the common case where the destination value was already materialized. Read-only relocations therefore serialized through one table-wide lock. (Azure DevOps 7747020.)

## Changes

**Strategy-partitioned write-once storage.** A strategy maps each affinity into a strategy partition. The table is sized once behind a `OnceLock`, and each partition owns an independent `OnceLock<sync::Arc<T>>`.

- The hit path reads the destination cell with a plain acquire load and carries no cell or table lock-word contention. Adopting and releasing `sync::Arc` values still update their strong counts, so callers that share an allocation can contend on its reference count.
- The miss path uses `get_or_init`, so one racer materializes the destination partition and every other racer adopts the published value. The factory runs at most once per strategy partition.

**Relocation correctness.** A cross-partition miss records the value the `Arc` moved away from in the source partition with a write-once `set`, allowing later relocation back to recover it without holding multiple cells simultaneously. Relocation within one strategy partition keeps the carried value, which preserves grouped strategies such as `PerProcess` and `PerNuma`.

**Initialization behavior.** Constructor functions, clone functions, and captured `ThreadAware` state run while the destination cell is initializing. They must not reenter that cell or create cyclic initialization dependencies. A panic propagates and leaves the cell empty for a later relocation to retry; there is no poisonable lock or partially published value.

**Out-of-range affinities.** Relocation to an affinity outside the table's coordinate space is a no-op that keeps the carried value and records the `thread_aware_arc_oob` event. Direct `Storage::insert` and `Storage::get` calls remain strict and panic when their affinity is out of range.

**Constructible storage handle.** The public `storage::Storage<T, S>` exposes affinity-keyed write-once insertion and lookup without exposing the internal cell layout. `Storage::insert` returns `Ok(())` when it publishes a value and `Err(value)` when the selected partition is already populated.

**API terminology.** `Arc`, `Storage`, `Strategy`, the built-in strategies, and the crate documentation consistently describe value identity and materialization in terms of strategy partitions. Affinity terminology is reserved for the source and destination coordinates that strategies map into those partitions.

[...]

PR #678, opened 2026-08-18
fix(thread_aware)!: derive ThreadAware for generics used only in PhantomData

wukchung:
&#x1F916; This description was written by an AI agent.

## Impact

`#[derive(ThreadAware)]` works for a type whose generic parameter appears only inside a
`PhantomData` field. That shape failed to compile, and the only way out was to widen the
author's own public API with a `Send` bound.

```rust
#[derive(ThreadAware)]
struct DirectPhantom<T, U>(T, PhantomData<U>);
```

`PhantomData<T>` now implements `ThreadAware` for every `T: ?Sized + Send`, which serves
hand-written generic code as much as derived impls.

Three narrower shapes also compile that did not: a marker nested inside a relocated field such
as `(PhantomData<T>,)`, an enum with no variants, and a type whose own items are spelled like
the identifiers the derive emits.

No crate in this workspace is affected: `anyspawn`, `bytesbuf`, `cachet_memory`, `fetch` and
`thread_aware_benchmarking` build and lint unchanged.

## Architecture

**The derive special-cases nothing.** Every field except a `#[thread_aware(skip)]` one is
relocated, and the type parameters the traversal reaches take a `ThreadAware` bound. That is
the whole rule. `is_phantom_data` is deleted, so the derive no longer decides anything from how
a field type is spelled.

**A marker relocates through an ordinary impl.**
`impl<T: ?Sized + Send> ThreadAware for PhantomData<T>` makes a marker a normal relocatable
value, as a field or nested in a tuple, with nothing for the macro to reason about. The bound
is `Send` rather than `ThreadAware` so that a marker naming a type this crate withholds
`ThreadAware` from - `PhantomData<Arc<i32>>` - still relocates; the marker holds no value, so
there is nothing to contend over.

**A skipped field is the one case the generated body does not touch,** so the impl gains
`where Self: Send` to keep the `ThreadAware: Send` supertrait satisfied. The predicate sits on
`Self` rather than on the field type: a type that is `Send` only through a manual `unsafe impl`
cannot prove a predicate such as `where *const T: Send`.

**The generated code cannot be captured by the author's own items.** Fields bind to
`__thread_aware_field_{i}`, the method's parameters are spelled the same way, and arms match
under `ref mut` so a residual collision is reported as `E0530` at the derive rather than as a
confusing type error. Macro hygiene does not prevent this - under `Span::mixed_site()` a
pattern still resolves against a `const` or `static` of the same name. Every path the derive
emits is fully qualified, `Option` included.

[...]

PR #694, opened 2026-08-25
chore(release): cascade thread_aware and ohno versions

sandersaares:
[Copilot speaking]

## Motivation

`thread_aware` 0.10.0 shipped the breaking per-affinity relocation-storage change from #681, while its published dependents still target the previous line. The public `PhantomData<T>` implementation and derive-bound correction from #678 require a new release.

The unreleased `ohno` macro changes likewise require `ohno` 0.5.0, `ohno_macros` 0.5.0, and `ohno_macros_impl` 0.5.1, followed by the same recursive treatment of their published dependents.

## Changes

Published 0.x dependents that expose a changed dependency through their public API receive a breaking minor increment. Dependents that keep the changed dependency private receive a patch increment.

```
thread_aware_macros_impl 0.11.0
  +-> thread_aware_macros 0.11.0
        +-> thread_aware 0.11.0
              +-> anyspawn 0.8.0 -> anyspawn_azure 0.3.0
              +-> bytesbuf 0.9.0 -> bytesbuf_io 0.9.0, multitude 0.9.0
              +-> cachet_memory 0.7.0
              +-> observed 0.25.0 -> observed_utils 0.2.0
              +-> seatbelt 0.8.0
              +-> tick 0.6.0 -> arty_executor 0.1.1
              +-> uniflight 0.5.0

ohno_macros_impl 0.5.1
  +-> ohno_macros 0.5.0
        +-> ohno 0.5.0
              +-> cachet_tier 0.5.0 -> cachet_service 0.5.0 -> cachet 0.13.0
              +-> fetch_tls 0.4.0
              +-> templated_uri 0.5.0
              +-> http_extensions 0.10.0
              +-> fetch_hyper 0.7.0 -> fetch 0.16.0 -> fetch_azure 0.6.0
              +-> seatbelt_http 0.8.0
```

The `thread_aware` runtime and macro crates share the 0.11.0 compatibility line. The `templated_uri` runtime and macro crates likewise move together at 0.5.0, preserving the compatibility boundary that prevents Cargo from unifying generated code across incompatible runtime lines.

The release notes retain the historical 0.10.0 `thread_aware` change separately from the new 0.11.0 work and preserve already-published `observed`, `observed_utils`, and `ohno_macros_impl` entries under their original versions.


crates/cachet_memory/src/tier.rs
```diff
@@ -208,13 +209,13 @@ where
         Ok(())
     }
 
-    async fn clear(&self) -> Result<(), Error> {
+    fn clear(&self) -> impl Future<Output = Result<(), Error>> {
         self.inner.invalidate_all();
-        Ok(())
+        ready(Ok(()))
     }
```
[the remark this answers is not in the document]
sandersaares: [Copilot speaking]

Moved invalidate_all into an async block so the cache is cleared when the future is polled, not when clear() is called.

crates/cachet_memory/src/tier.rs
```diff
@@ -208,13 +209,13 @@ where
...
 
-    async fn len(&self) -> Result<u64, SizeError> {
-        Ok(self.inner.entry_count())
+    fn len(&self) -> impl Future<Output = Result<u64, SizeError>> {
+        ready(Ok(self.inner.entry_count()))
     }
```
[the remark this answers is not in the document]
sandersaares: [Copilot speaking]

Moved entry_count into an async block so the length is read when the future is polled, not when len() is called.

crates/arty_executor/src/wake.rs
```diff
@@ -366,8 +366,8 @@ mod tests {
         // This is fine because `WakeSignal` only uses the `TaskRef`, not the task behind it.
         let fake_task_ref = unsafe { TaskRef::fake() };
 
-        // Capacity is 0 so the queue is not allowed to allocate (== is never used).
-        let awakened_queue = Arc::new(Mutex::new(VecDeque::with_capacity(0)));
+        // Empty with no reserved capacity so the queue is not allowed to allocate (== is never used).
```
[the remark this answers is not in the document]
sandersaares: [Copilot speaking]

Updated the comment to say there is no up-front allocation and that this test never uses the queue.

Vaiz (changes requested):
could you add `ohno 0.5`, `ohno_macros 0.5`, `ohno_macros_impl 0.5.1` to that list?

crates/cachet_tier/src/testing.rs
```diff
@@ -226,54 +227,54 @@ where
...
-            return Err(Error::from_message("mock: get failed"));
+            return ready(Err(Error::from_message("mock: get failed")));
         }
         self.record(op);
-        Ok(self.data.lock().get(key).cloned())
+        ready(Ok(self.data.lock().get(key).cloned()))
     }
 
-    async fn insert(&self, key: K, entry: CacheEntry<V>) -> Result<InsertOutcome, Error> {
+    fn insert(&self, key: K, entry: CacheEntry<V>) -> impl Future<Output = Result<InsertOutcome, Error>> {
```
martin-kolinek: 🤖: `MockCache`'s `CacheTier` impl is eager, while the same problem was fixed one commit earlier in `cachet_memory` (67e1b86 — "Keep the CacheTier work inside an async block so dropping the future without polling does not clear the cache"). Here `insert` (:240), `invalidate` (:254) and `clear` (:265) call `record(op)` and mutate `self.data` *before* `ready(...)` builds the future, and `get` (:230) evaluates `should_fail`, records the op and snapshots the value at call time (`len` at :276 likewise reads the map at call time). So constructing one of these futures and dropping it un-polled — a cancellation, a `select!` loser, a timeout race — records the operation and mutates the map anyway. That is exactly the semantics you rejected for `InMemoryCache`, applied to the mock whose whole purpose is to let tests assert which operations reached the tier.

This is public surface, not an internal detail: `pub use testing::{CacheOp, MockCache}` at `crates/cachet_tier/src/lib.rs:77`, under `feature = "test-util"`.

The fix is already in this PR's vocabulary — the same `#[expect(clippy::manual_async_fn, reason = …)]` + `async { … }` shape as `crates/cachet_memory/src/tier.rs:211-228`.

Worth a regression test while you are in there (`crates/cachet_tier/tests/tier.rs`): build `insert`/`invalidate`/`clear` futures, drop them un-polled, and assert that neither the cached state nor `operations()` changed. Every existing test awaits immediately, so none of them distinguishes the two behaviours.

crates/bytesbuf_io/src/testing/fake_read.rs
```diff
@@ -65,22 +66,21 @@ impl FakeRead {
     ///
     /// This call never fails.
     #[cfg_attr(test, mutants::skip)] // Mutations easily lead to infinite loops, not worth the effort.
-    #[expect(clippy::unused_async, reason = "API compatibility between trait and inherent fn")]
-    pub async fn read_at_most_into(&mut self, len: usize, mut into: BytesBuf) -> Result<(usize, BytesBuf), Infallible> {
+    pub fn read_at_most_into(&mut self, len: usize, mut into: BytesBuf) -> impl Future<Output = Result<(usize, BytesBuf), Infallible>> {
```
martin-kolinek: 🤖: Same eager-effect relocation as in `cachet_tier::MockCache`, here on `pub` `test-util` surface (`pub mod testing;` at `crates/bytesbuf_io/src/lib.rs:45`): `into.put_bytes(data_to_read)` and `self.contents.advance(bytes_to_read)` now run when `read_at_most_into` is *called*, not when the returned future is polled — so a test that builds the future and drops it un-polled has already consumed bytes from the fake source and dropped the buffer holding them. `FakeWrite::write` (`fake_write.rs:57`) has the same shape: the stream gains the data at call time.

The trait paths stay lazy — `impl Read for FakeRead` (:177) and `impl Write for FakeWrite` (`fake_write.rs:102`) both delegate with `.await` — so the same object now has different operation semantics depending on which call syntax you use. These are fakes for scripting I/O in tests, including cancellation tests, so "the work happens at construction" is precisely the wrong contract for them. Please keep the bodies lazy with `async { … }` + `#[expect(clippy::manual_async_fn, …)]`, matching `crates/cachet_memory/src/tier.rs:211-228`. (Checked the degenerate paths: the `bytes_to_read == 0` early return at :74-76 mutates nothing, and an empty `contents` or a `max_read_size` of 1 goes down that same branch, so only the non-empty read shifts.)

A regression test would lock it in: drop an unpolled `read_at_most_into` future and assert the next awaited read still returns the complete input; drop an unpolled `FakeWrite::write` future and assert `contents()` is still empty.

crates/cachet_memory/src/tier.rs
```diff
@@ -208,13 +208,23 @@ where
...
 
-    async fn clear(&self) -> Result<(), Error> {
-        self.inner.invalidate_all();
-        Ok(())
+    #[expect(
+        clippy::manual_async_fn,
+        reason = "async fn without .await trips clippy::unused_async_trait_impl on 1.98; the async block keeps the work lazy until polled"
+    )]
+    fn clear(&self) -> impl Future<Output = Result<(), Error>> {
```
martin-kolinek: 🤖: Non-blocking — the laziness this `async` block exists to preserve has no test behind it, so it can regress silently the next time a lint moves these bodies around. Consider locking it in from `crates/cachet_memory/tests/cache.rs`: dropping an unpolled `clear()` future leaves the entries intact, and a `len()` future created before an insert but polled after it reports the post-insert count. The current tests await both methods immediately, so the eager form this commit replaced would still pass them.

martin-kolinek (approved):
🤖: Approved based on human decision.

sandersaares:
[Copilot speaking]

Added `ohno` 0.5.0, `ohno_macros` 0.5.0, and `ohno_macros_impl` 0.5.1, together with their recursive published-dependent cascade. The `templated_uri` runtime and macro crates move together at 0.5.0 to preserve their generated-code compatibility boundary.

The release set also now includes `thread_aware` 0.11.0 and its macro packages because #678 merged after the already-published 0.10.0 version bump.

crates/cachet_tier/src/testing.rs
sandersaares: [Copilot speaking]

Removed the unrelated Rust source and test changes from this release PR. The branch now changes only release metadata, so the eager-future behavior and proposed regressions are no longer part of this change.

crates/bytesbuf_io/src/testing/fake_read.rs
sandersaares: [Copilot speaking]

Removed the unrelated Rust source and test changes from this release PR. The branch now changes only release metadata, so the eager-future behavior and proposed regressions are no longer part of this change.

crates/cachet_memory/src/tier.rs
sandersaares: [Copilot speaking]

Removed the unrelated Rust source and test changes from this release PR. The branch now changes only release metadata, so the eager-future behavior and proposed regressions are no longer part of this change.

crates/cachet_tier/src/testing.rs
```diff
@@ -226,54 +227,54 @@ where
     K: Clone + Eq + Hash + Send + Sync,
     V: Clone + Send + Sync,
 {
-    async fn get(&self, key: &K) -> Result<Option<CacheEntry<V>>, Error> {
+    fn get(&self, key: &K) -> impl Future<Output = Result<Option<CacheEntry<V>>, Error>> {
         let op = CacheOp::Get(key.clone());
         if self.should_fail(&op) {
             self.record(op);
-            return Err(Error::from_message("mock: get failed"));
+            return ready(Err(Error::from_message("mock: get failed")));
```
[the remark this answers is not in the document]
sandersaares: [Copilot speaking]

Removed the unrelated Rust source and test changes from this release PR. The branch now changes only release metadata, so the eager-future behavior and proposed regressions are no longer part of this change.

crates/bytesbuf_io/src/testing/fake_read.rs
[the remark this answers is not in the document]
sandersaares: [Copilot speaking]

Removed the unrelated Rust source and test changes from this release PR. The branch now changes only release metadata, so the eager-future behavior and proposed regressions are no longer part of this change.

crates/bytesbuf_io/src/testing/fake_write.rs
```diff
@@ -53,10 +54,9 @@ impl FakeWrite {
     /// # Errors
     ///
     /// This call never fails.
-    #[expect(clippy::unused_async, reason = "API compatibility between trait and inherent fn")]
-    pub async fn write(&mut self, data: BytesView) -> Result<(), Infallible> {
+    pub fn write(&mut self, data: BytesView) -> impl Future<Output = Result<(), Infallible>> {
         self.contents.put_bytes(data);
-        Ok(())
+        ready(Ok(()))
```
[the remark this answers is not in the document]
sandersaares: [Copilot speaking]

Removed the unrelated Rust source and test changes from this release PR. The branch now changes only release metadata, so the eager-future behavior and proposed regressions are no longer part of this change.

Vaiz:
> thread_aware_macros_impl 0.7.5
>  +-> thread_aware_macros 0.7.6
>        +-> thread_aware 0.11.0

this part doesn't look right, I think they should have the same minor version

sandersaares:
[Copilot speaking] Raised `thread_aware_macros_impl` to 0.8.0 for the removed public implementation APIs and updated the dependent release note, lockfile, and PR description.

crates/thread_aware_macros/Cargo.toml
```diff
@@ -4,7 +4,7 @@
 [package]
 name = "thread_aware_macros"
 description = "Macros for the thread_aware crate."
-version = "0.7.5"
+version = "0.7.6"
```
Vaiz: should be `0.11.0`

crates/thread_aware_macros_impl/Cargo.toml
```diff
@@ -4,7 +4,7 @@
 [package]
 name = "thread_aware_macros_impl"
 description = "Macros for the thread_aware crate."
-version = "0.7.4"
+version = "0.8.0"
```
Vaiz: should be 0.11.0

Vaiz (changes requested):

crates/thread_aware_macros/Cargo.toml
sandersaares: [Copilot speaking]

Aligned this package with the 	hread_aware 0.11.0 compatibility line. The manifests, lockfile, changelogs, generated README links, and PR description now use 0.11.0.

crates/thread_aware_macros_impl/Cargo.toml
sandersaares: [Copilot speaking]

Aligned this package with the 	hread_aware 0.11.0 compatibility line. The manifests, lockfile, changelogs, generated README links, and PR description now use 0.11.0.

sandersaares:
[Copilot speaking] Aligned `thread_aware`, `thread_aware_macros`, and `thread_aware_macros_impl` at 0.11.0 in e760c729. The PR description was also republished from a Markdown file so its paragraph and diagram newlines render correctly.

Vaiz (approved):

PR #695, opened 2026-08-26
test(tick): improve mutation coverage

martintmk:
Add focused boundary and state-transition tests for clock, delay, periodic timer, fast instant, and timer internals.


wukchung (approved):

Vaiz (approved):

crates/tick/src/delay.rs
```diff
@@ -211,6 +211,20 @@ mod tests {
...
+        let mut delay = Delay::new(&clock, Duration::from_millis(1));
+        let waker = Waker::noop();
+        assert_eq!(delay.register_timer(waker), Poll::Pending);
+        assert!(delay.current_timer.is_some());
+
+        delay.duration = Duration::MAX;
+        assert_eq!(delay.register_timer(waker), Poll::Pending);
+
+        assert_eq!(delay.current_timer, None);
+    }
```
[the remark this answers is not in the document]
psandana: 🤖 Pato's Pull Request Agent: Small correction here — `Waker::noop()` returns `&'static Waker`, so `delay.register_timer(waker)` is already passing `&Waker` and compiles fine; the "passes a `Waker` by value" note doesn't apply.

The valid part is that this state isn't reachable through `Delay::poll` — once `current_timer` is `Some`, `register_timer` is never called again. But calling it directly is actually justified for mutation coverage: `delay_close_to_max_ensure_timer_not_registered` already hits the overflow branch, yet it starts from `current_timer == None`, so a mutant that turns `self.current_timer = None` into a no-op survives there. Starting from `Some` is precisely what kills that mutant.

Suggested tweak: keep the direct call, but add a one-line comment noting it targets the overflow-clear branch for mutation coverage, and rename to reflect the actual assertion (e.g. `overflow_reregistration_clears_current_timer`). Worth noting the first registered timer is left in the clock's timer set here — harmless for the test, but confirms this path is synthetic.

sgalkin (approved):

crates/tick/src/clock_control.rs
```diff
@@ -749,6 +749,58 @@ mod tests {
...
 
+    #[test]
+    fn manual_advance_evaluates_auto_advance_timers() {
+        let mut state = State {
+            auto_advance_timers: true,
+            ..State::default()
+        };
+        let start = state.instant;
+        state.timers.register(start + Duration::from_secs(2), Waker::noop().clone());
+
```
martintmk, line 760: **Posted by an AI agent · Non-blocking**

**`ClockControl::advance` can exceed its documented duration**

**Problem**
The public docs say manual advancement moves by the specified duration and fires timers within that period, but this test codifies a one-second advance jumping two seconds when auto-advance timers are enabled.

**Why this matters**
Callers cannot tell whether manual advancement is bounded, and the test makes the undocumented overshoot part of the expected behavior.

**Suggested fix**
Clarify the interaction in the public method documentation, or change the behavior and test so manual advancement remains bounded.

crates/tick/src/clock_control.rs
```diff
@@ -749,6 +749,58 @@ mod tests {
...
+        assert_eq!(after_limit, after_timer);
+    }
+
+    #[test]
+    fn advance_time_fires_ready_timers() {
+        let mut state = State::default();
+        state.timers.register(state.instant + Duration::from_secs(1), Waker::noop().clone());
+
+        state.advance_time(Duration::from_secs(1), TimeFlow::Forward);
+
```
martintmk, line 790: **Posted by an AI agent · Non-blocking**

**`advance_time_fires_ready_timers` does not verify task notification**

**Problem**
The test registers `Waker::noop()` and only asserts that the timer entry is removed. It still passes if ready timers are deleted without waking their tasks.

**Why this matters**
Timer bookkeeping could report completion while the awaiting task remains permanently pending.

**Suggested fix**
Use a counting `Wake` implementation and assert one wake at the deadline and none before it.

martintmk (commented):
**Posted by an AI agent**

Reviewed the complete pinned test-only change at `81f48351dede70d47962f3d59a36497ae99364b6` across all ten required areas. I found two new non-blocking gaps: the manual-advance test codifies behavior that conflicts with the public duration contract, and the ready-timer test does not verify that the task is actually woken. Existing discussions about discriminator coupling, synthetic delay re-registration, platform-time flakiness, and the delay-test name remain applicable and were not duplicated.

The targeted `tick` suite passed. A non-required Linux ARM `pr-fast` check remains red, but its log exposes no diagnostic beyond the failed action step.

martintmk (commented):
**Posted by an AI agent**

Reviewed the complete pinned change with all ten Review Lens areas and three independent correctness reviewers. I found two non-blocking test/documentation issues; the shipping public API and rustdoc surface are unchanged. Focused validation passed: `cargo check -p tick --all-features` and the specialist `tick` test runs.

crates/tick/src/delay.rs
```diff
@@ -211,6 +211,20 @@ mod tests {
...
+    #[test]
+    fn overflowing_reregistration_clears_existing_timer() {
+        let clock = Clock::new_system_frozen();
+        let mut delay = Delay::new(&clock, Duration::from_millis(1));
+        let waker = Waker::noop();
+        assert_eq!(delay.register_timer(waker), Poll::Pending);
+        assert!(delay.current_timer.is_some());
+
+        delay.duration = Duration::MAX;
+        assert_eq!(delay.register_timer(waker), Poll::Pending);
```
martintmk, line 223: **Posted by an AI agent · Non-blocking**

**Overflow re-registration leaves the prior timer registered**

**Problem**
`overflowing_reregistration_clears_existing_timer` verifies only that `current_timer` becomes `None`. The exercised overflow branch drops the existing key without calling `unregister_timer`, so the clock still retains the timer and `Drop` can no longer remove it.

**Why this matters**
The test currently blesses an internally inconsistent state and would not protect a future re-arm/reset path from leaking a timer registration.

**Suggested fix**
Take and unregister the existing key before clearing the registration, then assert that the clock's timer count returns to zero.

crates/tick/src/clock_control.rs
```diff
@@ -749,6 +749,58 @@ mod tests {
...
+        state.timers.register(start + Duration::from_secs(2), Waker::noop().clone());
+
+        state.advance(Duration::from_secs(1), TimeFlow::Forward);
+
+        assert_eq!(state.instant.saturating_duration_since(start), Duration::from_secs(2));
+        assert_eq!(state.timers.len(), 0);
+    }
+
+    #[test]
+    fn auto_advanced_timer_consumes_limit() {
```
martintmk, line 768: **Posted by an AI agent · Non-blocking**

**`auto_advance_limit` docs exclude timer-driven advancement**

**Problem**
This test establishes that timer-driven fast-forwarding consumes the same limit, but the public `auto_advance_limit` docs say the limit only has an effect after `auto_advance` is configured with a non-zero duration.

**Why this matters**
Users of `auto_advance_timers` can incorrectly conclude that the limit cannot bound timer-driven advancement, so the documented configuration contract disagrees with the covered behavior.

**Suggested fix**
Document that the limit caps advancement from both `auto_advance` and `auto_advance_timers`, and state that at least one automatic advancement mode must be enabled.

crates/tick/src/state.rs
```diff
@@ -145,4 +145,16 @@ mod tests {
     fn clock_state_send_and_sync() {
         static_assertions::assert_impl_all!(ClockState: Send, Sync);
     }
+
+    #[test]
+    fn shared_timer_uniqueness_tracks_clones() {
```
geeknoid: **The new tests cannot improve mutation coverage for the skipped implementations** — *Testing · Medium · High*

This test exercises `SynchronizedTimers::is_unique`, but that method is annotated with `#[cfg_attr(test, mutants::skip)]`, so `cargo-mutants` will never generate the mutations this assertion is meant to kill. The same mismatch appears for the new [`platform_time` test](https://github.com/microsoft/oxidizer/blob/81f48351dede70d47962f3d59a36497ae99364b6/crates/tick/src/fast_instant.rs#L124-L127), [`advance_timers` boundary test](https://github.com/microsoft/oxidizer/blob/81f48351dede70d47962f3d59a36497ae99364b6/crates/tick/src/timers.rs#L168-L177), and [`PeriodicTimer::poll_next` deadline test](https://github.com/microsoft/oxidizer/blob/81f48351dede70d47962f3d59a36497ae99364b6/crates/tick/src/periodic_timer.rs#L228-L237); each target function is explicitly skipped. These tests may add ordinary coverage, but they do not support the PR's mutation-coverage claim for those behaviors.

**Direction:** either remove/narrow the relevant mutation exclusions so these assertions can kill specific mutants, or retarget the tests at unskipped code and describe the skipped tests as non-mutation coverage.

**Done when:** the diff-scoped mutation report contains targeted mutants killed by these tests, or the PR no longer claims mutation coverage for functions excluded from mutation.

geeknoid (commented):
Static review found one cross-cutting testing issue: several added tests are presented as mutation-coverage improvements, but the implementation functions they exercise are explicitly excluded with `mutants::skip`, so those tests cannot strengthen the mutation gate for the targeted behavior.

I also checked the changed boundary/state-transition assertions against the timer, delay, clock-control, fast-clock, and shared-state implementations. Existing discussion already covers the hard-coded timer discriminators, direct internal-map/helper coupling, missing wake assertions, overflow re-registration leak, `platform_time() != 0` assumption, and the clock-control limit/documentation mismatches; I did not duplicate those points.

All five taxonomies were assessed (820 classes) with mutation/oracle emphasis. This was static-only: no project code, tests, builds, benchmarks, profilers, Miri, coverage, mutation, fuzzing, or live probes were run.

PR #696, opened 2026-08-26
chore: reference the public nightly toolchain and rename the alignment-cap cfg

wukchung:
&#x1F916;

## Problem

The VS Code settings template and the Clippy prompt point contributors at a toolchain that public `rustup` cannot install, so following either verbatim fails on a clean clone.

Separately, the `multitude` test gate for over-aligned types is named after one particular codegen backend instead of the property that actually decides whether those tests can run, and the flag has no written contract anywhere in-tree.

## Change

- `.vscode/settings.template.jsonc` and `.github/prompts/clippy-nightly.prompt.md` now reference the public nightly toolchain. rustfmt is pinned to `+nightly-2026-05-30`, matching `RUST_NIGHTLY` in `constants.env`, which is what CI's format check runs.
- `scripts/update_rust_toolchain.ps1` rewrites that pin whenever it bumps `RUST_NIGHTLY`, so the two cannot drift apart, and `scripts/tests/Pester/unit/toolchain/PinnedNightly.Tests.ps1` asserts they agree regardless of who performs the bump. The template also tells contributors to re-copy the line and install the new toolchain afterwards, since their own `settings.json` does not move with a bump.
- The Clippy prompt names `RUST_NIGHTLY` rather than a floating `nightly`, and records that CI's Clippy gate is the stable `RUST_LATEST`, so nightly-only lints are advisory and should not be silenced with `#[expect(...)]` before confirming they fire on the gating toolchain.
- The test cfg is renamed to `align_capped_backend` across `Cargo.toml` and five `multitude` test files (65 occurrences), the comment in `crates/multitude/tests/arena.rs` states the constraint the gate encodes, and the `check-cfg` entry now documents the flag's contract: when to set it, that nothing in-tree sets it, and that it is all-or-nothing across the 32 KiB, 64 KiB and 128 KiB cases.

No test logic changes; the `check-cfg` declaration moves with the rename, so the gate behaves exactly as before.

## Drive-by: `chunks_exact` -> `as_chunks` in `alloc_growable.rs`

Unrelated to the above, but it blocks CI on this branch.

`clippy::chunks_exact_to_as_chunks` is newer than the pinned `RUST_LATEST` (1.96.1) and fires on the two UTF-16 byte-pairing sites in `crates/multitude/src/arena/alloc_growable.rs`, which `-D warnings` turns into an error. The lint is already latent on `main`; editing the workspace `Cargo.toml` invalidated the lint cache and re-linted the whole crate, which is why `pr-fast` fails here and not on branches opened earlier.

Both sites now use `as_chunks::<2>().0.iter()`. `as_chunks` discards the trailing odd byte exactly as `chunks_exact(2)` did, so the lossy constructors still append their own replacement character and behaviour is unchanged.

## Validation

- `cargo check -p multitude --tests --all-features` - clean.
- The same command with `RUSTFLAGS=--cfg align_capped_backend` - clean, no `unexpected_cfgs` warning.
- `cargo test -p multitude --all-features utf16` - 216 passed, 0 failed.
- `scripts/tests/Pester/Run-Tests.ps1 -Path unit` - 414 passed, 0 failed, 1 skipped. The new pin test was also confirmed to fail when the template pin is drifted by hand.

WI: https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7783302

### DIFF
--- a/.github/prompts/clippy-nightly.prompt.md
+++ b/.github/prompts/clippy-nightly.prompt.md
@@ -1,8 +1,8 @@
 # Fixing Nightly Clippy Lints with MCP Tools
 
-You are a Rust developer working on a project that uses the `ms-nightly` toolchain and enforces strict Clippy lints.
+You are a Rust developer working on a project that uses the `nightly` toolchain and enforces strict Clippy lints.
 Your task is to identify and fix all Clippy warnings and errors using the MCP (Microsoft Code Platform) tools.
-For all requests, you shall use `ms-nightly` toolchain and MCP tools to ensure compatibility with the latest features and lints.
+For all requests, you shall use `nightly` toolchain and MCP tools to ensure compatibility with the latest features and lints.
 
 ## Instructions
 
@@ -16,7 +16,7 @@ For all requests, you shall use `ms-nightly` toolchain and MCP tools to ensure c
     - For each crate in your workspace, perform the following steps before moving to the next crate.
 
 2. **Run Clippy with Nightly Toolchain**
-    - Use the `ms-nightly` toolchain to run Clippy on the current crate. Use the `#cargo-clippy` tool for this, targeting only the specific crate.
+    - Use the `nightly` toolchain to run Clippy on the current crate. Use the `#cargo-clippy` tool for this, targeting only the specific crate.
 
 3. **Automatically Fix Lints with `fix`**
     - Use the `fix` parameter with `#cargo-clippy` to automatically apply suggested fixes for applicable lints in the current crate.
@@ -24,7 +24,7 @@ For all requests, you shall use `ms-nightly` toolchain and MCP tools to ensure c
 
 4. **Analyze Clippy Output**
     - Review the Clippy output for any remaining warnings or errors in the current crate.
-    - Pay special attention to lints that are only available on `ms-nightly`.
+    - Pay special attention to lints that are only available on `nightly`.
 
 5. **Manually Fix Remaining Lints**
     - For each remaining lint:
@@ -44,7 +44,7 @@ For all requests, you shall use `ms-nightly` toolchain and MCP tools to ensure c
 
 ## Tips
 
-- Always use the latest `ms-nightly` toolchain and MCP tools for best results.
+- Always use the latest `nightly` toolchain and MCP tools for best results.
 - Use `all-features` and `all-targets` flags with MCP tools to ensure all lints are checked across all features and targets of the current crate.
 - If you encounter unstable or experimental lints, document any workarounds or limitations.
   Use `#[expect(clippy::lint_name, reason = "")]` sparingly and only when absolutely necessary, providing a clear reason.
--- a/.github/prompts/clippy-nightly.prompt.md
+++ b/.github/prompts/clippy-nightly.prompt.md
@@ -1,7 +1,7 @@
 # Fixing Nightly Clippy Lints with MCP Tools
 
 You are a Rust developer working on a project that uses the `nightly` toolchain and enforces strict Clippy lints.
-Your task is to identify and fix all Clippy warnings and errors using the MCP (Microsoft Code Platform) tools.
+Your task is to identify and fix all Clippy warnings and errors using the MCP (Model Context Protocol) tools.
 For all requests, you shall use `nightly` toolchain and MCP tools to ensure compatibility with the latest features and lints.
 
 ## Instructions
--- a/.github/prompts/clippy-nightly.prompt.md
+++ b/.github/prompts/clippy-nightly.prompt.md
@@ -2,7 +2,13 @@
 
 You are a Rust developer working on a project that uses the `nightly` toolchain and enforces strict Clippy lints.
 Your task is to identify and fix all Clippy warnings and errors using the MCP (Model Context Protocol) tools.
-For all requests, you shall use `nightly` toolchain and MCP tools to ensure compatibility with the latest features and lints.
+For all requests, you shall use the nightly toolchain named by `RUST_NIGHTLY` in `constants.env` and MCP tools to ensure compatibility with the latest features and lints.
+Do not use a bare `nightly`: it floats, and no gate in this repo enforces it.
+
+Note that CI's Clippy gate runs the *stable* toolchain named by `RUST_LATEST`, not nightly.
+Nightly-only lints are therefore advisory - they surface problems ahead of the gating
+toolchain. Do not add `#[expect(...)]` for a warning only nightly emits; confirm it fires on
+`RUST_LATEST` first, or the expectation becomes unfulfilled once nightly moves on.
 
 ## Instructions
 
@@ -16,7 +22,7 @@ For all requests, you shall use `nightly` toolchain and MCP tools to ensure comp
     - For each crate in your workspace, perform the following steps before moving to the next crate.
 
 2. **Run Clippy with Nightly Toolchain**
-    - Use the `nightly` toolchain to run Clippy on the current crate. Use the `#cargo-clippy` tool for this, targeting only the specific crate.
+    - Use the toolchain named by `RUST_NIGHTLY` to run Clippy on the current crate. Use the `#cargo-clippy` tool for this, targeting only the specific crate.
 
 3. **Automatically Fix Lints with `fix`**
     - Use the `fix` parameter with `#cargo-clippy` to automatically apply suggested fixes for applicable lints in the current crate.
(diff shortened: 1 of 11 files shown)


Vaiz (approved):

psandana:
🤖 Pato's Pull Request Agent: Renaming `utc_backend` → `align_capped_backend` here only covers the *consumers* of the cfg. Nothing in this repo sets `--cfg utc_backend` (code search finds it only in the files this PR renames), so the flag must be injected by the internal UTC/codegen build. Unless that injection is updated to pass `--cfg align_capped_backend` in the same rollout, these `#[cfg(not(align_capped_backend))]` gates will stop matching and the 128 KiB-alignment tests will run — and abort — on a backend that caps alignment. Can you confirm the external flag setter is being updated alongside this rename?

.vscode/settings.template.jsonc
```diff
@@ -58,12 +58,14 @@
...
     // with the regular toolchain. This enables those capabilities.
     //
-    // Requires the ms-nightly toolchain to be installed. You may need to install it manually, as
-    // it may not be installed by default (`msrustup toolchain install ms-nightly`).
-    "rust-analyzer.rustfmt.extraArgs": ["+ms-nightly", "--config-path", "${workspaceFolder}/unstable-rustfmt.toml"],
+    // The pinned nightly must match `RUST_NIGHTLY` in `constants.env`, which is what CI's
+    // format check runs; a different nightly can format the same file differently. Update
+    // both together. Install it with
+    // `rustup toolchain install nightly-2026-05-30 --component rustfmt`, or run `just install-tools`.
+    "rust-analyzer.rustfmt.extraArgs": ["+nightly-2026-05-30", "--config-path", "${workspaceFolder}/unstable-rustfmt.toml"],
```
martin-kolinek, line 72: 🤖: Non-blocking — the pin is right, but nothing maintains it. These two literals (`:67` install hint, `:68` `+nightly-2026-05-30`) are now the only hand-copied nightly outside `constants.env:13`, and the routine bump path is automated: `scripts/update_rust_toolchain.ps1` computes yesterday's nightly and rewrites `constants.env` only (`-ConstantsFile` is the sole file it targets). Nothing in it touches `.vscode/`. So the next auto-bump silently leaves this template on a nightly CI no longer uses — exactly the drift the pin was added to prevent — with the "Update both together" comment as the only safeguard.

That script is already a regex rewriter over a named file; extending it to rewrite `nightly-\d{4}-\d{2}-\d{2}` in this template (covered by its test) would make the invariant hold by construction instead of by memory.

One knock-on worth a line in the comment either way: `DEVELOPMENT.md:44` tells contributors to *copy* this template to `settings.json` once, and `just install-tools` installs whatever `$env:RUST_NIGHTLY` says at the time. After a bump, a contributor's copied `+nightly-2026-05-30` names a toolchain their own setup recipe no longer installs, so format-on-save fails outright rather than formatting differently — a harder failure than the drift being fixed. A note telling them to re-copy (or re-run `rustup toolchain install <RUST_NIGHTLY> --component rustfmt`) after a bump would cover it.

Cargo.toml
```diff
@@ -291,7 +291,7 @@ rust.unexpected_cfgs = { level = "warn", check-cfg = [
     'cfg(miri_race_coverage)',
     'cfg(miri_strict_provenance)',
     'cfg(miri_tree_borrows)',
-    'cfg(utc_backend)',
+    'cfg(align_capped_backend)',
```
martin-kolinek, line 299: 🤖: Non-blocking — follow-up to the external-setter question already on this PR (psandana's comment: https://github.com/microsoft/oxidizer/pull/696#issuecomment-5427615629); that ask stands and this doesn't repeat it. The separate point is the new name itself: it doesn't say *what* cap qualifies. The same switch suppresses test types aligned to 32, 64 and 128 KiB (e.g. `crates/multitude/tests/arena.rs:2244-2255` and `:6066-6069`), so a backend capped at 64 KiB can't tell from the name whether it should set the flag — and setting it loses the still-valid 32 KiB coverage. There is no prose for it anywhere in-tree (no `.md` under `crates/multitude` mentions it), and after the rename `git grep utc_backend` returns nothing, so this `check-cfg` line is the flag's only in-tree home. A sentence next to it naming the threshold semantics and who is expected to pass it would make the contract legible to the external builder that has to act on the rename.

.github/prompts/clippy-nightly.prompt.md
```diff
@@ -44,7 +44,7 @@ For all requests, you shall use `ms-nightly` toolchain and MCP tools to ensure c
 
 ## Tips
 
-- Always use the latest `ms-nightly` toolchain and MCP tools for best results.
+- Always use the latest `nightly` toolchain and MCP tools for best results.
```
martin-kolinek: 🤖: Non-blocking — this PR takes two different positions on the same value. `.vscode/settings.template.jsonc:64-68` gets a hard pin with the rationale that a contributor's toolchain must match `RUST_NIGHTLY`, while this line (and `:3`/`:5`) still tells the agent to use *the latest* unpinned `nightly`. Neither toolchain is what gates the repo: CI's clippy job runs `RUST_LATEST`, not nightly (`.github/workflows/main.yml:167-173`), and `RUST_NIGHTLY` is what the nightly-dependent checks pin to (`justfiles/anvil/versions.just:37-38`, "Pin only -- do not use bare `nightly` here").

So an agent following this prompt chases lints from a floating nightly that no gate enforces, and may miss ones that do fire on the gating toolchain — then lands `#[expect(...)]`s for warnings CI never emits. Pointing the prompt at the same source of truth ("use the nightly named by `RUST_NIGHTLY` in `constants.env`") keeps it correct across bumps without carrying its own literal. If the floating nightly is deliberate here — the point being to surface lints *before* they reach the pinned toolchain — a half-sentence saying so would settle it.

martin-kolinek (approved):
🤖: Approved.

wukchung:
&#x1F916; Checked this, and the premise does not hold: nothing sets the cfg, under either name.

`--cfg utc_backend` appears nowhere outside the files this PR renames - not in this repository, and not in any build tooling we control. The gates have therefore never matched, which is exactly why the over-aligned tests compile in and fail on a backend that caps alignment.

The build script that would have set the flag (a capability probe emitting `cargo::rustc-cfg=...`) lived in #692, which is now closed unmerged. So there is no external setter to update and no rollout to coordinate: the rename is a no-op at build time.

The only way it bites is a developer passing `RUSTFLAGS=--cfg utc_backend` by hand as a local workaround. That stops applying, with an `unexpected_cfgs` warning rather than a silent skip. Nothing in CI does it.

The underlying compile failure is tracked separately and is untouched here. Your question did land a change though: the flag's contract is now written next to its `check-cfg` entry in `Cargo.toml`, so whoever does set it can tell what qualifies.

.vscode/settings.template.jsonc
wukchung: &#x1F916; Fixed by construction. `scripts/update_rust_toolchain.ps1` now rewrites `nightly-\d{4}-\d{2}-\d{2}` in this template whenever it bumps `RUST_NIGHTLY`, so the pin travels with the bump instead of relying on the comment.

One correction: that script has no test today. `git grep update_rust_toolchain` returns only the script itself and `README.md:245`, and no workflow invokes it. So instead of extending a test that was not there, I added `scripts/tests/Pester/unit/toolchain/PinnedNightly.Tests.ps1`, which asserts the template's pin equals `RUST_NIGHTLY`. That holds whoever performs the bump, script or not. Verified both ways: it passes as-is, and fails when I drift the pin by hand.

Also took the `DEVELOPMENT.md:44` knock-on you flagged - the comment now tells contributors to re-copy the line and install the new toolchain (or re-run `just install-tools`) after a bump, since their copied `settings.json` does not move on its own.

Cargo.toml
wukchung: &#x1F916; Agreed - the name alone does not carry the threshold. The `check-cfg` entry now states the contract: what condition qualifies, that nothing in-tree sets it so a build using such a backend passes it itself, and that it is all-or-nothing across the 32 KiB, 64 KiB and 128 KiB cases, so a backend capped at 64 KiB still loses the coverage it could have run.

On the external setter question you are deferring to: there is no external setter. Nothing sets this flag under either name, in this repository or in any build tooling we control, and the build script that would have set it went away with #692. Details in my reply to psandana above.

.github/prompts/clippy-nightly.prompt.md
wukchung: &#x1F916; Right, that was inconsistent, and the floating nightly here was not deliberate. The prompt now names the toolchain from `RUST_NIGHTLY` in `constants.env` at every point it used a bare `nightly`, with an explicit instruction not to use the floating one.

I also took your point about the gate rather than leaving it implicit: the prompt now says CI's Clippy gate runs the stable `RUST_LATEST`, that nightly-only lints are advisory, and that an `#[expect(...)]` should not be landed for a warning only nightly emits without confirming it fires on `RUST_LATEST` first - otherwise the expectation goes unfulfilled once nightly moves on.

Vaiz (approved):

PR #656, opened 2026-08-11
feat(ohno_macros): rewrite the crate from requirements and design

Vaiz:
🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

## What this PR does

This PR rewrites `crates/ohno_macros` from zero. It adds two documents that state what the crate has to deliver and how it is arranged to deliver it, then implements the crate from them.

- `docs/requirements.md` — the behavior the crate owes: the derive (R1), `#[ohno::error]` (R2), `#[enrich_err]` (R3), diagnostics (R4), quality gates (R5).
- `docs/design.md` — the phases, the module tree, the types the phases exchange, how each generated item is produced, the diagnostics anchors, and the testing plan.
- `src/` — the implementation.

## The shape

The derive runs three phases: `TokenStream -> parse -> Ast -> validate -> Model -> generate -> TokenStream`.

- **parse** answers "can I read this". It decodes the crate's own attributes and reports what it cannot read.
- **validate** answers "is this allowed". It is the only phase that reports rule violations.
- **generate** returns `TokenStream`, not `Result`. It cannot fail.

That last point is the design. R4 forbids a diagnostic that points into generated code, and that is only structural if a `Model` which could make generation fail cannot be built. `Shape` splits the field list around the core instead of carrying an index, so "exactly one core" is unrepresentable. `Message` can only be obtained through validation, so an argument that is not rooted in a field is unrepresentable.

Where an invariant is not structural, the design says so plainly. The alignment between a conversion's initializers and the field list is a relation between two values; it is held by a private field and a single constructor, not pretended into the type system.

`#[ohno::error]` keeps no model and `#[enrich_err]` keeps no `Ast`, because neither has enough input to warrant one.

## Diagnostics

Faults accumulate through one `Errors` type. Its `add` takes tokens rather than a `Span`, so R4's rule that a multi-token diagnostic uses `syn::Error::new_spanned` is enforced by the signature.

Accumulation runs across independent concerns. A concern whose own input failed to decode is skipped, not guessed at: a template that cannot be split reports one fault, instead of that fault plus field errors invented by repairing it. This is what the separate parse phase buys.

## Tests

`crates/ohno/tests/**` is untouched and is the sole behavioral spec. It passes unchanged, including all twelve `.stderr` compile-fail snapshots.

[...]

PR #697, opened 2026-08-26
docs(ohno_macros): reconcile the design with the two-crate split

Vaiz:

The design and requirements documents still described a single `ohno_macros/src` implementation tree, phase-local tests, and a fallible `Shape::new`. None of the three has been true since #683 moved the implementation into `ohno_macros_impl` — the docs were last touched by the original rewrite in #656.

- New "Two crates" section stating the shim/implementation boundary and why it exists: a `proc-macro = true` crate can export nothing but macros, and its own tests cannot call them.
- Modules tree re-rooted at `ohno_macros_impl/src/`, with a note on what `lib.rs` owns there.
- Testing section rewritten around the public expansion functions and the grouped snapshots in `ohno_macros_impl/tests/public_api.rs`, replacing the per-phase table. The cost of dropping phase isolation is named rather than glossed over: there are no phase-local tests, so a `Model` no real input produces cannot be covered.
- `Shape::new` documented as infallible, with the reasoning for the panic — `validate` derives the core index from the field list it maps, so a disagreement is a bug in this crate rather than an input a diagnostic could describe, and R4 forbids inventing one.
- `prettyplease` corrected to `testing_aids` in the dev-dependency note.
- `template.rs`'s reference to `docs/design.md` repointed after the move.

Docs only, apart from that one doc-comment path fix.

Validated locally with `just anvil-fmt`, an `unstable-rustfmt.toml` check on the nightly pin, and `cargo +1.96.1 check -p ohno_macros -p ohno_macros_impl --all-targets`.

Related work item: AB#7776884

---

**Second commit (`de1f9a7`) — a separate concern, flagged for review.**

The design document had drifted from describing how the crate is arranged into arguing for it: alternatives weighed and rejected, counter-arguments answered, and a `Decided` list recording settled debates. That is out, and the facts it carried are kept.

- The "Is `Ast` worth its own type?" section is gone. What it established — `Ast` holds decoded payloads with their spans, `Model` drops spans, `Ast` re-houses neither `Generics` nor `Fields` — is now stated in the pipeline section as fact.
- The `Decided` list is gone. Every entry restated the body; its one unique fact, the ADO reference behind `pub(crate)`, moved to where the constructors are described.
- Alternatives-weighing prose removed throughout — including from the sections added in the first commit. Constraints and mechanisms stay.

Net −79 lines. No claim about the implementation changes, so the two commits can be reviewed independently or the second dropped.

### DIFF
--- a/crates/ohno_macros/docs/design.md
+++ b/crates/ohno_macros/docs/design.md
@@ -1,12 +1,12 @@
 # Design
 
 How the `ohno` macros are built. `requirements.md` says what they have to
-deliver; this document says how they are arranged to deliver it. Every decision
-below is argued from a requirement, and the requirement is named.
+deliver; this document says how they are arranged to deliver it. A requirement is
+named wherever a rule below implements one.
 
 The implementation spans two packages — `ohno_macros`, the proc-macro crate the
 compiler sees, and `ohno_macros_impl`, the ordinary library behind it. The Two
-crates section explains the split; until then, read "the crate" as both.
+crates section describes the split; until then, read "the crate" as both.
 
 ## The pipeline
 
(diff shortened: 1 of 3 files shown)


Vaiz:

Closing in favour of #698, which carries the identical two commits from a correctly named branch.

This PR's head branch was `u/eshutov/ohno-macros-docs-reconcile`, but commits on `microsoft/oxidizer` are authored as `Evgenii <vaiz.git@gmail.com>` and the repository's convention for those is `u/vaiz/<yyyy>/<mm>/<dd>/<topic>`. The branch prefix contradicted the commit author on the same branch and put a corporate alias in a public ref.

PR #700, opened 2026-08-26
fix: silence newly-firing clippy lints from Rust 1.98

Vaiz:

## What

Rust 1.98 ships a new pedantic lint, [`clippy::unused_async_trait_impl`](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#unused_async_trait_impl), which fires on an `async fn` inside an `impl` block whose body contains no `.await`. The workspace enables `clippy::pedantic` and `anvil-clippy` runs with `-D warnings`, so every one of these is a hard build failure.

This is currently blocking unrelated PRs. #698 is a documentation-only change, but its affected-package set pulls in `cachet_tier`, so `anvil-pr / pr-fast` fails on it with 5 errors in code the PR does not touch. This PR carries the lint fixes on their own so those PRs can go green.

## Lints fixed

| Lint | Sites |
|---|---|
| `clippy::unused_async_trait_impl` (new in 1.98) | 55, across 22 files |
| `clippy::chunks_exact_to_as_chunks` | 2, in `multitude` |
| `unused_imports` (rustc) | 3 `use ... as _;` trait imports |

## How

For genuine trait impls, apply clippy's own suggestion: drop `async` and return `impl Future<Output = ...>` built from `std::future::ready`. The traits (`CacheTier`, `layered::Service`, `bytesbuf_io::Read` / `Write`, the OTel exporter traits) already declare a future return type, so this is not a public API change, and the bodies were already synchronous.

Two places keep `async` instead:

* The inherent `pub async fn` mirrors in `bytesbuf_io::testing` (`Null`, `FakeRead`, `FakeWrite`) already carry `#[expect(clippy::unused_async, reason = "API compatibility between trait and inherent fn")]`. The `async` there is deliberate and the signatures are public API, so the new lint name was added to the existing `expect` list rather than changing them.
* `fundle/tests/bundle_setters.rs` has a crate-level allow list that already includes `clippy::unused_async`; the new lint name was added alongside it.

The `#[expect]` entries name a lint that does not exist before 1.98. That matches existing repo practice — `tick`'s `UnixSeconds::MAX` already `#[expect]`s `clippy::duration_suboptimal_units` the same way.

`multitude`'s `as_chunks::<2>()` change also removes the manual `let raw = [pair[0], pair[1]];` rebuild, since the iterator now yields `[u8; 2]` directly. `as_chunks` is stable since 1.88, below the 1.93.1 MSRV.

## Effects

* `anvil-fmt` and `anvil-clippy` (`--workspace --all-targets --all-features -- -D warnings`) both pass on Rust 1.98.0; before this change the workspace produced 60 clippy errors.
* Tests pass for all 13 touched packages (`--all-features`): 106 suites, 0 failures.
* The three removed imports were verified to still compile on the pinned `RUST_LATEST` 1.96.1 and on the 1.93.1 MSRV — 1.96.1 does not report them as unused, so their removal is a 1.98-only detection improvement rather than a resolution change.
* No public API changes: every rewritten signature is a trait impl whose shape the trait already fixes.
* Behaviour is unchanged apart from the timing inherent in clippy's suggested fix — the (synchronous, non-blocking) body now runs when the method is called rather than when the returned future is first polled. Every one of these methods is awaited at its call site.

## Not in scope

Lint fixes only. No refactors and no unrelated cleanups; `crates/ohno_macros/docs/` is untouched.

---

## Update — 25b3cceb

The claim above that a bare `#[expect]` naming a 1.98-only lint "matches existing repo practice" was wrong, and CI proved it:

* `static-analysis` and `anvil-pr / pr-fast (linux)` run the pinned `RUST_LATEST` 1.96.1, where `clippy::unused_async_trait_impl` does not exist. Naming it makes `unknown_lints` fire, and `-D warnings` turned that into 6 hard errors in `bytesbuf_io`. The `tick` precedent cited above does not transfer. The idiom that does is the `#![allow(unknown_lints)]` guard already carried by `rest_over_grpc_examples` and `rest_over_grpc_tests` for this exact toolchain split, now applied to `bytesbuf_io::testing` and the `fundle` setter test.
* The lint also fires on **inherent** `impl` blocks, not only trait impls, so the lint names themselves have to stay — the `bytesbuf_io` inherent mirrors genuinely need them.
* `anvil-pr / pr-fast (linux-arm)` runs 1.98 and reached `seatbelt/benches/observability_cg.rs`, which the first commit missed: the whole benchmark is `#[cfg(target_os = "linux")]`, so it never compiles on Windows. It now carries the same rewrite as its wall-clock twin `observability.rs`.

Verified locally on both toolchains: `cargo clippy --workspace --all-targets --all-features -- -D warnings` passes on stable 1.98.0 **and** on 1.96.1. `observability_cg.rs` remains unverifiable locally (Linux-only, no cross C toolchain here); it is a line-for-line mirror of the already-passing `observability.rs`.

---

## Update — 0c69a477

Review feedback from @martintmk: an example is read by a human, so `impl Future<Output = ...>` plus `std::future::ready` costs more in readability than the lint is worth there.

The five `cachet` examples are restored to plain `async fn`, and `clippy::unused_async_trait_impl` is suppressed at each example's crate root instead — behind the same `#![allow(unknown_lints)]` guard, since the pinned 1.96.1 gates do not know the lint name.

Non-example code (library `src/`, tests, benches) keeps the `impl Future` rewrite.

### DIFF
--- a/crates/cachet/examples/error_handling.rs
+++ b/crates/cachet/examples/error_handling.rs
@@ -8,6 +8,7 @@
 //! - Attach recovery information with `with_recovery()`
 //! - Extract and handle typed errors from cache operations
 
+use std::future::ready;
 use std::io::{self, ErrorKind};
 
 use cachet::{Cache, CacheEntry, CacheTier, Error, InsertOutcome};
@@ -19,21 +20,23 @@ use tick::Clock;
 struct FailingCache;
 
 impl CacheTier<String, i32> for FailingCache {
-    async fn get(&self, _key: &String) -> Result<Option<CacheEntry<i32>>, Error> {
+    fn get(&self, _key: &String) -> impl Future<Output = Result<Option<CacheEntry<i32>>, Error>> + Send {
         // Wrap the IO error and attach recovery information
-        Err(Error::from_source(io::Error::new(ErrorKind::TimedOut, "connection timed out")).with_recovery(RecoveryInfo::retry()))
+        ready(Err(
+            Error::from_source(io::Error::new(ErrorKind::TimedOut, "connection timed out")).with_recovery(RecoveryInfo::retry())
+        ))
     }
 
-    async fn insert(&self, _key: String, _entry: CacheEntry<i32>) -> Result<InsertOutcome, Error> {
-        Ok(InsertOutcome::Accepted)
+    fn insert(&self, _key: String, _entry: CacheEntry<i32>) -> impl Future<Output = Result<InsertOutcome, Error>> + Send {
+        ready(Ok(InsertOutcome::Accepted))
     }
 
-    async fn invalidate(&self, _key: &String) -> Result<(), Error> {
-        Ok(())
+    fn invalidate(&self, _key: &String) -> impl Future<Output = Result<(), Error>> + Send {
+        ready(Ok(()))
     }
 
-    async fn clear(&self) -> Result<(), Error> {
-        Ok(())
+    fn clear(&self) -> impl Future<Output = Result<(), Error>> + Send {
+        ready(Ok(()))
     }
 }
 
--- a/crates/cachet/examples/error_handling.rs
+++ b/crates/cachet/examples/error_handling.rs
@@ -8,7 +8,12 @@
 //! - Attach recovery information with `with_recovery()`
 //! - Extract and handle typed errors from cache operations
 
-use std::future::ready;
+#![allow(unknown_lints, reason = "the pinned and latest Clippy versions expose different async lints")]
+#![expect(
+    clippy::unused_async_trait_impl,
+    reason = "examples are written for a human reader, and a plain `async fn` shows the shape of a tier impl better than `impl Future` plus `std::future::ready`"
+)]
+
 use std::io::{self, ErrorKind};
 
 use cachet::{Cache, CacheEntry, CacheTier, Error, InsertOutcome};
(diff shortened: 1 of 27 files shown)


sandersaares (approved):

kate-shine (approved):

crates/cachet/examples/error_handling.rs
```diff
@@ -19,21 +20,23 @@ use tick::Clock;
 struct FailingCache;
 
 impl CacheTier<String, i32> for FailingCache {
-    async fn get(&self, _key: &String) -> Result<Option<CacheEntry<i32>>, Error> {
```
martintmk, line 22: for examples, we should not complicate them as this. better to suppress this lint (apply to all examples)

Vaiz: examples are for AI, right? I think suppressing lints will give them the wrong idea

martintmk: I still think we should be aware of humans reading this, we are not yet fully in AI world (and hopefully it will stay like this for a bit)

Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Agreed — reverted in `0c69a477`. All five `cachet` examples are back to plain `async fn`, with the lint suppressed at each example's crate root instead:

```rust
#![allow(unknown_lints, reason = "the pinned and latest Clippy versions expose different async lints")]
#![expect(
    clippy::unused_async_trait_impl,
    reason = "examples are written for a human reader, and a plain `async fn` shows the shape of a tier impl better than `impl Future` plus `std::future::ready`"
)]
```

The `unknown_lints` line is needed because the gates do not agree on a toolchain: `static-analysis` and `anvil-pr / pr-fast (linux)` run the pinned `RUST_LATEST` 1.96.1, which does not know the lint name, while `pr-fast (linux-arm)` runs 1.98, which does. Without the guard the older jobs fail on the suppression itself.

Verified: `clippy --workspace --all-targets --all-features -- -D warnings` passes on both 1.98.0 and 1.96.1.

Non-example code (library `src/`, tests, benches) keeps the `impl Future` rewrite — that is where the lint is doing real work.

psandana (approved):

PR #701, opened 2026-08-27
build: pin workspace Rust toolchain to 1.95.0

sandersaares:
[Copilot speaking]

## Motivation

Workspace commands otherwise use each environment's active default Rust toolchain. Compiler and lint updates can therefore introduce findings unrelated to a pull request, as seen in #700, rather than providing reproducible results.

## Changes

The workspace and `RUST_LATEST` select the exact Rust 1.95.0 release, matching the upstream version of the internal `ms-prod-1.95` release toolchain. Cross-linked comments define the three-way alignment contract.

The toolchain refresh script treats the internal release channel as the source of truth: maintainers select that channel first, then the script updates both public pins to the latest upstream patch in the selected minor.

The internal release-template protection is delivered by [ox-sdk PR 5587019](https://o365exchange.visualstudio.com/O365%20Core/_git/ox-sdk/pullrequest/5587019), which must merge before this pull request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,7 +242,8 @@ We strive to deliver high-quality code and as such, we've put in place a number 
 We pin the version of the tools we use in CI to ensure hermetic builds as much as possible. We routinely update
 the versions of everything to stay up to date using three scripts:
 
-- `scripts/update_rust_toolchain.ps1` which updates the Rust toolchain versions in `constants.env` and `rust-toolchain.toml`.
+- `scripts/update_rust_toolchain.ps1` which requires an internal `ox-sdk` checkout and updates the public
+  Rust toolchain versions after the internal release toolchain has been selected.
 - `scripts/update_tool_versions.ps1` which updates the version of tools in the `constants.env` file.
 - `scripts/update_action_versions.ps1` which updates the version of GitHub actions in the various files in `.github/workflows`.
 
--- a/README.md
+++ b/README.md
@@ -244,6 +244,7 @@ the versions of everything to stay up to date using three scripts:
 
 - `scripts/update_rust_toolchain.ps1` which requires an internal `ox-sdk` checkout and updates the public
   Rust toolchain versions after the internal release toolchain has been selected.
+  Example: `.\scripts\update_rust_toolchain.ps1 -InternalToolchainFile ..\ox-sdk\.pipelines\variables\publish.yml`.
 - `scripts/update_tool_versions.ps1` which updates the version of tools in the `constants.env` file.
 - `scripts/update_action_versions.ps1` which updates the version of GitHub actions in the various files in `.github/workflows`.
 
--- a/README.md
+++ b/README.md
@@ -242,7 +242,7 @@ We strive to deliver high-quality code and as such, we've put in place a number 
 We pin the version of the tools we use in CI to ensure hermetic builds as much as possible. We routinely update
 the versions of everything to stay up to date using three scripts:
 
-- `scripts/update_rust_toolchain.ps1` which updates the version of the Rust toolchain in the `constants.env` file.
+- `scripts/update_rust_toolchain.ps1` which updates the Rust toolchain versions in `constants.env` and `rust-toolchain.toml`.
 - `scripts/update_tool_versions.ps1` which updates the version of tools in the `constants.env` file.
 - `scripts/update_action_versions.ps1` which updates the version of GitHub actions in the various files in `.github/workflows`.
 
--- a/constants.env
+++ b/constants.env
@@ -7,7 +7,7 @@
 
 # used for testing, ensures the MSRV promise is kept; must match Cargo.toml [workspace.package].rust-version
 RUST_MSRV=1.93.1
-# used for static analysis & mutation testing
+# used for static analysis & mutation testing; must match rust-toolchain.toml
 RUST_LATEST=1.96.1
 # used for coverage and extended analysis; update on a regular basis
 RUST_NIGHTLY=nightly-2026-05-30
--- a/constants.env
+++ b/constants.env
@@ -7,8 +7,9 @@
 
 # used for testing, ensures the MSRV promise is kept; must match Cargo.toml [workspace.package].rust-version
 RUST_MSRV=1.93.1
-# used for static analysis & mutation testing; must match rust-toolchain.toml
-RUST_LATEST=1.96.1
+# used for static analysis & mutation testing; must match rust-toolchain.toml; its major/minor must match:
+# https://o365exchange.visualstudio.com/O365%20Core/_git/ox-sdk?path=/.pipelines/variables/publish.yml
+RUST_LATEST=1.95.0
 # used for coverage and extended analysis; update on a regular basis
 RUST_NIGHTLY=nightly-2026-05-30
 # used for external type exposure checks; update alongside updates to cargo-check-external-types
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,3 @@
+[toolchain]
+# Keep workspace builds and diagnostics reproducible across users.
+channel = "1.95.0"
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,6 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
 [toolchain]
 # Must match RUST_LATEST in constants.env so local and CI diagnostics use the same compiler.
 channel = "1.96.1"
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-# Keep workspace builds and diagnostics reproducible across users.
-channel = "1.95.0"
+# Must match RUST_LATEST in constants.env so local and CI diagnostics use the same compiler.
+channel = "1.96.1"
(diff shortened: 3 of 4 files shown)


Vaiz (approved):

constants.env
```diff
@@ -7,7 +7,7 @@
 
 # used for testing, ensures the MSRV promise is kept; must match Cargo.toml [workspace.package].rust-version
 RUST_MSRV=1.93.1
-# used for static analysis & mutation testing
+# used for static analysis & mutation testing; must match rust-toolchain.toml
```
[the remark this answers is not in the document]
sandersaares: [Copilot speaking]

Updated `scripts/update_rust_toolchain.ps1` so each stable toolchain update writes the same version to `RUST_LATEST` and `rust-toolchain.toml`, including correction of toolchain-only drift. The updater documentation now identifies both managed files.

scripts/update_rust_toolchain.ps1
```diff
@@ -146,6 +151,18 @@ function Update-ConstantsEnv {
...
 }
 
+function Update-RustToolchain {
+    param(
+        [string]$FilePath,
+        [string]$Version
+    )
+    $content = Get-Content $FilePath -Raw
+    $content = Get-Content $FilePath -Raw
+    $replacement = '${1}"' + $Version + '"'
```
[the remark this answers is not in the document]
sandersaares: [Copilot speaking]

Removed the redundant `Get-Content` assignment in commit `b7649af`; `Update-RustToolchain` now reads the file once before replacing its channel.

kate-shine (approved):

martintmk (approved):

rust-toolchain.toml
```diff
@@ -0,0 +1,6 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+[toolchain]
+# Must match RUST_LATEST in constants.env so local and CI diagnostics use the same compiler.
+channel = "1.96.1"
```
martin-kolinek: This will most likely break the internal release build

Vaiz (changes requested):
this will cause issues with publising

rust-toolchain.toml
sandersaares: Good point. I will adjust the release pipeline accordingly, as well.

scripts/update_rust_toolchain.ps1
```diff
@@ -4,40 +4,55 @@
...
 param(
     [string]$ConstantsFile = (Join-Path $PSScriptRoot ".." "constants.env"),
+    [string]$ToolchainFile = (Join-Path $PSScriptRoot ".." "rust-toolchain.toml"),
+    [Parameter(Mandatory = $true)]
+    [string]$InternalToolchainFile,
     [switch]$DryRun
 )
 
 $ErrorActionPreference = "Stop"
+$ToolchainChannelPattern = '(?m)^(channel[ \t]*=[ \t]*)"([^"]*)"'
```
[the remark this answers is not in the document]
sandersaares: [Copilot speaking]

Expanded the shared channel pattern in commit `ae10ffb` so its preserved capture includes leading horizontal whitespace. Parsing and replacement now accept an indented `channel` key without changing its alignment.

README.md
```diff
@@ -242,7 +242,8 @@ We strive to deliver high-quality code and as such, we've put in place a number
 We pin the version of the tools we use in CI to ensure hermetic builds as much as possible. We routinely update
 the versions of everything to stay up to date using three scripts:
 
-- `scripts/update_rust_toolchain.ps1` which updates the version of the Rust toolchain in the `constants.env` file.
+- `scripts/update_rust_toolchain.ps1` which requires an internal `ox-sdk` checkout and updates the public
+  Rust toolchain versions after the internal release toolchain has been selected.
```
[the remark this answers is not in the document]
sandersaares: [Copilot speaking]

Added a root-level invocation example in commit `ae10ffb`, including the mandatory `-InternalToolchainFile` argument and the expected `ox-sdk` publish-variable path.

sandersaares:
[Copilot speaking]

Addressed the suppressed review finding in commit `3c3183c`. `Get-InternalToolchainMinorVersion` now stores an explicit `Regex.Match` result and reads its named `minor` group instead of relying on automatic `$Matches` state. The parser also accepts consistently quoted YAML channel values.

scripts/update_rust_toolchain.ps1
```diff
@@ -4,40 +4,55 @@
...
+.PARAMETER InternalToolchainFile
+    Path to ox-sdk's .pipelines/variables/publish.yml file. Update its MSRUSTUP_TOOLCHAIN values first;
+    this script then aligns the public toolchain files to the latest upstream patch in that Rust minor.
+
 .PARAMETER DryRun
     If specified, shows what would be updated without actually modifying the files.
 
 .EXAMPLE
-    .\update_rust_toolchain.ps1
+    .\update_rust_toolchain.ps1 -InternalToolchainFile ..\ox-sdk\.pipelines\variables\publish.yml
```
AdomasBekeras, line 29: ox-sdk might not be the name of the folder under a different worktree

AdomasBekeras (approved):

scripts/update_rust_toolchain.ps1
sandersaares: Well, it's just an example. Or do you see the problem also in some non-example locations?

martin-kolinek (changes requested):
Rejecting this until it's clarified that this is the right approach. Most Rust libraries don't use rust-toolchain.toml file, we should be sure that it's desirable before we add it. We also should make sure that our internal build pipelines can handle it first.

rust-toolchain.toml
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+[toolchain]
+# Must match RUST_LATEST in constants.env. Its major/minor must match MSRUSTUP_TOOLCHAIN in:
+# https://o365exchange.visualstudio.com/O365%20Core/_git/ox-sdk?path=/.pipelines/variables/publish.yml
```
psandana, line 6: DANGER: Remove this reference to internal repos!

psandana (changes requested):

constants.env
```diff
@@ -7,8 +7,9 @@
 
 # used for testing, ensures the MSRV promise is kept; must match Cargo.toml [workspace.package].rust-version
 RUST_MSRV=1.93.1
-# used for static analysis & mutation testing
-RUST_LATEST=1.96.1
+# used for static analysis & mutation testing; must match rust-toolchain.toml; its major/minor must match:
+# https://o365exchange.visualstudio.com/O365%20Core/_git/ox-sdk?path=/.pipelines/variables/publish.yml
```
psandana, line 11: DANGER: remove references to internal repo

scripts/update_rust_toolchain.ps1
sgalkin: should we better use some fake name in example?

geeknoid (commented):
Static review found one material testing gap: the new internal-toolchain parser and two-file alignment workflow have no Pester coverage, so the contract can regress without an assertion failing. No exploitable security issue, production correctness defect, performance regression, unsafe-code concern, or conformance violation met the reporting bar.

The five passes covered security (150 classes), correctness (162), testing (182), performance (183), and conformance (143). The exact `1.95.0` pins agree across `constants.env` and `rust-toolchain.toml`; CI toolchain inputs still resolve explicitly from `constants.env`; and both toolchain/config/script changes are workspace-wide delta trip wires. Existing comments about internal URLs, the release pipeline, regex state, indentation, documentation, and encoding were not repeated.

The claimed internal `ms-prod-1.95` alignment and prerequisite ox-sdk PR cannot be independently verified from this public clone. The current PR head is also reported as conflicting with `main`, so merge-resolution effects are outside this head-vs-merge-base review.

Static-only limitation: I did not run tests, builds, linters, scripts, benchmarks, profilers, Miri, coverage, mutation testing, fuzzers, binaries, or live probes.

scripts/update_rust_toolchain.ps1
```diff
@@ -146,6 +156,69 @@ function Update-ConstantsEnv {
...
+        [string]$Version
+    )
+
+    $content = Get-Content $FilePath -Raw
+    $replacement = '${1}"' + $Version + '"'
+
+    return $content -replace $ToolchainChannelPattern, $replacement
+}
+
+function Get-InternalToolchainMinorVersion {
```
geeknoid: **The new toolchain parser and alignment update have no regression test** — *Testing · Medium · High confidence*

This function now defines the release-toolchain contract: it parses conditional YAML values, accepts quoted and backend-qualified channels, rejects mixed minors, and drives coordinated updates to both `constants.env` and `rust-toolchain.toml`. The repository's Pester job discovers every `*.Tests.ps1`, but no test file references this updater. Mutations such as deleting the mixed-version rejection, narrowing the accepted channel forms, or updating only one public pin therefore have no assertion that would fail.

**Direction:** add fixture-driven Pester coverage for the parser and an end-to-end dry-run/update scenario using mocked release responses and synthetic public/internal toolchain files.

**Done when:** valid quoted/unquoted/backend variants, malformed and mixed-minor inputs, one-file drift, two-file drift, and no-op alignment each have an assertion on both exit behavior and resulting file contents.

PR #702, opened 2026-08-27
fix(scripts): build semver-checks baselines under a short target directory

kate-shine:
Follow-up to #642 (AB#7790786 is still open).

Pointing the baseline build at `rust-lld` cleared `LNK1104`, but it only moved the `MAX_PATH` ceiling one step earlier in the toolchain. The work item filed four days after that merge is a different failure: six `C1083` errors from `cl.exe`, with no real `LNK1104` in the log.

## Why the linker fix doesn't cover this

`-sys` crates compile C through the `cc` crate, and MSVC `cl.exe` resolves its `-Fo` argument against `MAX_PATH`. Unlike the linker there is no long-path-aware drop-in guaranteed to be present — `clang-cl` ships with neither a default Rust nor a default Visual Studio install.

Measured on this machine with a 296-character output path:

| Compiler | Result |
|---|---|
| `cl.exe` 14.44.35207 | `fatal error C1083: Cannot open compiler generated file: '': Invalid argument` |
| `clang-cl` (LLVM) | wrote the object |

`cc`'s own 8.3 short-name fallback does not rescue it either; the field log shows it tried `OXIDIZ~1.WOR\...\SEMVER~1\GIT-7C~1\` and still failed.

## The lever that remains

`cargo semver-checks` nests baseline builds under the workspace target directory and exposes no flag to move them, but it derives that location from cargo metadata — so `CARGO_TARGET_DIR` does reach it.

`Invoke-SemverChecksCli` now scopes `CARGO_TARGET_DIR` to `<volume>\oxi-sc\<digest of repository root>` around the cargo call, alongside the existing linker override. The path budget:

```
C:\Source\oxidizer.worktrees\release-threadaware-fallout   56  (repository root)
  \target\semver-checks\git-<40 sha>\local-fetch-...\...   216 (fixed by the tools)
                                                          = 272  fails
C:\oxi-sc\1a2b3c4d                                         18
                                                          = 234  fits
```

The 216-character suffix is set by `cargo-semver-checks` and `aws-lc-sys` and cannot be shortened from here, which is why the repository path itself cannot be relied on to leave room.

The volume is the repository's own, so the build stays on the filesystem the developer chose, and the digest keeps sibling clones apart. The path is deterministic rather than unique so consecutive runs reuse the baseline rustdoc they just built; concurrent runs are safe because cargo locks that directory exactly as it does `target/`.

As with the linker, an explicit `CARGO_TARGET_DIR` is respected rather than overridden, and a directory that cannot be created degrades to a warning rather than aborting a release. Non-Windows platforms are unchanged.

The error hint that previously told the caller to shorten the repository path now names `C1083` as well and points at the remaining lever, since shortening the path is no longer the caller's job.

## Scope

Only the release library is changed, which covers `release-packages.ps1` (where the failure was reported) and `scripts/ci/semver-report.ps1`. `anvil-semver-check` calls `cargo semver-checks` directly and is `cargo-anvil`-managed, so it is untouched here; its Windows `pr-fast` leg has the same latent exposure and needs a change in `ox-tools`. It has not failed because GitHub runner workspaces (`D:\a\oxidizer\oxidizer`, 22 chars) leave about 22 characters of headroom.

## Testing

New integration test compiles a C file with the real MSVC toolchain at a >260-character output path, asserts `C1083` specifically so the premise cannot be satisfied by an unrelated failure, then repeats the compile under the relocated root and asserts it succeeds. It skips cleanly when no MSVC toolchain is present.

Twelve new unit tests cover the path shape (volume, length, determinism, clone separation, equivalent spellings, UNC fallback, non-Windows) and the environment-variable lifecycle (applied, created, restored, restored on throw, explicit value respected, unwritable directory warns).

Full Pester suite: 544 passed, 0 failed (unit 427, integration 89, scenarios 28).


scripts/lib/releasing.ps1
```diff
@@ -1152,7 +1233,7 @@ function ConvertFrom-SemverChecksOutput {
     }
 
     $pathHint = if ($IsWindows) {
-        ' If the output contains LNK1104 or a path-length error, a MAX_PATH-bound tool was reached; shorten the repository path.'
+        ' If the output contains LNK1104, C1083 or a path-length error, a MAX_PATH-bound tool was reached despite the short build directory these scripts select; setting CARGO_TARGET_DIR to a shorter path, or moving the repository closer to the volume root, buys back the remaining characters.'
```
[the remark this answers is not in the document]
kate-shine: GHCP: Good catch - the hint asserted a fact that isn't always true, since the short directory is skipped when CARGO_TARGET_DIR is already set and is only best-effort if the directory can't be created. Reworded in 85b52945 to state the constraint rather than the action taken: it now names the ~216-character budget below the target directory and points at CARGO_TARGET_DIR (with the short path as an example) as the lever, without claiming anything about what the build actually used.

docs/releasing.md
```diff
@@ -318,6 +318,41 @@ be read from `rustc -vV` the default linker stands (with a warning).
 Other hosts are unaffected — `*-windows-gnu` and non-Windows targets
 already default to linkers that handle long paths.
 
+#### Windows: baseline builds run under a short target directory
```
sandersaares, line 321: Have we filed an upstream issue for this? It would be better if cargo-semver-checks could be explicitly given a shorter root or if it acted more intelligently by default (would `%TEMP%` be a better default?). Let's file an issue of some appropriate breadth if not already filed and link it here.

docs/releasing.md
```diff
@@ -318,6 +318,41 @@ be read from `rustc -vV` the default linker stands (with a warning).
...
+
+Redirecting the linker clears `LNK1104`, but it does not help the C
+compilers that `-sys` crates drive through the `cc` crate. MSVC `cl.exe`
+resolves its `-Fo` argument against `MAX_PATH` and fails with `C1083`
+("Cannot open compiler generated file"), and unlike the linker there is
+no long-path-aware drop-in guaranteed to be installed — `clang-cl` is
+not part of a default Rust or Visual Studio install. So the remaining
+lever is the length of the path itself.
+
+`cargo semver-checks` nests baseline builds under the workspace target
```
sandersaares, line 331: The PR description contains some nice examples of path length calculation to illustrate the problem. Would be nice to also have it here.

docs/releasing.md
```diff
@@ -318,6 +318,41 @@ be read from `rustc -vV` the default linker stands (with a warning).
...
+directory and offers no flag to move them, but it derives that location
+from cargo metadata, so `CARGO_TARGET_DIR` does reach it. On Windows the
+release scripts therefore point it at
+`<volume>\oxi-sc\<digest-of-repository-root>` for the duration of the
```
sandersaares, line 335: Does it make sense to skip this if the original path is already short or is that pretty much never going to happen?

sandersaares (approved):
It's a nasty one but I guess we have no good choice. Feels dirty.

docs/releasing.md
kate-shine: Sadly, temp still gives long paths. At the end, it's quite an endless battle, unless it's fixed at the correct upstream - Microsoft's build tooling.

kate-shine: But I will check and fill an issue if it's not there already :)

kate-shine:
> It's a nasty one but I guess we have no good choice. Feels dirty.

Yea, I hate it.

docs/releasing.md
kate-shine: GHCP: Checked upstream - there is no existing issue for this. `cargo-semver-checks` has nothing matching MAX_PATH, long paths, or Windows path length, and the closest neighbours (#402 `Reuse existing target directory`, #401, #1272) are about caching and metadata rather than path length. So it would be a new report.

We have decided to leave it out of scope for this PR rather than file one now. Two reasons: `%TEMP%` is not the escape it looks like (`C:\Users\<user>\AppData\Local\Temp` is 35+ characters on a normal profile, worse than the 18 we use), and the nesting that actually hurts is only partly cargo-semver-checks' - of the 216 characters it is `\semver-checks\git-<40>\local-<name>-<ver>-<triple>-<16hex>` = 118, with the rest coming from cargo, `aws-lc-sys` and `cc`. The real fix is long-path awareness in the MSVC toolchain, not a shorter layout in one tool.

A `--target-dir` flag upstream would still be an improvement over having to go through the environment variable, so it is worth raising eventually - just not blocking this.

docs/releasing.md
kate-shine: GHCP: Added in 8cb6c5f5, as a table of the segment lengths plus the two worked examples.

While writing it up I found the PR description had one number wrong: it said the relocation brings the failing build to 234 characters. The correct figure is 227 - the 216 is measured from the repository root and so includes its `\target` (7 characters), which our short root replaces rather than adds to. The 272 in the field report reproduces exactly (56 + 216), so the breakdown is verified rather than estimated.

The docs now also state the CI headroom, since that seemed the more actionable half: a GitHub-hosted runner checks out at `D:\a\oxidizer\oxidizer` (22 characters), landing the same build at 238 - passing today with about 21 characters to spare.

docs/releasing.md
kate-shine: GHCP: It would not be never - a short checkout genuinely fits. The budget below the repository root is 216, so anything up to roughly 43 characters of root clears 260 unaided, which covers `C:\Users\<user>\src\oxidizer` and most normal layouts. The failing case was a 56-character worktree path.

We kept it unconditional anyway, and documented why in 8cb6c5f5. The threshold would have to be that 216, and it is not a constant anyone here controls: it moves with crate and target names (`local-fetch-0_15_0-x86_64_pc_windows_msvc-<16hex>` alone is 59, and a longer crate name lengthens it), with dependency versions, and with hash widths. A threshold fitted to one measurement would silently stop covering the case it exists for, and the failure mode is an opaque C1083 during a release rather than anything self-explanatory.

The second reason is test exposure: making it conditional would reserve the relocation path for the deepest checkouts, which are the rarest configurations and the least likely to be exercised before someone is mid-release. Unconditional keeps one path in use on every Windows run.

The cost of applying it needlessly is small and is called out in the docs - artifacts land outside the repository, so `cargo clean` and `git clean` miss them and `<volume>\oxi-sc` has to be deleted by hand. Happy to add the threshold if you would rather trade that off the other way.

kate-shine: GHCP: Implemented in e84618b7 - relocation is now conditional.

Rather than a bare length threshold, the decision projects the actual build path from three inputs: the repository root, the 216-character nesting, and the name of the package being checked. That third one matters more than I expected - the nesting embeds the package name once, so `http_client_api_with_templated_uri` (our longest workspace crate) reaches 29 characters further than `fetch` did. A threshold on the root alone would have been wrong for exactly the crates most likely to fail.

Behaviour across representative checkouts, with `fetch`:

| Root | Len | Decision |
| --- | ---: | --- |
| `D:\oxidizer` (Dev Drive) | 11 | default target dir |
| `C:\Source\oxidizer` | 18 | default target dir |
| `C:\src\repos\oxidizer` | 21 | relocate |
| `C:\Users\<user>\src\oxidizer` | 31 | relocate |
| the AB#7790786 worktree | 56 | relocate |

`D:\oxidizer` flips to relocate once the package is `http_client_api_with_templated_uri`, which is the projection doing its job.

I kept a 24-character margin, documented as a judgement call rather than derived: the package name is projected for explicitly, but the version string and the depth of a dependency's own build output are not ours to predict, and a dependency upgrade lengthens either. Your point that we can always tell them why it failed is what makes that safe - the hint now says the relocation only applies when the path looks long enough to need it, and names `CARGO_TARGET_DIR` as the manual override.

UNC roots skip the projection and always relocate, since the prefix is itself the length we are trying to shed.

Full Pester suite green: 547 passed, 0 failed. Added tests for the short-path skip, the package-name-driven flip, and that one clone still gets one directory whichever package is checked (so a multi-package release shares baseline artifacts).

scripts/lib/releasing.ps1
```diff
@@ -1001,8 +1001,66 @@ function Get-SemverChecksLinkerEnvName {
...
+
+    if (-not $IsWindows) {
+        return $null
+    }
+
+    # Normalise first: the digest has to be stable across callers that spell the
+    # same repository root differently (trailing separator, relative segments,
+    # or casing, none of which Windows treats as distinct).
+    $full = [System.IO.Path]::GetFullPath($RepoRoot).TrimEnd('\')
+
```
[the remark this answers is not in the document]
kate-shine: GHCP: Agreed, and fixed in ae961f6e. Aborting here would strand exactly the long paths the function exists for.

`GetFullPath` is now wrapped: on failure it warns, keeps the raw string, and marks the result unnormalised. The unnormalised flag also suppresses the projection - a path we could not parse is not evidence that it is short, so the build relocates rather than being left in place. `GetPathRoot` is guarded the same way and falls back to the system drive.

Note this is the opposite of the `return \` option you offered: failing open to the default target directory would hand back the C1083 we are trying to avoid. Failing open to the *relocation* keeps the protection, and the relocation itself already fails open if the directory cannot be created.

Covered by a new unit test using an embedded NUL, which is an input `GetFullPath` genuinely rejects.

scripts/tests/Pester/unit/releasing/SemverChecksLinker.Tests.ps1
```diff
@@ -77,11 +77,82 @@ LLVM version: 20.1.0
...
+    It 'roots the build at the volume of the repository on Windows' -Skip:(-not $IsWindows) {
+        $path = Get-SemverChecksTargetDirPath -RepoRoot 'C:\Source\oxidizer.worktrees\some-long-branch-name'
+
+        $path | Should -BeLike 'C:\oxi-sc\*'
+    }
+
+    It 'stays short enough to leave the MAX_PATH budget to the build itself' -Skip:(-not $IsWindows) {
+        # The observed worst case nests ~209 characters of cargo-semver-checks
+        # and aws-lc-sys build output beneath the target directory, so the root
+        # has to stay well clear of 260.
```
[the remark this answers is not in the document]
kate-shine: GHCP: Both numbers are correct but the comment did not say so, which is a fair thing to flag. 216 is measured from the repository root and includes its `\target` (7 characters); 209 is what remains below the target directory, because the short root replaces that `\target` rather than adding to it. This test asserts the length of the root itself, so 209 is the relevant figure.

Reworded in ae961f6e to state the relationship explicitly rather than leaving the reader to reconcile two numbers. The docs table already totals 216 and labels it 'below the repository root'.

scripts/tests/Pester/integration/SemverChecksLinker.Tests.ps1
```diff
@@ -84,3 +84,88 @@ edition = "2021"
...
+            # C1083 specifically: a laxer pattern would let an unrelated failure
+            # satisfy the premise and make the assertion below vacuous.
+            $deepResult.Output | Should -Match 'C1083'
+
+            # Now the same compile beneath the directory the release scripts pick.
+            $shortRoot = Get-SemverChecksTargetDirPath -RepoRoot $root
+            $shortDir = Join-Path $shortRoot 'probe'
+            $null = New-Item -ItemType Directory -Path $shortDir -Force
+            try {
+                $shortResult = & $compile (Join-Path $shortDir $leaf)
```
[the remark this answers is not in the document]
kate-shine: GHCP: Fixed in ae961f6e. The creation is now wrapped and calls `Set-ItResult -Skipped` naming the underlying error, so a locked-down machine reports a skip rather than a failure. That matches production, which warns and builds in place in the same situation.

The first half of the test still asserts C1083 before this point, so the skip cannot hide a regression in the premise - it only skips the half that needs the writable volume root.

scripts/lib/releasing.ps1
```diff
@@ -1001,8 +1001,107 @@ function Get-SemverChecksLinkerEnvName {
...
+    )
+
+    if (-not $IsWindows) {
+        return $null
+    }
+
+    # Normalise first: the digest has to be stable across callers that spell the
+    # same repository root differently (trailing separator, relative segments,
+    # or casing, none of which Windows treats as distinct).
+    $full = [System.IO.Path]::GetFullPath($RepoRoot).TrimEnd('\')
```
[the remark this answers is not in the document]
kate-shine: GHCP: This one was a real bug, not just wording - thanks. `GetFullPath('C:\')` returns `C:\`, and `TrimEnd('\')` reduced that to `C:`, which names the current directory on that drive rather than its root. It also gave `C:\` and `C:` different digests, defeating the normalisation the trim exists for.

Fixed in ae961f6e: the trimmed value is only taken when it is not a bare drive specifier, so `C:\` stays `C:\`. Verified that `C:\` and `C:\\` now produce the same directory.

On trimming both slash types - not needed here, and I checked rather than assuming: `GetFullPath` already rewrites forward slashes on Windows, so `C:/Source/oxidizer/` comes back as `C:\Source\oxidizer\` with only a backslash left to trim. Added a comment saying so, since it is the kind of thing worth not re-litigating later.

scripts/lib/releasing.ps1
```diff
@@ -1152,7 +1274,7 @@ function ConvertFrom-SemverChecksOutput {
     }
 
     $pathHint = if ($IsWindows) {
-        ' If the output contains LNK1104 or a path-length error, a MAX_PATH-bound tool was reached; shorten the repository path.'
+        " If the output contains LNK1104, C1083 or a path-length error, a MAX_PATH-bound tool was reached. Build artifacts can nest over 200 characters below the target directory, and these scripts relocate the build to a short path only when the repository path looks long enough to need it. Set CARGO_TARGET_DIR to a short path (for example ${env:SystemDrive}\$script:SemverChecksTargetDirName) to force that, or move the repository closer to the volume root."
```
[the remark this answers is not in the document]
kate-shine: GHCP: Correct, and this became wrong only an hour ago when the projection started factoring in the package name. Reworded in ae961f6e to `the repository path and the package name together project past that limit`, which is what the code actually does - a short checkout really can overflow on a long-named crate such as `http_client_api_with_templated_uri`.

scripts/lib/releasing.ps1
```diff
@@ -1001,8 +1001,128 @@ function Get-SemverChecksLinkerEnvName {
...
+    # -- which would strand exactly the long paths this exists for -- fall back
+    # to the raw string and let the relocation proceed unprojected.
+    $normalised = $true
+    try {
+        $full = [System.IO.Path]::GetFullPath($RepoRoot)
+    } catch {
+        Write-Warning "Could not normalise '$RepoRoot' ($($_.Exception.Message)); relocating the baseline build on the strength of the path as given."
+        $full = $RepoRoot
+        $normalised = $false
+    }
```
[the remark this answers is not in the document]
kate-shine: GHCP: Right - the fallback inherited an assumption from the branch it replaces. Fixed in be76db59: the catch now applies `-replace '/', '\'` itself, standing in for the one part of `GetFullPath` that still applies when the rest of it could not run.

Worth noting the blast radius is small, since the consequence is a second build directory for one clone rather than a wrong or unsafe path, and only for a root malformed enough that `GetFullPath` rejects it. Still worth fixing: the whole point of the digest is that a rerun reuses the baseline rustdoc it just built, and this path would silently forfeit that.

Pinned by a test asserting that `C:\repo<NUL>broken` and `C:/repo<NUL>broken` resolve to the same directory. 249 unit tests pass.

scripts/lib/releasing.ps1
```diff
@@ -1046,6 +1170,26 @@ function Invoke-SemverChecksCli {
...
+                Write-Verbose 'CARGO_TARGET_DIR is already set; building where it points.'
+            } else {
+                try {
+                    $null = New-Item -ItemType Directory -Path $targetDir -Force -ErrorAction Stop
+                    $env:CARGO_TARGET_DIR = $targetDir
+                    $targetDirApplied = $true
+                } catch {
+                    # Fail open -- a probe must never break a release. Name the
+                    # symptom, though, so a later C1083 is not a mystery.
+                    Write-Warning "Could not create '$targetDir'; building under the default target directory. If the repository path is long, the baseline build may fail with C1083. ($($_.Exception.Message))"
```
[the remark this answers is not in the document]
kate-shine: GHCP: Fixed in c684eb6a, along with the suppressed comment on line 1302, since the two share a root cause.

For the `New-Item` warning: by the time we reach that catch, `Get-SemverChecksTargetDirPath` has already returned a directory, which only happens when relocation was judged necessary. So the conditional framing was doubly wrong - it named the wrong input *and* hedged about something already decided. It now states it plainly: the default target directory 'was projected to be too long for MSVC and so may fail with C1083'.

For the hint on 1302, the suppressed comment made the better point: enumerating conditions is what keeps making these strings wrong. Relocation also applies to UNC roots and to paths that cannot be normalised, neither of which is a projection at all, so `repository path and package name together` was incomplete the moment I wrote it. Reworded to describe the mechanism rather than its inputs - the scripts relocate 'when they judge the default one too long, but that judgement is a projection and can fall short' - which stays true whichever way the decision is reached, and keeps `CARGO_TARGET_DIR` as the stated override.

Note the branch had `main` merged into it remotely while I was working, so this is rebased onto 0363afb9. Re-ran the full unit suite on the rebased tree: 434 passed, 0 failed.

docs/releasing.md
sandersaares: > A --target-dir flag upstream would still be an improvement over having to go through the environment variable, so it is worth raising eventually - just not blocking this.

So let's file it. I do not think submitting a single GitHub issue is something that deserves to be deferred.

scripts/lib/releasing.ps1
```diff
@@ -1001,8 +1001,132 @@ function Get-SemverChecksLinkerEnvName {
...
+        $volume = "$env:SystemDrive\"
+    }
+
+    # UNC has already lost the argument -- its own prefix is long enough that
+    # the projection below is not worth consulting -- and an unnormalised path
+    # is not a length worth trusting.
+    if (-not $isUnc -and $normalised) {
+        $nameShift = $PackageName.Length - $script:SemverChecksNestingPackage.Length
+        $projected = $full.Length + $script:SemverChecksNestingLength + $nameShift
+        if (($projected + $script:SemverChecksNestingSlack) -le $script:MaxPathLength) {
```
martin-kolinek, line 1111: 🤖: Non-blocking — the per-package relocation decision can split one release's baseline artifacts between `<repo>\target` and `<volume>\oxi-sc`, duplicating dependency builds and leaving only some artifacts reachable by `cargo clean`.

**Choose relocation once per clone, or document that short checkouts can use both caches.** The tests demonstrate the split: `D:\oxidizer` stays local for `fetch` but relocates for `http_client_api_with_templated_uri`. This also makes `docs/releasing.md:385` too broad when it says that checkout keeps its build output inside the repository.

scripts/lib/releasing.ps1
```diff
@@ -1001,8 +1001,132 @@ function Get-SemverChecksLinkerEnvName {
...
+
+    $sha = [System.Security.Cryptography.SHA256]::Create()
+    try {
+        $digest = $sha.ComputeHash([System.Text.Encoding]::UTF8.GetBytes($full.ToLowerInvariant()))
+    } finally {
+        $sha.Dispose()
+    }
+    $token = [System.BitConverter]::ToString($digest[0..3]).Replace('-', '').ToLowerInvariant()
+
+    return (Join-Path $volume (Join-Path $script:SemverChecksTargetDirName $token))
```
martin-kolinek, line 1125: 🤖: Non-blocking — on a shared Windows host, another authenticated user can pre-create `<volume>\oxi-sc` and control the ACL of the Cargo target directory where build-script executables are stored and run.

**Document that relocation assumes a trusted machine, or verify directory ownership before using it and fall back with a warning.** This trust boundary is new: the repository-local target directory is normally protected by the repository's ACL, while the deterministic volume-root path is guessable and its parent can be created by other local users under the default volume-root ACL.

docs/releasing.md
```diff
@@ -318,6 +318,85 @@ be read from `rustc -vV` the default linker stands (with a warning).
...
+to 227, leaving about 32 characters of headroom.
+
+That headroom is what the relocation buys, and it is worth noting how
+little of it there is by default. A GitHub-hosted runner checks out at
+`D:\a\oxidizer\oxidizer`, which is only 22 characters, so the same build
+lands at 238 and passes today with roughly 21 characters to spare — a
+longer crate name or a dependency version bump would be enough to
+consume it.
+
+As with the linker, an explicit `CARGO_TARGET_DIR` is respected rather
```
martin-kolinek, line 375: 🤖: Non-blocking — `scripts/release-packages.ps1:94`-`:99` still documents only the linker mitigation, so its operator-facing help omits the new out-of-repository artifact location.

**Update the entry-point help to describe conditional `CARGO_TARGET_DIR` relocation, explicit-variable precedence, and manual cleanup of `<volume>\oxi-sc`.**

martin-kolinek (approved):
🤖: Approved.

scripts/lib/releasing.ps1
kate-shine: GHCP: Tracked in #708. We are deferring the non-blocking split-cache documentation follow-up from this PR.

scripts/lib/releasing.ps1
kate-shine: GHCP: Tracked in #708. We are deferring the non-blocking volume-root ACL trust-boundary follow-up from this PR.

docs/releasing.md
kate-shine: GHCP: Tracked in #708. We are deferring the non-blocking operator-help update from this PR.

scripts/lib/releasing.ps1
```diff
@@ -1001,8 +1001,132 @@ function Get-SemverChecksLinkerEnvName {
...
+    # current directory on that drive rather than its root.
+    $trimmed = $full.TrimEnd('\')
+    if ($trimmed -notmatch '^[A-Za-z]:$') {
+        $full = $trimmed
+    }
+
+    # A UNC root has no drive letter to anchor to and would keep the very
+    # length this function exists to shed, so fall back to the system drive.
+    $volume = try { [System.IO.Path]::GetPathRoot($full) } catch { '' }
+    $isUnc = $volume.StartsWith('\\')
```
martin-kolinek, line 1100: 🤖: Non-blocking — `$volume.StartsWith('\\')` can dereference `$null` before the whitespace guard runs.

**Guard `$volume` before calling `StartsWith`.** A whitespace-only `$RepoRoot` makes `GetPathRoot` return `$null` without throwing, so this aborts the release instead of taking the function's fail-open path.

docs/releasing.md
kate-shine: Yep https://github.com/obi1kenobi/cargo-semver-checks/issues/1720

scripts/lib/releasing.ps1
kate-shine: GHCP: Tracked in #708. We are deferring this non-blocking null-guard follow-up from this PR.

PR #704, opened 2026-08-27
test(multitude): drive alignment guards through an injectable cap

AdomasBekeras:
## The bug

[ADO 7707893](https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7707893): `cargo test` fails on a clean checkout.

The arena refuses allocations whose alignment reaches a cap. `CHUNK_ALIGN` is 64 KiB, and the smart-pointer cap is half of it, 32 KiB, because a smart pointer recovers its chunk header by masking the value pointer's offset within its chunk tile — a value aligned that far can land outside the first tile, where the mask finds a different chunk's header.

To test the rejection, the tests had to instantiate a type aligned at or above the cap. So they declared `#[repr(align(32768))]`, `#[repr(align(65536))]` and `#[repr(align(131072))]` types. Some codegen backends cap type alignment at 8192 and refuse to compile such a type at all. The library built fine; three test binaries (`arena`, `audit_repro`, `pin_support`) and one doctest failed codegen.

## Why the previous fix didn't hold

#501 gated the tests behind `#[cfg(not(utc_backend))]` (since renamed `align_capped_backend`), with the flag set by an out-of-tree CI pipeline. Three problems:

- Nothing in this repo sets it, so the **default** build was the broken one. A contributor cloning the repo hits the failure; the gate only helps a build that already knows to opt in.
- It's all-or-nothing. A backend that sets it loses the guard coverage entirely, on a live memory-safety check.
- It missed the doctest, because rustdoc ignores `RUSTFLAGS`. Setting `RUSTFLAGS` would also have clobbered `.cargo/config.toml`'s `-C target-cpu=x86-64-v3`.

## This change

The test needs the type's alignment and the cap to meet. The old approach raised the alignment to the cap. This one lowers the cap to an alignment every backend compiles.

`Arena` gets a `#[cfg(test)]` alignment cap that the guards read:

```rust
#[cfg(not(test))]
fn chunk_align_cap(&self) -> usize { CHUNK_ALIGN }

#[cfg(test)]
fn chunk_align_cap(&self) -> usize { self.align_cap.get() }
```

Tests call `capped_arena()`, which sets the cap to 8192, and use shared helper types aligned to 4096 and 8192. Both boundaries stay reachable, and the production 2:1 ratio between the chunk cap and the smart-pointer cap is preserved, so each test still exercises the cap its entry point actually consults.

The `cfg(test)` field and the whole knob disappear from production builds — `Arena`'s layout is unchanged.

Along the way:

- Three duplicate `MAX_SMART_PTR_ALIGN` constants collapse into one accessor. All nine guard sites now route through `rejects_smart_ptr_align` / `rejects_chunk_align`.
- ~46 over-alignment tests move from `tests/` into `#[cfg(test)]` modules in `src/` (`arena/align_guard_tests.rs`, `bytemuck.rs`, `zerocopy.rs`) so they can reach the knob.
- The `align_capped_backend` cfg is deleted from the workspace `Cargo.toml`.
- Net −48 lines.

## Coverage

Nothing was dropped. Two additions beyond parity:

- One test asserts the error is specifically `is_alignment_too_large()`, which nothing outside the deleted doctest checked before.
- `try_alloc_slice_fill_iter`'s guard had no over-alignment test at all; it does now.

Mutation-checked by hand: forcing `rejects_smart_ptr_align` to `false` fails 39 tests, forcing `rejects_chunk_align` to `false` fails 9.

## Things worth a reviewer's attention

**The guards are no longer `const { }`.** They were `if const { align_of::<T>() >= MAX_SMART_PTR_ALIGN }`, folded at compile time by construction. They are now ordinary comparisons against an `#[inline(always)]` accessor. In release under `cfg(not(test))` that accessor returns a literal and `align_of::<T>()` is a constant, so LLVM folds it; debug builds pay a compare. The guarantee is gone, the behaviour isn't. Restoring the guarantee would need a macro expanding to the `const { }` form under `cfg(not(test))` — happy to add it if you'd rather have the certainty.

**`buffer_freezable` still reads the real cap.** It's used inside `const { }` on the `Vec` hot path, so making it cap-aware would put a runtime branch there. The consequence is that a capped arena is not a faithful model for `Vec`/`String` growth and freeze tests — documented on `capped_arena()` and `set_align_cap()`, and `set_align_cap` now asserts the cap can only be *lowered*. A new lib test asserts the arena's default caps equal `CHUNK_ALIGN` and `max_smart_ptr_align()`, so the two sources can't drift apart silently.

**One over-aligned type survives.** `non_freezable_overaligned_vec_grows_via_oversized_path` in `tests/arena.rs` still declares `#[repr(align(32768))]`. It's the one place where the alignment *is* the subject — it's what makes the element non-freezable — and it compiles because `try_reserve` never materialises the layout. That's an emergent property, not a guarantee, so there's now a comment naming it as the one fragile declaration left, to make a future failure diagnosable.

## Verification

- `cargo test -p multitude --all-features` passes on **both** the alignment-capped backend and an LLVM-backend toolchain.
- `cargo clippy -p multitude --all-features --all-targets -- -D warnings` clean.
- Formatted with the pinned nightly.
- `cargo spellcheck` could not be run — the binary is broken in my environment (missing DLL). Please let CI cover it.

## Not fixed here

A clean-checkout `cargo build --workspace` on the internal toolchain also fails in `zeroize` 1.9.0 (reached via `fetch*` → `rustls` → `aws-lc-rs`) with `codegen not yet implemented for Terminator_InlineAsm`. Unrelated to alignment and not fixable in this repo. Worth tracking separately.

### DIFF
--- a/crates/multitude/src/arena/align_guard_tests.rs
+++ b/crates/multitude/src/arena/align_guard_tests.rs
@@ -1,0 +1,211 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Every allocation entry point rejects a type aligned at or above the
+//! arena's cap.
+//!
+//! Smart pointers recover their chunk header by masking the value
+//! pointer's offset within its chunk tile. A value aligned at or above
+//! the cap can land outside the first tile, where the mask recovers a
+//! different chunk's header and `Drop` corrupts it. The guards make that
+//! unreachable from safe code, and these tests hold every entry point to
+//! it.
+//!
+//! The tests run against [`capped_arena`], whose caps are lowered so the
+//! boundary is reachable by an alignment every codegen backend accepts.
+
+use crate::Arena;
+use crate::internal::constants::{CHUNK_ALIGN, max_smart_ptr_align};
+use crate::tests_support::{ChunkOverAligned, SmartPtrOverAligned, SmartPtrOverAlignedDrop, capped_arena};
+
+// The guards read their cap from the arena; `buffer_freezable` still reads
+// `max_smart_ptr_align()` directly. The two must agree, or a `Vec` would
+// freeze in place for an element the smart-pointer path rejects.
+#[test]
+fn default_caps_match_the_constants() {
+    let arena = Arena::new();
+    assert_eq!(arena.chunk_align_cap(), CHUNK_ALIGN);
+    assert_eq!(arena.smart_ptr_align_cap(), max_smart_ptr_align());
+}
+
+#[test]
+#[should_panic(expected = "the cap may only be lowered")]
+fn set_align_cap_rejects_raising_the_cap() {
+    Arena::new().set_align_cap(CHUNK_ALIGN * 2);
+}
+
+#[test]
+fn rejection_reports_alignment_too_large() {
+    let arena = capped_arena();
+    let err = arena.try_alloc_with(|| SmartPtrOverAligned(0)).unwrap_err();
+    assert!(err.is_alignment_too_large());
+
+    let err = arena.try_alloc_arc_with(|| SmartPtrOverAlignedDrop(0)).unwrap_err();
+    assert!(err.is_alignment_too_large());
+
+    // A local binding, not `&[…][..]`: an over-aligned temporary that gets
+    // const-promoted lands in a section the linker rejects.
+    let src = [ChunkOverAligned(0)];
+    let err = arena.try_alloc_slice_copy(&src[..]).unwrap_err();
+    assert!(err.is_alignment_too_large());
+}
+
+#[test]
+fn try_alloc_with_rejects_smart_ptr_alignment() {
+    let arena = capped_arena();
+    arena.try_alloc_with(|| SmartPtrOverAligned(0)).unwrap_err();
+}
+
+#[test]
+fn try_alloc_arc_with_rejects_smart_ptr_alignment() {
+    let arena = capped_arena();
+    arena.try_alloc_arc_with(|| SmartPtrOverAlignedDrop(0)).unwrap_err();
+}
+
+#[test]
+fn try_alloc_box_with_rejects_smart_ptr_alignment() {
+    let arena = capped_arena();
+    arena.try_alloc_box_with(|| SmartPtrOverAlignedDrop(0)).unwrap_err();
+}
+
+#[test]
+fn try_alloc_uninit_box_rejects_smart_ptr_alignment() {
+    let arena = capped_arena();
+    arena.try_alloc_uninit_box::<SmartPtrOverAlignedDrop>().unwrap_err();
+}
+
+#[test]
+fn try_alloc_uninit_arc_rejects_smart_ptr_alignment() {
+    let arena = capped_arena();
+    arena.try_alloc_uninit_arc::<SmartPtrOverAlignedDrop>().unwrap_err();
+}
+
+#[test]
+fn try_alloc_slice_copy_arc_rejects_smart_ptr_alignment() {
+    let arena = capped_arena();
+    let src = [SmartPtrOverAligned(0), SmartPtrOverAligned(1)];
+    arena.try_alloc_slice_copy_arc(&src[..]).unwrap_err();
+}
+
+#[test]
+fn try_alloc_arc_with_no_drop_rejects_smart_ptr_alignment() {
+    let arena = capped_arena();
+    arena.try_alloc_arc_with(|| SmartPtrOverAligned(0)).unwrap_err();
+}
+
+#[test]
+fn try_alloc_slice_fill_with_arc_rejects_smart_ptr_alignment() {
+    let arena = capped_arena();
+    arena.try_alloc_slice_fill_with_arc(1, |_| SmartPtrOverAlignedDrop(0)).unwrap_err();
+}
+
+#[test]
+fn try_alloc_slice_fill_with_arc_no_drop_rejects_smart_ptr_alignment() {
+    let arena = capped_arena();
+    arena.try_alloc_slice_fill_with_arc(2, |_| SmartPtrOverAligned(0)).unwrap_err();
+}
+
+#[test]
+fn try_alloc_slice_fill_with_box_rejects_smart_ptr_alignment() {
+    let arena = capped_arena();
+    arena.try_alloc_slice_fill_with_box(1, |_| SmartPtrOverAlignedDrop(0)).unwrap_err();
+}
+
+#[test]
+fn try_alloc_uninit_slice_arc_rejects_smart_ptr_alignment() {
+    let arena = capped_arena();
+    arena.try_alloc_uninit_slice_arc::<SmartPtrOverAlignedDrop>(1).unwrap_err();
+}
+
+#[test]
+fn try_alloc_uninit_slice_box_rejects_smart_ptr_alignment() {
+    let arena = capped_arena();
+    arena.try_alloc_uninit_slice_box::<SmartPtrOverAlignedDrop>(1).unwrap_err();
+}
+
+// A pinned box hands out a smart pointer, so it rejects at the smart-pointer
+// cap like the other box entry points.
+#[test]
+fn try_alloc_uninit_box_pin_rejects_smart_ptr_alignment() {
+    let arena = capped_arena();
+    arena.try_alloc_uninit_box_pin::<SmartPtrOverAligned>().unwrap_err();
+}
+
+// Simple-reference slices hand back a plain `&mut [T]` with no header
+// recovery, so they use the looser chunk cap and only reject at it.
+#[test]
+fn try_alloc_slice_fill_with_rejects_chunk_alignment() {
+    let arena = capped_arena();
+    arena.try_alloc_slice_fill_with(1, |_| ChunkOverAligned(0)).unwrap_err();
+}
+
+#[test]
+fn try_alloc_slice_copy_rejects_chunk_alignment() {
+    let arena = capped_arena();
+    let src = [ChunkOverAligned(0)];
+    arena.try_alloc_slice_copy(&src[..]).unwrap_err();
+}
+
+#[test]
+fn try_alloc_slice_fill_iter_rejects_chunk_alignment() {
+    let arena = capped_arena();
+    arena.try_alloc_slice_fill_iter((0..1).map(|_| ChunkOverAligned(0))).unwrap_err();
+}
+
+// A type aligned below the chunk cap is accepted by the simple-reference
+// slice paths: the boundary the tighter smart-pointer cap does not apply to.
+#[test]
+fn alloc_slice_ref_accepts_below_chunk_alignment_for_non_drop() {
+    let arena = capped_arena();
+    let filled = arena.alloc_slice_fill_with(1, |_| SmartPtrOverAligned(0));
+    assert_eq!(filled.len(), 1);
+
+    let src = [SmartPtrOverAligned(0)];
+    let cloned = arena.alloc_slice_clone(&src[..]);
+    assert_eq!(cloned.len(), 1);
+}
+
+// Panicking entry points route through the same guards and surface the
+// rejection as the arena's allocation-failure panic.
+#[test]
+#[should_panic(expected = "multitude: allocator returned AllocError")]
+fn alloc_with_panics_on_over_alignment() {
+    let arena = capped_arena();
+    let _ = arena.alloc_with(|| SmartPtrOverAligned(0));
+}
+
+#[test]
+#[should_panic(expected = "multitude: allocator returned AllocError")]
+fn alloc_arc_with_panics_on_over_alignment() {
+    let arena = capped_arena();
+    let _ = arena.alloc_arc_with(|| SmartPtrOverAligned(0));
+}
+
+#[test]
+#[should_panic(expected = "multitude: allocator returned AllocError")]
+fn alloc_box_with_panics_on_over_alignment() {
+    let arena = capped_arena();
+    let _ = arena.alloc_box_with(|| SmartPtrOverAligned(0));
+}
+
+#[test]
+#[should_panic(expected = "multitude: allocator returned AllocError")]
+fn alloc_uninit_box_panics_on_over_alignment() {
+    let arena = capped_arena();
+    let _ = arena.alloc_uninit_box::<SmartPtrOverAligned>();
+}
+
+#[test]
+#[should_panic(expected = "multitude: allocator returned AllocError")]
+fn alloc_uninit_arc_panics_on_over_alignment() {
+    let arena = capped_arena();
+    let _ = arena.alloc_uninit_arc::<SmartPtrOverAligned>();
+}
+
+#[test]
+#[should_panic(expected = "multitude: allocator returned AllocError")]
+fn alloc_slice_copy_panics_on_over_alignment() {
+    let arena = capped_arena();
+    let src = [ChunkOverAligned(0)];
+    let _ = arena.alloc_slice_copy(&src[..]);
+}
--- a/crates/multitude/src/arena/align_guard_tests.rs
+++ b/crates/multitude/src/arena/align_guard_tests.rs
@@ -16,7 +16,7 @@
 
 use crate::Arena;
 use crate::internal::constants::{CHUNK_ALIGN, max_smart_ptr_align};
-use crate::tests_support::{ChunkOverAligned, SmartPtrOverAligned, SmartPtrOverAlignedDrop, capped_arena};
+use crate::tests_support::{ChunkOverAligned, SmartPtrOverAligned, SmartPtrOverAlignedDrop, TEST_CHUNK_ALIGN, capped_arena};
 
 // The guards read their cap from the arena; `buffer_freezable` still reads
 // `max_smart_ptr_align()` directly. The two must agree, or a `Vec` would
@@ -31,7 +31,42 @@ fn default_caps_match_the_constants() {
 #[test]
 #[should_panic(expected = "the cap may only be lowered")]
 fn set_align_cap_rejects_raising_the_cap() {
+    // Must stay a power of two, or the earlier assert fires instead.
     Arena::new().set_align_cap(CHUNK_ALIGN * 2);
+}
+
+#[test]
+#[should_panic(expected = "alignment cap must leave room")]
+fn set_align_cap_rejects_a_degenerate_cap() {
+    Arena::new().set_align_cap(1);
+}
+
+// The guards compare `>=`, so they must reject alignments strictly above the
+// cap as well as those exactly at it. Every fixture is aligned exactly at a
+// cap, so reaching "strictly above" means lowering the cap further rather
+// than raising an alignment.
+#[test]
+fn guards_reject_alignments_above_the_cap() {
+    let arena = capped_arena();
+    arena.set_align_cap(TEST_CHUNK_ALIGN / 2);
+
+    assert!(arena.rejects_smart_ptr_align(align_of::<SmartPtrOverAligned>()));
+    assert!(arena.rejects_chunk_align(align_of::<ChunkOverAligned>()));
+
+    arena.try_alloc_with(|| SmartPtrOverAligned(0)).unwrap_err();
+
+    let src = [ChunkOverAligned(0)];
+    arena.try_alloc_slice_copy(&src[..]).unwrap_err();
+}
+
+// The counterpart to the above: an alignment strictly below the cap is
+// accepted, so the guards are not simply rejecting everything.
+#[test]
+fn guards_accept_alignments_below_the_cap() {
+    let arena = capped_arena();
+
+    assert!(!arena.rejects_smart_ptr_align(align_of::<u64>()));
+    assert!(!arena.rejects_chunk_align(align_of::<SmartPtrOverAligned>()));
 }
 
 #[test]
--- a/crates/multitude/src/arena/mod.rs
+++ b/crates/multitude/src/arena/mod.rs
@@ -595,6 +595,7 @@ impl<A: Allocator + Clone> Arena<A> {
             cap <= CHUNK_ALIGN,
             "the cap may only be lowered: raising it would let the guards accept alignments no chunk can satisfy"
         );
+        assert!(cap >= 4, "alignment cap must leave room for a non-zero smart-pointer cap");
         self.align_cap.set(cap);
     }
 
--- a/crates/multitude/src/arena/mod.rs
+++ b/crates/multitude/src/arena/mod.rs
@@ -18,7 +18,7 @@ use crate::internal::chunk::Chunk;
 use crate::internal::chunk_mutator::ChunkMutator;
 use crate::internal::chunk_provider::{ChunkProvider, ChunkProviderConfig};
 use crate::internal::chunk_ref::ChunkRef;
-use crate::internal::constants::{MAX_NORMAL_ALLOC, SizeClass};
+use crate::internal::constants::{CHUNK_ALIGN, MAX_NORMAL_ALLOC, SizeClass};
 use crate::internal::current_chunk::CurrentChunk;
 
 /// Surplus of chunk strong refs the arena pre-credits to the
@@ -40,6 +40,8 @@ use crate::internal::current_chunk::CurrentChunk;
 /// chunk retire.
 const LARGE_SHARED_REF_SURPLUS: u32 = 1 << 30;
 
+#[cfg(test)]
+mod align_guard_tests;
 mod alloc_growable;
 #[cfg(feature = "hashbrown")]
 mod alloc_hashbrown;
@@ -152,6 +154,12 @@ pub struct Arena<A: Allocator + Clone = Global> {
     /// Number of completed bulk resets.
     #[cfg(feature = "stats")]
     resets: Cell<u64>,
+
+    /// Test-only override for [`CHUNK_ALIGN`] in the alignment-rejection
+    /// guards. Production builds read the constant directly; see
+    /// [`Self::chunk_align_cap`].
+    #[cfg(test)]
+    align_cap: Cell<usize>,
 }
 
 // Fields make `Arena` sendable when `A: Send + Sync`, but `CurrentChunk` and
@@ -315,6 +323,8 @@ impl<A: Allocator + Clone> Arena<A> {
             relocations_since_reset: Cell::new(0),
             #[cfg(feature = "stats")]
             resets: Cell::new(0),
+            #[cfg(test)]
+            align_cap: Cell::new(CHUNK_ALIGN),
         })
     }
 
@@ -549,6 +559,71 @@ impl<A: Allocator + Clone> Arena<A> {
     #[inline]
     pub(crate) fn max_normal_alloc(&self) -> usize {
         self.provider.config().max_normal_alloc()
+    }
+
+    /// Exclusive upper bound on the alignment a chunk can satisfy.
+    ///
+    /// Production builds always report [`CHUNK_ALIGN`]. Tests lower it
+    /// with [`Self::set_align_cap`] so the rejection guards can be driven
+    /// by types whose alignment every codegen backend accepts — a type
+    /// aligned to the real cap does not compile everywhere.
+    #[cfg(not(test))]
+    #[inline(always)]
+    #[expect(clippy::unused_self, reason = "signature must match the cfg(test) arm, which reads a field")]
+    pub(crate) fn chunk_align_cap(&self) -> usize {
+        CHUNK_ALIGN
+    }
+
+    #[cfg(test)]
+    #[inline(always)]
+    pub(crate) fn chunk_align_cap(&self) -> usize {
+        self.align_cap.get()
+    }
+
+    /// Lower the alignment caps so tests can reach them with types the
+    /// compiler will accept. `cap` replaces [`CHUNK_ALIGN`]; the
+    /// smart-pointer cap stays at half of it, as in production.
+    ///
+    /// Only the allocation guards consult this. The `buffer_freezable`
+    /// predicate that decides whether a `Vec` can freeze in place still
+    /// compares against the real constant, so a capped arena is not a
+    /// faithful model for the growable-collection paths.
+    #[cfg(test)]
+    pub(crate) fn set_align_cap(&self, cap: usize) {
+        assert!(cap.is_power_of_two(), "alignment cap must be a power of two");
+        assert!(
+            cap <= CHUNK_ALIGN,
+            "the cap may only be lowered: raising it would let the guards accept alignments no chunk can satisfy"
+        );
+        self.align_cap.set(cap);
+    }
+
+    /// Exclusive upper bound on the alignment a smart-pointer allocation
+    /// can satisfy.
+    ///
+    /// Values at or above this cap can no longer be guaranteed to lie
+    /// strictly inside the first [`CHUNK_ALIGN`] bytes of their chunk,
+    /// which would break the header-recovery mask used by `Drop` and
+    /// `deallocate`.
+    #[inline(always)]
+    pub(crate) fn smart_ptr_align_cap(&self) -> usize {
+        self.chunk_align_cap() / 2
+    }
+
+    /// Whether `align` is too large for a smart-pointer allocation.
+    #[inline(always)]
+    pub(crate) fn rejects_smart_ptr_align(&self, align: usize) -> bool {
+        align >= self.smart_ptr_align_cap()
+    }
+
+    /// Whether `align` exceeds what any single chunk can satisfy.
+    ///
+    /// A looser cap than [`Self::rejects_smart_ptr_align`]: simple-reference
+    /// allocations hand back a plain reference with no header-recovery mask,
+    /// so they can use the full chunk.
+    #[inline(always)]
+    pub(crate) fn rejects_chunk_align(&self, align: usize) -> bool {
+        align >= self.chunk_align_cap()
     }
 
     /// True iff an allocation request of `min_payload` bytes must be routed
--- a/crates/multitude/src/bytemuck.rs
+++ b/crates/multitude/src/bytemuck.rs
@@ -450,3 +450,94 @@ impl<'a, A: Allocator + Clone> BytemuckView<'a, A> {
         self.arena.try_alloc_slice_fill_with_box::<T, _>(len, |_| T::zeroed())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    //! Every `BytemuckView` entry point rejects a type aligned at or above
+    //! the relevant cap. The smart-pointer paths and the single-value
+    //! reference paths reject at the smart-pointer cap; the reference-slice
+    //! paths reject only at the looser chunk cap.
+    //!
+    //! The tests run against [`capped_arena`], whose caps are lowered so the
+    //! boundary is reachable by an alignment every codegen backend accepts.
+
+    use crate::tests_support::{ChunkOverAligned, SmartPtrOverAligned, capped_arena};
+
+    #[test]
+    fn try_alloc_box_over_aligned_returns_err() {
+        let arena = capped_arena();
+        arena.bytemuck().try_alloc_box::<SmartPtrOverAligned>().unwrap_err();
+    }
+
+    #[test]
+    fn try_alloc_arc_over_aligned_returns_err() {
+        let arena = capped_arena();
+        arena.bytemuck().try_alloc_arc::<SmartPtrOverAligned>().unwrap_err();
+    }
+
+    #[test]
+    fn try_alloc_slice_box_over_aligned_returns_err() {
+        let arena = capped_arena();
+        arena.bytemuck().try_alloc_slice_box::<SmartPtrOverAligned>(4).unwrap_err();
+    }
+
+    #[test]
+    fn try_alloc_slice_arc_over_aligned_returns_err() {
+        let arena = capped_arena();
+        arena.bytemuck().try_alloc_slice_arc::<SmartPtrOverAligned>(4).unwrap_err();
+    }
+
+    #[test]
+    #[should_panic = "arena allocation failed"]
+    fn alloc_box_panics_on_over_aligned() {
+        let arena = capped_arena();
+        let _ = arena.bytemuck().alloc_box::<SmartPtrOverAligned>();
+    }
+
+    #[test]
+    #[should_panic = "arena allocation failed"]
+    fn alloc_arc_panics_on_over_aligned() {
+        let arena = capped_arena();
+        let _ = arena.bytemuck().alloc_arc::<SmartPtrOverAligned>();
+    }
+
+    #[test]
+    #[should_panic = "arena allocation failed"]
+    fn alloc_slice_box_panics_on_over_aligned() {
+        let arena = capped_arena();
+        let _ = arena.bytemuck().alloc_slice_box::<SmartPtrOverAligned>(4);
+    }
+
+    #[test]
+    #[should_panic = "arena allocation failed"]
+    fn alloc_slice_arc_panics_on_over_aligned() {
+        let arena = capped_arena();
+        let _ = arena.bytemuck().alloc_slice_arc::<SmartPtrOverAligned>(4);
+    }
+
+    #[test]
+    fn try_alloc_over_aligned_returns_err() {
+        let arena = capped_arena();
+        arena.bytemuck().try_alloc::<SmartPtrOverAligned>().unwrap_err();
+    }
+
+    #[test]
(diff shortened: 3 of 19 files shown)


crates/multitude/src/error.rs
```diff
@@ -91,19 +91,15 @@ impl AllocError {
 
     /// Report whether the request exceeded the arena's supported alignment.
     ///
-    /// Such a request can never
-    /// succeed, regardless of how much memory is available.
+    /// Such a request can never succeed, regardless of how much memory is
+    /// available. The arena rejects any value whose alignment reaches half
```
martin-kolinek: 🤖: Non-blocking — `is_alignment_too_large` presents the smart-pointer cap as universal. **Document both boundaries: smart-pointer and single-value allocations reject at 32 KiB; simple-reference slices reject at 64 KiB.**

crates/multitude/src/zerocopy.rs
```diff
@@ -447,3 +447,86 @@ impl<'a, A: Allocator + Clone> ZerocopyView<'a, A> {
...
+    #[should_panic = "arena allocation failed"]
+    fn alloc_slice_arc_panics_on_over_aligned() {
+        let arena = capped_arena();
+        let _ = arena.zerocopy().alloc_slice_arc::<SmartPtrOverAligned>(4);
+    }
+
+    #[test]
+    fn try_alloc_ref_over_aligned_returns_err() {
+        let arena = capped_arena();
+        arena.zerocopy().try_alloc::<ChunkOverAligned>().unwrap_err();
```
martin-kolinek: 🤖: Non-blocking — these scalar zerocopy tests use the chunk-cap fixture, so they cannot detect a guard loosened from 4 KiB to 8 KiB. **Use `SmartPtrOverAligned` in `try_alloc_ref_over_aligned_returns_err` and `alloc_ref_panics_on_over_aligned`, matching the bytemuck tests.**

crates/multitude/src/tests_support.rs
```diff
@@ -15,6 +15,95 @@ use std::sync::atomic::{AtomicUsize, Ordering};
...
+/// — is a valid instance.
+macro_rules! impl_zero_valid {
+    ($ty:ty) => {
+        #[cfg(feature = "bytemuck")]
+        // SAFETY: a `u8` plus alignment padding; all-zeros is a valid instance.
+        unsafe impl bytemuck::Zeroable for $ty {}
+
+        #[cfg(feature = "zerocopy")]
+        // SAFETY: a `u8` plus alignment padding; every byte pattern is valid.
+        unsafe impl zerocopy::TryFromBytes for $ty {
```
martin-kolinek: 🤖: Non-blocking — these manual zerocopy implementations bypass the traits' derive-only seal and hidden internals. **Replace the zerocopy arms of `impl_zero_valid!` with feature-gated derives on the fixture types.** The current `u8` fixtures are bit-valid, but derives prevent this generic macro from silently asserting validity for a future niche-bearing type.

martin-kolinek (approved):
🤖: Approved.

crates/multitude/src/tests_support.rs
```diff
@@ -15,6 +15,95 @@ use std::sync::atomic::{AtomicUsize, Ordering};
...
+}
+
+/// Aligned exactly at the chunk cap: no chunk can satisfy it, so even the
+/// simple-reference paths reject it.
+#[derive(Clone, Copy, Debug)]
+#[repr(C, align(8192))]
+pub(crate) struct ChunkOverAligned(pub(crate) u8);
+
+// `repr(align)` takes a literal, so the helper types cannot name the cap
+// constants directly. Keep them in step.
```
sandersaares, line 74: [Copilot speaking]

# Replace the colloquial maintenance instruction with an explicit statement

The comment explaining the compile-time alignment assertions ends with “Keep them in step.” This idiom leaves both the antecedent of “them” and the precise maintenance requirement implicit, even though the following assertions enforce a specific relationship between the literal type alignments and the cap constants.

**Reproducible reasoning**: The English prose guideline requires professional language and explicitly asks authors to avoid colloquialisms. “Keep them in step” is an idiomatic expression rather than a direct description of the invariant. Reading the following assertions shows that the intended instruction is to keep the literal `repr(align)` values synchronized with `TEST_SMART_PTR_ALIGN` and `TEST_CHUNK_ALIGN`, so the comment can state that purpose directly.

**Consequence**: A maintainer, especially one unfamiliar with the idiom, may need to inspect and reverse-engineer the assertions to determine which values must remain synchronized when changing the test cap.

**Recommended action**: Replace the final sentence with direct wording such as: “The assertions below keep the literal type alignments synchronized with the cap constants.”

**Impacted locations**:

- `crates/multitude/src/tests_support.rs:69-70`

crates/multitude/src/bytemuck.rs
```diff
@@ -450,3 +450,94 @@ impl<'a, A: Allocator + Clone> BytemuckView<'a, A> {
         self.arena.try_alloc_slice_fill_with_box::<T, _>(len, |_| T::zeroed())
     }
 }
+
+#[cfg(test)]
+mod tests {
```
sandersaares, line 455: [Copilot speaking]

# Run the relocated bytemuck and zerocopy tests in default test builds

The relocated bytemuck and zerocopy alignment tests are nested in source modules that remain included only under their Cargo features. The Arena accessors and test fixture marker implementations they use are likewise feature-gated. Because Rust evaluates these enclosing gates before discovering the nested #[cfg(test)] modules, an ordinary cargo test build excludes the relocated coverage even though the optional crates are available as dev-dependencies.

**Reproducible reasoning**: A unit-test module can run only when every enclosing module and API it uses is compiled. crates/multitude/src/lib.rs excludes the bytemuck and zerocopy modules unless their features are active, crates/multitude/src/arena/mod.rs similarly excludes the corresponding Arena accessors, and crates/multitude/src/tests_support.rs excludes the required marker implementations. The inner #[cfg(test)] attributes therefore cannot make these tests participate in a default test build. This conflicts with the repository convention of making feature-dependent code available under cfg(any(test, feature = "...")) when dev-dependencies supply the optional crates.

**Consequence**: Regressions in the bytemuck or zerocopy alignment adapters can pass the ordinary test build; the relocated coverage runs only when callers explicitly enable the optional features or use all features.

**Recommended action**: Include the bytemuck and zerocopy module declarations and their Arena accessors under cfg(any(test, feature = "...")), while retaining feature-only doc(cfg(...)) annotations for published API documentation. Make the test-only fixture marker implementations available unconditionally within tests_support.rs, or use the matching any(test, feature = "...") conditions. Keep both optional crates available as non-optional dev-dependencies and enable zerocopy's derive feature for the dev-dependency as for the regular dependency.

**References**:

- crates/multitude/Cargo.toml feature and dev-dependency declarations

**Impacted locations**:

- `crates/multitude/src/bytemuck.rs:454-455`
- `crates/multitude/src/zerocopy.rs:451-452`
- `crates/multitude/src/tests_support.rs:83-99`
- `crates/multitude/src/lib.rs:518-529`
- `crates/multitude/tests/bytemuck_integration.rs:151-230`
- `crates/multitude/tests/zerocopy_integration.rs:124-196`
- `crates/multitude/src/bytemuck.rs:455-536`
- `crates/multitude/src/zerocopy.rs:452-533`
- `crates/multitude/src/lib.rs:518-529`
- `crates/multitude/src/arena/mod.rs:516-542`
- `crates/multitude/Cargo.toml:64-100`

crates/multitude/src/bytemuck.rs
```diff
@@ -450,3 +450,94 @@ impl<'a, A: Allocator + Clone> BytemuckView<'a, A> {
...
+    #[test]
+    fn try_alloc_arc_over_aligned_returns_err() {
+        let arena = capped_arena();
+        arena.bytemuck().try_alloc_arc::<SmartPtrOverAligned>().unwrap_err();
+    }
+
+    #[test]
+    fn try_alloc_slice_box_over_aligned_returns_err() {
+        let arena = capped_arena();
+        arena.bytemuck().try_alloc_slice_box::<SmartPtrOverAligned>(4).unwrap_err();
```
sandersaares, line 481: [Copilot speaking]

# Explain the non-empty slice length used by the adapter guard tests

The new bytemuck and zerocopy slice rejection tests repeatedly use the element count `4`, although the tested condition depends on element alignment and only needs a non-empty slice to select the initialization path. The test modules do not say whether that count is significant or arbitrary.

**Reproducible reasoning**: Each affected test changes the allocation family and aligned fixture while keeping the count fixed. The guard executes from `align_of::<T>()` before allocation size can affect the result, so no assertion depends on four elements. A reader extending the matrix must otherwise inspect the guard implementations to learn that any ordinary non-empty count is equivalent for this purpose.

**Consequence**: Future cases can preserve or change the repeated count inconsistently, and reviewers must reverse engineer whether slice length participates in the boundary being tested.

**Recommended action**: Use the simplest non-empty witness consistently, such as one element, and document once near the shared adapter matrix that slice length is intentionally minimal because only element alignment is under test. Introduce a named constant only if the length later gains semantics beyond being non-empty.

**Impacted locations**:

- `crates/multitude/src/bytemuck.rs:481-541`
- `crates/multitude/src/zerocopy.rs:470-530`

crates/multitude/src/tests_support.rs
```diff
@@ -15,6 +15,95 @@ use std::sync::atomic::{AtomicUsize, Ordering};
...
+/// The real cap is [`CHUNK_ALIGN`](crate::internal::constants::CHUNK_ALIGN),
+/// 64 KiB, and the smart-pointer cap is half of that. No test can name a
+/// type aligned that far — codegen backends differ in the maximum type
+/// alignment they accept, and 32 KiB is above the lowest of them. Lowering
+/// the caps puts both boundaries within reach of an alignment every backend
+/// compiles, so the rejection guards can be driven everywhere.
+pub(crate) const TEST_CHUNK_ALIGN: usize = 8192;
+
+/// Smart-pointer alignment cap under [`TEST_CHUNK_ALIGN`]. Half the chunk
+/// cap, as in production.
```
sandersaares, line 31: [Copilot speaking]

# Condense the new alignment-related documentation summaries

Several new item documentation blocks combine their summary and detailed rationale in one opening paragraph, so their rendered first paragraphs exceed the repository's 80-character summary limit. This affects `TEST_SMART_PTR_ALIGN`, `SmartPtrOverAlignedDrop`, `ChunkOverAligned`, the test-only `align_cap` field, and `set_align_cap`.

**Reproducible reasoning**: Rustdoc joins adjacent documentation-comment lines into one paragraph; wrapping the source therefore does not create a short rendered summary. Each affected item can state its role in one sentence, while details about production geometry, test-only thresholds, and `needs_drop` branching belong in a later paragraph. The change improves generated-documentation scanning without removing the rationale.

**Consequence**: Readers scanning item documentation receive wrapped implementation detail before a concise statement of each item's role.

**Recommended action**: Give each affected item a rendered opening sentence of at most 80 characters, add a blank documentation line, and move the remaining explanation into later paragraphs. Keep the substantive distinction between production geometry and synthetic test thresholds while shortening the summaries.

**Impacted locations**:

- `crates/multitude/src/tests_support.rs:30-31`
- `crates/multitude/src/tests_support.rs:52-53`
- `crates/multitude/src/tests_support.rs:63-64`
- `crates/multitude/src/arena/mod.rs:158-160`
- `crates/multitude/src/arena/mod.rs:583-585`

crates/multitude/src/arena/align_guard_tests.rs
```diff
@@ -0,0 +1,211 @@
...
+fn default_caps_match_the_constants() {
+    let arena = Arena::new();
+    assert_eq!(arena.chunk_align_cap(), CHUNK_ALIGN);
+    assert_eq!(arena.smart_ptr_align_cap(), max_smart_ptr_align());
+}
+
+#[test]
+#[should_panic(expected = "the cap may only be lowered")]
+fn set_align_cap_rejects_raising_the_cap() {
+    Arena::new().set_align_cap(CHUNK_ALIGN * 2);
```
sandersaares, line 35: [Copilot speaking]

# Explain why the raising-cap test preserves power-of-two alignment

The new `set_align_cap_rejects_raising_the_cap` test passes `CHUNK_ALIGN * 2`, but it does not document that this particular relationship is required to reach the intended upper-bound assertion. `set_align_cap` first rejects values that are not powers of two, so an apparently equivalent above-cap value can exercise a different validation branch.

**Reproducible reasoning**: The test expects the panic text from the `cap <= CHUNK_ALIGN` assertion. The preceding assertion requires `cap.is_power_of_two()`. Multiplying the power-of-two production cap by another power of two preserves that prerequisite while exceeding the production cap. This dependency is a non-obvious reason for the chosen literal multiplier and should be recorded where the value is selected.

**Consequence**: A maintainer can replace the expression with a simpler-looking above-cap value such as an incremented cap and unintentionally turn this into a test of the power-of-two check, weakening coverage of the lower-only constraint.

**Recommended action**: Add an inline comment explaining that the test input must remain a power of two while exceeding the production cap so execution reaches the lower-only assertion. Describe the relationship without restating the literal value.

**Impacted locations**:

- `crates/multitude/src/arena/align_guard_tests.rs:34`

crates/multitude/src/arena/mod.rs
```diff
@@ -551,6 +561,71 @@ impl<A: Allocator + Clone> Arena<A> {
...
+    /// Whether `align` exceeds what any single chunk can satisfy.
+    ///
+    /// A looser cap than [`Self::rejects_smart_ptr_align`]: simple-reference
+    /// allocations hand back a plain reference with no header-recovery mask,
+    /// so they can use the full chunk.
+    #[inline(always)]
+    pub(crate) fn rejects_chunk_align(&self, align: usize) -> bool {
+        align >= self.chunk_align_cap()
+    }
+
```
sandersaares, line 629: [Copilot speaking]

# Move the alignment-cap implementation into a focused arena submodule

The change adds the alignment-cap accessors, test override, and rejection predicates directly to `arena/mod.rs`. This review's module-layout rule reserves `mod.rs` for module documentation, declarations, metadata, and re-exports rather than implementation code. The neighboring allocation families already follow that structure through focused files such as `alloc_value.rs` and `alloc_slice_ref.rs`.

**Reproducible reasoning**: The added block at lines 562-628 defines the complete alignment-cap policy: production and test cap lookup, mutation of the test cap, derivation of the smart-pointer cap, and both rejection predicates. These functions form a cohesive implementation concern and do not need to live in the module entry point; Rust permits an inherent `impl Arena` block in a child module, as the existing `alloc_*.rs` files demonstrate. Therefore the new implementation can be placed in a focused module while `mod.rs` retains only the module declaration and the existing structural wiring.

**Consequence**: Keeping this policy in `mod.rs` further grows an already large structural entry point and makes the alignment-guard implementation less discoverable than the allocation implementations that consume it. Future changes are also more likely to continue accumulating unrelated implementation in the module root.

**Recommended action**: Add a focused module such as `arena/align_guard.rs`, move the new cap accessors, setter, and rejection predicates into an `impl Arena` block there, and leave only `mod align_guard;` in `mod.rs`. Keep any field and constructor wiring that cannot be moved without broadening this pull request's scope.

**Impacted locations**:

- `crates/multitude/src/arena/mod.rs:162`
- `crates/multitude/src/arena/mod.rs:327`
- `crates/multitude/src/arena/mod.rs:562-628`

crates/multitude/src/arena/mod.rs
```diff
@@ -551,6 +561,71 @@ impl<A: Allocator + Clone> Arena<A> {
...
     }
 
+    /// Exclusive upper bound on the alignment a chunk can satisfy.
+    ///
+    /// Production builds always report [`CHUNK_ALIGN`]. Tests lower it
+    /// with [`Self::set_align_cap`] so the rejection guards can be driven
+    /// by types whose alignment every codegen backend accepts — a type
+    /// aligned to the real cap does not compile everywhere.
+    #[cfg(not(test))]
+    #[inline(always)]
```
sandersaares, line 571: [Copilot speaking]

# Remove or justify forced inlining on the alignment-threshold helpers

The new crate-private `chunk_align_cap`, `smart_ptr_align_cap`, `rejects_smart_ptr_align`, and `rejects_chunk_align` helpers are all marked `#[inline(always)]`. The repository's performance guidance reserves inlining on non-public helpers for measured benefit and requires a specific explanation of why ordinary `#[inline]` is insufficient; the file-level statement that hot helpers must inline provides neither comparison.

**Reproducible reasoning**: The attributes force the accessor and comparison chain into every guarded caller. Constant folding on sized production paths is a separate concern that can be preserved explicitly with non-test constant-expression guards, while DST paths receive runtime `Layout` values. No cited benchmark or generated-code inspection shows that these private helpers need forced rather than ordinary inlining in either case. The current evidence therefore supports preserving the intended optimization result, but not this strongest attribute.

**Consequence**: The code establishes an unsupported forced-inlining precedent across a broad allocation surface and gives maintainers no evidence for deciding whether future changes may weaken or remove the attributes.

**Recommended action**: Restore any required compile-time production guards first, then remove these `#[inline(always)]` attributes or weaken them to the least directive form. If generated-code inspection or the existing allocation benchmarks show that forced inlining is necessary, retain it only on the demonstrated helper and document why ordinary `#[inline]` was insufficient. This trades a small amount of validation work for an evidence-based optimization contract.

**References**:

- crates/multitude/docs/PERF.md

**Impacted locations**:

- `crates/multitude/src/arena/mod.rs:570-580`
- `crates/multitude/src/arena/mod.rs:608-626`

crates/multitude/tests/arena.rs
```diff
@@ -3889,6 +3778,12 @@ mod refactor_coverage_gaps {
     // `try_reserve_local_slice`, and the non-freezable oversized arm. An
     // over-aligned element (align ≥ max_smart_ptr_align == CHUNK_ALIGN/2) is
     // non-freezable, and at 32 KiB each it forces the oversized refill.
+    //
+    // This is the last type in the crate aligned above 8192, which some
+    // codegen backends refuse to compile (see the arena's alignment-guard
+    // unit tests for why the others were removed). It compiles only because
+    // `try_reserve` never materializes the layout. If this file ever fails to
+    // build with an alignment error, this declaration is the cause.
     #[test]
```
sandersaares, line 3803: [Copilot speaking]

# Explain the remaining high-alignment reserve test accurately and locally

The new comment above `non_freezable_overaligned_vec_grows_via_oversized_path` describes removed declarations, calls this the crate's last type above a backend threshold, gives an absolute diagnosis for any future alignment build failure, and says `try_reserve` never materializes the layout. The durable local fact is narrower: the high alignment makes `Over` non-freezable and selects the oversized reserve path; the code computes its size and alignment and reserves aligned raw storage, but it never constructs an `Over` value or an over-aligned stack temporary.

**Reproducible reasoning**: `Vec::try_grow_to` uses `size_of::<T>()`, `align_of::<T>()`, and the non-freezable reservation path, so layout information is part of the operation. The test avoids `push` and leaves the vector empty, which means no typed `Over` value is created on the stack or written into the reserved storage. That distinction explains the intended portability behavior without maintaining a repository-wide census, narrating removed tests, or claiming every possible future alignment diagnostic has one cause.

**Consequence**: The current comment can become stale when other fixtures are added or moved and can send a maintainer toward allocation-layout construction even though the test deliberately preserves it and avoids only typed value materialization.

**Recommended action**: Replace the historical census and absolute troubleshooting claim with a present-state comment: the alignment intentionally makes the element non-freezable and drives oversized reservation; the path still computes and reserves aligned storage but does not construct an `Over` value or over-aligned stack temporary. Keep any backend caveat scoped to that typed-materialization constraint.

**References**:

- crates/multitude/src/vec/mod.rs (`Vec::try_grow_to`)

**Impacted locations**:

- `crates/multitude/tests/arena.rs:3777-3801`
- `crates/multitude/src/vec/mod.rs:129-183`

crates/multitude/tests/arena.rs
sandersaares, line 1: [Copilot speaking]

# Remove the now-empty Helpers section marker

The change removes the only helper type declarations beneath the three-line `Helpers` separator in `coverage_arena_gaps`, leaving the separator immediately followed by ordinary tests. The marker therefore no longer describes any content in the module.

**Reproducible reasoning**: Before this change, `HalfChunkAlign` and `ChunkAlign` appeared directly beneath the marker. The pull request deletes both declarations but retains the visual chapter heading. Reading the resulting module shows that the first item after the marker is `try_alloc_simple_ref_returns_mutable_reference`, so there is no helper section to identify. This also conflicts with the code-ordering guidance to avoid section-separator comments and instead make organization clear through code structure.

**Consequence**: Readers are directed to expect shared helper definitions and must scan onward to discover that the advertised section is empty, adding noise and making the module's organization misleading.

**Recommended action**: Delete the `Helpers` label and its surrounding separator lines. No replacement heading is needed because the remaining items are tests in the existing test module.

**Impacted locations**:

- `crates/multitude/tests/arena.rs:2236-2238`

crates/multitude/src/arena/mod.rs
```diff
@@ -551,6 +561,71 @@ impl<A: Allocator + Clone> Arena<A> {
         self.provider.config().max_normal_alloc()
     }
 
+    /// Exclusive upper bound on the alignment a chunk can satisfy.
+    ///
+    /// Production builds always report [`CHUNK_ALIGN`]. Tests lower it
+    /// with [`Self::set_align_cap`] so the rejection guards can be driven
+    /// by types whose alignment every codegen backend accepts — a type
+    /// aligned to the real cap does not compile everywhere.
```
sandersaares, line 569: [Copilot speaking]

# Define one authoritative alignment-policy model and use it consistently

The change uses the word "cap" for three different concepts whose relationships are not defined in one place: immutable production geometry (`CHUNK_ALIGN` and the first-tile masking invariant), test-only synthetic rejection thresholds lowered by `set_align_cap`, and the API-specific limits selected by each allocation family. The resulting documentation is internally inconsistent. Escape-capable `Box`, `Rc`, and `Arc` owners use the lower smart-pointer threshold; arena-bound `Alloc<[T]>` slices use the full chunk threshold; scalar `Alloc<T>` currently uses the lower threshold even though the design says arena-lifetime allocations are not subject to the smart-pointer cap; and all of these failures share `AllocError::is_alignment_too_large`. Local comments also describe synthetic thresholds as physical geometry and call `Alloc<[T]>` results plain references.

**Reproducible reasoning**: Production chunks remain aligned to `CHUNK_ALIGN`, and smart-pointer destruction recovers a chunk header by masking the payload address. The new `cfg(test)` field does not change that geometry: it only changes the values read by `rejects_smart_ptr_align` and `rejects_chunk_align`. Call-site inspection shows that smart owners and scalar `Alloc<T>` call the lower-threshold predicate, while arena-bound slice helpers call the full-chunk predicate; `buffer_freezable` remains a constant expression over the production threshold. The public error classifier therefore represents whichever limit applies to the selected API rather than one universal numeric boundary. Because no authoritative model names these layers and maps owner families to their policy, the helper rustdoc, test overview, public error docs, and design document each generalize from a different subset of the behavior. The scalar `Alloc<T>` call site additionally exposes a real unresolved contract question: its lower threshold is established behavior, but the nearby masking explanation applies to escape-capable owners and the design explicitly describes arena-lifetime allocations differently.

**Consequence**: Maintainers can use a synthetic test threshold in geometry-dependent code, apply the wrong guard to a new allocation entry point, or change the scalar `Alloc<T>` limit based on contradictory documentation. Callers can also infer a universal half-chunk limit from the shared error classifier even when an arena-bound slice API accepts alignments up to the full chunk threshold.

**Recommended action**: Create an affected-scope implementation section that defines production geometry, synthetic test thresholds, and API alignment policy as separate concepts. Include a table mapping scalar `Alloc<T>`, arena-bound `Alloc<[T]>`, and escape-capable owners to their selected threshold and rationale; explicitly decide and document the scalar `Alloc<T>` policy without presuming whether code or documentation should change. Record why the portable cases are crate unit tests and why `buffer_freezable` remains production-constant-based. Link this model from the existing design sections, then align the public error documentation and local comments with it: describe the shared classifier generically, use the established `Alloc` terminology, scope masking and chunk-capacity claims to production geometry, and describe the test override in present-state terms.

**References**:

- crates/multitude/docs/DESIGN.md, "Thin smart pointers: the alignment/masking trick", "Fragile invariants", and "Testing and verification"
- crates/multitude/docs/DESIGN.md, "Thin smart pointers: the alignment/masking trick" and "Sharp edges for users"
- crates/multitude/docs/DESIGN.md, "Thin smart pointers: the alignment/masking trick" and "The four allocation styles"
- crates/multitude/docs/DESIGN.md, "The four allocation styles"
- crates/multitude/docs/DESIGN.md, "The four allocation styles", "Thin smart pointers: the alignment/masking trick", and "Sharp edges for users"
- crates/multitude/docs/DESIGN.md, "Thin smart pointers: the alignment/masking trick"

**Impacted locations**:

- `crates/multitude/docs/DESIGN.md`
- `crates/multitude/src/tests_support.rs:21-42`
- `crates/multitude/src/arena/align_guard_tests.rs:4-28`
- `crates/multitude/src/arena/mod.rs:554-626`
- `crates/multitude/tests/arena.rs:3777-3801`
- `crates/multitude/src/error.rs:93-102`
- `crates/multitude/src/bytemuck.rs:91-129`
- `crates/multitude/src/zerocopy.rs:88-126`
- `crates/multitude/src/arena/align_guard_tests.rs:4-15`
- `crates/multitude/src/arena/align_guard_tests.rs:21-23`
- `crates/multitude/src/arena/align_guard_tests.rs:134-156`
- `crates/multitude/src/arena/alloc_value.rs:794-800`
- `crates/multitude/src/arena/mod.rs:601-616`
- `crates/multitude/src/arena/alloc_value.rs:916-919`
- `crates/multitude/src/arena/mod.rs:563-626`

crates/multitude/src/arena/align_guard_tests.rs
sandersaares, line 1: [Copilot speaking]

# Separate guard semantics from entry-point routing in the portable tests

The portable alignment tests are maintained as flat, hand-written API examples, although the change introduces two distinct verification responsibilities: proving the centralized predicates' comparison semantics and proving that every independently implemented allocation route calls the correct predicate before any fast return. The current suite tests only equality at both exclusive thresholds, omits distinct Box-copy, arena-bound clone, and DST guard implementations, no longer preserves typed-empty rejection coverage for several slice paths, and duplicates the bytemuck and zerocopy adapter matrix closely enough that the zerocopy scalar cases already use the wrong boundary fixture.

**Reproducible reasoning**: `rejects_smart_ptr_align` and `rejects_chunk_align` implement `align >= threshold`; all existing portable rejection fixtures are exactly equal to the injected thresholds, so changing either predicate to equality would preserve the suite. Call-site inspection then reveals independently breakable routes: Box copy calls `check_slice_box_layout` separately from Box fill, arena-bound clone calls `reject_over_aligned` separately from copy and fill, Box DST has its own runtime-layout guard, and Arc/Rc DST share another guard. These sites are not reached with the lowered cap. The slice guards execute before their zero-length fast returns, but relocated copy tests use non-empty inputs and the clone acceptance test uses a fixture below that call site's threshold, so those tests cannot detect a disconnected guard or reordered empty return. Finally, bytemuck and zerocopy scalar adapters both forward to `Arena::try_alloc_with`, yet only the bytemuck copy uses the lower-threshold fixture; the duplicated matrices therefore fail to encode their shared routing contract structurally. These are consequences of the same test organization rather than independent API defects.

**Consequence**: A predicate can regress from `>=` to `==`, a distinct allocation route can stop consulting the injected threshold, an empty slice can bypass validation, or one ecosystem adapter can drift to the wrong boundary while all portable tests that currently exercise the affected layer remain green.

**Recommended action**: Organize the portable coverage in two layers. First, add focused truth-table tests for both centralized predicates at equality and strictly above the injected threshold, without duplicating the full API matrix. Second, define an explicit routing matrix that maps every distinct guard implementation to its expected threshold fixture and input shape. Include typed-empty cases where validation must precede a fast return, cover Box copy and arena-bound clone separately, exercise both DST guard implementations with matching sized layouts under the `dst` feature, and generate the equivalent bytemuck and zerocopy cases from one shared mapping after correcting the zerocopy scalar fixture to `SmartPtrOverAligned`. This finding concerns semantic coverage when the relevant modules are compiled; feature-selection coverage remains a separate concern.

**References**:

- crates/multitude/src/arena/alloc_value.rs
- crates/multitude/docs/DESIGN.md, "Thin smart pointers: the alignment/masking trick" and "Safety invariants"
- crates/multitude/src/arena/align_guard_tests.rs
- crates/multitude/tests/dst.rs

**Impacted locations**:

- `crates/multitude/src/bytemuck.rs:451-543`
- `crates/multitude/src/zerocopy.rs:451-532`
- `crates/multitude/src/arena/align_guard_tests.rs:13-15`
- `crates/multitude/src/arena/align_guard_tests.rs:54-156`
- `crates/multitude/src/arena/mod.rs:609-626`
- `crates/multitude/src/arena/align_guard_tests.rs`
- `crates/multitude/src/arena/alloc_slice_box.rs:231-232`
- `crates/multitude/src/arena/alloc_slice_ref.rs:374-375`
- `crates/multitude/src/arena/alloc_unsized.rs:390-392`
- `crates/multitude/src/arena/alloc_unsized.rs:479-481`

crates/multitude/src/tests_support.rs
sandersaares: [Copilot speaking]

# Derive the bytemuck and zerocopy marker traits instead of implementing their internals

The new shared alignment fixtures manually implement `bytemuck::Zeroable`, `zerocopy::TryFromBytes`, and `zerocopy::FromZeros`. The zerocopy implementations define `TryFromBytes::is_bit_valid` and call hidden `only_derive_is_allowed_to_implement_this_trait` methods that its public documentation explicitly reserves for derive-generated code, even though this crate enables both dependencies' derive features and declares the fixtures in this module.

**Reproducible reasoning**: `crates/multitude/Cargo.toml` enables `bytemuck` and `zerocopy` with their `derive` features, and `Cargo.lock` selects zerocopy 0.8.52. `SmartPtrOverAligned` and `ChunkOverAligned` are `repr(C)` wrappers containing only a `u8`, so the supported derives can prove their marker contracts directly. Zerocopy's documented contract directs implementers to derive `TryFromBytes`, permits callers to trust a successful validity result, and warns that hidden internals can change in otherwise compatible releases in ways that affect soundness. `FromZeros` extends `TryFromBytes`, and deriving `FromZeros` emits both implementations after checking the fields. Although the handwritten `true` validity result is reasonable for these fixtures, the macro unnecessarily assumes ownership of unsafe contracts and bypasses the dependencies' supported compile-time analysis.

**Consequence**: A routine zerocopy dependency update can change hidden trait machinery or validity requirements while these handwritten implementations either stop compiling or continue with an outdated proof of bit validity, potentially making the test-only code unsound and blocking dependency updates.

**Recommended action**: Apply conditional derives to both fixture structs, using `cfg_attr(feature = "bytemuck", derive(bytemuck::Zeroable))` and `cfg_attr(feature = "zerocopy", derive(zerocopy::FromZeros))`, then remove `impl_zero_valid` and its handwritten marker implementations. This uses the dependencies' supported compile-time-checked path; deriving zerocopy's `FromZeros` also generates the required `TryFromBytes` implementation.

**References**:

- crates/multitude/Cargo.toml
- [bytemuck 1.25.0 Zeroable derive](https://docs.rs/bytemuck/1.25.0/bytemuck/derive.Zeroable.html)
- [zerocopy 0.8.26 FromZeros implementation contract](https://docs.rs/zerocopy/0.8.26/zerocopy/trait.FromZeros.html)
- [zerocopy 0.8.26 TryFromBytes implementation contract](https://docs.rs/zerocopy/0.8.26/zerocopy/trait.TryFromBytes.html)
- [Zerocopy 0.8.52 TryFromBytes implementation contract](https://docs.rs/zerocopy/0.8.52/zerocopy/trait.TryFromBytes.html#implementation)

**Impacted locations**:

- `crates/multitude/src/tests_support.rs:77-105`
- `crates/multitude/src/tests_support.rs:87-100`

sandersaares (commented):
[Copilot speaking]

Published 12 findings. One finding follows up on an existing discussion thread.

| Diagnostic | Value |
|------------|-------|
| Cache | Miss |

</details>

sandersaares:
Thanks for doing this. I was worried about our previous attempts to solve the problem, good to see a new attempt. Third time is the charm!

martin-kolinek (approved):

crates/multitude/src/tests_support.rs
```diff
@@ -15,6 +15,95 @@ use std::sync::atomic::{AtomicUsize, Ordering};
...
+#[derive(Clone, Copy, Debug)]
+#[repr(C, align(8192))]
+pub(crate) struct ChunkOverAligned(pub(crate) u8);
+
+// `repr(align)` takes a literal, so the helper types cannot name the cap
+// constants directly. Keep them in step.
+const _: () = {
+    assert!(align_of::<SmartPtrOverAligned>() == TEST_SMART_PTR_ALIGN);
+    assert!(align_of::<SmartPtrOverAlignedDrop>() == TEST_SMART_PTR_ALIGN);
+    assert!(align_of::<ChunkOverAligned>() == TEST_CHUNK_ALIGN);
```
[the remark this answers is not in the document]
AdomasBekeras: 🤖 Not needed. `align_of` is in the Rust prelude, so this compiles as written. Verified under `-D warnings` on four platforms in CI.

crates/multitude/src/arena/align_guard_tests.rs
```diff
@@ -0,0 +1,246 @@
...
+//! different chunk's header and `Drop` corrupts it. The guards make that
+//! unreachable from safe code, and these tests hold every entry point to
+//! it.
+//!
+//! The tests run against [`capped_arena`], whose caps are lowered so the
+//! boundary is reachable by an alignment every codegen backend accepts.
+
+use crate::Arena;
+use crate::internal::constants::{CHUNK_ALIGN, max_smart_ptr_align};
+use crate::tests_support::{ChunkOverAligned, SmartPtrOverAligned, SmartPtrOverAlignedDrop, TEST_CHUNK_ALIGN, capped_arena};
```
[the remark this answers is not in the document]
AdomasBekeras: 🤖 Not needed. `align_of` is in the Rust prelude, so this compiles as written. Verified under `-D warnings` on four platforms in CI.

crates/multitude/src/tests_support.rs
```diff
@@ -15,6 +15,95 @@ use std::sync::atomic::{AtomicUsize, Ordering};
...
+/// guards reject it — the two disagree only here, never in production.
+pub(crate) fn capped_arena() -> Arena {
+    let arena = Arena::new();
+    arena.set_align_cap(TEST_CHUNK_ALIGN);
+    arena
+}
+
+/// Aligned exactly at the smart-pointer cap: rejected by every
+/// smart-pointer entry point, accepted by the simple-reference paths.
+#[derive(Clone, Copy, Debug)]
```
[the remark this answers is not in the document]
AdomasBekeras: 🤖 Fixed. The doc now says "simple-reference slice paths".

crates/multitude/src/arena/mod.rs
```diff
@@ -551,6 +561,71 @@ impl<A: Allocator + Clone> Arena<A> {
...
+    /// predicate that decides whether a `Vec` can freeze in place still
+    /// compares against the real constant, so a capped arena is not a
+    /// faithful model for the growable-collection paths.
+    #[cfg(test)]
+    pub(crate) fn set_align_cap(&self, cap: usize) {
+        assert!(cap.is_power_of_two(), "alignment cap must be a power of two");
+        assert!(
+            cap <= CHUNK_ALIGN,
+            "the cap may only be lowered: raising it would let the guards accept alignments no chunk can satisfy"
+        );
```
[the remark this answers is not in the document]
AdomasBekeras: 🤖 Already addressed. `set_align_cap` asserts `cap >= 4`, and `set_align_cap_rejects_a_degenerate_cap` covers it.

crates/multitude/src/error.rs
AdomasBekeras: 🤖 Done. `is_alignment_too_large` now documents both boundaries.

crates/multitude/src/zerocopy.rs
AdomasBekeras: 🤖 Done. Both tests now use `SmartPtrOverAligned`.

crates/multitude/src/tests_support.rs
AdomasBekeras: 🤖 Done in e2046498. `impl_zero_valid!` and its three `unsafe impl`s are gone, replaced by `cfg_attr` derives of `bytemuck::Zeroable` and `zerocopy::FromZeros` on the fixtures. You were right that `FromZeros` also emits `TryFromBytes` - deriving both was a conflicting-impl error. The zerocopy dev-dependency also gained the derive feature, since AGENTS.md requires the dev-dep to carry what the feature activates.

crates/multitude/src/tests_support.rs
AdomasBekeras: 🤖 Leaving as is. The assertions directly below say precisely what must match; restating them in prose adds nothing.

crates/multitude/src/bytemuck.rs
AdomasBekeras: 🤖 Out of scope. The bytemuck and zerocopy modules were already feature-gated before this PR, so the relocated tests inherit an existing gap rather than creating one - they run under --all-features, as they did as integration tests. Changing the feature gating is a separate change with its own compatibility surface.

crates/multitude/src/bytemuck.rs
AdomasBekeras: 🤖 Leaving as is. The length is arbitrary and the guard runs before it matters, so changing it does not affect what the tests prove.

crates/multitude/src/tests_support.rs
AdomasBekeras: 🤖 Leaving as is. These are test-only items, not published API.

crates/multitude/src/arena/align_guard_tests.rs
AdomasBekeras: 🤖 Already present: "Must stay a power of two, or the earlier assert fires instead."

crates/multitude/src/arena/mod.rs
AdomasBekeras: 🤖 Leaving as is. The accessors read the `align_cap` field that `mod.rs` owns, and they are about 25 lines. A separate file would split one small policy across two.

crates/multitude/src/arena/mod.rs
AdomasBekeras: 🤖 Leaving as is. These replaced `const {}` blocks that were guaranteed to fold, and forced inlining is what keeps the release-build result unchanged. Under `cfg(not(test))` the accessors return a literal.

crates/multitude/tests/arena.rs
AdomasBekeras: 🤖 Already tightened - the comment no longer claims a repository-wide census or an absolute diagnosis.

crates/multitude/tests/arena.rs
AdomasBekeras: 🤖 Already removed.

crates/multitude/src/arena/mod.rs
AdomasBekeras: 🤖 Not doing this here. This PR is a test-mechanics change; the caps and which path uses which are unchanged from before it. The scalar `Alloc<T>` threshold question you raise is real, but it predates this PR and deserves its own discussion rather than being settled in a documentation section written alongside a test refactor.

crates/multitude/src/arena/align_guard_tests.rs
AdomasBekeras: 🤖 Investigated with mutation testing rather than by inspection. Ran `cargo mutants` over the five guard files with `dst,bytemuck,zerocopy` enabled: the `check_slice_box_layout -> Ok(())` and `reject_over_aligned -> Ok(())` mutants - exactly the "this route stopped consulting the guard" scenario for Box copy and arena-bound clone - are both caught, as are `rejects_smart_ptr_align` and `rejects_chunk_align` returning a constant and their `>=` flipped to `<`. The `>=` versus `==` case is covered by `guards_reject_alignments_above_the_cap`, which lowers the cap so the fixtures land strictly above it. The zerocopy scalar fixture is fixed. The two DST guards in `alloc_unsized.rs` are genuinely untested, but they were untested before this PR as well - nothing in tests/dst.rs asserts ALIGNMENT_TOO_LARGE - and the diff there is a two-line swap from a local constant to the accessor. That gap is pre-existing and out of scope.

martin-kolinek (approved):

wukchung (approved):

PR #706, opened 2026-08-27
perf(ci): parallelize scheduled Miri and reduce resource outliers

st-dev-gh:
## Problem

Scheduled runtime analysis runs four full-workspace Miri profiles:

1. Plain Miri
2. Tree Borrows
3. Strict provenance
4. Race coverage

The profiles intentionally remain sequential, but `cargo miri test` also ran every test artifact sequentially within each profile. This left most runner CPUs idle and made a few extreme artifacts dominate runtime and memory consumption.

This builds on #674, which optimizes the PR stacked-borrows job. That PR deliberately leaves the scheduled full-workspace profiles unchanged; this PR targets those scheduled runs.

## Findings

A local Tree Borrows profile compiled the workspace once and measured all 226 test artifacts with four concurrent workers.

Most artifacts were small:

- Median duration: 3 seconds
- P95 duration: 2.37 minutes
- Median peak working set: 129.5 MiB
- P95 peak working set: 494.9 MiB

A few artifacts dominated resource use:

| Artifact | Duration | Peak working set | Finding |
|---|---:|---:|---|
| `routerama` | over 9 observed CPU-hours | 645 MiB | A 4,096-segment stack-safety test made Miri interpret thousands of trie nodes |
| `multitude::arena` | 27.5 minutes | 13.49 GiB | Large Drop slices caused extreme Miri metadata and destructor-processing costs |
| `bytesbuf` | 1.6 minutes | 4.27 GiB | Significant memory consumer, but not the largest outlier |
| `arena_vec` | 1.1 minutes | 1.32 GiB | Secondary memory outlier |

Arena profiling isolated the largest consumers:

- `u16::MAX + 1` Drop-slice tests reached over 20 GiB when run separately.
- Related overflow tests reached approximately 14 GiB.
- Oversized 3,000-element Drop-slice tests reached approximately 8 GiB.
- The zero-test harness baseline was only 193 MiB.

Runtime and memory were correlated across ordinary artifacts, but the extreme outliers were different problems: Routerama was CPU-bound while Arena was memory-bound.

### Compile once and execute artifacts concurrently

The shared Miri recipe now:

- Compiles all selected packages together with `--no-run`.
- Excludes `*_macros_impl` code-generation crates; emitted code remains covered through consuming crates.
- Parses Cargo's JSON `compiler-artifact` output.
- Runs independent artifacts concurrently using `cargo-miri runner`.
- Defaults to one worker per logical processor.
- Supports an explicit `ANVIL_MIRI_JOBS` override.
- Preserves workspace feature unification and each package's working directory.
- Groups artifact output and reports all failed artifacts.
- Starts historically long artifacts first to reduce the tail.

Compiling the selected workspace together is important: invoking Miri separately per package changes Cargo feature unification and previously broke packages relying on workspace-unified features.

The four Miri profiles themselves remain sequential.

### Add resource telemetry

Scheduled logs now report:

- Logical processor count
- Selected artifact concurrency
- GC-reported memory load and estimated headroom
- Memory telemetry when each artifact starts and completes

The telemetry is informational and does not silently reduce concurrency.

### Bound Routerama's deep-trie workload

`failed_build_discards_deep_source_trie_iteratively` retains 4,096 segments natively but uses 512 under Miri.

This preserves the full native stack-safety stress test while still exercising deep trie construction, conflict traversal, and iterative destruction under Miri.

Measured Tree Borrows runtime fell from the multi-hour outlier to approximately 5.7 minutes.

### Reduce Arena's Miri memory pressure

Tests whose only purpose is validating runtime length boundaries above `u16::MAX` are ignored under Miri while retaining their full native coverage.

Oversized Drop-slice tests still exercise the same allocation and destructor paths, but under Miri they use:

- 513 elements instead of 3,000
- A 4 KiB normal-allocation threshold to ensure the oversized path remains selected

The complete Arena artifact now measures:

| Metric | Before | After |
|---|---:|---:|
| Peak working set | 13.49 GiB | 5.16 GiB |
| Duration | 27.5 minutes | 10.0 minutes |
| Miri-enabled tests | 445 passed | 435 passed |
| Intentionally ignored tests | 3 | 13 |

The excluded cases remain covered at their original sizes by native tests.

### Exclude an unsupported strict-provenance dependency path

On Linux and Linux ARM, strict-provenance Miri rejected an integer-to-pointer conversion inside `parking_lot_core` while running:

`internity::freeze_races_writer_and_stays_prefix_consistent`

This is a platform-specific dependency implementation rather than an Internity provenance violation. The test is now ignored only for strict-provenance Miri and remains covered by:

- Native tests
- Plain Miri
- Tree Borrows
- Race coverage

## Results

The latest Windows scheduled runtime-analysis job completed successfully.

Linux and Linux ARM also completed the optimized Arena and Routerama artifacts without runner loss or memory termination. They subsequently exposed the separate `parking_lot_core` strict-provenance incompatibility addressed by this PR.

## Coverage trade-offs

- Native stress-test sizes are unchanged.
- All selected runtime workspace packages are still compiled together.
- `*_macros_impl` crates are excluded because their tests parse and compare generated tokens rather than execute runtime code; emitted code remains covered by consuming crates.
- Scheduled profiles continue to cover the full workspace.
- Only repetitive runtime-boundary breadth with no additional memory-safety coverage is excluded from Miri.
- The strict-provenance exclusion applies only to the dependency path that Miri cannot interpret.
- The four Miri profiles retain their existing order and configuration.

### DIFF
--- a/crates/multitude_macros_impl/src/lib.rs
+++ b/crates/multitude_macros_impl/src/lib.rs
@@ -1523,64 +1523,14 @@ fn enum_tokens(
 
 #[cfg(test)]
 mod tests {
-    use std::ops::Deref;
-
     use super::*;
 
-    struct Generated(String);
-
-    impl Generated {
-        fn new(value: &str) -> Self {
-            let mut compact = String::with_capacity(value.len());
-            let mut in_string = false;
-            let mut escaped = false;
-            for ch in value.chars() {
-                if in_string {
-                    compact.push(ch);
-                    if escaped {
-                        escaped = false;
-                    } else if ch == '\\' {
-                        escaped = true;
-                    } else if ch == '"' {
-                        in_string = false;
-                    }
-                } else if ch == '"' {
-                    in_string = true;
-                    compact.push(ch);
-                } else if !ch.is_whitespace() {
-                    compact.push(ch);
-                }
-            }
-            Self(compact)
-        }
-
-        fn contains(&self, expected: &str) -> bool {
-            self.0.contains(expected) || self.0.contains(&Self::new(expected).0)
-        }
-
-        fn match_count(&self, expected: &str) -> usize {
-            self.0.matches(&Self::new(expected).0).count()
-        }
-    }
-
-    impl Deref for Generated {
-        type Target = str;
-
-        fn deref(&self) -> &Self::Target {
-            &self.0
-        }
-    }
-
-    fn expand(source: &str) -> Generated {
+    fn expand(source: &str) -> String {
         let input: DeriveInput = syn::parse_str(source).unwrap();
         let root: Path = parse_quote!(::multitude::de);
         let generated = super::expand(&input, &root).unwrap();
-
-        #[cfg(not(miri))]
-        syn::parse2::<syn::File>(generated.clone()).unwrap();
-        let rendered = generated.to_string();
-
-        Generated::new(&rendered)
+        let file: syn::File = syn::parse2(generated).unwrap();
+        prettyplease::unparse(&file)
     }
 
     #[test]
@@ -1651,7 +1601,7 @@ mod tests {
         assert!(enumeration.contains("variant index 0 <= i < 2"));
         assert!(enumeration.contains("Variant2"));
         assert!(!enumeration.contains("Variant1,"));
-        assert!(compact_enumeration.contains("1u64=>::core::result::Result::Ok(__MultitudeVariantForE::Variant2)"));
+        assert!(compact_enumeration.contains("1u64=>{::core::result::Result::Ok(__MultitudeVariantForE::Variant2)"));
         assert!(!compact_enumeration.contains("2u64=>"));
     }
 
@@ -1816,7 +1766,7 @@ mod tests {
         assert!(!variant.contains("T: ::multitude::de::DeserializeIn"));
 
         let compact: String = variant.split_whitespace().collect();
-        assert_eq!(variant.match_count("let _: () = __value;"), 3);
+        assert_eq!(compact.matches("let_:()=__value;").count(), 3);
         let single_named_constructor = ["E::Single", "{", "value:__value", "}"].concat();
         assert!(compact.contains(&single_named_constructor));
         for constructor in [
@@ -1946,7 +1896,7 @@ mod tests {
         let shapes = expand("enum Shapes { New(u8), OneSkipped(#[serde(skip)] u8), Unit, Named { value: u8 } }");
         let compact: String = shapes.split_whitespace().collect();
         assert!(compact.contains("__MultitudeVariantForShapes::Variant0=>{let__value=__Serde::de::VariantAccess::newtype_variant_seed"));
-        assert!(compact.contains("__MultitudeVariantForShapes::Variant1=>__Serde::de::VariantAccess::tuple_variant(__access,0usize,"));
+        assert!(compact.contains("__MultitudeVariantForShapes::Variant1=>{__Serde::de::VariantAccess::tuple_variant(__access,0usize,"));
         assert!(shapes.contains("tuple_variant"));
         assert!(shapes.contains("struct_variant"));
         assert!(shapes.contains("unit_variant"));
--- a/crates/multitude_macros_impl/src/lib.rs
+++ b/crates/multitude_macros_impl/src/lib.rs
@@ -1523,14 +1523,64 @@ fn enum_tokens(
 
 #[cfg(test)]
 mod tests {
+    use std::ops::Deref;
+
     use super::*;
 
-    fn expand(source: &str) -> String {
+    struct Generated(String);
+
+    impl Generated {
+        fn new(value: &str) -> Self {
+            let mut compact = String::with_capacity(value.len());
+            let mut in_string = false;
+            let mut escaped = false;
+            for ch in value.chars() {
+                if in_string {
+                    compact.push(ch);
+                    if escaped {
+                        escaped = false;
+                    } else if ch == '\\' {
+                        escaped = true;
+                    } else if ch == '"' {
+                        in_string = false;
+                    }
+                } else if ch == '"' {
+                    in_string = true;
+                    compact.push(ch);
+                } else if !ch.is_whitespace() {
+                    compact.push(ch);
+                }
+            }
+            Self(compact)
+        }
+
+        fn contains(&self, expected: &str) -> bool {
+            self.0.contains(expected) || self.0.contains(&Self::new(expected).0)
+        }
+
+        fn match_count(&self, expected: &str) -> usize {
+            self.0.matches(&Self::new(expected).0).count()
+        }
+    }
+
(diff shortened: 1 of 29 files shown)


crates/multitude_macros_impl/src/lib.rs
```diff
@@ -1523,14 +1523,64 @@ fn enum_tokens(
 
 #[cfg(test)]
 mod tests {
+    use std::ops::Deref;
+
     use super::*;
 
-    fn expand(source: &str) -> String {
+    struct Generated(String);
```
Vaiz: from PR description

> Avoid formatting-only macro work under Miri
>
> multitude_macros_impl no longer runs generated syntax through prettyplease under Miri. > Tests compare normalized token streams instead.
>
> Non-Miri tests still parse the generated output as a syn::File, preserving syntax validation.

please, roll this back. The correct solution is to exclude `macros_impl` from miri check completely

justfiles/anvil/checks/miri-race-coverage.just
Vaiz, line 1: I though those changes must be done in cargo-anvil itself

Vaiz: > Managed by cargo-anvil. Update the corresponding template in the cargo-anvil
> crate unless the change is repository-specific.
> Update behaviour: https://github.com/microsoft/ox-tools/blob/main/crates/cargo-anvil/docs/design/updates.md

Vaiz (changes requested):

justfiles/anvil/checks/miri-race-coverage.just
st-dev-gh: Local repo can be customized, but you are right that the changes should eventually go to cargo-anvil itself. That will be done separately.

crates/multitude_macros_impl/src/lib.rs
st-dev-gh: Rolled back the `multitude_macros_impl` token-normalization workaround and restored its tests and `prettyplease` dev dependency to match `main`.

Scheduled Miri now excludes all eight workspace `*_macros_impl` packages. Their generated unsafe code is still interpreted through consuming crates, while the implementation crates' own token parsing/comparison tests are no longer selected as Miri artifacts. I also verified the compile manifest contains no runnable artifacts for the excluded packages; the restored `multitude_macros_impl` suite passes all 30 tests.

Vaiz (approved):

justfiles/anvil/checks/miri.just
```diff
@@ -10,6 +10,186 @@
...
+    $packageArgs += @(
+        '--exclude', 'data_privacy_macros_impl',
+        '--exclude', 'fundle_macros_impl',
+        '--exclude', 'internity_macros_impl',
+        '--exclude', 'multitude_macros_impl',
+        '--exclude', 'observed_macros_impl',
+        '--exclude', 'ohno_macros_impl',
+        '--exclude', 'templated_uri_macros_impl',
+        '--exclude', 'thread_aware_macros_impl'
+    )
```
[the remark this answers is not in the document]
st-dev-gh: [copilot speaking] Miri exclusions are now declared by each package through `[package.metadata.anvil.miri] exclude = true`. Workspace runs translate those declarations into `--exclude` arguments, while impact-scoped runs remove the complete `--package name@version` pairs. If every affected package is excluded, the recipe exits successfully without invoking Cargo.

justfiles/anvil/checks/miri.just
```diff
@@ -10,6 +10,186 @@
...
+        $seenArtifacts = [System.Collections.Generic.HashSet[string]]::new([System.StringComparer]::Ordinal)
+        foreach ($line in Get-Content -LiteralPath $artifactManifest) {
+            if (-not $line.TrimStart().StartsWith('{')) {
+                [Console]::Error.WriteLine($line)
+                continue
+            }
+            $message = $line | ConvertFrom-Json -ErrorAction Stop
+            if ($message.reason -eq 'compiler-message' -and $message.message.rendered) {
+                [Console]::Error.Write($message.message.rendered)
+            } elseif ($message.reason -eq 'compiler-artifact' -and $message.executable) {
```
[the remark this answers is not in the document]
st-dev-gh: [copilot speaking] The artifact collector now requires `profile.test == true`, so only artifacts Cargo compiled with the Rust test harness are scheduled through `cargo-miri runner`. This also excludes a real non-test `rallocator_cli` executable present in the captured workspace manifest.

crates/routerama/src/dyn_builder.rs
```diff
@@ -148,13 +148,18 @@ mod tests {
 
     use super::*;
 
+    #[cfg(miri)]
+    const DEEP_TRIE_SEGMENTS: usize = 512;
+    #[cfg(not(miri))]
+    const DEEP_TRIE_SEGMENTS: usize = 4_096;
```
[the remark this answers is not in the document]
st-dev-gh: [copilot speaking] This test module predates the PR; the change only parameterizes the existing stress-test size under Miri. Keeping this regression test included in coverage is useful, so I am leaving the module unchanged rather than introducing an unrelated coverage exclusion.

crates/multitude_macros_impl/src/lib.rs
```diff
@@ -1523,14 +1523,64 @@ fn enum_tokens(
 
 #[cfg(test)]
 mod tests {
+    use std::ops::Deref;
```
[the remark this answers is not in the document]
st-dev-gh: Exclusion from coverage is not subject of this PR.

justfiles/anvil/checks/miri.just
```diff
@@ -10,6 +10,208 @@
...
+#
+# ANVIL_MIRI_JOBS overrides the automatically selected process count. The
+# default uses one process per logical processor. GC memory information is
+# logged as telemetry but does not limit concurrency, allowing scheduled runs
+# to reveal whether memory pressure makes the CPU-based setting unsuitable for
+# a runner class.
+
+# Run selected Miri test artifacts with CPU-based parallelism.
+[script("pwsh")]
+_anvil-miri-test:
```
martin-kolinek, line 26: I guess this is intentional, but I really would like to already see an accompanying PR into cargo-anvil (or at least a cargo tool that does this).

martin-kolinek (approved):

wukchung (approved):

justfiles/anvil/checks/miri.just
st-dev-gh: The PR to cargo-anvil is already in preparation.

justfiles/anvil/checks/miri-race-coverage.just
st-dev-gh: microsoft/ox-tools#123

justfiles/anvil/checks/miri.just
st-dev-gh: microsoft/ox-tools#123

PR #709, opened 2026-08-28
fix(ci): bump cargo-semver-checks to 0.50.0 for rustdoc JSON v60

AdomasBekeras:
## Problem

`anvil-semver-check` fails for every library crate on every PR:

```
error: unsupported rustdoc format v60 for file: .../target/doc/anyspawn_azure.json
(supported formats are v55, v56, v57)
```

`justfiles/anvil/versions.just` pinned `cargo-semver-checks` 0.46.0, which reads rustdoc JSON v55/v56/v57. The recipe does not pin a stable toolchain, so it uses the runner default — now Rust 1.98.0, whose rustdoc emits **v60**. The tool aborts while parsing the JSON it just produced.

The recipe is advisory, so the crash was not surfaced as a failure. Instead each tool error was collected as a "finding" and posted as a sticky **:warning: Potential breaking changes detected** comment listing ~21 crates. Net effect: SemVer checking has been off repo-wide, and every PR carried a misleading breaking-change warning. Example: #705.

## Fix

Bump the pin to 0.50.0, whose `trustfall_rustdoc` features are `v57`, `v60`, `v61`.

`constants.env` (used by the `main.yml` `semver` job and `scripts/release-packages.ps1`) is bumped to the same version. That job pins `RUST_LATEST=1.96.1`, which emits v57, so it was not broken — but 0.48.0 reads only v56/v57, so a developer running `release-packages.ps1` on current stable would hit the identical error locally.

## Compatibility

Measured `format_version` per toolchain:

| toolchain | rustdoc JSON |
| --- | --- |
| 1.93.1 (MSRV) | v57 |
| 1.95.0 | v57 |
| 1.96.1 (`RUST_LATEST`) | v57 |
| 1.97.0 | v57 |
| 1.98.0 (runner stable) | **v60** |
| nightly-2026-05-30 | v57 |

0.50.0 covers v57 and v60, so every toolchain in play works. It drops v56, which corresponds to rustc below our MSRV. Install MSRV of 0.50.0 is Rust 1.93, which the repo already meets.

Verified locally with 0.50.0 under 1.98.0:

```
$ cargo semver-checks --package allocation_hints --baseline-rev origin/main
     Checked [   0.021s] 196 checks: 196 pass, 58 skip
     Summary no semver update required
```

The summary wording is unchanged, so `ConvertFrom-SemverChecksOutput` in `scripts/lib/releasing.ps1` still parses it. `just anvil-semver-check-validate-prereqs` passes.

## Expected side effects

Versions 0.47 through 0.50 added new lints, which has two consequences.

First, the advisory comment will now report breaking changes that 0.46.0 was never able to detect, so it may look busier than it used to. That is the point of the fix, but the first few pull requests after this merges will probably carry longer comments than usual.

Second, `release-packages.ps1` derives each crate's change type from the same tool, so a crate whose release would previously have been classified as a patch may now be classified as breaking. Both outcomes are the tool working correctly rather than a regression, but they are new behaviour and worth expecting rather than being surprised by.

## Follow-ups (not in this PR)

- The pin will drift again: the recipe tracks whatever stable the runner ships, while the tool supports a discrete set of format versions. Either pin a toolchain for this recipe or expect a periodic bump.
- `anvil-semver-check` renders a tool crash identically to a real finding. A parse or build failure should be reported as an operational problem, not as "potential breaking changes".
- `justfiles/anvil/versions.just` is anvil-managed; the same pin should be bumped in the `cargo-anvil` template so other adopting repos are not left broken.


justfiles/anvil/versions.just
```diff
@@ -66,7 +66,7 @@ cargo_heather_version := "0.2.1"
 cargo_llvm_cov_version := "0.8.4"
 cargo_mutants_version := "26.1.2"
 cargo_nextest_version := "0.9.122"
-cargo_semver_checks_version := "0.46.0"
+cargo_semver_checks_version := "0.50.0"
```
Vaiz, line 69: it should be done in cargo-anvil

> Managed by cargo-anvil. Update the corresponding template in the cargo-anvil

Vaiz (approved):

justfiles/anvil/versions.just
AdomasBekeras: ugh

PR #712, opened 2026-08-28
fix(observed): correct enrichment and test-harness behavior

Vaiz:

## Problem

Behavioral and ergonomics defects in `observed` and `observed_testing`, each confirmed against `main`. Unlike the sibling documentation pull requests, this one changes code.

## What changes

**Runtime**

- `Slot::push`, `Sink::push_enrichment` and `EnrichmentTransfer` installed an `EnrichmentNode` and touched every captured slot even when the enrichment layer was empty. An empty layer now returns a no-op guard and leaves the slot untouched. `push_for` takes the same early return, so an empty targeted layer allocates nothing either. Guard merge and drop semantics are unchanged; new tests cover the empty push, the merged-guard case, and the transfer path.
- The early return lives in `push` and `push_for`, where the layer is still owned. The private `EnrichmentTransfer::push_entries` briefly repeated the check, but by then both callers had already returned, so that half of its guard could never be true; the mutation gate caught it as a surviving `||`/`&&` mutant and it is removed.
- Each guard now records which cost it avoids. The transfer guards run while the entries are still an owned `Vec`, so they also skip the shared entry slice. The `Sink` and `Slot` guards run after `.enrich(...)` has already built that slice, so they avoid composite fan-out, thread-local initialization and slot writes, but not the slice itself.

**Test harness**

- `ExpectedEvent` stored the log expectation twice: once in `severity: Option<Severity>` and once in a separate `expect_log` flag. A captured event carries a severity exactly when it carries a log signal, so the flag added no valid state, and `without_severity().log()` was an expectation no event could satisfy. `expect_log` and the `log()` builder are removed, the log expectation now follows the severity, and the `.log()` call sites go with them. A fieldless-log test pins the behavior.
- The type claimed partial matching while `PartialEq` compares body, dimensions, metric and disabled exactly. The documentation now says so, and scopes the claim: source location and metric instrument metadata are outside the comparison, and are reached through the `CapturedEvent` accessors. `observed_testing` is not published, so this reaches no external consumer.

**Examples and benchmarks**

- `support/otel.rs` exported the emitting crate as `code.namespace`, an attribute OpenTelemetry has deprecated. The mapping is removed, the test that pinned it now asserts absence, and a comment beside the mapper records why file and line stay while the crate does not.
- `layered_app/token_issuer.rs` hand-built `DataClass::new("microsoft", "PublicNonPersonalData")` instead of using the shared taxonomy constant, and recorded validation failures as an up-down counter carrying a constant `failure_count: 1`. A failure count only increases, so it is now a fieldless counter.
- The `emit_context` benchmark group named operations its bodies do not perform: `attach_emitter` constructs and drops a `Sink`, and `emit_event_no_emitter` emits through `Sink::noop()`. The group, functions, identifiers and the module overview are renamed to match. This changes benchmark identifiers, so historical series under the old names do not carry over.

## Withdrawn after review

Two changes were reverted in `b9639949`:

- **Default no-op `EventProcessor::flush`.** A default body lets a processor that buffers telemetry silently skip flushing it, with no error and no compile-time signal. The repeated `Ok(())` in stateless processors is the cheaper cost. `flush` stays a required method, and no processor code is touched by this pull request.
- **Documenting `EnrichmentEntry`.** The type is public only because Rust cannot hide a type the derive expansion must reference. It is not a stable construction API, and `#[derive(Enrichment)]` remains the single supported route, so it is back to `#[doc(hidden)]`.

## Effects

- An empty enrichment layer installs no chain node and writes no slot. Pushed through a captured `Transfer` it also costs no entry-slice allocation; `.enrich(...)` and `.enrich_for(...)` still build their `Arc<[EnrichmentEntry]>` before the guard is reached.
- `ExpectedEvent::new` matches its documentation and `ExpectedEvent::log` no longer exists; assertions that relied on the previous default change meaning, and call sites that used `.log()` no longer compile.
- Exported telemetry no longer carries the deprecated `code.namespace` attribute.
- Benchmark identifiers in the sink-operations group change; comparisons against earlier runs need remapping.
- No published crate changes its API. The `observed` public surface, including `EventProcessor` and the `enrichment` re-exports, is as it was on `main`. The removed `ExpectedEvent::log` is a behavioral and signature change to the unpublished `observed_testing` harness contract.

## Validation

`just anvil-fmt`, `just anvil-clippy`, `just anvil-readme-check`, `just anvil-spellcheck`, `cargo +1.96.1 check --workspace --all-targets`, `cargo +1.96.1 test -p observed -p observed_testing -p fetch_winhttp_impl`, and a `cargo mutants` run scoped to `crates/observed/src/enrichment/slot.rs` all pass. The branch is rebased on `main`.

### DIFF
--- a/crates/observed/benches/observed_benchmarks.rs
+++ b/crates/observed/benches/observed_benchmarks.rs
@@ -277,11 +277,11 @@ fn entrypoint(c: &mut Criterion) {
         group.finish();
     }
 
-    // --- Context operation benchmarks ---
+    // --- Sink operation benchmarks ---
     {
-        let mut group = c.benchmark_group("emit_context");
-        bench_attach_emitter(&mut group, &allocs, &time);
-        bench_emit_event_direct(&mut group, &allocs, &time);
+        let mut group = c.benchmark_group("sink_operations");
+        bench_construct_processor_free_sink(&mut group, &allocs, &time);
+        bench_emit_to_noop_sink(&mut group, &allocs, &time);
         group.finish();
     }
 
@@ -468,11 +468,15 @@ fn bench_enrich_push_pop(group: &mut BenchmarkGroup<'_, WallTime>, allocs: &allo
 }
 
 // ---------------------------------------------------------------------------
-// Context operation benchmarks
+// Sink operation benchmarks
 // ---------------------------------------------------------------------------
 
-fn bench_attach_emitter(group: &mut BenchmarkGroup<'_, WallTime>, allocs: &alloc_tracker::Session, time: &all_the_time::Session) {
-    const ID: &str = "attach_emitter";
+fn bench_construct_processor_free_sink(
+    group: &mut BenchmarkGroup<'_, WallTime>,
+    allocs: &alloc_tracker::Session,
+    time: &all_the_time::Session,
+) {
+    const ID: &str = "construct_processor_free_sink";
     const EMPTY_PROCESSORS: Vec<Arc<dyn observed::processing::EventProcessor>> = Vec::new();
 
     bench_with_tracking(group, allocs, time, ID, || {
@@ -481,11 +485,10 @@ fn bench_attach_emitter(group: &mut BenchmarkGroup<'_, WallTime>, allocs: &alloc
     });
 }
 
-fn bench_emit_event_direct(group: &mut BenchmarkGroup<'_, WallTime>, allocs: &alloc_tracker::Session, time: &all_the_time::Session) {
-    const ID: &str = "emit_event_no_emitter";
+fn bench_emit_to_noop_sink(group: &mut BenchmarkGroup<'_, WallTime>, allocs: &alloc_tracker::Session, time: &all_the_time::Session) {
+    const ID: &str = "emit_to_noop_sink";
 
-    // Benchmark the emit path when no sink is registered - measures the
-    // cost of the context lookup + early return.
+    // Benchmark the emit path through an explicit no-op sink.
     let noop = Sink::noop();
     bench_with_tracking(group, allocs, time, ID, || {
         observed::emit!(&noop, SimpleLogEvent { status: 200, retries: 0 });
--- a/crates/observed/benches/observed_benchmarks.rs
+++ b/crates/observed/benches/observed_benchmarks.rs
@@ -7,7 +7,7 @@
 //! - Full emit pipeline (event -> log record via `OTel` provider)
 //! - Enrichment resolution (context lookup + Vec building)
 //! - Metric dimension building
-//! - Context operations (enrich)
+//! - Sink operations (construction, emit through a no-op sink)
 //!
 //! Run with:
 //! ```sh
--- a/crates/observed/examples/support/otel.rs
+++ b/crates/observed/examples/support/otel.rs
@@ -130,6 +130,11 @@ pub(crate) fn populate_log_record(record: &mut impl LogRecord, event: &EventView
         ControlFlow::Continue(())
     });
 
+    // File and line use stable OpenTelemetry code attributes. The emitting
+    // crate is deliberately not exported: `code.namespace` is deprecated with
+    // no standalone replacement, and its successor `code.function.name` needs a
+    // fully qualified function name that the event model does not capture.
+    // https://opentelemetry.io/docs/specs/semconv/registry/attributes/code/
     if let Some(file) = event.source_file() {
         record.add_attribute(opentelemetry::Key::from_static_str(CODE_FILE_PATH), AnyValue::String(file.into()));
     }
--- a/crates/observed/examples/support/otel.rs
+++ b/crates/observed/examples/support/otel.rs
@@ -31,9 +31,6 @@ const CODE_FILE_PATH: &str = "code.file.path";
 /// `OTel` attribute key for the source line a call site came from.
 const CODE_LINE_NUMBER: &str = "code.line.number";
 
-/// `OTel` attribute key for the crate a call site came from.
-const CODE_NAMESPACE: &str = "code.namespace";
-
 /// Converts a [`Text`] into an `OTel` string, preserving the borrowed-versus-
 /// shared distinction so neither representation copies.
 ///
@@ -142,12 +139,5 @@ pub(crate) fn populate_log_record(record: &mut impl LogRecord, event: &EventView
             AnyValue::Int(i64::from(line)),
         );
     }
-    if let Some(crate_name) = event.source_crate() {
-        record.add_attribute(
-            opentelemetry::Key::from_static_str(CODE_NAMESPACE),
-            AnyValue::String(crate_name.into()),
-        );
-    }
-
     true
 }
--- a/crates/observed/src/enrichment/enrichment_trait.rs
+++ b/crates/observed/src/enrichment/enrichment_trait.rs
@@ -16,22 +16,6 @@ use crate::enrichment::EnrichmentEntry;
 ///
 /// See the [`Enrichment` derive macro](crate::Enrichment) for field attributes
 /// and usage examples.
-///
-/// # Manual implementation
-///
-/// ```
-/// use observed::enrichment::{Enrichment, EnrichmentEntry};
-///
-/// struct RequestContext {
-///     request_id: &'static str,
-/// }
-///
-/// impl Enrichment for RequestContext {
-///     fn into_entries(self) -> Vec<EnrichmentEntry> {
-///         vec![EnrichmentEntry::unclassified("request.id", self.request_id)]
-///     }
-/// }
-/// ```
 pub trait Enrichment {
     /// Converts this enrichment struct into its [`EnrichmentEntry`] items.
     ///
--- a/crates/observed/src/enrichment/enrichment_trait.rs
+++ b/crates/observed/src/enrichment/enrichment_trait.rs
@@ -16,6 +16,22 @@ use crate::enrichment::EnrichmentEntry;
 ///
 /// See the [`Enrichment` derive macro](crate::Enrichment) for field attributes
 /// and usage examples.
+///
+/// # Manual implementation
+///
+/// ```
+/// use observed::enrichment::{Enrichment, EnrichmentEntry};
+///
+/// struct RequestContext {
+///     request_id: &'static str,
+/// }
+///
+/// impl Enrichment for RequestContext {
+///     fn into_entries(self) -> Vec<EnrichmentEntry> {
+///         vec![EnrichmentEntry::unclassified("request.id", self.request_id)]
+///     }
+/// }
+/// ```
 pub trait Enrichment {
     /// Converts this enrichment struct into its [`EnrichmentEntry`] items.
     ///
--- a/crates/observed/src/enrichment/slot.rs
+++ b/crates/observed/src/enrichment/slot.rs
@@ -69,6 +69,10 @@ impl Slot {
 
     /// Pushes entries onto the enrichment chain and returns a guard.
     pub(crate) fn push(&self, entries: Arc<[EnrichmentEntry]>) -> Guard {
+        if entries.is_empty() {
+            return Guard::empty();
+        }
+
         let prev = {
             let cell = self.0.get_or(|| RefCell::new(None));
             let next = Arc::new(EnrichmentNode {
@@ -100,6 +104,10 @@ pub(crate) struct Guard {
 }
 
 impl Guard {
+    pub(crate) fn empty() -> Self {
+        Self { slots: SmallVec::new() }
+    }
+
     /// Flattens several guards into one. Each input guard is consumed and its
     /// `Drop` is disarmed (`mem::take` empties its slots), so the merged guard
     /// owns the restoration responsibility.
@@ -171,25 +179,29 @@ impl EnrichmentTransfer {
     /// Broadcast within the transfer's known slots; transfers with no
     /// captured slots are left unchanged.
     pub(crate) fn push(&mut self, additional_enrichment: impl Enrichment) {
-        self.push_entries(&Arc::from(additional_enrichment.into_entries()));
+        let entries = additional_enrichment.into_entries();
+        if entries.is_empty() {
+            return;
+        }
+
+        self.push_entries(&Arc::from(entries));
     }
 
     /// Same as [`push`](Self::push), but marks every entry as targeted at
     /// `target`, so only that sink observes it. Mirrors the `with_target`
     /// mapping that `EnrichFutureExt::enrich_for` applies.
     pub(crate) fn push_for(&mut self, target: SinkId, additional_enrichment: impl Enrichment) {
-        self.push_entries(
-            &additional_enrichment
-                .into_entries()
-                .into_iter()
-                .map(|entry| entry.with_target(target))
-                .collect(),
-        );
+        let entries: Arc<[EnrichmentEntry]> = additional_enrichment
+            .into_entries()
+            .into_iter()
+            .map(|entry| entry.with_target(target))
+            .collect();
+        self.push_entries(&entries);
     }
 
     /// Layers `entries` onto every captured chain as a single new node.
     fn push_entries(&mut self, entries: &Arc<[EnrichmentEntry]>) {
-        if self.slots.is_empty() {
+        if self.slots.is_empty() || entries.is_empty() {
             return;
         }
 
@@ -218,6 +230,14 @@ mod coverage_tests {
     impl Enrichment for TestEnrichment {
         fn into_entries(self) -> Vec<EnrichmentEntry> {
             vec![EnrichmentEntry::unclassified(self.0, 1_i64)]
+        }
+    }
+
+    struct EmptyEnrichment;
+
+    impl Enrichment for EmptyEnrichment {
+        fn into_entries(self) -> Vec<EnrichmentEntry> {
+            Vec::new()
         }
     }
 
@@ -256,6 +276,20 @@ mod coverage_tests {
     }
 
     #[test]
+    fn empty_push_returns_noop_guard_without_touching_the_slot() {
+        let slot = Slot::new();
+        let empty_guard = slot.push(Arc::from(Vec::new()));
+        assert!(slot.current().is_none());
+
+        let real_guard = slot.push(Arc::from(vec![EnrichmentEntry::unclassified("k", 1_i64)]));
+        let merged = Guard::merge([empty_guard, real_guard]);
+        assert!(slot.current().is_some());
+
+        drop(merged);
+        assert!(slot.current().is_none());
+    }
+
+    #[test]
     fn transfer_captures_a_slot_and_layers_further_enrichment_onto_it() {
         let slot = Slot::new();
         let mut transfer = EnrichmentTransfer::default();
@@ -276,6 +310,22 @@ mod coverage_tests {
             entries.iter().any(|entry| entry.target() == Some(SinkId::new("target"))),
             "`push_for` must mark its entry as targeted"
         );
+
+        drop(guard);
+        assert!(slot.current().is_none());
+    }
+
+    #[test]
+    fn transfer_ignores_empty_enrichment_layers() {
+        let slot = Slot::new();
+        let mut transfer = EnrichmentTransfer::default();
+        transfer.add_slot(&slot);
+
+        transfer.push(EmptyEnrichment);
+        transfer.push_for(SinkId::new("target"), EmptyEnrichment);
+
+        let guard = transfer.apply();
+        assert!(slot.current().is_none());
 
         drop(guard);
         assert!(slot.current().is_none());
--- a/crates/observed/src/enrichment/slot.rs
+++ b/crates/observed/src/enrichment/slot.rs
@@ -191,12 +191,12 @@ impl EnrichmentTransfer {
     /// `target`, so only that sink observes it. Mirrors the `with_target`
     /// mapping that `EnrichFutureExt::enrich_for` applies.
     pub(crate) fn push_for(&mut self, target: SinkId, additional_enrichment: impl Enrichment) {
-        let entries: Arc<[EnrichmentEntry]> = additional_enrichment
-            .into_entries()
-            .into_iter()
-            .map(|entry| entry.with_target(target))
-            .collect();
-        self.push_entries(&entries);
+        let entries = additional_enrichment.into_entries();
+        if entries.is_empty() {
+            return;
+        }
+
+        self.push_entries(&entries.into_iter().map(|entry| entry.with_target(target)).collect());
     }
 
     /// Layers `entries` onto every captured chain as a single new node.
--- a/crates/observed/src/enrichment/slot.rs
+++ b/crates/observed/src/enrichment/slot.rs
@@ -68,6 +68,10 @@ impl Slot {
     }
 
     /// Pushes entries onto the enrichment chain and returns a guard.
+    ///
+    /// An empty layer installs no chain node and records no restoration, so the
+    /// storage boundary holds for callers that reach it without passing through
+    /// [`Sink::push_enrichment`](crate::Sink).
     pub(crate) fn push(&self, entries: Arc<[EnrichmentEntry]>) -> Guard {
         if entries.is_empty() {
             return Guard::empty();
@@ -178,6 +182,11 @@ impl EnrichmentTransfer {
     /// Pushes an additional enrichment node onto every captured chain.
     /// Broadcast within the transfer's known slots; transfers with no
     /// captured slots are left unchanged.
+    ///
+    /// Emptiness is tested while the entries are still an owned `Vec`, so an
+    /// empty layer skips the shared-slice allocation as well as the chain
+    /// nodes. [`push_entries`](Self::push_entries) relies on this and on the
+    /// same check in [`push_for`](Self::push_for) for its non-empty input.
     pub(crate) fn push(&mut self, additional_enrichment: impl Enrichment) {
         let entries = additional_enrichment.into_entries();
         if entries.is_empty() {
--- a/crates/observed/src/enrichment/slot.rs
+++ b/crates/observed/src/enrichment/slot.rs
@@ -201,7 +201,7 @@ impl EnrichmentTransfer {
 
     /// Layers `entries` onto every captured chain as a single new node.
     fn push_entries(&mut self, entries: &Arc<[EnrichmentEntry]>) {
-        if self.slots.is_empty() || entries.is_empty() {
+        if self.slots.is_empty() {
             return;
         }
 
--- a/crates/observed/src/processing/processor.rs
+++ b/crates/observed/src/processing/processor.rs
@@ -60,7 +60,7 @@ pub trait EventProcessor: Send + Sync {
     /// Forces any buffered telemetry produced by this processor out to its
     /// final destination, surfacing errors. Idempotent and non-terminating -
     /// the processor remains usable after `flush()` returns. Implementors
-    /// with nothing to flush should return `Ok(())`.
+    /// with nothing to flush use the default no-op implementation.
     ///
     /// [`Sink::flush`](crate::Sink::flush) iterates all registered
     /// processors and calls this; it reports every failure, not just the first.
@@ -73,7 +73,9 @@ pub trait EventProcessor: Send + Sync {
     ///
     /// Returns a [`FlushError`] if flushing buffered telemetry to the final
     /// destination fails.
-    fn flush(&self) -> Result<(), FlushError>;
+    fn flush(&self) -> Result<(), FlushError> {
+        Ok(())
+    }
 }
 
 impl<T: EventProcessor + ?Sized> EventProcessor for Arc<T> {
--- a/crates/observed/src/processing/processor.rs
+++ b/crates/observed/src/processing/processor.rs
@@ -60,7 +60,7 @@ pub trait EventProcessor: Send + Sync {
     /// Forces any buffered telemetry produced by this processor out to its
     /// final destination, surfacing errors. Idempotent and non-terminating -
     /// the processor remains usable after `flush()` returns. Implementors
-    /// with nothing to flush use the default no-op implementation.
+    /// with nothing to flush should return `Ok(())`.
     ///
     /// [`Sink::flush`](crate::Sink::flush) iterates all registered
     /// processors and calls this; it reports every failure, not just the first.
@@ -73,9 +73,7 @@ pub trait EventProcessor: Send + Sync {
     ///
     /// Returns a [`FlushError`] if flushing buffered telemetry to the final
     /// destination fails.
-    fn flush(&self) -> Result<(), FlushError> {
-        Ok(())
-    }
+    fn flush(&self) -> Result<(), FlushError>;
 }
 
 impl<T: EventProcessor + ?Sized> EventProcessor for Arc<T> {
--- a/crates/observed/src/sink/core.rs
+++ b/crates/observed/src/sink/core.rs
@@ -284,8 +284,12 @@ impl Sink {
     /// This is the entry point used by the `.enrich(&sink, ...)` API in
     /// [`EnrichFutureExt`](crate::enrichment::EnrichFutureExt) and
     /// [`EnrichFnExt`](crate::enrichment::EnrichFnExt). Composites with zero children
-    /// return a no-op guard.
+    /// and empty enrichment layers return a no-op guard.
     pub(crate) fn push_enrichment(&self, entries: Arc<[EnrichmentEntry]>) -> Guard {
+        if entries.is_empty() {
+            return Guard::empty();
+        }
+
         match &*self.inner {
             SinkInner::Single(state) => state.enrichment.push(entries),
             SinkInner::Composite { children } => Guard::merge(children.iter().map(|c| c.enrichment.push(Arc::clone(&entries)))),
--- a/crates/observed/src/sink/core.rs
+++ b/crates/observed/src/sink/core.rs
@@ -283,8 +283,7 @@ impl Sink {
     ///
     /// This is the entry point used by the `.enrich(&sink, ...)` API in
     /// [`EnrichFutureExt`](crate::enrichment::EnrichFutureExt) and
-    /// [`EnrichFnExt`](crate::enrichment::EnrichFnExt). Composites with zero children
-    /// and empty enrichment layers return a no-op guard.
+    /// [`EnrichFnExt`](crate::enrichment::EnrichFnExt).
     pub(crate) fn push_enrichment(&self, entries: Arc<[EnrichmentEntry]>) -> Guard {
         if entries.is_empty() {
             return Guard::empty();
--- a/crates/observed/src/sink/core.rs
+++ b/crates/observed/src/sink/core.rs
@@ -284,6 +284,10 @@ impl Sink {
     /// This is the entry point used by the `.enrich(&sink, ...)` API in
     /// [`EnrichFutureExt`](crate::enrichment::EnrichFutureExt) and
     /// [`EnrichFnExt`](crate::enrichment::EnrichFnExt).
+    ///
+    /// An empty layer returns early, which spares composite fan-out and every
+    /// downstream slot write. The entry slice is built by the caller, so this
+    /// does not avoid the slice's own allocation.
     pub(crate) fn push_enrichment(&self, entries: Arc<[EnrichmentEntry]>) -> Guard {
         if entries.is_empty() {
             return Guard::empty();
--- a/crates/observed_testing/src/mock_processor.rs
+++ b/crates/observed_testing/src/mock_processor.rs
@@ -209,9 +209,7 @@ impl CapturedEvent {
 /// # struct TestEvent { #[unredacted] value: i64 }
 /// emit!(sink, TestEvent { value: 1 });
 ///
-/// let expected = ExpectedEvent::new("test.event", Severity::Info)
-///     .dimension("value", 1i64)
-///     .log();
+/// let expected = ExpectedEvent::new("test.event", Severity::Info).dimension("value", 1i64);
 /// assert_eq!(processor.single_event(), expected);
 /// ```
 #[derive(Clone)]
@@ -364,14 +362,14 @@ fn passthrough_redaction_engine() -> data_privacy::RedactionEngine {
 }
 
 // ---------------------------------------------------------------------------
-// ExpectedEvent - partial-match builder for test assertions
+// ExpectedEvent - exact-match builder for test assertions
 // ---------------------------------------------------------------------------
 
-/// A partial-match event specification for test assertions.
+/// An exact-match event specification for test assertions.
 ///
-/// Construct with [`ExpectedEvent::new`] and chain builder methods for the fields
-/// you want to verify. Only specified fields are checked - unspecified fields are
-/// ignored during comparison.
+/// Construct with [`ExpectedEvent::new`] and chain builder methods for the
+/// fields and signals the event must carry. Omitted body and dimensions are
+/// expected to be absent, and metric and disabled flags default to `false`.
 ///
 /// # Example
 ///
@@ -382,7 +380,6 @@ fn passthrough_redaction_engine() -> data_privacy::RedactionEngine {
 /// let expected = ExpectedEvent::new("http.request", Severity::Info)
 ///     .body("Request handled")
 ///     .dimension("status", 200i64)
-///     .log()
 ///     .metric();
 /// ```
 #[derive(Debug)]
@@ -399,9 +396,10 @@ pub struct ExpectedEvent {
 impl ExpectedEvent {
     /// Creates a new expected **log** event with the given name and severity.
     ///
-    /// All other fields are unchecked by default. Use builder methods to add
-    /// constraints. For an event with no log signal - a metric-only or
-    /// signal-less event - use [`without_severity`](Self::without_severity).
+    /// Omitted body and dimensions are expected to be absent, the metric and
+    /// disabled flags default to `false`, and the log signal is expected. For an
+    /// event with no log signal - a metric-only or signal-less event - use
+    /// [`without_severity`](Self::without_severity).
     #[must_use]
     pub fn new(name: impl Into<String>, severity: Severity) -> Self {
         Self {
@@ -409,7 +407,7 @@ impl ExpectedEvent {
             severity: Some(severity),
             body: None,
             sorted_dimensions: Vec::new(),
-            expect_log: false,
+            expect_log: true,
             expect_metric: false,
             expect_disabled: false,
         }
@@ -690,7 +688,7 @@ impl PartialEq<EventDescription> for ExpectedEventDescription {
 #[cfg_attr(coverage_nightly, coverage(off))]
 #[cfg(test)]
 mod coverage_tests {
-    use observed::{Event, emit};
+    use observed::{Event, emit, event};
 
     use super::*;
     use crate::events::ProbeEvent;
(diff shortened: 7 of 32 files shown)


crates/observed/src/enrichment/enrichment_trait.rs
```diff
@@ -16,6 +16,22 @@ use crate::enrichment::EnrichmentEntry;
 ///
 /// See the [`Enrichment` derive macro](crate::Enrichment) for field attributes
 /// and usage examples.
+///
+/// # Manual implementation
```
Vaiz: > /// # Manual implementation

it wasn't documented on purpose. This API is not stable, and only supported option is to use derive macro. The only reason it's public is due to Rust limitations

Vaiz: same reason why EnrichmentEntry was `doc(hidden)`

crates/observed/src/processing/processor.rs
```diff
@@ -60,7 +60,7 @@ pub trait EventProcessor: Send + Sync {
     /// Forces any buffered telemetry produced by this processor out to its
     /// final destination, surfacing errors. Idempotent and non-terminating -
     /// the processor remains usable after `flush()` returns. Implementors
-    /// with nothing to flush should return `Ok(())`.
+    /// with nothing to flush use the default no-op implementation.
```
Vaiz: roll this back, every EventProcessor must implement `flush`. This is an important operation, and failure to implement it may result in telemetry loss

crates/observed/src/sink/core.rs
```diff
@@ -284,8 +284,12 @@ impl Sink {
     /// This is the entry point used by the `.enrich(&sink, ...)` API in
     /// [`EnrichFutureExt`](crate::enrichment::EnrichFutureExt) and
     /// [`EnrichFnExt`](crate::enrichment::EnrichFnExt). Composites with zero children
```
Vaiz: >  Composites with zero children

this is implementation details that should not be exposed to user

crates/observed/src/processing/processor.rs
Vaiz: Rolled back in b9639949. `flush` is a required method again, and the eleven `Ok(())` implementations across the examples are restored.

Your reasoning is the stronger one: a default no-op means a processor that buffers telemetry and forgets to override `flush` silently drops it at shutdown, and that failure is invisible. The repetition in stateless processors is the cheaper cost.

The tracking item for this is closed as by design with that rationale attached.

crates/observed/src/enrichment/enrichment_trait.rs
Vaiz: Agreed, and reverted in b9639949. `EnrichmentEntry` is back to `#[doc(hidden)]` and the `# Manual implementation` section is gone, so the derive macro is again the only documented way in.

Sorry for the noise here — the work item behind this (ADO 7757604) read the `#[doc(hidden)]` as an oversight because a manual `impl Enrichment` has to name the type. I have closed it as by design with your reasoning recorded, so it does not get re-raised on the next review pass.

crates/observed/src/sink/core.rs
Vaiz: Removed the whole sentence in b9639949, including the pre-existing "composites with zero children" half — flagging that it was not introduced by this PR, so revert that hunk if you wanted only my addition dropped.

The doc now stops at what the method is for and says nothing about how dispatch shortcuts internally.

crates/observed/src/enrichment/slot.rs
```diff
@@ -171,25 +179,29 @@ impl EnrichmentTransfer {
...
-                .into_iter()
-                .map(|entry| entry.with_target(target))
-                .collect(),
-        );
+        let entries: Arc<[EnrichmentEntry]> = additional_enrichment
+            .into_entries()
+            .into_iter()
+            .map(|entry| entry.with_target(target))
+            .collect();
+        self.push_entries(&entries);
```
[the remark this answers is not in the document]
Vaiz: Correct, fixed in b9639949. `push_for` now takes the same early return as `push`:

```rust
let entries = additional_enrichment.into_entries();
if entries.is_empty() {
    return;
}

self.push_entries(&entries.into_iter().map(|entry| entry.with_target(target)).collect());
```

The `push_entries` guard already stopped the slot writes, but the `Arc<[EnrichmentEntry]>` was still built before reaching it, so an empty targeted layer allocated where an empty untargeted one did not.

crates/observed/examples/support/otel.rs
```diff
@@ -31,9 +31,6 @@ const CODE_FILE_PATH: &str = "code.file.path";
 /// `OTel` attribute key for the source line a call site came from.
 const CODE_LINE_NUMBER: &str = "code.line.number";
 
-/// `OTel` attribute key for the crate a call site came from.
-const CODE_NAMESPACE: &str = "code.namespace";
```
evgenyfedorov2, line 35: Why removing this, but keeping `CODE_LINE_NUMBER`?

evgenyfedorov2 (approved):

crates/observed/examples/support/otel.rs
Vaiz: Because they have opposite status in the OpenTelemetry semantic conventions: `code.line.number` is Stable, `code.namespace` is Deprecated.

The [code attribute registry](https://opentelemetry.io/docs/specs/semconv/registry/attributes/code/) lists `code.file.path`, `code.line.number`, `code.column.number`, `code.function.name` and `code.stacktrace` as Stable. `code.namespace` sits in the Deprecated table with the note "namespace is now included into `code.function.name`", so it has no standalone replacement.

The confusing part is that this file already survived the earlier rename wave: `code.filepath` became `code.file.path` and `code.lineno` became `code.line.number`, and the constants here are already on the new names. `code.namespace` was the one attribute in that group that was retired outright rather than renamed, so it is the only one left to drop.

Worth noting what this costs: `EventView::source_crate()` still exposes the emitting crate, we simply stop mapping it onto a deprecated key. If it turns out to be worth exporting, the semconv-correct route is to fold it into `code.function.name` as a fully-qualified name, but this crate does not capture a function name today, so that would be a new feature rather than a rename.

evgenyfedorov2: thanks!

crates/observed_testing/src/mock_processor.rs
```diff
@@ -690,7 +688,7 @@ impl PartialEq<EventDescription> for ExpectedEventDescription {
 #[cfg_attr(coverage_nightly, coverage(off))]
```
martin-kolinek, line 688: 🤖: Non-blocking — **`crates/observed_testing/src/mock_processor.rs:612` (nearest commentable line; this is about line 612):** `ExpectedEventDescription::new` still has the contradictory matching contract this PR removes from `ExpectedEvent`: it accepts a severity but defaults `is_log` to `false`, so the severity is ignored when matching a non-log description, while the surrounding "partial-match" docs describe equality that actually compares the body and signal flags exactly.

**Please align this sibling with the corrected builder by making `new(name, severity)` imply `LOG` and updating its matching docs.**

martin-kolinek (approved):
🤖: Approved.

crates/observed/src/sink/core.rs
```diff
@@ -283,9 +283,12 @@ impl Sink {
     ///
     /// This is the entry point used by the `.enrich(&sink, ...)` API in
     /// [`EnrichFutureExt`](crate::enrichment::EnrichFutureExt) and
-    /// [`EnrichFnExt`](crate::enrichment::EnrichFnExt). Composites with zero children
-    /// return a no-op guard.
+    /// [`EnrichFnExt`](crate::enrichment::EnrichFnExt).
     pub(crate) fn push_enrichment(&self, entries: Arc<[EnrichmentEntry]>) -> Guard {
+        if entries.is_empty() {
+            return Guard::empty();
+        }
```
sandersaares, line 294: [Copilot speaking]

# Reconcile empty-enrichment allocation behavior with the stated fast path

The added checks in `Sink::push_enrichment` and `Slot::push` run after the public `EnrichFutureExt` and `EnrichFnExt` constructors have materialized an `Arc<[EnrichmentEntry]>`, a reference-counted shared slice. Both conversion forms used by `.enrich(...)` and `.enrich_for(...)` allocate even when `Enrichment::into_entries()` returns an empty vector. The new guards therefore prevent composite fan-out, thread-local initialization, per-slot linked-list nodes, and restoration records, but they do not prevent the earlier shared-slice allocation. This differs from the pull request description's broader statement that an empty enrichment layer costs no allocation.

**Reproducible reasoning**: The ordinary future and closure constructors convert the returned `Vec<EnrichmentEntry>` directly into `Arc<[EnrichmentEntry]>`; the targeted constructors map the vector and collect into the same type. The wrapper then calls `Sink::push_enrichment`, whose new empty check is necessarily later. A counting-global-allocator probe on Rust 1.96.1 independently recorded one allocation for both `Vec::new() -> Arc<[T]>` and `Iterator::empty().collect::<Arc<[T]>>()`. By contrast, `EnrichmentTransfer::push` and `push_for` inspect their vectors before conversion, so their empty paths avoid shared-slice allocation as well as node installation. The slot regression test begins with an already-created empty `Arc`, so it validates the storage boundary rather than the public constructor's allocation behavior.

**Consequence**: Code that repeatedly constructs conditional enrichment scopes which produce no entries still performs one heap allocation per wrapper construction. Future polls do not repeat that allocation, but they still enter the sink's no-op guard path, so the public synchronous and asynchronous APIs do not currently provide the allocation-free behavior described for empty layers.

**Recommended action**: Reconcile the implementation and the stated contract. If the public paths are intended to be allocation-free, keep the entries as a `Vec` long enough to test emptiness, store a nonallocating empty state such as `None`, and have the closure invocation or future poll call `push_enrichment` only for the non-empty state; the existing non-empty path and downstream defensive guards can remain unchanged. If the earlier shared-slice allocation is intentional, narrow the documentation and pull request wording to the costs actually removed. Add `alloc_tracker` coverage around empty `.enrich` and `.enrich_for` construction for both closure and future variants so the selected guarantee is measured at the public boundary.

**References**:

- docs/performance.md — Memory allocation is the root of all evil
- crates/observed/src/enrichment/enrich_ext.rs
- [Rust 1.96 `Arc` implementation](https://doc.rust-lang.org/1.96.0/src/alloc/sync.rs.html)

**Impacted locations**:

- `crates/observed/src/sink/core.rs:287-290`
- `crates/observed/src/enrichment/slot.rs:68-74`
- `crates/observed/src/enrichment/enrich_ext.rs:48-62`
- `crates/observed/src/enrichment/enrich_ext.rs:97-118`

crates/observed/src/enrichment/slot.rs
```diff
@@ -69,6 +69,10 @@ impl Slot {
 
     /// Pushes entries onto the enrichment chain and returns a guard.
     pub(crate) fn push(&self, entries: Arc<[EnrichmentEntry]>) -> Guard {
+        if entries.is_empty() {
+            return Guard::empty();
+        }
+
```
sandersaares, line 79: [Copilot speaking]

# Add an implementation guide for enrichment storage and transfer

The published `observed` library has a user-facing design document but no implementation guide describing how its enrichment machinery fits together. This change depends on several internal layers: extension methods create shared entry slices, `Sink` broadcasts entries to leaf sinks, each `Slot` maintains a per-thread linked-list head, `Guard` restores prior heads, and `EnrichmentTransfer` captures and augments heads for another thread. The cost avoided by an empty-layer check depends on which of these boundaries performs it, but that end-to-end model is not documented in one maintainable place.

**Reproducible reasoning**: A repository search finds no `implementation.md` or equivalent guide under `crates/observed`. The source has useful type and method comments, but they describe individual components rather than the ownership and control flow across components. That distinction matters here: the sink check avoids composite traversal, the slot check protects the storage boundary, transfer checks run while a vector is still owned, and the ordinary extension constructors currently create their shared slice before reaching the sink. A high-level implementation guide would let maintainers evaluate future refactors against the actual allocation, thread-local, and restoration guarantees without reconstructing the entire call graph or relying on pull request history.

**Consequence**: A refactor can preserve emitted enrichment values while moving an emptiness check across a boundary and unintentionally reintroducing shared-slice allocation, thread-local initialization, composite fan-out, linked-list nodes, or captured-head churn. Component-level tests may still pass because they assert values rather than explaining which layer owns each performance and restoration guarantee.

**Recommended action**: Create an implementation guide for `observed`, link it from `crates/observed/DESIGN.md`, and document the enrichment pipeline from `Enrichment::into_entries()` through shared entry storage, sink dispatch, slot nodes, guard restoration, transfer capture, and transfer application. Record which empty checks are defensive and which avoid a specific cost, explicitly distinguishing entry-slice allocation from per-slot node allocation and mutation. Keep the guide descriptive of the chosen implementation; if the allocation contract is changed, update the guide and code together.

**References**:

- crates/observed/DESIGN.md
- crates/observed/src/enrichment/enrich_ext.rs
- crates/observed/src/enrichment/slot.rs

**Impacted locations**:

- `crates/observed/src/enrichment/slot.rs:68-75`
- `crates/observed/src/enrichment/slot.rs:102-118`
- `crates/observed/src/enrichment/slot.rs:160-205`
- `crates/observed/src/sink/core.rs:278-296`

crates/observed/benches/observed_benchmarks.rs
sandersaares, line 1: [Copilot speaking]

# Reconcile the benchmark overview with the registered sink-operation group

The renamed benchmark group and functions describe processor-free sink construction and emission through an explicit no-op sink, while the module-level `Measures` overview still lists `Context operations (enrich)`. Enrichment already has its own preceding category, and the overview does not introduce the sink-operation category registered by `entrypoint`.

**Reproducible reasoning**: `entrypoint` creates the `sink_operations` group and registers `bench_construct_processor_free_sink` and `bench_emit_to_noop_sink`; the nearby section heading uses the same terminology. The file header is the reader's top-down map of the suite, but its final bullet retains the former context terminology and overlaps the enrichment category. The two descriptions therefore no longer identify the same measured area.

**Consequence**: Readers relying on the file overview can look for context attachment benchmarks that are not registered and overlook the sink construction and no-op emission baselines until they inspect the implementation.

**Recommended action**: Reconcile the module overview and the registered benchmark category so they use one intended concept. If the renamed group represents the desired suite, describe sink operations at the overview level; if context enrichment remains the intended category, align the group and benchmark registrations with that intent instead. Keep the overview general and let the individual benchmark names carry the detailed operations.

**Impacted locations**:

- `crates/observed/benches/observed_benchmarks.rs:10`
- `crates/observed/benches/observed_benchmarks.rs:278-285`

crates/observed/src/enrichment/slot.rs
```diff
@@ -255,6 +275,20 @@ mod coverage_tests {
         assert!(slot.current().is_none(), "dropping the guard must pop the pushed node");
     }
 
+    #[test]
+    fn empty_push_returns_noop_guard_without_touching_the_slot() {
+        let slot = Slot::new();
+        let empty_guard = slot.push(Arc::from(Vec::new()));
+        assert!(slot.current().is_none());
+
+        let real_guard = slot.push(Arc::from(vec![EnrichmentEntry::unclassified("k", 1_i64)]));
```
sandersaares, line 293: [Copilot speaking]

# Clarify non-semantic fixture literals in the empty-enrichment tests

The new empty-enrichment tests choose a key and value solely to create one non-empty guard, and choose a target sink identifier for a call whose empty enrichment produces no entries. The tests exercise empty-versus-non-empty control flow and targeted empty transfer; their assertions do not depend on the particular fixture values.

**Reproducible reasoning**: The variable name `real_guard` and the test names explain the broad scenarios, but they do not explicitly distinguish representative fixture data from a boundary value or an identifier synchronized with another declaration. Each literal can be replaced by another valid representative without changing the assertions. A short rationale would make the tested dimension explicit without adding another case or changing production behavior.

**Consequence**: Future maintainers may spend time preserving or coordinating fixture values that carry no semantics, or may change them expecting additional coverage when the tests only distinguish an empty layer from a non-empty one.

**Recommended action**: Add concise comments stating that the enrichment entry is an arbitrary non-empty representative used to create a guard with restoration work, and that the target identifier is arbitrary because emptiness is handled before targeting can affect an entry. Do not repeat the literal values in those comments.

**Impacted locations**:

- `crates/observed/src/enrichment/slot.rs:284`
- `crates/observed/src/enrichment/slot.rs:325`

crates/observed/src/enrichment/slot.rs
```diff
@@ -69,6 +69,10 @@ impl Slot {
 
     /// Pushes entries onto the enrichment chain and returns a guard.
     pub(crate) fn push(&self, entries: Arc<[EnrichmentEntry]>) -> Guard {
+        if entries.is_empty() {
+            return Guard::empty();
+        }
```
sandersaares, line 78: [Copilot speaking]

# Document why empty-enrichment guards exist at multiple boundaries

The new empty checks protect different layers of the enrichment pipeline, but the source does not explain their distinct roles. `Arc<[EnrichmentEntry]>` is the shared slice holding one enrichment layer, while an `EnrichmentNode` is the linked-list node installed in a sink's per-thread `Slot`. The sink check avoids broadcasting an empty shared slice to composite children, the slot check protects the storage boundary from installing a node or restoration record, and the transfer checks run while they still own a vector so they can avoid targeting, shared-slice construction, and captured-head layering.

**Reproducible reasoning**: For ordinary `.enrich` wrappers, an entry slice already exists when `Sink::push_enrichment` runs; its check prevents composite fan-out and all later slot work. `Slot::push` is currently reached through the sink in production, but its own guard preserves the storage invariant for direct internal callers and future refactors. `EnrichmentTransfer::push` and `push_for` instead start from a vector, reject emptiness before conversion, and establish the non-empty precondition relied on by private `push_entries`. Without this context, the branches look interchangeable or redundant even though moving one changes which allocation or mutation is avoided. Comments must also avoid claiming that the ordinary public wrapper path is globally allocation-free, because its shared slice is constructed earlier.

**Consequence**: A later cleanup can centralize or delete an apparently duplicate check and still pass value-oriented tests while restoring composite traversal, thread-local initialization, empty chain nodes, restoration entries, or transfer-side allocation.

**Recommended action**: Add a shared high-level explanation, with short references from the relevant methods, that names the entry-slice and chain-node layers and states which cost each empty check prevents. Document that `push_entries` relies on transfer callers to provide a non-empty slice, and distinguish the sink and slot guarantees from the earlier allocation behavior of the public extension constructors. If an implementation guide is added, point the local comments to its enrichment-storage section rather than duplicating the full explanation.

**References**:

- crates/observed/DESIGN.md, Context and enrichment
- crates/observed/src/enrichment/enrich_ext.rs

**Impacted locations**:

- `crates/observed/src/enrichment/slot.rs:68-75`
- `crates/observed/src/enrichment/slot.rs:172-205`
- `crates/observed/src/sink/core.rs:278-296`

crates/observed/examples/support/otel.rs
sandersaares, line 1: [Copilot speaking]

# Preserve the OpenTelemetry source-attribute rationale beside the mapper

`populate_log_record` continues to export source file and line metadata but deliberately stops exporting `EventView::source_crate()`. OpenTelemetry semantic conventions are the standard names and meanings for telemetry attributes: the current convention keeps `code.file.path` and `code.line.number`, but removes standalone `code.namespace` and folds namespace information into `code.function.name`, the fully qualified function name. The mapper contains no explanation for this non-obvious asymmetry, although an existing review thread asked exactly why namespace was removed while line number remained.

**Reproducible reasoning**: The OpenTelemetry migration guide and current attribute registry independently confirm that file and line are stable attributes and that `code.namespace` is deprecated because its value belongs in `code.function.name`. `EventView` still exposes the emitting crate, but it does not expose a function name from which this mapper could construct the required fully qualified Rust function value. Omitting the crate is therefore a deliberate mapping decision rather than a forgotten source field. The integration test records only that the old key is deprecated; the production example, which other consumers may copy, does not preserve why the crate cannot simply be moved to the replacement key. The resolved discussion demonstrates that this distinction was already unclear to a reviewer.

**Consequence**: A future completeness cleanup could restore `code.namespace` or populate `code.function.name` with only a crate name, teaching consumers an obsolete or incomplete source mapping and reopening the same investigation.

**Recommended action**: Add a concise comment beside the source-attribute block explaining that file and line use stable keys, while crate metadata is intentionally omitted because the removed namespace key has no standalone replacement and the event model lacks the function component required for `code.function.name`. Link the OpenTelemetry migration guide or registry so the decision can be revalidated when the convention evolves.

**References**:

- [OpenTelemetry code-attribute stability migration guide](https://opentelemetry.io/docs/specs/semconv/non-normative/code-attrs-migration/)
- [OpenTelemetry code attribute registry](https://opentelemetry.io/docs/specs/semconv/registry/attributes/code/)
- crates/observed/FEATURES.md, Source location
- [Existing discussion of the source-attribute decision](https://github.com/microsoft/oxidizer/pull/712#discussion_r3895150611)

**Impacted locations**:

- `crates/observed/examples/support/otel.rs:134-143`
- `crates/observed/tests/otel_log_record.rs:260-274`

crates/observed_testing/src/mock_processor.rs
sandersaares, line 1: [Copilot speaking]

# Reconcile the pull request metadata with the final reviewed scope

The pull request metadata describes the final state inconsistently. Its title still advertises a correction to processor behavior, although the final base-to-head diff contains no processor implementation change and the description records that work as withdrawn. The Effects section also says `No public API change` after describing a behavioral change to the public `ExpectedEvent::new` function. That phrase may intend to mean no signature or published-`observed` surface change, but it can also be read as saying that no callable API contract changed.

**Reproducible reasoning**: At the reviewed head, `ExpectedEvent::new` sets `expect_log` to `true`; at the reviewed base it set it to `false`. Existing callers that omit `.log()` can therefore produce different assertion outcomes without any signature change. `observed_testing` is unpublished, which limits compatibility exposure, but its public Rust items still form the workspace test-harness contract. Separately, the changed-file list and final diff confirm that `EventProcessor` is unchanged, matching the withdrawn-work section rather than the title. The detailed body does mention both facts, so the issue is summary ambiguity rather than missing technical explanation.

**Consequence**: A reviewer or maintainer who scans the title and Effects summary can misclassify the final processor scope or the compatibility significance of the `ExpectedEvent` default before reaching the detailed sections.

**Recommended action**: Reconcile the title and Effects wording with the intended final contract without assuming which wording is authoritative. Make the title identify processor behavior only if the final reviewed content changes it, and make the API statement distinguish signatures or published surface from behavioral contracts if that narrower meaning is intended. Keep the `ExpectedEvent` compatibility description consistent with the selected terminology.

**References**:

- [Pull request 712](https://github.com/microsoft/oxidizer/pull/712)
- crates/observed_testing/Cargo.toml

**Impacted locations**:

- `crates/observed_testing/src/mock_processor.rs:396-410`
- `crates/observed_testing/src/lib.rs:24-28`

crates/observed_testing/src/mock_processor.rs
```diff
@@ -723,4 +721,17 @@ mod coverage_tests {
...
         assert_eq!(expected_desc, ProbeEvent::DESCRIPTION);
     }
+
+    #[event("fieldless.log")]
+    #[info]
+    struct FieldlessLog;
+
+    #[test]
+    fn expected_event_new_matches_fieldless_log_without_log_builder() {
+        let (sink, processor) = test_emitter(observed::SinkId::new("fieldless_log"));
```
sandersaares: [Copilot speaking]

# Consolidate `SinkId` references in the coverage test module

The new fieldless-log regression test constructs `observed::SinkId` with an inline crate-qualified path even though the test module already groups its other `observed` dependencies in a `use` statement. An earlier test in the same module uses the same qualified form, so the new line follows nearby precedent but also extends the split between imported items and inline type paths.

**Reproducible reasoning**: There is no competing `SinkId` in `coverage_tests`, and both calls refer to the same public type. Bringing the type into the module import list makes the module's direct dependencies visible in one place and avoids maintaining two spelling styles. Updating only the newly added call would leave the adjacent existing call inconsistent, so the mechanically coherent change covers both test-local references.

**Consequence**: The mixed import pattern adds small navigation and maintenance friction: readers must scan test bodies for dependency paths, and a future path or import reorganization has more call sites to update.

**Recommended action**: Add `SinkId` to `use observed::{...};` in `coverage_tests` and use `SinkId::new(...)` for both sink constructions in that module. This is a local readability cleanup; it does not change the test behavior.

**Impacted locations**:

- `crates/observed_testing/src/mock_processor.rs:696-706`
- `crates/observed_testing/src/mock_processor.rs:729-731`

crates/observed_testing/src/mock_processor.rs
```diff
@@ -723,4 +721,17 @@ mod coverage_tests {
         let expected_desc = ExpectedEventDescription::new("test.probe", Severity::Info).log();
         assert_eq!(expected_desc, ProbeEvent::DESCRIPTION);
     }
+
+    #[event("fieldless.log")]
+    #[info]
+    struct FieldlessLog;
```
sandersaares: [Copilot speaking]

# Group the fieldless event declaration with the module's fixtures

The newly added `FieldlessLog` test event is declared between existing test functions in `coverage_tests`. A type-first module layout is useful here because helper event declarations are fixtures shared conceptually by tests, while functions are the executable cases; interleaving the two makes the module harder to scan as more fixtures are added.

**Reproducible reasoning**: The module currently has imports, two test functions, the `FieldlessLog` declaration, and then its test. Moving the declaration next to the imports gives readers one predictable place to find the module's types and keeps the test functions as a contiguous block. The change is mechanical and does not alter macro expansion or test behavior.

**Consequence**: Leaving new fixture types interspersed with test functions establishes an inconsistent growth pattern that makes declarations and executable cases progressively harder to locate.

**Recommended action**: Move the `FieldlessLog` declaration to immediately after the imports in `coverage_tests`, before the test functions, and leave its test with the other executable cases.

**Impacted locations**:

- `crates/observed_testing/src/mock_processor.rs:692-727`

crates/observed_testing/src/mock_processor.rs
```diff
@@ -364,14 +362,14 @@ fn passthrough_redaction_engine() -> data_privacy::RedactionEngine {
...
 
-/// A partial-match event specification for test assertions.
+/// An exact-match event specification for test assertions.
 ///
-/// Construct with [`ExpectedEvent::new`] and chain builder methods for the fields
-/// you want to verify. Only specified fields are checked - unspecified fields are
-/// ignored during comparison.
+/// Construct with [`ExpectedEvent::new`] and chain builder methods for the
+/// fields and signals the event must carry. Omitted body and dimensions are
+/// expected to be absent, and metric and disabled flags default to `false`.
```
sandersaares: [Copilot speaking]

# Reconcile metric and disabled terminology in `ExpectedEvent` documentation

The revised `ExpectedEvent` documentation groups the metric expectation and the disabled state together as `flags`, while the crate's conceptual model describes logs and metrics as signals and `disabled` as a separate flag. The same constructor paragraph already calls the log expectation a signal, so the public explanation uses two categories for the symmetric log and metric expectations.

**Reproducible reasoning**: `EventDescription` contains optional log and metric signal descriptions plus an independent disabled flag, and `ExpectedEvent` compares the presence of each signal separately from disabled state. Calling the internal Boolean expectations flags is understandable as an implementation description, but the surrounding API documentation otherwise explains event semantics rather than storage fields. The current wording therefore makes it harder to map `.metric()` to `EventDescription::contains_metrics()` and `.disabled()` to `EventDescription::is_disabled()`.

**Consequence**: Readers can interpret metrics as a status bit rather than one of the event's emitted signals, weakening the correspondence between the test builder and the terminology used throughout `observed`.

**Recommended action**: Make the `ExpectedEvent` paragraphs and the higher-level event model use one terminology. If the design vocabulary remains authoritative, describe the metric expectation as a signal and `disabled` as a flag; if `flags` is intended as a public builder concept, define that concept and apply it consistently to the log expectation as well.

**References**:

- crates/observed/DESIGN.md — EventDescription terminology

**Impacted locations**:

- `crates/observed_testing/src/mock_processor.rs:365-372`
- `crates/observed_testing/src/mock_processor.rs:396-400`

sandersaares: [Copilot speaking]

# Reconcile the exact-match claim with the properties actually compared

The revised type documentation calls `ExpectedEvent` an `exact-match event specification`, but its equality implementation does not compare every property retained by `CapturedEvent`. It compares name, severity, body, collected dimensions, log presence, metric presence, and disabled state; it does not compare source file or line, and metric presence does not verify event- or field-level instrument descriptions.

**Reproducible reasoning**: The exact-match wording correctly communicates that omitted body, dimensions, and signal expectations are not wildcards. However, `CapturedEvent` also exposes source-location accessors, an event-level metric description, and captured field metrics. Those properties remain outside `PartialEq<ExpectedEvent>`. Without defining the comparison boundary, `exact-match` can reasonably be read as equality over the complete captured snapshot rather than exactness over the subset modeled by `ExpectedEvent`.

**Consequence**: A test author can believe that `assert_eq!(captured, expected)` verifies source metadata or metric instrument kind and name even though those properties can change without failing the assertion.

**Recommended action**: Reconcile the documented contract and the comparison boundary without assuming which side should expand. Either scope `exact-match` to the properties `ExpectedEvent` models and direct readers to `CapturedEvent` accessors for source and metric metadata, or extend the expectation model and equality implementation if complete snapshot equality is intended.

**References**:

- crates/observed_testing/src/mock_processor.rs — CapturedEvent accessors and PartialEq<ExpectedEvent>

**Impacted locations**:

- `crates/observed_testing/src/mock_processor.rs:365-372`
- `crates/observed_testing/src/mock_processor.rs:470-478`

crates/observed_testing/src/mock_processor.rs
sandersaares, line 1: [Copilot speaking]

# Derive the expected log signal from severity instead of storing it twice

`ExpectedEvent` represents the same log-signal fact in two fields: `severity: Option<Severity>` records whether a severity is expected, while `expect_log: bool` separately records whether the event is expected to contain a log signal. A captured event does not have two independent facts here: `EventView::severity()` returns `Some` precisely when `EventDescription::log()` is present, and `EventDescription::is_log()` tests that same option. The corrected constructors now initialize the two expectation fields in lockstep, leaving the `log()` builder as the only API that can separate them.

**Reproducible reasoning**: `ExpectedEvent::new` produces `Some(severity)` with `expect_log = true`, and `ExpectedEvent::without_severity` produces `None` with `expect_log = false`; the other builders do not modify either field. On the captured side, both severity and log presence originate from `EventDescription::log()`. Consequently, `log()` after `new` is idempotent, while `log()` after `without_severity` creates an expectation that requires both no severity and a log signal, which no captured event can satisfy. The independent Boolean therefore carries no valid state that the severity option does not already express.

**Consequence**: The builder can construct a permanently unsatisfiable expectation, and future constructor or builder changes can accidentally desynchronize two fields that are meant to express one invariant. Retaining the duplicate state also leaves callers and documentation maintaining explicit `.log()` syntax that no longer contributes information.

**Recommended action**: Remove `expect_log` and the `ExpectedEvent::log` builder, remove the remaining redundant `.log()` calls from `ExpectedEvent::new` chains, and derive log presence from the severity option during comparison. Because `CapturedEvent::severity` and `EventDescription::is_log()` come from the same log-description option, the existing severity comparison already checks both presence and level; the separate log comparison can be removed. This causes mechanical call-site churn in the unpublished test-support crate, but it makes the invalid state unrepresentable and leaves one source of truth for the expectation.

**References**:

- crates/observed/src/processing/event_view.rs — EventView::severity
- crates/observed/src/metadata/event.rs — EventDescription::is_log
- crates/observed/DESIGN.md — Signal Routing: How Events Become Logs and Metrics

**Impacted locations**:

- `crates/observed_testing/src/mock_processor.rs:396-452`

sandersaares (approved):
[Copilot speaking]

Published 12 findings. No finding follows up on an existing discussion thread.

| Diagnostic | Value |
|------------|-------|
| Cache | Hit |

</details>

crates/observed/src/enrichment/slot.rs
Vaiz: Declined.

The observation behind this is sound and I took the cheap half of it: `slot.rs:71`, `slot.rs:185` and `sink/core.rs:287` now each state which cost their empty check avoids, and the `Sink` comment says explicitly that the entry slice is built by the caller and is therefore not avoided there. That is the part a refactor actually needs, and it sits where the refactor happens.

A separate implementation guide is the part I am not taking. It would be a new document whose subject is three `is_empty()` checks, kept in sync by hand with the code that already explains itself, and this repository has a standing preference for less documentation rather than more. A guide that drifts is worse than no guide: it reads as authoritative while describing an implementation that has moved on.

Leaving this open rather than resolving it, since it is a judgement call and you may weigh the maintenance cost differently.

crates/observed/src/enrichment/slot.rs
Vaiz: Declined.

The premise is true — the entry key, its value and the target sink identifier are arbitrary representatives, and the assertions do not depend on them. But the remedy is a comment whose only content is that the line below it means nothing in particular, and the test names (`push_empty_layer_leaves_slot_untouched`, and the targeted transfer case) already carry the tested dimension. Adding prose that says "this literal is arbitrary" is the kind of sentence that survives long after the test around it has changed, and this repository prefers fewer of them.

Leaving this open in case you disagree.

crates/observed/src/sink/core.rs
Vaiz: Verified and fixed in `5c5ae977`, though not by removing the allocation.

The premise is correct. `EnrichFutureExt::enrich` at `enrich_ext.rs:51` does `enrichment.into_entries().into()`, and `enrich_for` at `enrich_ext.rs:62` collects into the same type, so both materialize an `Arc<[EnrichmentEntry]>` before `Sink::push_enrichment` is ever called. `EnrichmentTransfer::push` and `push_for` do test emptiness while the entries are still an owned `Vec`, so the two paths genuinely differ. My Effects bullet claimed the broader guarantee and was wrong.

I narrowed the contract rather than widening the fast path. Making the public wrappers allocation-free means storing `Option<Arc<[EnrichmentEntry]>>` in `Enriched` and branching on it in every `poll`, which charges a branch to every user of `.enrich(...)` to save one allocation in the case where the enrichment layer is empty — a case that is rare precisely because a caller who knows the layer is empty does not wrap at all. The trade did not look worth it.

So the PR description now says an empty layer installs no chain node and writes no slot, that a layer pushed through a captured `Transfer` also costs no entry-slice allocation, and that `.enrich(...)` and `.enrich_for(...)` still build their `Arc` before the guard is reached. `Sink::push_enrichment` carries the same statement in its own doc comment, so the boundary is recorded where a future refactor will read it.

I did not add the `alloc_tracker` coverage, since there is no longer a public allocation-free guarantee to measure. Happy to reverse the decision and take the `Option` if you would rather have the fast path than the narrower contract.

crates/observed/benches/observed_benchmarks.rs
Vaiz: Valid, fixed in `5c5ae977`.

`observed_benchmarks.rs:10` still read `Context operations (enrich)` after the group became `sink_operations`, so the overview described a category that `entrypoint` does not register while duplicating the enrichment category directly above it. The bullet now reads `Sink operations (construction, emit through a no-op sink)`, which matches the registered group and the section heading.

crates/observed/src/enrichment/slot.rs
Vaiz: Partly taken, in `5c5ae977`.

Each of the three checks now states which cost it avoids, at the site of the check rather than in a shared document. `Sink::push_enrichment` (`sink/core.rs:287`) records that an empty layer spares composite fan-out and every downstream slot write, and that the entry slice is built by the caller so its allocation is not avoided here. `Slot::push` (`slot.rs:71`) records that an empty layer installs no chain node and no restoration record, which is what keeps the storage boundary correct for callers that do not arrive through the sink. `EnrichmentTransfer::push` (`slot.rs:185`) records that emptiness is tested while the entries are still an owned `Vec`, so it skips the shared slice as well, and that `push_entries` depends on that check and on the matching one in `push_for` for its non-empty input.

I did not add the separate implementation guide the sibling comment asks for; my reasoning is on that thread.

crates/observed/src/enrichment/slot.rs
Vaiz: Correction to my previous reply: I named the tests from memory and got them wrong. They are `empty_push_returns_noop_guard_without_touching_the_slot` (`slot.rs:288`) and `transfer_ignores_empty_enrichment_layers` (`slot.rs:328`). The point stands — both names state the tested dimension — but the citations in my reply above did not.

crates/observed_testing/src/mock_processor.rs
Vaiz: Valid, fixed.

The title said `processor` while the final diff touches no processor code, because that work was withdrawn in `b9639949`. It is now `fix(observed): correct enrichment and test-harness behavior`, and the withdrawn-work section says explicitly that no processor code is touched by this pull request.

The Effects line was ambiguous in the way you describe, and the ambiguity has since become sharper: `ExpectedEvent::log` is now removed outright, so the harness contract changes in signature as well as behavior. That bullet now reads that no published crate changes its API, that the `observed` public surface is as it was on `main`, and that the removal is a behavioral and signature change to the unpublished `observed_testing` harness contract.

crates/observed/examples/support/otel.rs
Vaiz: Taken, in `5c5ae977`.

A comment now sits above the source-attribute block in `examples/support/otel.rs`: file and line use stable OpenTelemetry code attributes, and the emitting crate is deliberately not exported because `code.namespace` is deprecated with no standalone replacement, while its successor `code.function.name` requires a fully qualified function name the event model does not capture. The attribute registry is linked from the comment so the decision can be rechecked when the convention moves.

You are right that the earlier discussion is the evidence this needed writing down: a reviewer asked the same question on the diff, and the answer lived only in that thread.

crates/observed_testing/src/mock_processor.rs
Vaiz: Valid, fixed in `5c5ae977`. The documentation now calls the metric expectation a signal and reserves "flag" for `disabled`, matching the way `EventDescription` models them. The log expectation is no longer described as a separate switch at all, because it is now derived from the severity.

crates/observed_testing/src/mock_processor.rs
Vaiz: Valid, fixed in `5c5ae977`. The `FieldlessLog` declaration now sits immediately after the imports, and the test functions in `coverage_tests` are a contiguous block. Its test also lost the `_without_log_builder` suffix, since the builder it referred to no longer exists.

crates/observed_testing/src/mock_processor.rs
Vaiz: Valid, fixed in `5c5ae977`. `SinkId` is now in the `use observed::{Event, SinkId, emit, event};` import and both constructions in `coverage_tests` use the short form, including the pre-existing one at what was line 706.

crates/observed_testing/src/mock_processor.rs
Vaiz: Valid, fixed in `5c5ae977` by scoping the claim rather than widening the comparison.

You are right that `PartialEq<ExpectedEvent> for CapturedEvent` leaves `source_file`, `source_line`, `field_metrics` and the event-level metric description outside the comparison, while the type documentation invited a reading of "exact" as equality over the whole captured snapshot.

The documentation now says exactness covers what the type models — name, severity, body, dimensions, the metric signal and the disabled flag — and directs readers to the `CapturedEvent` accessors for source and metric metadata. I did not extend `ExpectedEvent` to model those, because `ExpectedEventDescription` already exists for asserting instrument name and kind, and source location is captured for inspection rather than for assertion; folding both into this type would duplicate one and pin the other in every assertion that does not care about it.

crates/observed_testing/src/mock_processor.rs
Vaiz: Valid, taken in full in `5c5ae977`.

Confirmed against the code: `CapturedEvent::severity` is `event.severity()` (`mock_processor.rs:176`), and `EventView::severity` is `description().log().map(LogDescription::severity)` (`event_view.rs:178`), so a captured event carries a severity exactly when it carries a log signal. `expect_log` therefore carried no state the severity option did not already express, and `without_severity().log()` built an expectation nothing could satisfy.

`expect_log` and the `log()` builder are gone, the redundant `other.expect_log == self.description.is_log()` comparison is dropped from `PartialEq`, and the ~70 `.log()` call sites across `observed_testing`'s tests and `fetch_winhttp_impl` are removed. `ExpectedEventDescription::log` is untouched — it is a different type with a non-optional severity, and the sibling thread about it is answered separately.

The documentation now states the derivation rather than the flag, and the fieldless-log test is renamed to `expected_event_new_matches_fieldless_log`.

crates/observed_testing/src/mock_processor.rs
Vaiz: Confirmed the inconsistency, but declining it here and proposing a follow-up instead.

The premise checks out. `ExpectedEventDescription::new` takes a severity and sets `is_log: false`, and `PartialEq` at `mock_processor.rs:663-674` then compares `self.is_log() == other.is_log`, so a description built with `new(name, severity)` and no `.log()` can only ever match an event that carries no log signal — at which point the severity argument is inert.

What stops me making `new` imply LOG is that this type's `severity` is not an `Option`, so there is no `without_severity` counterpart to escape to. `derive_event.rs:259` and `fetch_winhttp_impl/src/telemetry.rs:148` both build metric-only descriptions as `new(name, Severity::Info)` with no `.log()`, and the latter carries a comment saying exactly that the severity is inert there. Flipping the default means either adding a `no_log()` opt-out — a new builder whose only job is to undo a default — or making the severity optional, and then updating roughly twenty call sites across two crates in a type this pull request does not otherwise touch.

Given it is non-blocking, I would rather keep it out of a change that already reworks `ExpectedEvent`, and do it as its own change where the `Option<Severity>` question can be decided on its merits. Leaving this thread open as the record. If you would rather have it here, say so and I will take the `Option<Severity>` route in this branch.

sgalkin (approved):

martin-kolinek (approved):

crates/observed_testing/tests/basic_emission.rs
```diff
@@ -133,10 +131,7 @@ fn emit_event_with_no_fields() {
 
     emit!(sink, Heartbeat);
 
-    assert_eq!(
-        processor.single_event(),
-        ExpectedEvent::new("internal.heartbeat", Severity::Trace).log(),
-    );
+    assert_eq!(processor.single_event(), ExpectedEvent::new("internal.heartbeat", Severity::Trace),);
```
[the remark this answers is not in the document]
Vaiz: Fixed in `befe9d69`. The call takes the two-argument form, so the trailing comma is gone:

```rust
assert_eq!(processor.single_event(), ExpectedEvent::new("internal.heartbeat", Severity::Trace));
```

The same comma in `metric_routing.rs:43` is removed in the same commit. The other `,);` occurrences under `crates/` are pre-existing and outside this pull request's diff, so they stay as they are.

crates/observed_testing/tests/metric_routing.rs
```diff
@@ -40,10 +40,7 @@ fn event_without_metrics_is_log_only() {
     emit!(sink, LogEvent { value: 1 });
 
     let event = processor.single_event();
-    assert_eq!(
-        event,
-        ExpectedEvent::new("test.probe", Severity::Info).dimension("value", 1i64).log(),
-    );
+    assert_eq!(event, ExpectedEvent::new("test.probe", Severity::Info).dimension("value", 1i64),);
```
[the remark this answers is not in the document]
Vaiz: Fixed in `befe9d69`, together with the same comma in `basic_emission.rs:134`:

```rust
assert_eq!(event, ExpectedEvent::new("test.probe", Severity::Info).dimension("value", 1i64));
```

martin-kolinek (approved):

PR #715, opened 2026-08-31
refactor(ohno_macros_impl): let the shape place the core and the grammar split a `#[from]` entry

Vaiz:

Mostly an internal refactor of `ohno_macros_impl`. One input that used to be accepted is now rejected, deliberately. **No code that compiles today stops compiling** — see Compatibility below.

## The corner this is really about

A `#[from(...)]` entry is a type, optionally followed by `(field: expression, ...)` overrides. That makes a `(` in the leading position genuinely ambiguous, and these all mean different things:

```rust
#[from(std::io::Error)]                         // a path type
#[from((std::io::Error))]                       // the same type, parenthesized
#[from((u8, u16))]                              // a tuple type
#[from(std::io::Error(path: "?".to_owned()))]   // a type, plus a field override
#[from((std::io::Error)(path: "?".to_owned()))] // parenthesized, plus an override
```

Telling them apart is the parser's whole job here. Before this PR, exactly one of them had a test. All of them do now.

## What you write, and what you get

```rust
#[derive(Error)]
#[from(std::io::Error)]
struct T {
    source: String,
    inner: ohno::OhnoCore,
}
```

produces this, byte-identically before and after the change:

```rust
impl ::core::convert::From<std::io::Error> for T {
    fn from(error: std::io::Error) -> Self {
        let __ohno_fields = (::core::default::Default::default(),);
        Self {
            source: __ohno_fields.0,
            inner: ::ohno::OhnoCore::from(error),
        }
    }
}
```

so ordinary usage is untouched:

```rust
let io_err = std::io::Error::new(std::io::ErrorKind::NotFound, "test error");
let my_err: MyError = io_err.into();

assert_error_message!(my_err, "test error");
assert_eq!(my_err.code, 0); // fields you did not name are defaulted
```

Field overrides, several source types in one attribute, generic sources whose commas belong to the type rather than to the list, tuple structs, and a core field placed anywhere in the struct all produce identical output too. Every pre-existing expansion snapshot is unchanged.

## `()` is now rejected

`#[from(())]` used to be accepted by the macro. It was never *usable*: `OhnoCore::from` requires `T: Into<Box<dyn Error + Send + Sync>>`, which `()` does not satisfy, so the generated `From<()>` failed to compile — pointing into code the user never wrote, which `design.md` R4 forbids.

Before, on `main`:

```
error[E0277]: the trait bound `(): std::error::Error` is not satisfied
    = note: required for `OhnoCore` to implement `From<()>`
```

After, pointing at what you actually wrote:

```
error: `#[from(...)]` cannot convert from `()`, which is not an error type.
       Name the type the conversion converts from, such as `#[from(std::io::Error)]`
  --> tests/ui/from_unit_source.rs:11:8
   |
11 | #[from(())]
   |        ^^
```

`#[from((()))]` is the same type and gets the same message. `((),)` is a genuine one-element tuple and is left alone.

The rejection is a *rule*, checked in the validation phase, and that placement matters. `#[from(...)]` parses its entries as a single list, so had the parser instead treated `()` as an empty override list and failed at decode time, one `()` would have discarded every other entry in the attribute. A new `tests/ui` fixture pins that it does not:

```rust
#[from((), std::io::Error(missing: 1))]
```

reports **both** the `()` and the unknown field, rather than losing the second behind the first.

## Compatibility

Not a breaking change, in the sense that matters: no source that compiles against `main` fails to compile after this.

- **`#[from(())]`** — verified empirically, not assumed. `OhnoCore` has exactly one `From` impl, bounded on `Into<Box<dyn StdError + Send + Sync>>`, and `OhnoCore::from(())` is rejected by `rustc` with `E0277`. So any crate containing `#[from(())]` already failed to build. Code that did not compile still does not compile; only the diagnostic improves.
- **Public API** — unchanged. `ohno_macros_impl` exports the same three functions with the same signatures, and no type in `ohno` is touched, so `cargo-semver-checks` has nothing to report.
- **Diagnostic wording** — one message changes for a malformed override list, which both versions reject. Message text is not part of the API contract, and no pre-existing `.stderr` snapshot covered it.

| input | before | after |
| --- | --- | --- |
| `#[from(())]` | accepted, then `E0277` in generated code | rejected, message at the `()` |
| `#[from(std::io::Error(path: ))]` | rejected, ``expected a type`` | rejected, ``unexpected token, expected ")"`` |
| everything else | — | identical |

## Diagnostics you see when you get it wrong

Otherwise unchanged, message for message and span for span — the compile-fail tests that pin them are untouched:

```
error: unknown field `missing` in `#[from(...)]`, available fields: `path`
error: `#[from(...)]` cannot initialize `inner`, which holds the OhnoCore and is built from the source error
error: `#[from(...)]` field keys for a tuple struct are field indexes, not names, so `path:` names no field
error: `#[from(...)]` needs at least one type, such as `#[from(std::io::Error)]`
```

## Why touch it at all

Three things were re-derived by hand that a type or a grammar already held. The entry parser restated part of Rust's path grammar as raw token peeking, so that `(std::io::Error)` would not read as a field named `std`; it now settles the question by trying to parse an override list and seeing whether that succeeds. Both code generators re-found the core field by comparing member names, and one threaded its own index alongside that walk guarded by an `expect` — a panic path in the generation phase, which `design.md` requires not to have one; the shape now reports each field's position directly. Finally, override keys were compared by formatting both sides to `String`, where `syn::Member` already implements `PartialEq`.

Implementation source shrinks by roughly 20 lines. The rest of the diff is test coverage, plus the `design.md` and `requirements.md` updates for the `()` rejection.

## Validation

`ohno_macros_impl` and `ohno` both pass on the pinned MSRV, including the `trybuild` compile-fail tests that pin the diagnostics. Clippy is clean under `-D warnings`, and both formatting gates pass on the pinned nightly. The parser change was checked against the previous implementation with a differential probe over 68 adversarial `#[from(...)]` entries; `()` is the only input whose outcome moves.

`darling` was evaluated first and rejected: its derives parse each attribute through `NestedMeta`, which cannot represent `#[display("{a}", self.x())]`, a `#[from(...)]` entry carrying generic arguments or `key: expr` pairs, or bare markers such as `#[no_debug]`. Adopting it would have added a dependency without removing any of the code above.

### DIFF
--- a/crates/ohno_macros_impl/src/derive_error/mod.rs
+++ b/crates/ohno_macros_impl/src/derive_error/mod.rs
@@ -18,7 +18,7 @@ pub(crate) mod parse;
 pub(crate) mod validate;
 
 use proc_macro2::TokenStream;
-use syn::DeriveInput;
+use syn::{DeriveInput, Member};
 
 use crate::diagnostics::Errors;
 
@@ -33,4 +33,16 @@ pub(crate) fn expand(input: DeriveInput) -> TokenStream {
         .map(|model| generate::generate(&model));
 
     expanded.unwrap_or_else(|| errors.into_compile_error())
+}
+
+/// Renders a member the way a diagnostic spells it, and the way `Debug` labels it: `path`, or `0`.
+///
+/// Shared by all three phases: a member reaches the user as text in a diagnostic, in a template
+/// resolution and in a `Debug` label, and the three have to agree on how it is spelled.
+#[must_use]
+pub(crate) fn member_name(member: &Member) -> String {
+    match member {
+        Member::Named(ident) => ident.to_string(),
+        Member::Unnamed(index) => index.index.to_string(),
+    }
 }
--- a/crates/ohno_macros_impl/src/derive_error/model.rs
+++ b/crates/ohno_macros_impl/src/derive_error/model.rs
@@ -11,8 +11,8 @@ use proc_macro2::Span;
 use quote::format_ident;
 use syn::{Expr, Generics, Ident, Member, Type};
 
-use super::ast::Style;
-use super::parse::member_name;
+use super::ast::{FromOverride, Style};
+use super::member_name;
 use crate::diagnostics::Errors;
 use crate::message::Message;
 
@@ -79,9 +79,30 @@ impl Shape {
         &self.core
     }
 
+    /// Every field, in declaration order, each with its [`Position`]. This is what generation walks.
+    ///
+    /// Pairing each field with its position here is what keeps "which field holds the core" a fact
+    /// of the shape. A generator that recovered it by comparing members would be re-deciding, at
+    /// every use, something the split around the core already settled.
+    pub(crate) fn positions(&self) -> impl Iterator<Item = (&ModelField, Position)> {
+        let after = self.before.len();
+
+        self.before
+            .iter()
+            .enumerate()
+            .map(|(index, field)| (field, Position::Data(index)))
+            .chain(std::iter::once((&self.core, Position::Core)))
+            .chain(
+                self.after
+                    .iter()
+                    .enumerate()
+                    .map(move |(index, field)| (field, Position::Data(after + index))),
+            )
+    }
+
     /// Every field, in declaration order. What `Debug` prints.
     pub(crate) fn all(&self) -> impl Iterator<Item = &ModelField> {
-        self.before.iter().chain(std::iter::once(&self.core)).chain(self.after.iter())
+        self.positions().map(|(field, _)| field)
     }
 
     /// Every field but the core, in declaration order.
@@ -91,6 +112,15 @@ impl Shape {
         self.before.iter().chain(self.after.iter())
     }
 }
+/// Where a field sits in a [`Shape`].
+#[derive(Clone, Copy, Debug)]
+pub(crate) enum Position {
+    /// The field holding the `OhnoCore`.
+    Core,
+    /// A field other than the core, numbered as [`Shape::data`] yields it.
+    Data(usize),
+}
+
 /// One field of a validated error type.
 #[derive(Debug)]
 pub(crate) struct ModelField {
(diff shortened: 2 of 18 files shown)


Vaiz:
🤖 **Clawpilot here!** Automated review comment from an AI agent -- not a human reviewer. Please verify before acting.

AI reviewer: Claude Opus 5

`design.md` justifies the new `()` rejection with "nothing converts from it, so a generated `From<()>` would fail inside code the user cannot see, which R4 forbids". That reason does not pick out `()`. This same commit adds the snapshot case `a tuple source type is not an override list`, which accepts `#[from((u8, u16))]` and expands to `impl ::core::convert::From<(u8, u16)> for T` with the body `inner: ::ohno::OhnoCore::from(error)`. `OhnoCore`'s only `From` impl is bounded on `Into<Box<dyn Error + Send + Sync>>` (`crates/ohno/src/core.rs:178`), so that body fails with the same `E0277`, in generated code, exactly as `From<()>` would. R4 allows recorded exceptions, but the `Limits` section lists only the `const fn`, format-spec and `map_err` cases, so this one is not an accepted limit either.

The paragraph above it already justifies the rule without appealing to R4: `()` is the one input both readings can claim, and reading it as an empty override list would fail at decode time and discard every sibling entry in the attribute. Drop the R4 clause and let that argument stand. The `from_unit_source.rs` module doc and the new `display_diagnostics.rs` comment repeat the same R4 framing and need the same edit. The `UNIT_SOURCE` message in `validate.rs` has the same problem: `(u8, u16)` is not an error type either.

The alternative is to widen `is_unit` to every `Type::Tuple` and update the `(u8, u16)` snapshot case. That is a larger rule than this PR set out to add. Either way, a reader currently has to guess whether `#[from((u8, u16))]` is an unfixed bug against a stated rule or evidence that the rule is decorative.

crates/ohno_macros_impl/src/derive_error/mod.rs
```diff
@@ -34,3 +34,15 @@ pub(crate) fn expand(input: DeriveInput) -> TokenStream {
 
     expanded.unwrap_or_else(|| errors.into_compile_error())
 }
+
+/// Renders a member the way a diagnostic spells it, and the way `Debug` labels it: `path`, or `0`.
+///
+/// Shared by all three phases: a member reaches the user as text in a diagnostic, in a template
```
Vaiz, line 40: 🤖 **Clawpilot here!** Automated review comment from an AI agent -- not a human reviewer. Please verify before acting.

AI reviewer: Claude Opus 5

`member_name` has no call site in `parse.rs`, and had none before this change either. The callers are `model.rs:203`, `model.rs:215`, `display/mod.rs:41`, `display/mod.rs:50` and `generate/traits.rs:138`. The `display` module is imported and called only from `validate.rs`, and `unknown_key` is a validation diagnostic, so the users are validate and generate — two of the three phases the module doc names above.

The count is load-bearing for a later reader deciding whether this helper can move back down into one module, so make it say two: "Shared by validate and generate: a member reaches the user as text in a diagnostic, in a template resolution and in a `Debug` label, and the three have to agree on how it is spelled." The list of three uses is correct; only the phase count is not.

crates/ohno_macros_impl/src/derive_error/model.rs
```diff
@@ -91,6 +112,15 @@ impl Shape {
         self.before.iter().chain(self.after.iter())
```
Vaiz, line 112: 🤖 **Clawpilot here!** Automated review comment from an AI agent -- not a human reviewer. Please verify before acting.

AI reviewer: Claude Opus 5

`Position::Data` is documented as "numbered as `Shape::data` yields it", and the generated conversions depend on that: `Conversion::new` lays out `initializers` by walking `data()`, while `construct` emits `__ohno_fields.#index` from the `Position::Data(index)` that `positions()` hands out. Those two orders come from two independent iterator expressions — `positions()` walks `before`/`core`/`after` with its own offset, this line walks `before`/`after`.

They agree today. If `data()` later gains a filter or a different chain order, `positions()` keeps numbering the old way and every generated `From<T>` assigns initializers to the wrong fields, with no compile error — only the snapshots would catch it. That is the second alignment the `positions()` doc argues against a few lines up. `all()` was rewritten in this PR to go through `positions()`; `data()` can do the same, and then the invariant stops being a comment:

```rust
pub(crate) fn data(&self) -> impl Iterator<Item = &ModelField> {
    self.positions()
        .filter_map(|(field, position)| matches!(position, Position::Data(_)).then_some(field))
}
```

Low priority as written, since the two definitions sit a few lines apart today.

crates/ohno_macros_impl/src/derive_error/model.rs
```diff
@@ -79,9 +79,30 @@ impl Shape {
         &self.core
     }
 
+    /// Every field, in declaration order, each with its [`Position`]. This is what generation walks.
+    ///
+    /// Pairing each field with its position here is what keeps "which field holds the core" a fact
+    /// of the shape. A generator that recovered it by comparing members would be re-deciding, at
+    /// every use, something the split around the core already settled.
+    pub(crate) fn positions(&self) -> impl Iterator<Item = (&ModelField, Position)> {
+        let after = self.before.len();
```
Vaiz, line 88: 🤖 **Clawpilot here!** Automated review comment from an AI agent -- not a human reviewer. Please verify before acting.

AI reviewer: Claude Opus 5

This binds the count of the fields *before* the core to `after`, which is also the name of a field on this struct. Eleven lines down it reads `Position::Data(after + index)` inside the closure that walks `self.after`, where it looks like "the number of fields after the core, plus this one's index" — the wrong quantity, and one that would give the wrong answer.

The value is correct: the `after` fields do start at `before.len()` in the data numbering. But this is the one function the generated conversions' field ordering depends on, so the name should not point the reader the other way. Rename it: `let data_before_core = self.before.len();`, used as `Position::Data(data_before_core + index)`.

PR #725, opened 2026-09-03
feat(observed): add event sampling support

evgenyfedorov2:
Adds opt-in event sampling to `observed`.
 
 - Calls the sampler once for every event a `Sink` is interested in.
 - Sampling implementations have access to the event description, Sink identity, and timestamp through a read-only context.
 - Supports single, composite, and no-op `Sink`s while preserving borrowed event emission.


crates/observed/src/sampling/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+/// sampler instance may be attached to multiple Sinks; use
+/// [`EventContext::sink_id`] to distinguish them.
+pub trait EventSampler: Send + Sync + 'static {
+    /// Decides whether `event` continues to this Sink's processors.
+    ///
+    /// Called exactly once for every event the attached Sink is interested in.
+    /// A Sink is interested when at least one of its processors reports
+    /// interest. [`EventSamplingDecision::Continue`] runs normal processing;
+    /// [`EventSamplingDecision::Drop`] runs no processor for that Sink.
+    ///
```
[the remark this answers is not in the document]
evgenyfedorov2: no, this is too much of implementation detail

crates/observed/src/sink/core.rs
```diff
@@ -446,20 +515,34 @@ struct SingleSinkState {
...
-    /// hands it to every interested processor.
+    /// Offers the event context to this leaf's [`EventSampler`], then builds an
+    /// [`EventView`] rooted at this leaf's enrichment slot and hands it to each
+    /// interested processor unless the sampler dropped it.
     fn dispatch(&self, event: &dyn DynEvent, description: &EventDescription) {
         // Reading the leaf's clock keeps timestamps off `SystemTime::now()`,
         // so frozen clocks make this Miri-safe.
         let timestamp = self.clock.system_time();
+
+        if let Some(sampler) = &self.sampler
```
psandana, line 541: 🤖 Pato's Pull Request Agent: The sampler gate runs *after* the event is fully constructed — `emit_impl` calls `state.evaluate()` (running the user closure and every field initializer) before `dispatch` consults the sampler. Since `EventContext` only exposes data available pre-construction (`description`, `sink_id`, `timestamp`), a `Drop` never avoids construction cost — it only skips downstream processors. For the flagship "drop most events" case (e.g. the `DropHealthChecks` example, which decides purely on `description().name()`), every dropped event still pays full construction. Is post-construction placement intentional (e.g. to expose event fields to samplers later)? If so, worth saying; if not, a single-leaf fast path could sample before `evaluate()`. Either way, DESIGN.md notes the ordering but the public docs (`with_event_sampler`, `EventSampler::sample`, FEATURES.md) don't carry the cost caveat.

crates/observed/src/sampling/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+//! ```
+
+use std::time::SystemTime;
+
+use crate::SinkId;
+use crate::metadata::EventDescription;
+
+/// Read-only inputs for one [`EventSampler`] decision.
+#[derive(Debug)]
+pub struct EventContext<'a> {
```
psandana: 🤖 Pato's Pull Request Agent: `observed` already has a `context` module (`observed::context`, cross-thread enrichment `Transfer`), so `sampling::EventContext` overloads "context" with a second, unrelated meaning in the same crate. Since this is a public type name — cheap to change now, breaking later — consider something sampling-specific like `SamplingContext` or `SampleInputs`.

crates/observed/src/sampling/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+    #[must_use]
+    pub const fn sink_id(&self) -> SinkId {
+        self.sink_id
+    }
+
+    /// Returns the timestamp this Sink assigned to the event.
+    #[must_use]
+    pub fn timestamp(&self) -> SystemTime {
+        self.timestamp
+    }
```
psandana, line 97: 🤖 Pato's Pull Request Agent: `timestamp()` is the only `EventContext` accessor that isn't `const fn` — `description()`, `sink_id()`, and `new()` all are. `SystemTime` is `Copy`, so this compiles as `const` and keeps the accessor set uniform.

```suggestion
    /// Returns the timestamp this Sink assigned to the event.
    #[must_use]
    pub const fn timestamp(&self) -> SystemTime {
        self.timestamp
    }
```

crates/observed/src/sampling/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+}
+
+/// What a leaf sink does with an event offered to its [`EventSampler`].
+#[expect(
+    clippy::exhaustive_enums,
+    reason = "event sampling is a predicate with exactly two states, not an extensible workflow"
+)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum EventSamplingDecision {
```
psandana, line 106: 🤖 Pato's Pull Request Agent: `EventSamplingDecision` being exhaustive is reasonable for the decision itself, but it also closes the door on a future weight-carrying outcome (e.g. `Continue { weight }`) that production samplers often need so downstream can un-bias counts of retained events — that'd be a breaking change here. If downstream reweighting (or sampler self-metrics) is a plausible future, consider reserving room now (`#[non_exhaustive]`, or a small struct); if it's an explicit non-goal, a one-line doc note would set expectations. (Noting you deliberately marked it exhaustive, so feel free to wave off.)

psandana:
🤖 Pato's Pull Request Agent: The routing flowchart in DESIGN.md (the `:::mermaid` diagram: emit → Interest pass → Collect dimensions → per-processor Redaction → exporters) wasn't updated for sampling — it has no `EventSampler` node, though the surrounding prose now describes the gate. Consider inserting a sampler decision node between "Interest pass" and "Collect dimensions" (Drop ⇒ stop the leaf, Continue ⇒ proceed) so the diagram matches the code.

crates/observed/src/sampling/mod.rs
evgenyfedorov2: - It could be `EventSamplingContext` but it would be 3 words, wdyt?
- It cannot be SamplingContext - it is ambiguous  because I am going to introduce (late) log sampling which will have some context too, so it is important to distinguish event context vs log sampling context

crates/observed/src/sampling/mod.rs
evgenyfedorov2: We can afford breaking changes here, it is a brand new feature with 0 users.

crates/observed/src/sampling/mod.rs
evgenyfedorov2: Done

crates/observed/src/sink/core.rs
evgenyfedorov2: > a Drop never avoids construction cost — it only skips downstream processors

not sure what that means, the Drop immediately stops any further event processing per each Sink. 

>  but the public docs (with_event_sampler, EventSampler::sample, FEATURES.md) don't carry the cost caveat.

Public benchmarks will show the cost (will add later). It would be too much implementation detail to mention in the public docs

evgenyfedorov2:
> 🤖 Pato's Pull Request Agent: The routing flowchart in DESIGN.md (the `:::mermaid` diagram: emit → Interest pass → Collect dimensions → per-processor Redaction → exporters) wasn't updated for sampling — it has no `EventSampler` node, though the surrounding prose now describes the gate. Consider inserting a sampler decision node between "Interest pass" and "Collect dimensions" (Drop ⇒ stop the leaf, Continue ⇒ proceed) so the diagram matches the code.

Fixed, thank you!

crates/observed/src/sampling/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+}
+
+impl<'a> EventContext<'a> {
+    pub(crate) const fn new(description: &'a EventDescription, sink_id: SinkId, timestamp: SystemTime) -> Self {
+        Self {
+            description,
+            sink_id,
+            timestamp,
+        }
+    }
```
martintmk: [AI AGENT]: non-blocking: `EventSampler` is an unsealed extension point, but `EventContext` - its only input - has no public constructor, so a downstream implementor cannot unit-test their own sampler. To observe a single decision they must stand up a `Sink`, register a processor that reports interest, and emit an event through it. The crate's own tests don't feel this, because `pub(crate)` already reaches them.

All three inputs are public already (`EventDescription::new` and `SinkId::new` are both `pub const fn`, `SystemTime` is std), so the restriction reads as incidental rather than deliberate. Keeping the public constructor behind `test-util` also avoids freezing the parameter list into the semver surface, so you keep the freedom to add inputs later.

Verified: an external test in `observed_testing` calling `EventContext::new(&description, SinkId::new("svc"), SystemTime::UNIX_EPOCH)` fails to compile with `error[E0624]: associated function 'new' is private`. The suggestion below mirrors how `Sink::current_enrichments` is already gated; I checked it builds clean both with and without `--features test-util`.

```suggestion
    pub(crate) const fn new(description: &'a EventDescription, sink_id: SinkId, timestamp: SystemTime) -> Self {
        Self {
            description,
            sink_id,
            timestamp,
        }
    }

    /// Creates a context for unit-testing an [`EventSampler`] implementation
    /// without emitting through a [`Sink`](crate::Sink).
    ///
    /// This function is gated behind the `test-util` feature and is intended
    /// for testing and diagnostics only.
    #[cfg(any(test, feature = "test-util"))]
    #[must_use]
    pub const fn for_test(description: &'a EventDescription, sink_id: SinkId, timestamp: SystemTime) -> Self {
        Self::new(description, sink_id, timestamp)
    }
```

crates/observed_testing/tests/event_sampling.rs
```diff
@@ -0,0 +1,500 @@
...
+        }
+    );
+
+    assert_eq!(calls.load(Ordering::Relaxed), 1);
+    assert!(logs.is_empty());
+    assert!(metrics.is_empty());
+}
+
+#[test]
+fn event_sampler_is_not_called_without_processor_interest() {
```
martintmk, line 173: [AI AGENT]: non-blocking: `with_event_sampler` documents "A sink that is not interested in an event never calls the sampler", and nothing pins that for a composite leaf. This test covers the single-sink case, but there `emit_impl`'s own `is_interested_in` gate (`sink/core.rs` L392) already returns before dispatch - so the per-leaf gate in `dispatch_to_processors` (`sink/core.rs` L433) is the real enforcement point and is never exercised.

That gate is easy to lose in a later refactor precisely because dropping it changes nothing else: the per-processor loop re-checks interest, so delivery stays correct and only the sampler is over-called. For a stateful sampler - a rate limiter, a counter, a reservoir - that is a silent behavioral bug. Tooling can't cover it here either: `pr-mutants` is diff-scoped (`--in-diff diff.txt`) and `dispatch_to_processors` is unchanged by this PR, so it is out of scope; nightly `Mutate All` is unscoped and would surface it only after merge.

Verified: deleting `if child.is_interested(description)` from `dispatch_to_processors` leaves `cargo test -p observed -p observed_testing` fully green - 271 tests, 0 failed, including all 15 in this file. Adding the test below makes that mutant fail (`left: 2, right: 1`); with the gate restored it passes, taking this file to 16. Please add it to `crates/observed_testing/tests/event_sampling.rs`:

```rust
/// [`Sink::with_event_sampler`] documents that a sink which is not interested
/// in an event never calls the sampler. Pin that for a composite leaf whose
/// sibling *is* interested - the only path that exercises the per-leaf gate.
#[test]
fn composite_does_not_sample_an_uninterested_leaf() {
    let interested_id = SinkId::new("interested");
    let uninterested_id = SinkId::new("uninterested");
    let interested = MockProcessor::new();
    let uninterested = MockProcessor::with_filter(|_| false);
    let (sampler, calls) = probe_sampler(move |event| {
        assert_eq!(event.sink_id(), interested_id);
        EventSamplingDecision::Continue
    });

    let composite = Sink::composite([
        Sink::new(
            interested_id,
            vec![Arc::new(interested.clone())],
            tick::SimpleClock::new_frozen(),
        ),
        Sink::new(
            uninterested_id,
            vec![Arc::new(uninterested.clone())],
            tick::SimpleClock::new_frozen(),
        ),
    ])
    .with_event_sampler(sampler);

    emit!(
        composite,
        UserAction {
            user: PiiString("Alice".into()),
            action_code: 3,
        }
    );

    assert_eq!(calls.load(Ordering::Relaxed), 1);
    assert_eq!(interested.len(), 1);
    assert!(uninterested.is_empty());
}
```

crates/observed/src/sink/core.rs
```diff
@@ -201,6 +203,73 @@ impl Sink {
         }
     }
 
+    /// Attaches one [`EventSampler`] to this sink, replacing any sampler
+    /// already attached.
+    ///
+    /// For a composite, the same sampler is attached to every leaf. A
+    /// [`Sink::noop`] value is returned unchanged. A sink that is not
+    /// interested in an event never calls the sampler.
```
martintmk: [AI AGENT]: nit: worth stating that composite attachment *discards* samplers already attached to the leaves. `composite_attachment_applies_one_sampler_to_every_interested_leaf` pins exactly that (it asserts `replaced_calls == 0`), but from the prose alone "replacing any sampler already attached" reads as being about this sink, and "the same sampler is attached to every leaf" reads as additive rather than destructive.

The consequence is quiet: someone composing pre-configured sinks that each carry their own sampling policy loses all of them, and the symptom is a change in telemetry volume rather than an error.

```suggestion
    /// For a composite, the same sampler is attached to every leaf,
    /// discarding any sampler those leaves already had. A
    /// [`Sink::noop`] value is returned unchanged. A sink that is not
    /// interested in an event never calls the sampler.
```

martintmk (approved):
[AI AGENT]: Reviewing as an AI agent, with most attention on the new public sampling contract. I took this PR's green CI as the baseline rather than re-running it, and verified the two substantive findings below locally against `c8062d44` with targeted probes.

**Public surface reviewed:** `observed::sampling` (`EventContext`, `EventSamplingDecision`, `EventSampler`) and `Sink::with_event_sampler`. One finding, on `EventContext`'s constructor. The rest of the surface holds up: `Arc<dyn EventSampler>` matches the crate's existing `Sink::new(.., Vec<Arc<dyn EventProcessor>>, ..)` family and is what makes one sampler shareable across leaves; `EventContext`'s private fields leave room to add inputs without a break; and an unsealed single-method trait is the right shape for an intended extension point. I did not re-raise the `EventContext` name or `#[non_exhaustive]` points already settled in earlier threads.

**Correctness - no defects found.** I traced every changed dispatch path. Two things worth recording as sound rather than assumed: the sampler runs *inside* the reentrancy guard taken in `emit_impl` (`sink/core.rs` L403), so a sampler that emits telemetry is dropped rather than recursing, which is what makes the "not supported" wording safe rather than merely discouraged; and `with_event_sampler` clones `SingleSinkState`, whose `Slot` is `Arc<ThreadLocal<..>>`, so leaf identity and the shared enrichment slot genuinely survive attachment exactly as the doc claims.

**Tests:** one gap, commented inline. Otherwise the suite is behaviour-shaped and covers the semantics I would want pinned - interest gating, composite independence, clone-before/clone-after, replacement, borrowed and `DynEvent` emission, and the noop/empty/processorless corners.

**Performance, telemetry, resilience, dependencies:** no findings. The added hot-path cost is one `Option` check plus one virtual call; the clock read that precedes it is contract, not waste, since `EventContext::timestamp` exposes it. No new dependencies, no new features, no telemetry emitted.

**Verdict: approve with non-blocking comments.** Nothing here blocks the merge. The composite test is the one item I would like to see land with the feature, since nothing else in CI can catch that regression.

sgalkin (approved):
🤖 Co-authored with Copilot.

crates/observed/src/sink/core.rs
evgenyfedorov2: Done

crates/observed_testing/tests/event_sampling.rs
evgenyfedorov2: Done

crates/observed/src/sampling/mod.rs
evgenyfedorov2: Done

crates/observed/src/sampling/mod.rs
evgenyfedorov2: Renamed to EventSamplingContext for now

martintmk (approved):

PR #726, opened 2026-09-03
fix(observed_macros): resolve renamed dependencies and reject mutable-reference fields

Vaiz:

Two independent hygiene fixes in the `observed` procedural macros, both in `crates/observed_macros_impl`. One commit each, so they can be reviewed separately.

## Resolve a renamed `observed` dependency in generated paths (AB#7757619)

Both generators emitted a hard-coded `::observed` into every path they produce. `extern crate self as observed` in the runtime crate only helps expansions inside that crate, so a consumer that renames its dependency:

```toml
telemetry = { package = "observed", version = "0.24" }
```

had no `observed` in scope and failed to compile for both `#[event(...)]` and `#[derive(Enrichment)]`.

The runtime crate is now resolved once through `proc-macro-crate` and the resolved path is threaded through event generation, enrichment generation and the `where`-clause predicates the two share. The new `resolver` module follows the local pattern in `routerama_build::macro_impl::resolver`, including its handling of `FoundCrate::Itself`.

Behaviour is unchanged for every crate that does not rename the dependency: `FoundCrate::Itself` (the runtime crate expanding its own events, which reaches itself via `extern crate self as observed`) and an unreadable manifest both keep `::observed`. All 32 pre-existing expansion snapshots pass untouched, which is the evidence for that.

Coverage: a new expansion snapshot drives both entry points against a renamed path and pins every emitted path — the trait, the descriptors, `Value`, the visitor function type, the `__private` items and the spelled-out bounds — so a single hard-coded `::observed` left anywhere would show up in it. Unit tests cover the path derivation itself (rename, dash-containing alias, keyword alias needing a raw identifier, `Itself`, unresolvable). Hidden `*_with_runtime_path` entry points exist only so the tests can vary a path the production entry points read from the caller's manifest.

## Reject mutable-reference event fields (AB#7757620)

`is_reference_type` matched any `syn::Type::Reference` and never inspected `TypeReference::mutability`, so `#[event(...)]` accepted a `&mut T` field and generated a visit body for it. An event is visited through `&self`, so that body cannot hand out the exclusive borrow it asks for — the input was rejected only later, by the compiler, against generated code the author never wrote.

Mutability is now checked during field parsing and reported against the offending field type, saying that event fields support only shared references. `Option<&mut T>` is checked too, because the visit body dereferences the inner type rather than the field type, and the check recurses through nested references so `&&mut T` is caught as well. A companion snapshot proves shared references (`&'a T`, `&'a &'a T`, `Option<&'a T>`) still expand exactly as before.

## Deliberately not fixed here

`AB#7757617` ("Expose fields of signal-less events to name-based processors") lives in the same file — `field_reaches_signal` in `src/event.rs` — but needs a design decision about what the contract for signal-less fields should be. It is left untouched, and the `expansion__an_event_with_no_signal_at_all_visits_nothing` snapshot is unchanged.

## Validation

Run for `observed`, `observed_macros`, `observed_macros_impl` and `observed_testing`:

| Check | Result |
| --- | --- |
| `cargo +nightly-2026-05-30 fmt --check` | pass |
| `cargo +1.96.1 clippy --all-targets` (`-D warnings`) | pass |
| `cargo +1.96.1 test` | pass (all suites, incl. doctests) |

No pre-existing snapshot was re-blessed; the only new `.snap` files are the three added by the new tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3381,6 +3381,7 @@ name = "observed_macros_impl"
 version = "0.24.0"
 dependencies = [
  "insta",
+ "proc-macro-crate 3.5.0",
  "proc-macro2",
  "quote",
  "syn 3.0.3",
--- a/crates/observed_macros_impl/CHANGELOG.md
+++ b/crates/observed_macros_impl/CHANGELOG.md
@@ -20,3 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `proc-macro-crate` instead of emitting a hard-coded `::observed`. A crate that
   renames its dependency (`telemetry = { package = "observed", ... }`) can now
   use `#[event(...)]` and `#[derive(Enrichment)]`.
+- `#[event(...)]` now rejects a field holding a mutable reference (`&mut T` or
+  `Option<&mut T>`) while parsing, naming the offending field, rather than
+  accepting it and failing later inside the generated code. Event fields are read
+  through `&self` when the event is visited, so only shared references work.
--- a/crates/observed_macros_impl/CHANGELOG.md
+++ b/crates/observed_macros_impl/CHANGELOG.md
@@ -13,3 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `observed` procedural macros (`#[event(...)]` and `#[derive(Enrichment)]`).
   `observed_macros` is now a thin `proc-macro` shim that delegates here. Use the
   re-exports from `observed` rather than depending on this crate directly.
+
+### Fixed
+
+- Both generators now resolve the `observed` runtime crate through
+  `proc-macro-crate` instead of emitting a hard-coded `::observed`. A crate that
+  renames its dependency (`telemetry = { package = "observed", ... }`) can now
+  use `#[event(...)]` and `#[derive(Enrichment)]`.
--- a/crates/observed_macros_impl/Cargo.toml
+++ b/crates/observed_macros_impl/Cargo.toml
@@ -27,6 +27,7 @@ allowed_external_types = ["proc_macro2::*", "syn::*"]
 all-features = true
 
 [dependencies]
+proc-macro-crate = { workspace = true }
 proc-macro2 = { workspace = true }
 quote = { workspace = true }
 syn = { workspace = true, features = [
--- a/crates/observed_macros_impl/src/enrichment.rs
+++ b/crates/observed_macros_impl/src/enrichment.rs
@@ -124,13 +124,13 @@ fn parse_enrichment_field_def(field: &Field) -> Result<EnrichmentFieldDef> {
 // ================================================================================================
 
 /// Entry point: generates the `Enrichment` trait implementation from `DeriveInput`.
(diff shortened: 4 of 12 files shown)


Vaiz:
Closing this PR — not because anything here is wrong, but because the approach it takes is now a team decision rather than a settled one.

## Why

`AB#7757619` (resolve renamed `observed` dependencies) has been moved to **Blocked** pending a team decision on the mechanism. This PR implements one of three viable options, and it happens to be the one the ecosystem uses least.

A survey of the ten most-downloaded proc-macro crates (crates.io 90-day downloads, mechanism read from each crate's source, 2026-09-04):

| Approach | Count | Examples |
| --- | --- | --- |
| Manual `crate = …` attribute | 5 of 10 | `serde_derive`, `tokio-macros`, `zerocopy-derive`, `strum_macros`, `darling` |
| Hard-coded absolute path, no escape hatch | 3 of 10 | `thiserror-impl`, `tracing-attributes`, `clap_derive` |
| `$crate` shim through a `macro_rules!` wrapper | 1 of 10 | `futures-macro` |
| No runtime dependency at all | 1 of 10 | `async-trait` |
| **`proc-macro-crate`** | **0 of 10** | — |

Across the whole local cargo registry, 441 `proc-macro = true` crate versions contain exactly four callers of `proc_macro_crate`. The `$crate` shim is the strongest guarantee but is only available to function-like macros invoked through a `macro_rules!` wrapper, so it is not an option for an attribute macro and a derive.

`proc-macro-crate` is still defensible — it is what `routerama_build::macro_impl::resolver` already does in this repo, and it needs no action from the consumer. But it reads `CARGO_MANIFEST_DIR` and parses the manifest, so it degrades under non-Cargo build systems (Buck, Bazel) and cannot see a rename expressed as `use foo as observed`. In both cases it falls back to `::observed` — today's behaviour, so no regression, but a silent non-fix. That tradeoff is a contract decision for `observed`, not something this PR should settle on its own.

The three options are written up in full on the work item.

## Nothing is lost

The branch `u/eshutov/observed-macro-hygiene` stays pushed. Both commits are intact and validated (`fmt --check`, `clippy -D warnings`, and the full test suite all pass for `observed`, `observed_macros`, `observed_macros_impl` and `observed_testing`; no pre-existing snapshot was re-blessed). Whichever option the team picks, the path is already threaded through both generators and the shared `where`-clause predicates, so switching to the attribute form is a small follow-up rather than a rewrite.

## Note on the second commit

`AB#7757620` (reject mutable-reference event fields) is the second commit here and is **completely independent** of the approach question — it adds a mutability check during field parsing so `&mut T` and `Option<&mut T>` are rejected with a clear diagnostic instead of failing later inside generated code. If that fix is wanted before the resolver question is settled, say so and it can go up as its own PR off the same branch.

Vaiz:
Follow-up: the independent half of this PR is now open on its own as https://github.com/microsoft/oxidizer/pull/730 — `AB#7757620`, reject mutable-reference event fields. It is rebuilt on top of current `main` rather than cherry-picked, carries no `proc-macro-crate` dependency and no `Cargo.lock` change, and both of its snapshots are byte-identical to the ones reviewed here.

This PR stays closed. Branch `u/eshutov/observed-macro-hygiene` is still pushed and holds the resolver work for whenever `AB#7757619` is unblocked.

PR #728, opened 2026-09-04
ci(anvil): update to cargo-anvil 0.7.0

Vaiz:

Fixes AB#7802687. Replaces #718, which adopted cargo-anvil 0.6.0 and is superseded by this one.

## Problem

`anvil-semver-check` has been non-functional repo-wide. Every library crate aborts with:

```
error: unsupported rustdoc format v60 for file: .../target/doc/anyspawn_azure.json
(supported formats are v55, v56, v57)
```

The recipe installs the pin from `justfiles/anvil/versions.just`, which was `cargo-semver-checks` 0.46.0 (rustdoc JSON v55/v56/v57). The recipe does not pin a toolchain for doc building, so it uses the runner default stable — now Rust 1.98, whose rustdoc emits **v60**. The tool aborts while parsing the JSON it just produced.

Because the recipe is advisory, the crash was never surfaced as a failure. Each tool error was collected as a *finding* and posted as a sticky ":warning: Potential breaking changes detected" comment listing ~21 crates. Net effect: SemVer checking has been off across the workspace since the runner rolled to 1.98, and every PR carried a misleading breaking-change warning.

#709 fixed this by editing the pin directly and was closed unmerged, because `versions.just` is anvil-generated and a repository-local edit stops it receiving automatic updates. This PR takes the sanctioned route instead.

## Change

Regenerate the anvil-managed tree with `cargo anvil` from **cargo-anvil 0.7.0** (2026-09-03; previously 0.3.0, #596, 2026-07-24). The 0.7.0 catalog pins `cargo-semver-checks` **0.50.0**, which reads rustdoc JSON v57, v60 and v61.

A second goal of this PR is to **shrink the repository-specific surface** and let the anvil templates own as much of the tree as they can. Three artifacts that were customized before this PR are now the template verbatim; see "Custom configuration retired" below.

## Why 0.7.0 rather than 0.6.0

#718 adopted 0.6.0 and hit two problems that had to be worked around or left open. **0.7.0 fixes both upstream**, which is why that PR is closed in favour of this one.

| Problem on #718 (0.6.0) | State on 0.7.0 |
| --- | --- |
| `cargo anvil` appended an `anvil-lints` region without noticing an existing *unmanaged* `[lints]` table, producing **two** `[lints]` tables in 20 crates and a workspace cargo rejects outright. #718 had to delete the pre-existing block by hand in every one. | Fixed. The generator adopts the existing table into the managed region. The diff in those 20 crates is now just the `# >>> anvil-managed` fences (and, where the table sat mid-file, a move to the end). No manual edit. |
| The new `pr-msrv` group ran `cargo test --all-targets`, and `--all-targets` implies `--benches`, so `cargo test` **executed** the Callgrind bench harnesses. Those are `harness = false` and hand off to `gungraun-runner`, which the group never installs — `anvil-msrv-test` failed on linux and linux-arm. | Fixed. `msrv-test.just` now uses `--tests`, with a comment stating exactly this reason. Benches/examples stay compile-only under `anvil-bench` / `anvil-examples`. |

## Compatibility

rustdoc JSON `format_version` against the toolchains in play:

| toolchain | rustdoc JSON | read by 0.50.0 |
| --- | --- | --- |
| 1.95 (`RUST_MSRV`) | v57 | yes |
| 1.96.1 (`RUST_LATEST`) | v57 | yes |
| 1.98 (runner stable) | v60 | yes |
| nightly-2026-05-30 | v57 | yes |

0.50.0 drops v55 and v56, both of which correspond to a rustc below the MSRV. Its install MSRV is Rust 1.93, below this repo's 1.95.

## What else the 0.7.0 catalog brings

* **Impact analysis moves to an artifact, and scoping moves into the recipe tree.** `ANVIL_INCLUDE_MODIFIED` / `ANVIL_INCLUDE_AFFECTED` / `ANVIL_INCLUDE_REQUIRED` are gone as job outputs threaded into every group. The impact jobs upload an `anvil-impact-<os>` artifact; each group job downloads it into `target/anvil/impact`, and every scoped check reads its own scope through the new `justfiles/anvil/impact.just` helper `_anvil-impact-include <tier>` — the same cache a local run reads.
* **One generic group runner.** The eight per-group composite actions (`anvil-pr-fast`, `anvil-pr-test`, …) are replaced by a single `anvil-run-group` plus `anvil-report-status`, and `anvil-setup` gains a `just-problem-matcher.json` so `just` errors annotate the diff.
* **New MSRV check and group:** `justfiles/anvil/checks/msrv-test.just` plus a `pr-msrv` group, wired into `anvil-pr-impl.yml` across the four legs.
* **Container artifacts.** A generated `.anvil/container/Dockerfile` + `.dockerignore` and `justfiles/anvil/container.just`, imported with `import?` so the documented `without_artifact` opt-out cannot break Justfile parsing.
* **Stable-toolchain resolution.** `versions.just` gains `_anvil_stable_toolchain_args`, a bootstrap that resolves the repository toolchain file (or, absent one, the root MSRV) before cargo is available to parse the manifest.
* **Tool pins bumped:** cargo-aprz 1.0.0 → 1.1.0, cargo-careful 0.4.9 → 0.4.10, cargo-deny 0.19.0 → 0.20.2, cargo-doc2readme 0.6.4 → 0.7.3, cargo-hack 0.6.41 → 0.6.45, cargo-llvm-cov 0.8.4 → 0.9.0, cargo-mutants 26.1.2 → 27.1.0, cargo-nextest 0.9.122 → 0.9.143, cargo-semver-checks 0.46.0 → 0.50.0, cargo-sort 2.0.2 → 2.1.4, cargo-spellcheck 0.15.1 → 0.15.7, cargo-udeps 0.1.60 → 0.1.61, plus a new `cargo-each` pin.
* `.gitattributes` gains `*.sh text eol=lf`.

## Custom configuration retired

cargo-anvil refuses to overwrite an artifact an adopter has edited; it writes a `.anvil-proposed` sidecar instead (not committed). Eight artifacts were in that state after regeneration. **Three are now the template verbatim**, so anvil owns them again and they will pick up future catalog updates automatically:

* **`justfiles/anvil/checks/bolero.just`.** The template discovers targets per package rather than in one call, raises a discovery failure as an error instead of swallowing it as "no targets", and builds with `--profile release`. Strictly better than the repository copy.
* **`deny.toml [anvil-deny-bans]`.** `wildcards = "warn"` → `wildcards = "deny"` with `allow-wildcard-paths = true`. The adopter comment justified `warn` by "internal workspace crates use wildcard version requirements" — those are path dependencies, which `allow-wildcard-paths` permits, so the stricter setting costs nothing and now catches a registry wildcard. `just anvil-deny` reports `advisories ok, bans ok, licenses ok, sources ok`.
* **`.github/skills/code-review/SKILL.md`.** The template is a superset of the repository copy: same rules, plus the evidence rule and seven worked failure cases.

`CARGO_DENY_VERSION` in `constants.env` follows the anvil pin to **0.20.2** so both consumers of the now-generated `deny.toml` use the same tool.

That leaves **five** customized artifacts, down from eight. (`Cargo.toml [anvil-workspace-lints]`, `deny.toml [anvil-deny-advisories]` / `[anvil-deny-licenses]` and `clippy.toml [anvil-clippy]` also appear as "leave alone" in the dry run, but those are adopter-owned by design — anvil never overwrites them.)

## Custom configuration retained, with reasons

**`justfiles/anvil/checks/fmt.just` — the template silently disables the check in this repository.** The proposed recipe runs `cargo {{_anvil_stable_toolchain_args}} each --workspace -- cargo +<nightly> fmt --manifest-path '{manifest}' --check`. This repository has **no `rust-toolchain.toml`**, so the bootstrap falls through to the root MSRV and the *outer* `cargo +1.95` exports `RUSTUP_TOOLCHAIN=1.95` into the *inner* nightly `cargo fmt`. Measured locally: 2 300 lines of

```
Warning: can't set `imports_granularity = Module`, unstable features are only available in nightly channel.
Warning: can't set `group_imports = StdExternalCrate`, unstable features are only available in nightly channel.
```

and **exit 0** — rustfmt runs on stable, ignores every unstable option in `rustfmt.toml`, and the check passes without enforcing what it exists to enforce. Running the same `cargo each … -- cargo +<nightly> fmt` line by hand, without the outer toolchain argument, is silent. The repository recipe (`cargo +<nightly> fmt --package $p --check` per package) is kept, rewired to the new impact helper. Worth an upstream issue against `microsoft/ox-tools`.

**The four `miri` recipes.** Adopting the templates would drop two deliberate things:

* the parallel artifact runner from #706 — compile once, then run the Miri test artifacts across all cores. Miri is on the PR critical path;
* `[package.metadata.anvil.miri] exclude = true`, honoured by `_anvil-miri-test` and set by 8 `*_macros_impl` crates. The template ignores the key, so Miri would start running on all of them. The code-review skill adopted *in this PR* describes that list as "a Miri exclusion list that intentionally covers every `*_macros_impl` crate as a cost policy".

They are kept as-is and rewired to the new impact helper (see below).

**`.delta.toml`.** cargo-anvil reports it left the managed region empty because the repository defines top-level `trip_wire_patterns`. Adopting the managed region would *narrow* impact scoping: the managed list is `Cargo.lock` / `Cargo.toml` / `rust-toolchain.toml`, while the repository list also trips on `.github/*`, `*.just`, `deny.toml`, `scripts/*` and more — including the paths this very PR touches. Not equivalent, so the repository key stays.

## Other manual changes regeneration cannot make

**`constants.env` follows three anvil pins.** The file is not anvil-managed, and these tools check the *same checked-in artifacts* the anvil recipes check, so they cannot disagree:

* `CARGO_SEMVER_CHECKS_VERSION` 0.48.0 → **0.50.0**, so the `main.yml` `semver` job and `scripts/release-packages.ps1` install what the recipes install. That path pins `RUST_LATEST=1.96.1`, which emits v57, so it was never broken — but 0.48.0 reads only v55/v56/v57, so a developer running `release-packages.ps1` on current stable would hit the identical error locally.
* `CARGO_DOC2README_VERSION` 0.7.2 → **0.7.3**. 0.7.3 builds docs.rs URLs from a dependency's **package** name rather than its **lib** name. `main.yml`'s `static-analysis` job and `anvil-readme-check` both check the same READMEs, and only one generator can be right for a single file.
* `CARGO_DENY_VERSION` 0.19.8 → **0.20.2**, as above.

**READMEs regenerated** with the pinned 0.7.3 and the workspace `crates/README.j2` template. The diff is limited to the embedded dependency-info blob and, in `seatbelt`, `http_extensions` and `rest_over_grpc`, the docs.rs URLs of the three dependencies whose package and lib names differ (`tower-service`, `http-body`).

**The five retained recipes rewired to the new impact helper.** Left untouched they would have read the now-removed `ANVIL_INCLUDE_*` variables, silently fallen back to `--workspace`, and `anvil-fmt` would never have seen its `--skip` short-circuit. Each now takes an `anvil-impact` dependency and calls `_anvil-impact-include`, and each moves to `[script("pwsh", "-NoProfile")]` so all script recipes in the tree are consistent and none load the invoking user's PowerShell profile.

`constants.env` and `versions.just` still disagree on cargo-llvm-cov (0.8.7 vs 0.9.0), cargo-mutants (27.0.0 vs 27.1.0), cargo-nextest (0.9.137 vs 0.9.143) and cargo-spellcheck (0.15.1 vs 0.15.7). None of those pairs of consumers check a shared checked-in artifact, so they are left alone here rather than bumped speculatively. **cargo-spellcheck is the one to watch:** the 0.3.0 `versions.just` pinned 0.15.1 with a comment calling 0.15.7 an em-dash regression, and the 0.7.0 catalog now pins 0.15.7 while `constants.env` still says 0.15.1 — if `anvil-spellcheck` reports em-dash findings that `main.yml` does not, that is why.

## Verification

* `cargo anvil --dry-run` after the change reports **136 unchanged items and zero pending writes** — the tree is fully in sync with the 0.7.0 catalog, and the customized surface is down from 12 reported items to 9 (of which 4 are adopter-owned by design).
* `cargo metadata --no-deps --locked` succeeds, so the 20 regenerated `[lints]` regions parse.
* `just --list` parses the regenerated recipe tree (265 recipes).
* `just anvil-fmt` passes and is silent, exercising the new `anvil-impact` → `_anvil-impact-include` path end to end on a retained recipe.
* `just anvil-readme-check` passes over the whole workspace with the pinned 0.7.3.
* `just anvil-deny` passes with the adopted `[bans]` region and cargo-deny 0.20.2.
* No `.rs` files are touched, so formatting and lint output cannot change.

The workflow restructure — the impact artifact hand-off, the generic `anvil-run-group` action and the new `pr-msrv` group — can only really be validated by CI on this PR.

## Follow-up: `cachet_memory` tokio `macros` feature

`PR Job / Check Group: Runtime Analysis` failed on linux and linux-arm inside `anvil-bolero`, on `cargo test --target x86_64-unknown-linux-gnu --profile release --package cachet_memory --no-run`, with 30x `error[E0433]: cannot find 'test' in 'tokio'`.

This is a pre-existing latent manifest defect, not a regression from this upgrade. `crates/cachet_memory` uses `#[tokio::test]` 30 times but declared its tokio dev-dependency as `features = ["rt", "rt-multi-thread"]`. The workspace tokio is `default-features = false`, so `macros` was only ever supplied by workspace-wide feature unification from the 18 sibling crates that do declare it — every single-package build of the crate has been broken. It reproduces with no bolero involved at all: `cargo test --no-run -p cachet_memory` fails identically.

The 0.7.0 `anvil-bolero` template surfaces it because it discovers fuzz targets with a per-package `cargo bolero list --package <name>`. The previous repository recipe swallowed a discovery failure as "no targets" and exited 0, which is exactly the kind of silent pass the template removes.

Fix: declare the missing feature in `crates/cachet_memory/Cargo.toml`, one line, `Cargo.lock` unchanged. `cachet_memory` is the only crate in the workspace with this defect — every other crate using `#[tokio::test]` already declares `macros`. Re-customizing `bolero.just` to restore the swallow was rejected: it would leave the crate unbuildable standalone and re-add repository-specific surface this PR is removing.

## Behavioural change: `anvil-aprz` moves from the PR tier to the scheduled tier

Flagged in review and confirmed against `main`. Under cargo-anvil 0.3.0 the `anvil-pr-fast` group ended with `anvil-aprz`; under 0.7.0 it does not, and `anvil-aprz` now runs only in `anvil-scheduled-advisories`. Dependency feature and version-requirement analysis therefore moves from every pull request to the scheduled run.

This is an upstream template decision, not a repository choice. `justfiles/anvil/groups/pr-fast.just` and `justfiles/anvil/groups/scheduled-advisories.just` are both anvil-generated, and this PR takes the 0.7.0 templates verbatim. Restoring `anvil-aprz` to the PR group here would mean hand-editing a generated file, which stops automatic updates for that file and is the exact reason PR #709 was closed unmerged. If PR-tier aprz coverage is wanted, the fix belongs in cargo-anvil upstream so every adopter gets it.

Recording it here so the change is deliberate and visible rather than an unnoticed side effect of the version bump.

## Update: all shared `constants.env` pins now aligned

The PR originally bumped only three pins in `constants.env` — `CARGO_SEMVER_CHECKS_VERSION`, `CARGO_DOC2README_VERSION` and `CARGO_DENY_VERSION` — on the reasoning that those are the ones whose two consumers check the same committed artifact, so a mismatch there can produce a conflicting result.

Raised in review: that reasoning does not cover the other kind of divergence. `constants.env` feeds the pre-anvil `main.yml` workflow and `justfiles/anvil/versions.just` feeds the anvil recipes, so any drift means the two paths install different versions of the same tool, whatever they check. The remaining four are now aligned as well:

| Pin | Was | Now |
| --- | --- | --- |
| `CARGO_LLVM_COV_VERSION` | 0.8.7 | 0.9.0 |
| `CARGO_MUTANTS_VERSION` | 27.0.0 | 27.1.0 |
| `CARGO_NEXTEST_VERSION` | 0.9.137 | 0.9.143 |
| `CARGO_SPELLCHECK_VERSION` | 0.15.1 | 0.15.7 |

Every tool pinned in both files now carries the same version; the only entries left without a counterpart are ones that exist in just one file (`cargo-machete` and `cargo-workspaces` in `constants.env`; `cargo-aprz`, `cargo-audit`, `cargo-coverage-gate` and `cargo-each` in `versions.just`).

`constants.env` is not anvil-managed, so this is a repository edit and not a generated-file change. `cargo anvil --dry-run` still reports zero pending writes, and `just --list` still parses — the root justfile loads this file through `set dotenv-path`, so a malformed edit would break every recipe.

### Adopted: the generated `anvil-delta` trip-wire patterns

The repository-level `trip_wire_patterns` key in `.delta.toml` is removed, so `cargo anvil` now populates the managed `anvil-delta` region with the 0.7.0 template list.

This reverses an earlier decision in this PR to keep the repository list. That decision was made against the pre-0.7.0 `[delta]` region shape and was never revisited; re-checking it against the actual 0.7.0 template shows the generated list is a **strict superset** of the repository list. It keeps all 18 repository entries and adds `.pipelines/**`, `Justfile` and `justfiles/**`, and it uses recursive `**` where the repository used single-level `*` for `.ado`, `.cargo`, `.github` and `scripts`. That last difference is not cosmetic: the repository's `.github/*` does not match `.github/workflows/anvil-pr.yml`, while `.github/**` does — so the generated list scopes strictly more conservatively, not less, including on paths this PR touches.

Adopting it also removes one more repository-specific customization, which is the stated goal of this change. `cargo anvil --dry-run` reports zero pending writes afterwards.


.anvil/container/Dockerfile
```diff
@@ -0,0 +1,112 @@
...
+# `${HOME}/.netrc` would expand to `/.netrc` and leave the file root actually
+# uses. Both spellings are named.
+#
+# `registry` and `git` must exist before the `chmod`. The run mounts a named
+# volume over each, and an engine seeds a new volume from the image path it
+# covers; a path that does not exist seeds as root-owned 0755, which the
+# `--user` mapping cannot write, so the first cargo fetch fails with EACCES.
+WORKDIR /opt/anvil
+COPY justfiles ./justfiles
+COPY rust-toolchain.toml ./
```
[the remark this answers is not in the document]
Vaiz: must be fixed upstream

martintmk (approved):

constants.env
st-dev-gh, line 1: Could we update the remaining shared tool pins here to match the versions generated by cargo-anvil 0.7.0 in `justfiles/anvil/versions.just`? For reference:

- `CARGO_LLVM_COV_VERSION`: `0.8.7` → `0.9.0`
- `CARGO_MUTANTS_VERSION`: `27.0.0` → `27.1.0`
- `CARGO_NEXTEST_VERSION`: `0.9.137` → `0.9.143`
- `CARGO_SPELLCHECK_VERSION`: `0.15.1` → `0.15.7`

Keeping the duplicate pins aligned avoids the legacy workflows and anvil recipes exercising different tool versions.

justfiles/anvil/groups/pr-fast.just
```diff
@@ -23,8 +23,7 @@ anvil-pr-fast: anvil-pr-fast-validate-prereqs \
     anvil-audit \
     anvil-udeps \
     anvil-semver-check \
-    anvil-external-types \
-    anvil-aprz
+    anvil-external-types
```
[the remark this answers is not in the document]
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Confirmed, and thank you — I checked it against `main` rather than taking it on trust. On `main` the `anvil-pr-fast` group ends with `anvil-aprz`; on this branch it does not, and `anvil-aprz` appears only in `justfiles/anvil/groups/scheduled-advisories.just`. So it is a real behavioural change, not a rename or a relocation I had already accounted for.

I have taken your second option and documented it in the PR description, under "Behavioural change: `anvil-aprz` moves from the PR tier to the scheduled tier".

I have not taken the first option — restoring it here. Both `pr-fast.just` and `scheduled-advisories.just` are anvil-generated, and this PR deliberately takes the 0.7.0 templates verbatim; editing a generated file stops automatic updates for it, which is why PR #709 was closed unmerged. Group composition is an upstream decision, so if PR-tier aprz coverage is still wanted the change belongs in cargo-anvil, where every adopter gets it, rather than as a local divergence here.

constants.env
Vaiz: thanks! they should be in sync

.delta.toml
st-dev-gh, line 1: On `trip_wire_patterns` (line 23): could we adopt cargo-anvil 0.7.0’s generated list instead of retaining this repository copy? The upstream list is a superset: it additionally covers `.pipelines/**`, `Justfile`, and `justfiles/**`, and uses recursive `**` patterns for `.ado`, `.cargo`, `.github`, and `scripts`. Keeping the template-owned list would provide the broader coverage and allow future cargo-anvil updates to maintain it automatically.

constants.env
Vaiz: 🤖 **Clawpilot here!** Posted automatically by Clawpilot (an AI agent), not by a human. Please verify before acting.

Agreed and done in `b3fd5cc4`. All four are aligned exactly as you listed: `CARGO_LLVM_COV_VERSION` 0.8.7 → 0.9.0, `CARGO_MUTANTS_VERSION` 27.0.0 → 27.1.0, `CARGO_NEXTEST_VERSION` 0.9.137 → 0.9.143, `CARGO_SPELLCHECK_VERSION` 0.15.1 → 0.15.7.

Your reasoning is what changed my mind. The PR had bumped only the three pins whose two consumers check the same committed artifact, on the basis that a mismatch elsewhere cannot produce a conflicting result. That argument is about artifacts and simply does not address yours, which is about the two install paths exercising different tool versions — and that is true regardless of what they check.

I checked the whole table rather than just the four: every tool pinned in both `constants.env` and `justfiles/anvil/versions.just` now carries the same version. The only entries without a counterpart are ones that exist in a single file — `cargo-machete` and `cargo-workspaces` in `constants.env`, and `cargo-aprz`, `cargo-audit`, `cargo-coverage-gate` and `cargo-each` in `versions.just`.

`constants.env` is not anvil-managed, so this stays a repository edit rather than a generated-file change. `cargo anvil --dry-run` still reports zero pending writes, and `just --list` still parses — worth checking here because the root justfile loads this file through `set dotenv-path`, so a malformed edit would break every recipe rather than just the legacy workflow. The reasoning is recorded in the PR description too.

st-dev-gh (approved):

.delta.toml
Vaiz: You are right, and I was wrong to keep the repository list — thanks for pushing on it.

I checked the 0.7.0 template (`templates/regions/delta.toml`) against what is here. The generated list is a strict superset: it keeps all 18 repository entries and adds `.pipelines/**`, `Justfile` and `justfiles/**`, and it uses recursive `**` where the repository used single-level `*` for `.ado`, `.cargo`, `.github` and `scripts`. That last part is the one that actually bites: the repository's `.github/*` does not match `.github/workflows/anvil-pr.yml`, while `.github/**` does. So the repository list was scoping *less* conservatively than the template, on paths this very PR touches.

The earlier "keep it" call was made against the pre-0.7.0 region shape and never revisited after the upgrade. Adopted in `db4814be`: the repository-level `trip_wire_patterns` key is gone, `cargo anvil` populates the managed `anvil-delta` region, `cargo anvil --dry-run` is back to zero pending writes, and `cargo delta snapshot` parses the new config. Rationale recorded in the PR description.

martin-kolinek (approved):

sgalkin (approved):

PR #729, opened 2026-09-04
fix(rest_over_grpc): bound query field path nesting depth

Vaiz:

## Summary

A dotted query parameter key (`?a.b.c=1`) built one nesting level per `.`-separated segment, and both that build and the matching deserialization descended once per level. Neither had an upper bound, so the nesting depth was limited only by the input.

This caps the depth at 64 levels. A deeper field path is now rejected with `Code::InvalidArgument` instead of being decoded.

## Effects

- Query field paths nested more than 64 levels deep are rejected as an invalid request structure.
- No change for any realistic message shape; proto nesting stays far below the bound.
- No public API change.

## Validation

- `just anvil-fmt`, `just anvil-clippy`
- `cargo test` for `rest_over_grpc` and `rest_over_grpc_tests`
- Unit tests cover both sides of the bound, including through the public `decode_request` entry point.

## Design note

The 64-level cap is deliberately fixed: it is not configurable, not runtime-tunable, and not exposed in any public API, because a security bound a caller can raise does not bound anything. The cap is documented on `MAX_QUERY_FIELD_DEPTH` as a known limitation — field paths nested deeper than it are rejected with a clear error, never silently truncated and never allowed to overflow the stack. Making the depth configurable is out of scope for this fix.

The limit is documented for crate users in the crate-level `# Limitations` section of `lib.rs` (and the generated `README.md`), because `MAX_QUERY_FIELD_DEPTH` is private and `codegen_helpers` is `#[doc(hidden)]`, so a doc comment on either would never reach rustdoc.

### DIFF
--- a/crates/rest_over_grpc/src/transcode/overlay.rs
+++ b/crates/rest_over_grpc/src/transcode/overlay.rs
@@ -439,6 +439,11 @@ enum QueryNode {
 /// build and the matching deserialization descend once per level. The bound
 /// keeps an untrusted request from driving that descent past the stack, while
 /// staying far above the nesting any real proto message uses.
+///
+/// Known limitation: a query parameter whose field path nests deeper than this
+/// is rejected with `Code::InvalidArgument`, never truncated and never decoded.
+/// The cap is fixed and deliberately not configurable — a depth a caller could
+/// raise would not bound anything.
 const MAX_QUERY_FIELD_DEPTH: usize = 64;
 
 fn decode_tree<T: DeserializeOwned>(body: Option<&[u8]>, query: &[(&str, &str)]) -> Result<T, TranscodeError> {
--- a/crates/rest_over_grpc/src/transcode/overlay.rs
+++ b/crates/rest_over_grpc/src/transcode/overlay.rs
@@ -433,6 +433,14 @@ enum QueryNode {
     Leaf(SmallVec<[String; 2]>),
 }
 
+/// The deepest `.`-separated field path a query parameter may name.
+///
+/// A dotted query key builds one [`QueryNode`] level per segment, and both that
+/// build and the matching deserialization descend once per level. The bound
+/// keeps an untrusted request from driving that descent past the stack, while
+/// staying far above the nesting any real proto message uses.
+const MAX_QUERY_FIELD_DEPTH: usize = 64;
+
 fn decode_tree<T: DeserializeOwned>(body: Option<&[u8]>, query: &[(&str, &str)]) -> Result<T, TranscodeError> {
     let body = match body {
         Some(bytes) if !bytes.is_empty() => match from_slice::<BodyTop<'_>>(bytes) {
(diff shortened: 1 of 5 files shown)


wukchung:
I would expect the limitation get documented in README

sandersaares (approved):

Vaiz:
Thanks — both points are noted, and both are declined for this PR. Answering here because they arrived as suppressed comments in review 5112095065 rather than as threads.

**Empty path segments (`a..b`, `.a`)** — out of scope, and not a hole. This PR fixes one CVE: unbounded recursion over nested query field paths. Empty-segment handling is pre-existing behaviour in code this PR does not change (your own note marks it "previously missed — in code that hasn't changed since the last review"). It is also not exploitable: an empty segment inserts a field named `""`, which matches no proto field, so the request is already rejected by the deserializer as an unknown field. There is no recursion, no truncation and no panic on that path. Tightening it is a separate, non-security change and belongs in its own PR.

**`#[cfg_attr(coverage_nightly, coverage(off))]` on `mod tests`** — incorrect for this crate. `rest_over_grpc` does not use that attribute on any of its ~25 `#[cfg(test)] mod tests` blocks; it applies `coverage(off)` only to individually unreachable functions such as `expecting`/`visit_str` in this very file (overlay.rs:220, 870, 887, 948, 953, 1061). The coverage gate passes on the crate as it stands, so the new tests need no exemption, and adding one to a single module would make this file inconsistent with the rest of its own crate. `crates/tick` follows a different local style; that is not this crate's convention.

The depth cap itself remains fixed and deliberately not configurable, for the reason given under "Design note" in the PR description.

crates/rest_over_grpc/src/transcode/overlay.rs
```diff
@@ -449,16 +462,26 @@ fn decode_tree_entries<T: DeserializeOwned>(body: Vec<(String, &RawValue)>, quer
...
+        insert_query_node(&mut query_root, key.split('.'), value.into_owned(), MAX_QUERY_FIELD_DEPTH)?;
     }
     T::deserialize(TreeDeserializer { body, query: query_root }).map_err(TranscodeError::deserialize)
 }
 
+/// Inserts `value` at the `.`-separated field `path`, descending one level per
+/// segment.
+///
+/// `remaining_depth` is the number of levels still allowed; it starts at
+/// [`MAX_QUERY_FIELD_DEPTH`] and is spent one per segment so an untrusted key
```
Vaiz, line 474: This is not visible in public documentation

Vaiz: You are right — `MAX_QUERY_FIELD_DEPTH` is a private constant in a private module, and `codegen_helpers` is `#[doc(hidden)]`, so neither its doc comment nor `decode_request`'s reached rustdoc or the README. Fixed in 5389189c: the limit is now stated in the crate-level `# Limitations` section in `lib.rs`, next to the streaming and JSON-buffering limitations, and the README paragraph is regenerated to match. The constant keeps its doc comment for the implementation rationale, and stays private — the cap remains fixed and not configurable.

One note for whoever regenerates READMEs next: `just readme` rewrote all 40 crate READMEs on my machine because the locally installed `cargo-doc2readme` is 0.7.3 while CI uses the pinned 0.7.2, which churns the `__cargo_doc2readme_dependencies_info` blob and the `tower_service` link paths. I reverted that churn and hand-applied only the prose paragraph, so this diff touches `rest_over_grpc` alone.

Vaiz: Correction to my note above: the README churn was my own doing, not a 0.7.2-vs-0.7.3 mismatch. CI pins `cargo-doc2readme` to **=0.6.4** (`Installing cargo-doc2readme =0.6.4` in the pr-fast log), while my machine had 0.7.3 — the anvil prereq check only requires `>= 0.6.4`, so it accepted the newer one. Hand-applying the prose kept the stale input hash in the `__cargo_doc2readme_dependencies_info` blob, which is what `anvil-readme-check` compares, so CI answered "Input has changed" and `static-analysis` reported the same thing. Fixed in e66d2bd0: installed 0.6.4 and regenerated, which moves only that blob and leaves the prose and every `tower_service` link exactly as committed. `just anvil-readme-check` now passes locally.

geeknoid (approved):

PR #737, opened 2026-09-07
feat(arty_io_core): add injectable I/O driver contracts

martintmk:
## Summary

Introduces the initial stable `arty_io_core` contract for injecting I/O drivers after runtime startup.

- `DriverContext` selects its associated provider, so `get_context::<MyContext>()` needs no registration value.
- The first lookup initializes one driver instance on every active worker before returning the caller-local context.
- Later lookups reuse the cached context without creating more drivers.
- Multiple unrelated and semver-incompatible driver types can coexist in one runtime.
- `Driver` remains thread-local, boxable as `Box<dyn Driver<Context = C>>`, and exposes only shared-reference callbacks.
- `Parker` defines the latched wait/wake contract.
- `SystemTasks` is a cloneable crate-owned handle that hides runtime shared-ownership plumbing.
- Shutdown is initiated exactly once by the runtime and returns an object-safe boxed completion future.
- Contexts and in-flight operations retain state through safe ownership handles rather than an unsafe `is_inert` contract.

The crate includes `REQUIREMENTS.md`, `DESIGN.md`, public-surface tests, and a modular two-worker example. The example registers two independent driver types after worker startup, performs in-memory I/O through both contexts, caches repeated lookups, drains all four driver instances, and rejects operations through retained contexts after shutdown.

## Review

- Multi-model reviews with Opus, Sonnet, and GPT found no remaining issues.
- Pragmatic Rust public API review found no remaining issues after removing speculative `Clone` from `DriverInit` and replacing exposed `Arc<dyn ...>` plumbing with `SystemTasks`.
- Package coverage is 100%.

## Validation

- `just package=arty_io_core clippy`
- `just package=arty_io_core format`
- `just package=arty_io_core readme`
- `just package=arty_io_core spellcheck`
- `cargo test -p arty_io_core`
- `cargo run -p arty_io_core --example two_thread_runtime`
- `cargo coverage-gate --package arty_io_core`
- `cargo public-api --all-features --manifest-path crates/arty_io_core/Cargo.toml`

## CI infrastructure blocker

The Windows coverage job fails after all 6,867 tests pass. `cargo llvm-cov` generates a 33,956-character `llvm-cov.exe export` command for 445 objects, exceeding the Windows 32,767-character process limit (`os error 206`). The failed job was retried once and reproduced identically. Linux and Linux ARM coverage, Codecov, mutation tests, static analysis, MSRV, semver, external-type exposure, spelling, and other completed checks pass.


crates/arty_io_core/src/driver.rs
```diff
@@ -0,0 +1,76 @@
...
+    /// return [`Poll::Ready`].
+    fn poll_shutdown(&self, cx: &mut TaskContext<'_>) -> Poll<()>;
+
+    /// Returns a future that begins and then polls graceful shutdown.
+    ///
+    /// Runtime implementations that erase driver types can call
+    /// [`begin_shutdown`](Self::begin_shutdown) and [`poll_shutdown`](Self::poll_shutdown)
+    /// directly. This adapter is the ergonomic form for callers holding a concrete driver. The
+    /// caller ensures no previous shutdown was started for the same driver.
+    fn shutdown(&self) -> Shutdown<'_, Self>
```
martintmk: [AI AGENT]: `Driver::shutdown` cannot uphold the new exactly-once initiation contract. Because it takes `&self`, callers can create two independent `Shutdown` futures; each future starts with `started = false`, so polling both invokes `begin_shutdown` twice even though implementations are now explicitly allowed not to tolerate that.

Verified: a focused `public_surface` probe created two futures from one `LocalDriver`; `cargo test -p arty_io_core --test public_surface multiple_shutdown_futures_begin_shutdown_more_than_once -- --exact` failed with `begin_calls` equal to `2`, expected `1`. Please either restore idempotence as the driver requirement, or remove/reshape this convenience API so the type system gives one owner of shutdown initiation (for example, consume a shutdown handle or the driver).

martintmk (commented):
[AI AGENT]: I reviewed all eight added public contracts (`Driver`, `DriverContext`, `DriverInit`, `DriverProvider`, `Parker`, `Shutdown`, `SystemTask`, and `SystemTaskSpawner`), their rustdoc, the example runtime, dependencies, naming, object safety, shutdown behavior, and the new public-surface tests. Two targeted probes reproduced lifecycle contradictions below. The remaining public surface is documented and intentionally narrow; I found no additional dependency, naming, performance, resilience, or test-weakening issue. CI was still running with no reported failures at the final check.

crates/arty_io_core/examples/two_thread_runtime/sample_driver.rs
```diff
@@ -0,0 +1,139 @@
...
+    shutdown_complete: AtomicBool,
+}
+
+impl SampleContext {
+    pub(super) fn driver_thread(&self) -> thread::ThreadId {
+        self.state.driver_thread
+    }
+
+    pub(super) fn perform_io(&self, input: usize) -> usize {
+        let operation = self.state.operations.fetch_add(1, Ordering::Relaxed) + 1;
```
martintmk, line 44: [AI AGENT]: The example now demonstrates an I/O operation but not the public contract that contexts may outlive drivers and later operations fail safely. `shutdown_complete` is set during shutdown, yet `perform_io` ignores it and continues to report success and mutate the operation count after the runtime has stopped, teaching driver authors the opposite lifecycle.

Verified: retaining `SampleContext`, calling `runtime.shutdown()`, then calling `perform_io(41)` printed `in-memory I/O operation #2` and returned `42`; the probe failed because the operation count became `2`, expected `1`. Make `perform_io` return an error once shutdown has started/completed and keep this post-shutdown sequence as focused example coverage.

crates/arty_io_core/src/driver.rs
martintmk: &#x1F916; Fixed in `d413ede3`: `Driver::begin_shutdown` now returns `Pin<Box<dyn Future<Output = ()> + '_>>` directly. The duplicate convenience shutdown API and standalone `Shutdown` type were removed, the runtime owns the sole initiation call, and `Box<dyn Driver<Context = C>>` is covered by a public-surface test.

---

crates/arty_io_core/examples/two_thread_runtime/sample_driver.rs
martintmk: &#x1F916; Fixed in `f57236c4`: the sample context now shares shutdown state with its driver, rejects operations after shutdown with `SampleIoError`, and verifies the rejected operation does not mutate the operation count.

---

martintmk (commented):
[AI AGENT]: Pragmatic Rust public API review completed with no findings. Verified the all-features `cargo public-api` surface, generated rustdoc JSON, package tests, clippy, spelling, the two-driver example, and the 100% coverage gate.

Coverage: reviewed `Driver`, `DriverContext`, `DriverProvider`, `DriverInit`, `Parker`, `SystemTasks`, and `SystemTask` from the downstream consumer and evolution perspective. The audit cannot prove runtime behavior or soundness from API output alone; those were covered separately by tests and multi-model review.

The resulting surface removes speculative `Clone` from `DriverInit`, hides `Arc<dyn ...>` behind the cloneable `SystemTasks` handle, and documents the boxed shutdown future as the deliberate object-safe shape for `Box<dyn Driver<Context = C>>`.

crates/arty_io_core/src/provider.rs
```diff
@@ -0,0 +1,32 @@
...
+    /// This method runs on the thread that will own the returned driver. It must return promptly
+    /// and must not wait for async workers to make progress. Consuming the relocated provider clone
+    /// makes the one-creation-per-worker lifecycle explicit.
+    ///
+    /// # Panics
+    ///
+    /// Panics when this worker's driver instance cannot be initialized. Driver registration is
+    /// runtime-fundamental: after one worker fails to initialize, the runtime cannot continue in a
+    /// coherent partially registered state.
+    fn create(self, context: DriverContext) -> Self::Driver;
```
martintmk, line 31: **Posted by an AI agent**

**DriverProvider::create makes environmental initialization failures fatal**

**Problem**
`DriverProvider::create` returns `Self::Driver` and its contract requires implementations to panic whenever a worker instance cannot be initialized. Driver initialization can fail for runtime conditions such as exhausted OS resources or a permission change after the separately suggested capability check, so the public contract provides no typed way to report a normal initialization failure.

**Why this matters**
A runtime cannot roll back registration or report the affected driver reliably; with `panic = abort` the mandated path terminates the process, and even unwinding forces every runtime to treat ordinary setup failure as a panic across a foundational extension boundary.

**Suggested fix**
Make `create` return `Result<Self::Driver, DriverInitializationError>` (or the crate's canonical extensible error type), and let the runtime abort and roll back the partial registration explicitly after receiving the error.

crates/arty_io_core/examples/two_thread_runtime/echo_driver.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    fn process_completions(&mut self, _max_wait: Duration) {}
+
+    fn interruptor(&self) -> Waker {
+        Waker::noop().clone()
+    }
+
+    fn shutdown(self) -> Result<(), ShutdownError> {
+        // Release publishes admission closure before graceful cleanup starts.
+        self.state.shutdown_started.store(true, Ordering::Release);
```
martintmk, line 117: **Posted by an AI agent**

**Example shutdown returns before admitted I/O completes**

**Problem**
`EchoContext::perform_io` checks `shutdown_started` and then performs the operation without acquiring an active-operation lease, while `EchoDriver::shutdown` only stores the flag and immediately returns; `SampleContext` uses the same check-then-act pattern at `sample_driver.rs:38`. A focused probe using the exact head modules started a 256 MiB uppercase operation, waited for it to enter `perform_io`, and then shut down the runtime; shutdown completed and printed `shutdown returned while operation was still active`, contradicting R6's requirement that shutdown close admission and wait for active operations to drain.

**Why this matters**
The canonical runtime example demonstrates that a successful graceful shutdown may return while previously admitted I/O is still running, so driver authors following it can release resources or report shutdown completion before their operations and callbacks are finished.

**Suggested fix**
Give each example state a synchronized closed/active-operation lifecycle: atomically acquire a lease while admission is open, release it when the operation finishes, and have `shutdown` close admission before waiting with a bounded deadline for the active count to reach zero. Add a focused concurrent example test that proves shutdown cannot return before an admitted operation completes.

crates/arty_io_core/tests/public_surface.rs
```diff
@@ -0,0 +1,478 @@
...
+    let system_tasks = SystemTasks::new(move |task| {
+        accepted_by_callback.fetch_add(1, Ordering::Relaxed);
+        task();
+    });
+    let worker = worker_thread();
+    let context = DriverContext::new(worker.clone(), system_tasks);
+
+    context.system_tasks().spawn(|| {});
+
+    assert_eq!(context.thread(), &worker);
```
martintmk, line 38: **Posted by an AI agent**

**The public-surface test never exercises `SystemTasks::Debug`**

**Problem**
The compile-time assertion proves that `SystemTasks` implements `Debug`, but this test formats only `DriverContext` and `ProviderContext`. The matching package coverage run reports `system_tasks.rs:39-41`—the complete `Debug::fmt` implementation—as uncovered, leaving `arty_io_core` at 94.23% line coverage and matching Codecov's three missing lines.

**Why this matters**
The new test suite does not preserve the observable debug representation it advertises, and the uncovered implementation currently contributes to the failing `codecov/patch` and `codecov/project` checks.

**Suggested fix**
Add an assertion that formats the reachable `SystemTasks` handle and checks its stable type name, for example `assert!(format!("{:?}", context.system_tasks()).contains("SystemTasks"));`.

crates/arty_io_core/src/driver.rs
```diff
@@ -0,0 +1,100 @@
...
+
+    /// Returns a handle that causes the current or next completion wait to return.
+    ///
+    /// Interrupts are latched: an interrupt raised before a blocking wait makes the next blocking
+    /// wait behave like a non-blocking poll. An interrupt only ends the wait; pending completions
+    /// are still processed. Same-thread interrupts are honored. Redundant interrupts may be
+    /// coalesced, but an interrupt is never dropped. The returned interruptor remains safe to
+    /// invoke after the driver is dropped.
+    #[must_use]
+    fn interruptor(&self) -> Waker;
```
martintmk, line 74: **Posted by an AI agent · Non-blocking**

**`Driver::interruptor` obscures that the handle is a standard `Waker`**

**Problem**
This method returns `std::task::Waker`, while the workspace names the same returned concept `waker` (for example, `arty_executor::wake::Wake::waker` at `crates/arty_executor/src/wake.rs:152`). Renaming it to the newly coined `interruptor` makes this trait the only sibling to hide the foundational type behind a different noun.

**Why this matters**
Driver implementers and runtime authors must learn a crate-specific synonym, and the divergence propagates through the public trait, tests, examples, requirements, design document, generated README, and spelling dictionary.

**Suggested fix**
Rename the method back to `waker` and update those references; the current PR title does not mention either term and needs no change.

```suggestion
    fn waker(&self) -> Waker;
```

martintmk (commented):
**Posted by an AI agent**

Reviewed the exact PR head across API design, correctness, tests, performance, naming, telemetry, resilience, consistency, public API, and public documentation. I found three blocking issues and two non-blocking issues. Four are anchored inline; the remaining non-blocking issue is that the PR description still documents the obsolete `DriverContext`/`Parker`/boxed-future API and should be refreshed to match the current `IoContext`, `interruptor`, and blocking `shutdown(self)` contract.

Focused validation: `cargo check -p arty_io_core` passed; specialist runs also passed the 13 public-surface tests and the two-thread example. The focused correctness probe reproduced shutdown returning while admitted I/O was still active.

**Posted by an AI agent · Non-blocking**

**The PR summary describes an obsolete arty_io_core API**

**Problem**
The PR body still says `DriverContext` selects the provider, `Driver` exposes only shared-reference callbacks, `Parker` defines waiting, and shutdown returns an object-safe boxed future. At the pinned head, `IoContext::provider` selects the provider, `process_completions` takes `&mut self`, `interruptor` returns `Waker`, and `shutdown(self)` blocks and returns `Result<(), ShutdownError>`.

**Why this matters**
Reviewers and future readers are given a materially different registration, completion, and shutdown contract from the code and the new requirements/design documents.

**Suggested fix**
Refresh the PR summary to name `IoContext`, `process_completions`/`interruptor`, and the consuming blocking `shutdown` result, removing the obsolete `Parker`, shared-reference-callback, and boxed-future claims.

crates/arty_io_core/src/system_tasks.rs
```diff
@@ -0,0 +1,42 @@
...
+/// A cloneable handle for running I/O system work on runtime-owned threads.
+///
+/// An I/O driver may use this facility instead of creating threads of its own. Submitted work is
+/// not an async application task and never runs on an async worker. The runtime callback must
+/// permit the work to block.
+///
+/// The facility remains available until every driver that received it has completed shutdown, so
+/// cleanup work submitted during shutdown can still run.
+#[derive(Clone)]
+pub struct SystemTasks {
```
martintmk, line 19: **Posted by an AI agent · Non-blocking**

**SystemTasks names the submitted work instead of the spawning capability**

**Problem**
The type is documented as a cloneable handle and its only operation is `spawn`, but `SystemTasks` reads as a collection of tasks. The workspace names this role `anyspawn::Spawner` (`crates/anyspawn/src/spawner.rs:99`) and the boxed work `BoxedBlockingTask` (`crates/anyspawn/src/custom.rs:51`).

**Why this matters**
This is a new stable public contract, so consumers will otherwise have to learn an unexplained synonym and a later correction would require a breaking rename.

**Suggested fix**
Rename the handle to `SystemTaskSpawner` (and update `DriverContext::system_tasks`, docs, tests, examples, and generated README references); retain the narrower wrapper rather than exposing the broader `anyspawn::Spawner` capability.

martintmk (commented):
**Posted by an AI agent**

Reviewed the current PR head across API design, correctness, tests, performance, naming, telemetry, resilience, consistency, public API, and public documentation. Three blocking issues and two non-blocking issues remain applicable.

Four are already represented by unresolved current threads: provider creation mandates panic, graceful shutdown can return before admitted I/O drains, `SystemTasks::Debug` remains untested, and `Driver::interruptor` obscures its `Waker` contract. One new non-blocking naming issue is posted inline: `SystemTasks` names submitted work rather than its spawning capability.

Focused validation passed all 13 `arty_io_core` public-surface tests and the two-thread runtime example. Paired `cargo public-api` and rustdoc JSON extraction completed for the exact base/head snapshot.

crates/arty_io_core/src/provider.rs
martintmk: keep panic

martintmk: &#x1F916; Preserved the panic-based registration contract as requested ("keep panic"). DriverProvider::create remains infallible at the type level; the integration does not introduce a Result return.

---

crates/arty_io_core/examples/two_thread_runtime/echo_driver.rs
martintmk: &#x1F916; Superseded by the requested integration in 7fc2f37b: the two-worker I/O example was removed and replaced with a single-thread registration and peer-discovery example that performs no I/O operations. The existing public-surface shutdown_waits_for_active_operations_not_contexts test remains. Closing this thread against the removed example, rather than claiming its old race was repaired.

---

crates/arty_io_core/tests/public_surface.rs
martintmk: &#x1F916; Fixed in 2fdc4807: system_task_spawner_debug now asserts the complete Debug output of SystemTaskSpawner, the renamed SystemTasks handle. The focused test passes; fresh CI will confirm the resulting coverage.

---

crates/arty_io_core/src/driver.rs
martintmk: &#x1F916; Included in the requested branch integration, merged in 7fc2f37b: Driver now exposes waker(), and the public tests and example driver implementations use that name.

---

crates/arty_io_core/src/system_tasks.rs
martintmk: &#x1F916; Included in the requested branch integration, merged in 7fc2f37b: the public handle is now SystemTaskSpawner, and registration APIs, documentation, tests, and the runtime example use the renamed handle.

---

crates/arty_io_core/docs/COMPLETION_COORDINATION.md
```diff
@@ -0,0 +1,379 @@
+# Completion coordination across I/O drivers
```
martintmk: i want you to review latest api surface against this doc to check if all requirements can be met. the idea is that individual drivers are aware of each other and can coordinate the completioitions if possible.

martintmk: &#x1F916; Review of the current API against the completion-coordination proposal.

**Posted by an AI agent**

**Bottom line: the API enables compatible drivers to coordinate privately, but it does not provide every guarantee in the broader runtime-owned-coordinator proposal.** The proposal explicitly describes an alternative, not the current contract. No API changes are requested or made by this review.

### What the latest API enables

`DriverOptions::drivers()` exposes earlier drivers before the new driver is created; `Driver::on_peer_registered()` then notifies earlier drivers before registration is acknowledged. `DriverHandle` exposes `Any`, allowing a recognized handle to be downcast and independently owned state to be cloned. This gives both registration orders a discovery opportunity without retaining references to another driver.

Sources: [construction-time discovery](https://github.com/microsoft/oxidizer/blob/50d7cbd73ed712fa132fc66433eff255eb67f020/crates/arty_io_core/src/driver_options.rs#L38-L45), [peer notification](https://github.com/microsoft/oxidizer/blob/50d7cbd73ed712fa132fc66433eff255eb67f020/crates/arty_io_core/src/driver.rs#L22-L41), [type-erased handle](https://github.com/microsoft/oxidizer/blob/50d7cbd73ed712fa132fc66433eff255eb67f020/crates/arty_io_core/src/driver_handle.rs#L7-L26).

**Inference:** cooperating implementations could expose a common owned completion-domain handle, join it during creation, and privately manage native registrations, record routing, notification latching, and retirement. They must agree on that handle's concrete Rust type/version and its protocol. The first participant must already support shared routing before later participants join; notifying an existing driver cannot by itself retrofit an incompatible native resource arrangement. Discovery is therefore sufficient as an extension point, not proof that arbitrary drivers compose.

**Posted by an AI agent · Design note, no change requested**

**`process_completions` does not expose the proposal's generic parking guarantees**

**Problem**
[`process_completions`](https://github.com/microsoft/oxidizer/blob/50d7cbd73ed712fa132fc66433eff255eb67f020/crates/arty_io_core/src/driver.rs#L43-L66) returns `()` and receives a maximum wait and cycle timestamp, not a service budget/result, remaining-work indication, source deadline, or prepare-to-wait result. Its waker interrupts a driver's wait; the contract does not connect every peer's native readiness to whichever wait the runtime selects. Compare [the proposed service/parking protocol](https://github.com/microsoft/oxidizer/blob/50d7cbd73ed712fa132fc66433eff255eb67f020/crates/arty_io_core/docs/COMPLETION_COORDINATION.md#L209-L253).

**Why this matters**
A generic runtime cannot infer that it is safe to sleep after zero-duration passes. In the proposal's sequence, B receives a completion while the worker waits in A; discovery alone does not wake A. A compatible private coordination protocol can supply that connection, but it must also own routing, budgets, rearming, and submission progress. `cycle_start` alone is not such a protocol.

**Posted by an AI agent · Design note, no change requested**

**Consuming blocking shutdown excludes general cross-driver cooperative draining**

**Problem**
[`Driver::shutdown`](https://github.com/microsoft/oxidizer/blob/50d7cbd73ed712fa132fc66433eff255eb67f020/crates/arty_io_core/src/driver.rs#L68-L83) owns the worker until it returns and explicitly forbids depending on work available only afterward. The proposal instead needs the runtime to close admission across participants and continue servicing them while draining ([proposal](https://github.com/microsoft/oxidizer/blob/50d7cbd73ed712fa132fc66433eff255eb67f020/crates/arty_io_core/docs/COMPLETION_COORDINATION.md#L255-L291)).

**Why this matters**
If A's shutdown needs B's next owner-thread service call, the existing interface cannot schedule that call through the generic runtime while A is blocked. Sharing native collection is still possible when the shared component can make all required progress independently; it must not introduce this forbidden dependency. General runtime-driven draining would require an explicit lifecycle-contract decision, not just exchanged handles.

**Posted by an AI agent · Design note, no change requested**

**Provider options do not negotiate runtime-controlled hosting policies**

**Problem**
[`ProviderOptions`](https://github.com/microsoft/oxidizer/blob/50d7cbd73ed712fa132fc66433eff255eb67f020/crates/arty_io_core/src/provider_options.rs#L4-L17) is empty, and `DriverOptions` supplies a thread, spawner, and prior-driver handles. Neither exposes the proposal's capability selection, exclusive-host approval, active-polling requirements, or fallback budgets. The current requirements also expressly permit [private threads](https://github.com/microsoft/oxidizer/blob/50d7cbd73ed712fa132fc66433eff255eb67f020/crates/arty_io_core/docs/REQUIREMENTS.md#L57-L74).

**Why this matters**
A driver can negotiate with a known peer inside `create`, but the generic runtime cannot select and enforce every hosting arrangement described in the proposal from this surface alone. Supplying future optional facilities is possible; claiming the policy is already enforced is not.

### Existing policy decision

The proposal's recoverable native-attachment requirement also differs from the current panic-on-registration-failure contract. Your explicit **"keep panic"** decision remains authoritative; this review does not reopen it. Consequently, **all proposal requirements cannot be satisfied as written under the current policy**. Initialization still needs safe resource ownership and cleanup even when failure is fatal.

### Scope and remaining evidence

Reviewed all exported types/traits, their bounds and constructors, dependency/feature exposure, the registration example, and public-surface tests at `50d7cbd7`. The single-thread example demonstrates discovery, not native completion coordination. Native adapters, platform correctness, lost-wakeup behavior, late-packet retirement, and performance were not implemented or executed here. The proposal's old names (`interruptor`, `ProviderContext`, `DriverContext`) and removed two-thread-example link also predate the integrated API.

For the narrower goal "drivers know about each other and coordinate when compatible," the discovery hooks are the appropriate foundation. The decisive next evidence would be two compatible drivers sharing a real completion domain while preserving the current no-peer-dependent-shutdown rule, rather than requiring every conceptual operation in the alternative proposal to become public API now.

---

crates/arty_io_core/docs/COMPLETION_COORDINATION.md
```diff
@@ -0,0 +1,379 @@
...
+
+Unify notification, scheduling, and lifetime rules, not drivers' operation
+types, buffer layouts, or private completion queues. Sharing the coordination
+contract still requires compatible versions of `arty_io_core`; it does not
+provide interoperability between incompatible copies of the contract itself.
+
+## Why the current boundary is insufficient
+
+The current [driver interface](../src/driver.rs) combines completion processing
+and waiting in `process_completions(max_wait)`. Its `interruptor` is a handle for
```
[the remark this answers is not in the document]
martintmk: &#x1F916; Agreed - updating the proposal to use the current waker and options names, including the current completion-processing signature.

---

crates/arty_io_core/docs/COMPLETION_COORDINATION.md
```diff
@@ -0,0 +1,379 @@
...
+source.
+
+Finite waits can bound this delay, but introduce polling and latency.
+Zero-duration scans avoid blocking on the wrong driver but consume CPU while
+idle. A dedicated observer solves progress at the cost of threads and possible
+cross-thread completion delivery.
+
+The [two-thread example](../examples/two_thread_runtime/runtime.rs) demonstrates
+registration and retained contexts, not a native completion loop. Its workers
+receive control commands, and its example operations execute synchronously.
```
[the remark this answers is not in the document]
martintmk: &#x1F916; Agreed - updating the link and description to the existing single-thread registration example, and correcting the stale requirements anchor.

---

crates/arty_io_core/docs/COMPLETION_COORDINATION.md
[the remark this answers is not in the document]
martintmk: &#x1F916; Fixed: the proposal now uses waker, ProviderOptions, and DriverOptions, and includes cycle_start when spelling the current process_completions signature.

---

crates/arty_io_core/docs/COMPLETION_COORDINATION.md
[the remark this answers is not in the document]
martintmk: &#x1F916; Fixed: the proposal links to examples/single_thread_runtime/runtime.rs, describes its registration and peer-discovery behavior without claiming I/O operations, and targets the current R7 registration-failure heading.

---

crates/arty_io_core/src/driver.rs
```diff
@@ -0,0 +1,82 @@
...
+    /// indefinitely, and must not depend on work that can run only after this method returns.
+    ///
+    /// [`Drop::drop`] runs after this method returns. Shared cleanup must therefore be idempotent
+    /// or otherwise guarded. Regardless of the result, the driver must remain safe to drop and its
+    /// contexts must reject new operations.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if graceful cleanup cannot be completed.
+    fn shutdown(self) -> Result<(), ShutdownError>;
```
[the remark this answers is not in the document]
martintmk: &#x1F916; No API change is needed here. The [Rust Reference's dyn-compatibility rules](https://doc.rust-lang.org/reference/items/traits.html#dyn-compatibility) explicitly state that a receiver of type `Self` (`self`) implies the non-dispatchable `Self: Sized` restriction. A trait can remain dyn-compatible while its consuming method is not callable through the trait object.

That distinction is intentional in this crate: [DESIGN.md](https://github.com/microsoft/oxidizer/blob/24e0d9f2fb716537e20977f615fe73514cbcf560/crates/arty_io_core/docs/DESIGN.md#L48-L50) and [R4](https://github.com/microsoft/oxidizer/blob/24e0d9f2fb716537e20977f615fe73514cbcf560/crates/arty_io_core/docs/REQUIREMENTS.md#L64-L66) specify a private owning shim for boxed storage. The [example's ErasedDriver/RegisteredDriver adapter](https://github.com/microsoft/oxidizer/blob/24e0d9f2fb716537e20977f615fe73514cbcf560/crates/arty_io_core/examples/single_thread_runtime/runtime.rs#L21-L53) accepts `self: Box<Self>` privately and then consumes the concrete driver. The public-surface test asserts `Box<dyn Driver>` construction and shared-method dispatch, not direct trait-object shutdown.

The documented contract and test are therefore consistent. Keeping the public consuming receiver unchanged, as approved by the author.

---

martintmk (commented):
**Posted by an AI agent**

Reviewed the current PR head across API design, correctness, tests, performance, naming, telemetry, resilience, consistency, public API, and public documentation. Three new non-blocking issues were found: standard default construction for `ProviderOptions`, avoidable allocation/dispatch in repeated example context lookups, and lost recovery classification in `ShutdownError`.

The previously reported panic-based registration contract remains an explicit accepted design decision (`keep panic`) and is not re-raised. The prior waker, system-task naming, debug coverage, and two-thread shutdown findings are resolved or no longer applicable.

Focused package tests passed, public API extraction and isolated documentation filtering completed for the exact base/head snapshot, and all ten specialist passes completed.

crates/arty_io_core/src/provider_options.rs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+/// Options for creating a driver provider.
+///
+/// This type is currently empty. Its private representation allows compatible versions to add
+/// optional runtime facilities.
+#[derive(Debug)]
+#[non_exhaustive]
+pub struct ProviderOptions;
```
martintmk: **Posted by an AI agent · Non-blocking**

**ProviderOptions lacks the standard default-construction trait**

**Problem**
`ProviderOptions` is a zero-field, `#[non_exhaustive]` options type with a zero-argument `new()`, but it does not implement `Default`.

**Why this matters**
Consumers cannot use the type with standard default-based generic APIs or `unwrap_or_default`; they must special-case its inherent constructor even though there is only one default value.

**Suggested fix**
Implement `Default` for `ProviderOptions` by delegating to `ProviderOptions::new()`, while retaining `new()` for discoverability.

crates/arty_io_core/examples/single_thread_runtime/runtime.rs
```diff
@@ -0,0 +1,204 @@
...
+        C: IoContext,
+    {
+        self.try_get_context::<C>().unwrap_or_else(|| self.initialize_context::<C>())
+    }
+
+    fn try_get_context<C>(&self) -> Option<C>
+    where
+        C: IoContext,
+    {
+        let (reply_tx, reply_rx) = mpsc::channel();
```
martintmk: **Posted by an AI agent · Non-blocking**

**Repeated context lookups allocate and round-trip through the worker**

**Problem**
Every `Runtime::get_context` call creates a reply channel, boxes a command, scans every registered driver, and boxes the cloned context—even after that context is initialized.

**Why this matters**
The canonical runtime example makes an otherwise cheap cached-context lookup scale with driver count and worker scheduling, with at least two explicit heap allocations per lookup.

**Suggested fix**
Cache initialized contexts by `TypeId` on `Runtime`; serve hits by downcasting and cloning the cached value, and use the worker round-trip only for a serialized cache miss.

crates/arty_io_core/src/shutdown_error.rs
```diff
@@ -0,0 +1,53 @@
...
+    #[must_use]
+    pub fn from_message(message: impl Into<String>) -> Self {
+        Self {
+            kind: ShutdownErrorKind::Message(message.into().into_boxed_str()),
+        }
+    }
+
+    /// Creates an error with the given source.
+    #[must_use]
+    pub fn from_source(source: impl Error + Send + Sync + 'static) -> Self {
```
martintmk: **Posted by an AI agent · Non-blocking**

**ShutdownError discards recovery classification from shutdown failures**

**Problem**
`ShutdownError::from_source` retains only `dyn Error`; the wrapper does not preserve `recoverable::Recovery` metadata, so transient I/O failures, unavailability, and delay hints are erased at this runtime boundary.

**Why this matters**
Although the consumed driver cannot repeat its own shutdown, the runtime may still need the classification to choose process-level retry, driver recreation, fallback, or escalation. Callers currently have to downcast arbitrary source chains themselves.

**Suggested fix**
Capture `RecoveryInfo` when constructing `ShutdownError`, implement `Recovery`, and provide an I/O-source convenience path that derives classification through the source chain and `ErrorKind`.

crates/arty_io_core/src/provider_options.rs
martintmk: &#x1F916; Keeping ProviderOptions without Default, as explicitly reaffirmed by the author. Explicit new() construction is intentional; the narrowly scoped new_without_default expectation records that decision.

---

crates/arty_io_core/examples/single_thread_runtime/runtime.rs
martintmk: &#x1F916; Agreed, and the author approved this change: I am adding a TypeId-keyed cache on the example Runtime so cache hits clone locally and only serialized misses communicate with the worker.

---

martintmk: &#x1F916; Fixed: Runtime now stores contexts in a TypeId-keyed cache. Hits downcast and clone locally without command/reply allocation or a worker round-trip; misses remain serialized until registration completes. Two focused example tests cover a hit while the worker is occupied and concurrent requests registering each context only once.

---

martintmk (commented):
**Posted by an AI agent**

Reviewed the current PR head across API design, correctness, tests, performance, naming, telemetry, resilience, consistency, public API, and public documentation.

The new context cache has one blocking concurrency defect: it invokes an unconstrained context `Clone` while holding the runtime-wide cache mutex, so one context can stall or deadlock every cached lookup. Two additional non-blocking issues remain in the new surface: the worker-side context lookup is now redundant with the authoritative host cache, and `DriverOptions::new` unnecessarily requires callers to construct a `Vec`.

The previously reported `ShutdownError` recovery-classification issue still applies but is not posted again. The `ProviderOptions: Default` suggestion is omitted because the current discussion records the explicit decision to keep `new()` without `Default`. The earlier context lookup allocation/worker-roundtrip issue is resolved.

All ten specialist passes completed for the exact base/head snapshot. Focused package and example tests passed; the cache-lock defect was additionally reproduced with a blocked clone and an unrelated cached lookup timing out.

crates/arty_io_core/examples/single_thread_runtime/runtime.rs
```diff
@@ -0,0 +1,267 @@
...
+            commands: commands_tx,
+            thread,
+        }
+    }
+
+    pub(super) fn get_context<C>(&self) -> C
+    where
+        C: IoContext,
+    {
+        let mut contexts = self
```
martintmk: **Posted by an AI agent**

**Context cloning holds the global runtime cache lock**

**Problem**
`Runtime::get_context` retains `contexts` while invoking the context's unconstrained `Clone` implementation. A focused probe blocked one cached context clone for 500 ms; an unrelated cached `EchoContext` lookup timed out after 100 ms.

**Why this matters**
One driver context can stall every context lookup in the runtime. It can also deadlock through lock inversion if `Clone` acquires context-owned synchronization while another thread holds that synchronization and requests any context.

**Suggested fix**
Store independently synchronized cache entries, clone the selected entry handle under the map lock, release the map lock, and only then invoke `C::clone`. Preserve single initialization with per-entry once/registration state and add the reproduced concurrent cached-lookup regression test.

crates/arty_io_core/examples/single_thread_runtime/runtime.rs
```diff
@@ -0,0 +1,267 @@
...
+            .clone()
+    }
+
+    fn initialize_context<C>(&self) -> C
+    where
+        C: IoContext,
+    {
+        let (reply_tx, reply_rx) = mpsc::channel();
+        self.run(move |worker, spawner, drivers| {
+            if let Some(context) = find_context::<C>(drivers) {
```
martintmk: **Posted by an AI agent · Non-blocking**

**The worker-side context lookup is now unreachable**

**Problem**
The new `Runtime::contexts` cache serializes every miss until it has inserted the initialized context. Consequently, a later miss for the same `TypeId` cannot reach `initialize_context`, so `find_context` and the `ErasedDriver::context_type`/`context` storage path can no longer return an existing context.

**Why this matters**
The canonical example now maintains two context registries with different ownership models, even though only the host-side cache can serve repeated lookups. That makes the registration flow harder to follow and risks the two copies diverging as the example evolves.

**Suggested fix**
Make the host cache the sole context registry: remove the worker-side context lookup and stop storing the context in `RegisteredDriver` after returning the value needed for the cache.

crates/arty_io_core/src/driver_options.rs
```diff
@@ -0,0 +1,56 @@
...
+///
+/// These options may borrow thread-local drivers and are therefore neither [`Send`] nor [`Sync`].
+pub struct DriverOptions<'a> {
+    thread: Thread,
+    spawner: SystemTaskSpawner,
+    drivers: Vec<DriverHandle<'a>>,
+}
+
+impl<'a> DriverOptions<'a> {
+    /// Creates options for `thread`.
```
martintmk: **Posted by an AI agent · Non-blocking**

**DriverOptions::new unnecessarily fixes callers to Vec**

**Problem**
The public constructor accepts `drivers: Vec<DriverHandle<'a>>`, although the options type only exposes those handles as a slice.

**Why this matters**
Runtime implementations with arrays or iterator-based handle collection must allocate and collect into this exact container before construction, making `Vec` an avoidable part of the public contract.

**Suggested fix**
Accept `impl IntoIterator<Item = DriverHandle<'a>>` (or another suitably broad input) and collect internally.

crates/arty_io_core/examples/single_thread_runtime/runtime.rs
```diff
@@ -0,0 +1,267 @@
...
+    let driver = driver.handle();
+    for existing_driver in existing_drivers {
+        existing_driver.on_peer_registered(driver);
+    }
+}
+
+fn shutdown_drivers(drivers: DriverStore) -> ShutdownResult {
+    let mut failure = None;
+
+    for driver in drivers {
```
geeknoid: **A first shutdown error can stop later drivers without any test failing** — *Testing · Medium · High confidence*

The loop correctly keeps shutting down after recording the first error, which is the R6 guarantee that one driver's failure must not prevent remaining drivers from shutting down. The current fixtures only return `Ok(())`, so replacing this loop with an early `return Err(error)` (or breaking after the first failure) survives every test while leaving later drivers unshut.

**Direction:** add an example-runtime test with at least two registered drivers where the first shutdown returns `Err` and the later driver records that its shutdown still ran.

**Done when:** an early-return-on-first-error mutation makes the focused test fail, while the returned result still reports an error.

geeknoid (commented):
Static-only review found one new testing gap: the example runtime's required "continue shutting down remaining drivers after an error" behavior is implemented but not protected against regression.

The security and production-correctness reads were otherwise clean for the new contract surface: no unsafe/FFI exists in scope; the ownership model keeps contexts, wakers, tasks, and callback-visible state independent of thread-local drivers; shutdown consumes the driver and the contract requires admission closure, bounded draining, and drop safety. Registration ordering and same-worker peer discovery match the requirements and design documents.

I assessed all 820 taxonomy classes. I did not duplicate the existing threads covering the context-cache lock around arbitrary `Clone`, unreachable worker-side context lookup, coverage annotation, `DriverOptions` container flexibility, deliberate panic-based registration, or shutdown error classification. Generated `README.md` content was excluded per repository policy. Nothing was executed: no project, tests, builds, benchmarks, profilers, Miri, coverage, mutation, fuzzing, or live probes.

PR #740, opened 2026-09-08
fix(thread_aware_macros): express derive bounds on the relocated field type

psandana:
## What & why

`#[derive(ThreadAware)]` inferred its bounds by collecting every generic parameter a relocated field's type reaches and bounding each by `ThreadAware`. That predicate is on the **parameter**, not on the obligation the generated body actually has — the body relocates the **whole field**, so what it needs is `<field type>: ThreadAware`.

Bounding the parameter is too strong for a wrapper that implements `ThreadAware` unconditionally:

```rust
struct Wrapper<T>(PhantomData<fn() -> T>);
impl<T> ThreadAware for Wrapper<T> {
    fn relocate(&mut self, _: Option<&Thread>, _: &Thread) {}
}

#[derive(ThreadAware)]
struct Outer<T>(Wrapper<PhantomData<T>>);
```

The body only relocates the whole `Wrapper<PhantomData<T>>` field, which the wrapper's own impl already satisfies. The old derive additionally emitted `T: ThreadAware`, so `Outer<Rc<()>>` was rejected even though `Rc<()>` never needs to be `ThreadAware`.

## Change

`add_bounds` now emits, for each relocated field whose type reaches a generic parameter, a `where <field type>: ThreadAware` predicate — the exact obligation the generated body discharges — replacing the per-parameter predicates.

- **Slice traversal added:** the traversal now also descends into `Type::Slice`, so a `Box<[T]>` field emits `where Box<[T]>: ThreadAware` (`[T]` has a conditional `ThreadAware` impl in `thread_aware_core`). A marker payload behind a function pointer (`PhantomData<fn(*const T)>`) still owes no bound, so the raw-pointer variance idiom stays bound-free.
- **Recursive generics:** a field whose type names the type being derived (or `Self`) would make `where <field type>: ThreadAware` self-referential and overflow, so it falls back to bounding the parameters it reaches — recursive types (`struct Node<T> { children: Vec<Node<T>> }`) still derive.
- **Alias escape hatch:** a type alias (`type Children<T> = Vec<Node<T>>`) can hide that self-reference from the syntactic derive, which can't resolve aliases. `#[thread_aware(bound = "...")]` replaces the inferred bounds with the author's own for that case.
- **De-duplication:** a field that is exactly a parameter the author already bounded by `ThreadAware` emits nothing, and predicates the author already wrote in a `where` clause are not repeated, avoiding `clippy::trait_duplication_in_bounds`.
- The `where Self: Send` predicate for skipped fields is unchanged.

## Acceptance criteria

- [x] Compile regression for the unconditional-wrapper case.
- [x] Compile regressions for recursive generics, a slice-bearing sibling, and the alias-hidden recursion with the `bound` escape hatch.
- [x] Existing `derive_compiles.rs` coverage passes unchanged.
- [x] Snapshots and the derive's "Generic Bounds" / "Attributes" documentation updated.
- [x] `thread_aware`, `thread_aware_macros_impl`, and the four consuming crates build and lint clean locally.

Refs AB#7783612. Follow-up to #678 (raised in review by Sander Saares).

### DIFF
--- a/crates/thread_aware/src/lib.rs
+++ b/crates/thread_aware/src/lib.rs
@@ -161,22 +161,29 @@ mod thread;
 /// * `#[thread_aware(skip)]`: Prevents a field from being recursively transferred.
 ///
 /// # Generic Bounds
-/// Every field except a `#[thread_aware(skip)]` one is relocated by calling its own
-/// `ThreadAware::relocate`, and the bounds follow from that:
-/// * each relocated field whose type reaches a generic parameter contributes a
-///   `where <field type>: ::thread_aware::ThreadAware` predicate - the obligation the generated
-///   body actually has, discharged by that field type's own impl;
-/// * if any field carries `#[thread_aware(skip)]`, a `where Self: Send` predicate is added,
-///   since the `ThreadAware: Send` supertrait still has to hold.
+/// The derive adds one bound per relocated field: `where <field type>: ThreadAware`. A
+/// `#[thread_aware(skip)]` field isn't relocated, so it contributes `where Self: Send` instead,
+/// which is all the `ThreadAware: Send` supertrait needs.
 ///
-/// The predicate lands on the field type as written, not on the parameters inside it. A
-/// `PhantomData<U>` field yields `where PhantomData<U>: ThreadAware` (which reduces to `U: Send`
-/// through the no-op `impl<T: ?Sized> ThreadAware for PhantomData<T> where Self: Send`), and a
-/// wrapper `W<U>` yields `where W<U>: ThreadAware`, governed by `W`'s own impl rather than by a
-/// bound on `U` - so a wrapper that implements the trait unconditionally keeps deriving for
-/// arguments no bound on `U` would admit. The traversal does not enter a function pointer, so
-/// writing a marker payload as one carries the parameter for variance, stays `Send` for every
-/// argument, and emits no bound at all:
+/// Bounds are on the whole field type, not the parameters inside it. A `Wrapper<T>` field bounds
+/// `Wrapper<T>: ThreadAware` and defers to that wrapper's own impl, so a wrapper that is
+/// `ThreadAware` for every `T` keeps deriving even where `T` isn't.
+///
+/// The derive doesn't look inside function pointers, so a variance marker written as
+/// `PhantomData<fn(*const T)>` adds no bound at all:
+///
+/// ```rust
+/// # use core::marker::PhantomData;
+/// # use thread_aware::ThreadAware;
+/// #[derive(ThreadAware)]
+/// struct Marked<T> {
+///     // `PhantomData<*const T>` would make `Marked` `!Send`; this does not.
+///     marker: PhantomData<fn(*const T)>,
+/// }
+/// ```
+///
+/// If you already wrote a bare `ThreadAware` bound yourself, the derive assumes it's this trait
+/// and skips the duplicate; qualify the path to disambiguate.
 ///
 /// ```rust
 /// # use core::marker::PhantomData;
--- a/crates/thread_aware/src/lib.rs
+++ b/crates/thread_aware/src/lib.rs
@@ -158,14 +158,20 @@ mod thread;
 /// Unions are not supported and will produce a compile error.
 ///
 /// # Attributes
-/// * `#[thread_aware(skip)]`: Prevents a field from being recursively transferred.
+/// * `#[thread_aware(skip)]` (on a field): moves the field without calling `relocate` on it.
+/// * `#[thread_aware(bound = "...")]` (on the type): replaces the derive's inferred `where`
+///   predicates with your own. An escape hatch for a field whose bound the derive can't infer -
+///   most often a type alias that hides a recursive self-reference, where the inferred field-type
+///   bound would be circular. Write `#[thread_aware(bound = "T: ThreadAware")]` to bound the
+///   parameter instead.
 ///
 /// # Generic Bounds
 /// The derive bounds each relocated field by its own type - `where <field type>: ThreadAware` -
 /// not by the type parameters inside it. A field that refers back to the type being derived is the
-/// exception: it bounds the parameters it reaches instead, so recursive types still compile. A
-/// `#[thread_aware(skip)]` field isn't relocated, so it contributes `where Self: Send` instead,
-/// which is all the `ThreadAware: Send` supertrait needs.
+/// exception: it bounds the parameters it reaches instead, so recursive types still compile (a
+/// recursion hidden behind a type alias is invisible to the derive - use `#[thread_aware(bound =
+/// "...")]` there). A `#[thread_aware(skip)]` field isn't relocated, so it contributes
+/// `where Self: Send` instead, which is all the `ThreadAware: Send` supertrait needs.
 ///
 /// Bounding the field type means a `Wrapper<T>` field defers to that wrapper's own impl, so a
 /// wrapper that is `ThreadAware` for every `T` keeps deriving even where `T` isn't. The derive
--- a/crates/thread_aware/src/lib.rs
+++ b/crates/thread_aware/src/lib.rs
@@ -161,15 +161,15 @@ mod thread;
 /// * `#[thread_aware(skip)]`: Prevents a field from being recursively transferred.
 ///
 /// # Generic Bounds
-/// The derive adds one bound per relocated field: `where <field type>: ThreadAware`. A
+/// The derive bounds each relocated field by its own type - `where <field type>: ThreadAware` -
+/// not by the type parameters inside it. A field that refers back to the type being derived is the
+/// exception: it bounds the parameters it reaches instead, so recursive types still compile. A
 /// `#[thread_aware(skip)]` field isn't relocated, so it contributes `where Self: Send` instead,
 /// which is all the `ThreadAware: Send` supertrait needs.
 ///
-/// Bounds are on the whole field type, not the parameters inside it. A `Wrapper<T>` field bounds
-/// `Wrapper<T>: ThreadAware` and defers to that wrapper's own impl, so a wrapper that is
-/// `ThreadAware` for every `T` keeps deriving even where `T` isn't.
-///
-/// The derive doesn't look inside function pointers, so a variance marker written as
+/// Bounding the field type means a `Wrapper<T>` field defers to that wrapper's own impl, so a
+/// wrapper that is `ThreadAware` for every `T` keeps deriving even where `T` isn't. The derive
+/// doesn't look inside function pointers, so a variance marker written as
 /// `PhantomData<fn(*const T)>` adds no bound at all:
 ///
 /// ```rust
@@ -182,21 +182,8 @@ mod thread;
 /// }
 /// ```
 ///
-/// If you already wrote a bare `ThreadAware` bound yourself, the derive assumes it's this trait
-/// and skips the duplicate; qualify the path to disambiguate.
-///
-/// ```rust
-/// # use core::marker::PhantomData;
-/// # use thread_aware::ThreadAware;
-/// #[derive(ThreadAware)]
-/// struct Marked<T> {
-///     // `PhantomData<*const T>` would make `Marked` `!Send`; this does not.
-///     marker: PhantomData<fn(*const T)>,
-/// }
-/// ```
-///
-/// A trait referred to by the bare name `ThreadAware` in your own bounds is assumed to be this
-/// crate's and suppresses the generated bound; qualify the path to disambiguate.
+/// If you already wrote a bare `ThreadAware` bound yourself, the derive assumes it's this trait and
+/// skips the duplicate; qualify the path to disambiguate.
 ///
 /// # Example
 /// ```rust
--- a/crates/thread_aware/src/lib.rs
+++ b/crates/thread_aware/src/lib.rs
@@ -161,18 +161,22 @@ mod thread;
 /// * `#[thread_aware(skip)]`: Prevents a field from being recursively transferred.
 ///
 /// # Generic Bounds
-/// Every field except a `#[thread_aware(skip)]` one is relocated, and the bounds follow from
-/// that:
-/// * a generic type parameter the traversal reaches through a relocated field receives a
-///   `::thread_aware::ThreadAware` bound;
+/// Every field except a `#[thread_aware(skip)]` one is relocated by calling its own
+/// `ThreadAware::relocate`, and the bounds follow from that:
+/// * each relocated field whose type reaches a generic parameter contributes a
+///   `where <field type>: ::thread_aware::ThreadAware` predicate - the obligation the generated
+///   body actually has, discharged by that field type's own impl;
 /// * if any field carries `#[thread_aware(skip)]`, a `where Self: Send` predicate is added,
 ///   since the `ThreadAware: Send` supertrait still has to hold.
 ///
-/// A `PhantomData<..>` field is no exception: it relocates through the no-op
-/// `impl<T: ?Sized + Send> ThreadAware for PhantomData<T>`, and a parameter the traversal
-/// reaches inside it - the `U` of `PhantomData<U>` - is bound like any other. The traversal
-/// does not enter a function pointer, so writing the payload as one carries the parameter for
-/// variance, stays `Send` for every argument, and emits no bound at all:
+/// The predicate lands on the field type as written, not on the parameters inside it. A
+/// `PhantomData<U>` field yields `where PhantomData<U>: ThreadAware` (which reduces to `U: Send`
+/// through the no-op `impl<T: ?Sized> ThreadAware for PhantomData<T> where Self: Send`), and a
+/// wrapper `W<U>` yields `where W<U>: ThreadAware`, governed by `W`'s own impl rather than by a
+/// bound on `U` - so a wrapper that implements the trait unconditionally keeps deriving for
+/// arguments no bound on `U` would admit. The traversal does not enter a function pointer, so
+/// writing a marker payload as one carries the parameter for variance, stays `Send` for every
+/// argument, and emits no bound at all:
 ///
 /// ```rust
 /// # use core::marker::PhantomData;
--- a/crates/thread_aware_macros_impl/src/lib.rs
+++ b/crates/thread_aware_macros_impl/src/lib.rs
@@ -111,11 +111,10 @@ fn add_bounds(input: &DeriveInput, root_path: &Path) -> syn::Result<syn::Generic
     // is a recursive self-reference, so the inferred `Children<T>: ThreadAware` becomes circular.
     // Writing `#[thread_aware(bound = "T: ThreadAware")]` restores the parameter bound instead.
     if let Some(bounds) = parse_container_attrs(&input.attrs)?.bound {
-        if !bounds.is_empty() {
-            let where_clause = generics.make_where_clause();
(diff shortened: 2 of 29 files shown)


psandana:
Addressed the suggestion in fcf1ce2e: `add_bounds` now seeds its emitted-predicate set from the author's own `where` clause, so a generated `where <field type>: ThreadAware` predicate that duplicates one the author already wrote (e.g. a `Vec<T>` field with `where Vec<T>: ThreadAware`) is suppressed rather than repeated. Added a snapshot pinning it (`user_where_clause_thread_aware_bound_is_not_duplicated`).

crates/thread_aware_macros_impl/src/lib.rs
```diff
@@ -104,33 +104,81 @@ pub(crate) fn param_idents() -> (syn::Ident, syn::Ident) {
...
+        // A field that is exactly a parameter the author already bounded by `ThreadAware` needs
+        // no generated predicate: emitting one would duplicate the author's own bound and trip
+        // `clippy::trait_duplication_in_bounds` at their declaration.
+        if let Some(param) = as_bare_param(bound_ty, &generic_idents)
+            && param_has_thread_aware_bound(&generics, param, &thread_aware_path)
+        {
+            continue;
+        }
+
+        predicates.push(parse_quote!(#bound_ty: #thread_aware_path));
```
wukchung: &#x1F916; **Recursive generic types no longer derive: the emitted predicate is self-referential**

When a relocated field''s type mentions the type being derived, `where <field type>: ThreadAware` *is* the impl being defined, so trait selection cycles.

```rust
#[derive(ThreadAware)]
pub struct Node<T> {
    value: T,
    children: Vec<Node<T>>,
}
```

Expansion at this commit:

```rust
impl<T> ::thread_aware::ThreadAware for Node<T>
where
    T: ::thread_aware::ThreadAware,
    Vec<Node<T>>: ::thread_aware::ThreadAware,   // requires Node<T>: ThreadAware
{ /* ... */ }
```

`assert_ta::<Node<u32>>()` then fails with `error[E0275]: overflow evaluating the requirement Vec<Node<u32>>: ThreadAware`. Enums are hit the same way: `enum List<T> { Nil, Cons(T, Box<List<T>>) }` gives `E0275 ... Box<List<u32>>: ThreadAware`.

Both compile at the merge base (`efd8178`), where the header was `impl<T: ThreadAware>`: a bound on the *parameter* bottoms out at the concrete argument, so the body''s `Vec<Node<T>>: ThreadAware` obligation is discharged inductively from the impl under construction. Verified by building the same source against both revisions.

Worth noting that the same type spelled `Vec<Self>` still derives and instantiates, because `Self` is not a generic ident so no predicate is emitted - identical types, opposite outcomes, decided purely by spelling.

Suggested direction: skip the field-type predicate when the field type names the deriving type (or `Self`) and fall back to bounding the parameters that field reaches. A `derive_compiles.rs` case that actually requires `Node<u32>: ThreadAware` would pin this down - nothing in the current suite exercises a recursive generic.

crates/thread_aware_macros_impl/src/lib.rs
```diff
@@ -172,80 +220,116 @@ fn is_same_trait(candidate: &Path, emitted: &Path) -> bool {
...
+                    .iter()
+                    .any(|b| matches!(b, syn::TypeParamBound::Trait(t) if is_same_trait(&t.path, thread_aware_path))))
+    })
+}
+
+/// Reports whether `ty` reaches a generic parameter through a shape the generated body relocates
+/// through: a path's type arguments, a reference, a tuple, an array, or a `Group`/`Paren` wrapper.
+///
+/// When it does, the field's `ThreadAware`-ness depends on that parameter and the impl owes
+/// `<field type>: ThreadAware`. The shapes deliberately left out - `Slice`, `Ptr`, `BareFn`,
```
wukchung: &#x1F916; **A slice-bearing sibling field loses the bound it previously got for free**

`type_reaches_param` does not traverse `Type::Slice`, so `Box<[T]>` reaches no parameter and owes no predicate. Under the old per-parameter model that was harmless: a sibling `Vec<T>` field emitted `T: ThreadAware` on the impl, which also discharged the slice field''s obligation. A field-type predicate does not do that - `Vec<T>: ThreadAware` does not imply `T: ThreadAware` inside the body.

```rust
#[derive(ThreadAware)]
pub struct Mixed<T> {
    values: Vec<T>,
    rest: Box<[T]>,
}
```

Expansion at this commit:

```rust
impl<T> ::thread_aware::ThreadAware for Mixed<T>
where
    Vec<T>: ::thread_aware::ThreadAware,   // nothing emitted for `rest`
{ /* ... */ }
```

This fails with `error[E0277]: the trait bound T: ThreadAware is not satisfied` (`required for [T] to implement ThreadAware`, then `Box<[T]>`), plus the matching `T: Send` error. It compiles at the merge base `efd8178`. Verified against both revisions.

A lone `Box<[T]>` field fails on both sides, so the traversal gap itself is pre-existing - but this shape is a regression, because the sibling bound that used to cover it is gone.

Relatedly, the comment just above states that the omitted shapes are ones "the rest have no impl at all". That is not true of `Slice`: `thread_aware_core`''s `impls.rs` has `impl<T: ThreadAware> ThreadAware for [T]`, a *conditional* impl - precisely the case the maintenance rule removed in this diff said required a matching arm here. Adding a `Type::Slice` arm makes `Mixed` emit `where Box<[T]>: ThreadAware` and compile again.

crates/thread_aware_macros_impl/src/lib.rs
psandana: Great catch — fixed in 535e5e73. A field whose type names the deriving type (or `Self`) now bounds the **parameters it reaches** rather than the field type, so `Node<T>` / `List<T>` derive again without the `E0275` overflow — the parameter bound bottoms out at the concrete argument. Added a `derive_compiles.rs` regression covering a recursive struct and enum, plus a snapshot. `Vec<Self>` still emits nothing (it reaches no parameter), which stays correct: the impl is then unconditional.

crates/thread_aware_macros_impl/src/lib.rs
psandana: Fixed in 535e5e73 — `type_reaches_param` now traverses `Type::Slice`, so `Box<[T]>` emits `where Box<[T]>: ThreadAware` and the sibling case compiles again. Also corrected the stale comment that claimed the omitted shapes have no impl (`[T]` has a conditional one). Added a `SliceSibling` case to `derive_compiles.rs` and a snapshot.

crates/thread_aware/src/lib.rs
```diff
@@ -204,18 +204,22 @@ mod thread;
...
+///   body actually has, discharged by that field type's own impl;
 /// * if any field carries `#[thread_aware(skip)]`, a `where Self: Send` predicate is added,
 ///   since the `ThreadAware: Send` supertrait still has to hold.
 ///
-/// A `PhantomData<..>` field is no exception: it relocates through the no-op
-/// `impl<T: ?Sized + Send> ThreadAware for PhantomData<T>`, and a parameter the traversal
-/// reaches inside it - the `U` of `PhantomData<U>` - is bound like any other. The traversal
-/// does not enter a function pointer, so writing the payload as one carries the parameter for
-/// variance, stays `Send` for every argument, and emits no bound at all:
+/// The predicate lands on the field type as written, not on the parameters inside it. A
```
martintmk: this reads like AI mumbo-jumbo. Make it more concise and understandable for humans?

martintmk (approved):

crates/thread_aware/src/lib.rs
psandana: Good call — rewrote it (19088864). Trimmed the mechanism-heavy prose down to the three things a user actually needs: one `where <field type>: ThreadAware` bound per relocated field, `where Self: Send` for skipped fields, and the `PhantomData<fn(*const T)>` marker escape hatch. Also updated the branch onto current `main`. Thanks for the approve!

crates/thread_aware/src/lib.rs
```diff
@@ -161,18 +161,29 @@ mod thread;
...
 /// # Generic Bounds
-/// Every field except a `#[thread_aware(skip)]` one is relocated, and the bounds follow from
-/// that:
-/// * a generic type parameter the traversal reaches through a relocated field receives a
-///   `::thread_aware::ThreadAware` bound;
-/// * if any field carries `#[thread_aware(skip)]`, a `where Self: Send` predicate is added,
-///   since the `ThreadAware: Send` supertrait still has to hold.
+/// The derive adds one bound per relocated field: `where <field type>: ThreadAware`. A
+/// `#[thread_aware(skip)]` field isn't relocated, so it contributes `where Self: Send` instead,
+/// which is all the `ThreadAware: Send` supertrait needs.
```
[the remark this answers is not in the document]
psandana: Fixed in 8cb57699:
- Dropped the "one bound per relocated field" wording (misleading given the dedup and the recursive-type fallback). It now says the derive bounds each relocated field by its own type, and calls out the exception in plain language: a field that refers back to the type being derived bounds the parameters it reaches instead, so recursive types still compile.
- Removed the duplicated `Marked<T>` example and the duplicated "bare `ThreadAware` bound" sentence the previous edit left behind — one of each now.

crates/thread_aware_macros_impl/src/lib.rs
```diff
@@ -104,33 +104,107 @@ pub(crate) fn param_idents() -> (syn::Ident, syn::Ident) {
...
 
-        if usage.relocated.contains(&ty_param.ident) {
-            let already = ty_param
-                .bounds
+        // Choose what to bound. A field whose type names the type being derived (or `Self`) cannot
+        // be bounded by its own type: `where <field type>: ThreadAware` would be a bound on the impl
+        // under construction and the trait solver overflows on it. Bound the parameters it reaches
+        // instead - those bottom out at the concrete argument, proving the recursive impl
+        // inductively. Every other field is bounded by its own type.
+        let targets: Vec<Type> = if type_reaches_ident(field_ty, &self_idents) {
```
Vaiz, line 183: 🤖 **Clawpilot here!** Automated review comment from an AI agent -- not a human reviewer. Please verify before acting.

AI reviewers: Claude Opus 5, GPT-6 Astra

The recursion guard misses aliases: with `type Children<T> = Vec<Node<T>>;` and `struct Node<T> { value: T, children: Children<T> }`, it emits `Children<T>: ThreadAware`. This creates a circular obligation through `Vec<Node<T>>`, whereas the previous derive requires only `T: ThreadAware`. Please provide a way to replace inferred bounds without skipping relocation, and add a compile regression for this alias case.

Vaiz:
🤖 **Clawpilot here!** Automated review comment from an AI agent -- not a human reviewer. Please verify before acting.

AI reviewer: Claude Opus 5

The description says the traversal shape-set is unchanged, but this PR adds `Type::Slice` traversal so `Box<[T]>` receives its required bound. Please update the “Change” section to mention this addition.

Vaiz:
@psandana, could you provide a realistic example of what exactly doesn't work with current version

crates/thread_aware_macros_impl/src/lib.rs
psandana: Good catch — a proc-macro can't resolve the alias, so it genuinely can't see that `Children<T>` is `Vec<Node<T>>` and self-referential. Added the escape hatch Clawpilot suggested: a container attribute `#[thread_aware(bound = "...")]` that replaces the inferred `where` predicates with the author's own (bdef8810). For the alias case:

```rust
type Children<T> = Vec<Node<T>>;

#[derive(ThreadAware)]
#[thread_aware(bound = "T: ThreadAware")]
struct Node<T> { value: T, children: Children<T> }
```

emits `impl<T> ThreadAware for Node<T> where T: ThreadAware` instead of the circular `Children<T>: ThreadAware`, and still relocates `children`. Added a `derive_compiles.rs` regression (`alias_hidden_recursion_compiles_with_bound_override`) that compiles and relocates it, a macro snapshot, and unit tests for the attribute parser.

psandana:
You're right — the description was stale. This PR does add `Type::Slice` traversal (so `Box<[T]>` gets its bound). Updated the "Change" section to call that out explicitly, alongside the recursive-generic handling and the new `bound` escape hatch.

psandana:
Sure. The motivating case is a wrapper that implements `ThreadAware` **unconditionally** (ignoring its type parameter) — the merge-base derive rejects deriving through it for arguments the wrapper itself never constrains.

```rust
use std::rc::Rc;
use core::marker::PhantomData;
use thread_aware::ThreadAware;

// ThreadAware for every T — it holds no T, so relocation is a no-op.
struct Wrapper<T>(PhantomData<fn() -> T>);
impl<T> ThreadAware for Wrapper<T> {
    fn relocate(&mut self, _: Option<&thread_aware::Thread>, _: &thread_aware::Thread) {}
}

#[derive(ThreadAware)]
struct Outer<T>(Wrapper<PhantomData<T>>);

fn main() {
    // Rc<()> is neither Send nor ThreadAware.
    fn assert_ta<X: ThreadAware>() {}
    assert_ta::<Outer<Rc<()>>>();
}
```

**Merge base (per-parameter bounds):** the derive walks into `Wrapper<PhantomData<T>>`, reaches `T`, and emits `impl<T: ThreadAware> ThreadAware for Outer<T>`. `Outer<Rc<()>>` is then rejected with `Rc<()>: !ThreadAware` — even though the only field, `Wrapper<PhantomData<T>>`, is `ThreadAware` for **every** `T` and asks nothing of it.

**This PR (field-type bound):** it emits `where Wrapper<PhantomData<T>>: ThreadAware`, which the wrapper's own blanket impl satisfies, so `Outer<Rc<()>>` compiles. That's the exact obligation the generated body has (it only relocates the whole field).

This isn't hypothetical: it's the shape a lot of typed-handle / phantom-typed wrappers take (an id or handle that is generic for type-safety but carries no `T`). `derive_compiles.rs::field_type_bound_lets_an_unconditional_wrapper_derive_for_any_argument` pins it. The story that motivated it is AB#7783612 (raised by Sander in review of #678).

martin-kolinek (approved):

Vaiz (changes requested):
> Sure. The motivating case is a wrapper that implements `ThreadAware` **unconditionally** (ignoring its type parameter) — the merge-base derive rejects deriving through it for arguments the wrapper itself never constrains.
> 
> ```rust
> use std::rc::Rc;
> use core::marker::PhantomData;
> use thread_aware::ThreadAware;
> 
> // ThreadAware for every T — it holds no T, so relocation is a no-op.
> struct Wrapper<T>(PhantomData<fn() -> T>);
> impl<T> ThreadAware for Wrapper<T> {
>     fn relocate(&mut self, _: Option<&thread_aware::Thread>, _: &thread_aware::Thread) {}
> }
> 
> #[derive(ThreadAware)]
> struct Outer<T>(Wrapper<PhantomData<T>>);
> 
> fn main() {
>     // Rc<()> is neither Send nor ThreadAware.
>     fn assert_ta<X: ThreadAware>() {}
>     assert_ta::<Outer<Rc<()>>>();
> }
> ```
> 
> **Merge base (per-parameter bounds):** the derive walks into `Wrapper<PhantomData<T>>`, reaches `T`, and emits `impl<T: ThreadAware> ThreadAware for Outer<T>`. `Outer<Rc<()>>` is then rejected with `Rc<()>: !ThreadAware` — even though the only field, `Wrapper<PhantomData<T>>`, is `ThreadAware` for **every** `T` and asks nothing of it.
> 
> **This PR (field-type bound):** it emits `where Wrapper<PhantomData<T>>: ThreadAware`, which the wrapper's own blanket impl satisfies, so `Outer<Rc<()>>` compiles. That's the exact obligation the generated body has (it only relocates the whole field).
> 
> This isn't hypothetical: it's the shape a lot of typed-handle / phantom-typed wrappers take (an id or handle that is generic for type-safety but carries no `T`). `derive_compiles.rs::field_type_bound_lets_an_unconditional_wrapper_derive_for_any_argument` pins it. The story that motivated it is AB#7783612 (raised by Sander in review of #678).


that was done on purpose. if you want to make your type `ThreadAware`, please use correct form of `PhantomData` that doesn't depends on `T`

for example `PhantomData<fn() -> T>`

crates/thread_aware_macros_impl/src/lib.rs
```diff
@@ -172,80 +260,119 @@ fn is_same_trait(candidate: &Path, emitted: &Path) -> bool {
...
+                if targets.contains(&segment.ident) {
+                    return true;
                 }
                 if let PathArguments::AngleBracketed(ab) = &segment.arguments {
                     for arg in &ab.args {
-                        if let syn::GenericArgument::Type(t) = arg {
-                            collect_generics_in_type(t, generic_idents, acc)?;
+                        if let syn::GenericArgument::Type(t) = arg
+                            && type_reaches_ident(t, targets)
+                        {
```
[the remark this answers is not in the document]
psandana: Good catch — real bug. Fixed in 3a18647a: `type_reaches_ident` now traverses `TypePath::qself`, so a field written as `<T as Provider>::Item` is seen to reach `T` and gets `where <T as Provider>::Item: ThreadAware`. Added a snapshot and a `derive_compiles.rs` regression that relocates such a field. (The self-reference check benefits too — a `<Self as Trait>::Item` field is now correctly treated as self-referential.)

psandana:
@Vaiz thanks — that's a fair point, and you're right that `PhantomData<fn() -> T>` is a workaround: written that way the merge-base traversal doesn't enter the fn pointer, so no `T: ThreadAware` bound is emitted and `Outer<Rc<()>>` compiles there too.

Where I'd push back: relying on that asks the author to know an arcane rule — "spell your phantom as `fn() -> T` so the derive's traversal doesn't reach `T`" — to avoid a bound the generated body never actually needs. The body relocates the whole field by calling `<field type>::relocate`, so its real obligation is `<field type>: ThreadAware`, which is what this PR emits. For a wrapper that is `ThreadAware` for every `T` (a typed handle, a phantom-typed id, …), the per-parameter bound is strictly stronger than the body requires, and there's no `#[thread_aware(skip)]`-free way to say "trust the field's own impl" under the merge-base model.

I do take the point that the field-type model costs more machinery — the self-reference guard, the `bound` escape hatch for alias-hidden recursion, and the qself/slice traversal — that the per-parameter model didn't need. That's a genuine tradeoff.

@martin-kolinek approved the field-type direction and it's the shape Sander asked for when he raised this in the #678 review (AB#7783612). But you've raised a real design question, and I'd rather settle the direction explicitly than merge over a change request. Happy to discuss here or offline — and if the team would rather keep the per-parameter model and document the `fn() -> T` idiom instead, I'm fine closing this in favour of that.

psandana:
Closing as **won't fix**, per discussion with @Vaiz (Evgenii).

The case this PR addressed — a wrapper that is `ThreadAware` for every `T` being rejected when reached through a `PhantomData<T>` — has an easy author-side workaround: spell the marker as `PhantomData<fn() -> T>` so the derive's traversal doesn't reach `T`, and it derives fine under the current (per-parameter) model.

Moving the derive to field-type bounds to remove that workaround turned out to cost more than it's worth: it needs a self-reference guard, a `#[thread_aware(bound = ...)]` escape hatch for alias-hidden recursion, and qself/slice traversal — a lot of machinery for a case users can already handle. The per-parameter model stays.

Thanks @Vaiz and @martin-kolinek for the review. Abandoning the branch; AB#7783612 is closed as won't fix.

PR #742, opened 2026-09-08
docs(thread_aware): add a thread-aware authoring guide

psandana:
## What & why

Adds a `_documentation` module to `thread_aware` with a **task-oriented authoring guide**, the piece the crate''s (otherwise reference-style) docs are missing. Follows the `recoverable::_documentation` pattern and renders on docs.rs.

Addresses **7552151** (publish the thread-aware authoring guide) and folds in **7722787** (the 3S / oxidizer-spawner migration experience).

## Contents

- **Why thread-awareness exists** — the thread-per-core motivation and the "relocation is a performance cooperation, not a correctness guarantee" principle.
- **Authoring a type** — prefer the derive; `#[thread_aware(skip)]`; what the field-type bounds mean; hand-written impls.
- **Choosing an implementation** — derive vs. manual vs. [`Unaware`] vs. strategy [`Arc`], as a decision table.
- **Anti-patterns** (from the 3S migration) — `Clone` copies stored affinity instead of relocating; `#[thread_aware(skip)]` on a sole field is a silent no-op; don''t trust inherited markings; relocate the whole graph once at the boundary.
- **Testing** — the `Tracker` observation pattern (assert relocation reaches non-skipped fields and not skipped ones).
- **Debugging & telemetry** — `*.thread_mismatch`-style warnings, and why their absence doesn''t prove correctness.
- **Validating correctness** — what the compiler/derive/clippy check, and what only your tests can.

## Notes

- All four code examples are doctested and pass under **default** and **all-feature** builds.
- The "what the generated bounds mean" section describes the field-type bound model implemented in #740 (sibling story under the same feature); the two are best merged together.
- This is an initial draft — happy to adjust depth, structure, or voice to the team''s preference.

Refs AB#7552151, AB#7722787.

### DIFF
--- a/crates/thread_aware/src/_documentation/mod.rs
+++ b/crates/thread_aware/src/_documentation/mod.rs
@@ -68,8 +68,9 @@
 //! ```
 //!
 //! `skip` is a claim that a field genuinely has nothing to rebind. It is not an escape hatch for
-//! "this field does not implement `ThreadAware` yet" - reach for [`Unaware`](crate::Unaware) or
-//! [`Arc`](crate::Arc) for that, so the intent is visible in the type.
+//! "this field does not implement `ThreadAware` yet" - reach for [`Unaware`](crate::Unaware) or the
+//! strategy-partitioned `Arc` (with the `std` feature) for that, so the intent is visible in the
+//! type.
 //!
 //! ## What the generated bounds mean
 //!
--- a/crates/thread_aware/src/_documentation/mod.rs
+++ b/crates/thread_aware/src/_documentation/mod.rs
@@ -26,9 +26,9 @@
 //! ## Prefer the derive
 //!
 //! In almost all cases, implement [`ThreadAware`](crate::ThreadAware) with
-//! [the derive macro](derive@crate::ThreadAware). It generates a
-//! [`relocate`](crate::ThreadAware::relocate) that forwards the notification to every field, which
-//! is exactly what a compound type owes its parts:
+//! [the derive macro](https://docs.rs/thread_aware/latest/thread_aware/derive.ThreadAware.html). It
+//! generates a [`relocate`](crate::ThreadAware::relocate) that forwards the notification to every
+//! field, which is exactly what a compound type owes its parts:
 //!
 //! ```rust
 //! use thread_aware::{Thread, ThreadAware};
@@ -68,7 +68,8 @@
 //! ```
 //!
 //! `skip` is a claim that a field genuinely has nothing to rebind. It is not an escape hatch for
-//! "this field does not implement `ThreadAware` yet" - reach for [`Unaware`](crate::Unaware) or the
+//! "this field does not implement `ThreadAware` yet" - reach for
+//! [`Unaware`](https://docs.rs/thread_aware/latest/thread_aware/struct.Unaware.html) or the
 //! strategy-partitioned `Arc` (with the `std` feature) for that, so the intent is visible in the
 //! type.
 //!
@@ -77,7 +78,8 @@
 //! You rarely need to reason about this: the derive adds exactly the `ThreadAware` bounds its
 //! generated body needs and no more, so a correct type "just derives". When it matters - a generic
 //! wrapper, or a marker field that should stay bound-free - the derive's
-//! [Generic Bounds](derive@crate::ThreadAware#generic-bounds) reference has the rules.
+//! [Generic Bounds](https://docs.rs/thread_aware/latest/thread_aware/derive.ThreadAware.html#generic-bounds)
+//! reference has the rules.
 //!
 //! ## Implementing the trait by hand
 //!
@@ -105,7 +107,8 @@
 //!
 //! When several workers share a value but each should keep its *own* instance - a per-core cache, a
 //! pool you do not want contended across cores - wrap it in the strategy-partitioned `Arc<T, S>`
-//! that the crate's `std` feature provides (`thread_aware::Arc`). With the `PerThread` strategy,
+//! ([`thread_aware::Arc`](https://docs.rs/thread_aware/latest/thread_aware/struct.Arc.html), with
+//! the crate's `std` feature). With the `PerThread` strategy,
 //! relocation materializes a separate `T` for the destination worker (lazily, on first use there),
 //! so the sharing is per-worker instead of process-wide. Use `PerProcess`, which behaves as a
 //! vanilla `Arc`, when one shared instance is what you want, and `PerNumaNode` for one instance per
@@ -117,11 +120,12 @@
 //! |---|---|---|
 //! | A compound of thread-aware fields | `#[derive(ThreadAware)]` | Forwards relocation to each field. |
 //! | A field with genuine per-core behavior | a hand-written impl | Only you know what "rebind" means. |
-//! | A foreign type that carries no affinity | [`Unaware<T>`](crate::Unaware) | A `MoveAsIs<T>`: implements the trait as a no-op. |
+//! | A foreign type that carries no affinity | [`Unaware<T>`](https://docs.rs/thread_aware/latest/thread_aware/struct.Unaware.html) | A `MoveAsIs<T>`: implements the trait as a no-op. |
 //! | Shared state that should differ per worker | `Arc<T, PerThread>` (`std`) | Materializes a separate `T` per destination. |
 //! | Shared state that is the same everywhere | `Arc<T, PerProcess>` (`std`) | Behaves as a vanilla `Arc`. |
 //!
-//! [`Unaware`](crate::Unaware) wraps a value and satisfies `ThreadAware` without reacting to
+//! [`Unaware`](https://docs.rs/thread_aware/latest/thread_aware/struct.Unaware.html) wraps a value
+//! and satisfies `ThreadAware` without reacting to
 //! relocation - use it for inert, foreign, or allocation-free values that legitimately do not care
 //! which worker they are on. Wrapping a type that *does* implement the trait is discouraged: it
 //! silences that type's own relocation (a performance loss, not a correctness bug).
@@ -219,8 +223,10 @@
 //! }
 //! ```
 //!
-//! The `test-utils` feature adds a `Relocator` (`thread_aware::Relocator`) that drives relocations
-//! without hand-built [`Thread`](crate::Thread) values, which is usually what a real test wants.
+//! The `test-utils` feature adds a
+//! [`Relocator`](https://docs.rs/thread_aware/latest/thread_aware/struct.Relocator.html) that drives
+//! relocations without hand-built [`Thread`](crate::Thread) values, which is usually what a real
+//! test wants.
 //!
 //! # Validating correctness
 //!
--- a/crates/thread_aware/src/_documentation/mod.rs
+++ b/crates/thread_aware/src/_documentation/mod.rs
@@ -26,7 +26,7 @@
 //! ## Prefer the derive
 //!
 //! In almost all cases, implement [`ThreadAware`](crate::ThreadAware) with
-//! [the derive macro](macro@crate::ThreadAware). It generates a
+//! [the derive macro](derive@crate::ThreadAware). It generates a
 //! [`relocate`](crate::ThreadAware::relocate) that forwards the notification to every field, which
 //! is exactly what a compound type owes its parts:
 //!
@@ -76,7 +76,7 @@
 //! You rarely need to reason about this: the derive adds exactly the `ThreadAware` bounds its
 //! generated body needs and no more, so a correct type "just derives". When it matters - a generic
 //! wrapper, or a marker field that should stay bound-free - the derive's
-//! [Generic Bounds](macro@crate::ThreadAware#generic-bounds) reference has the rules.
+//! [Generic Bounds](derive@crate::ThreadAware#generic-bounds) reference has the rules.
 //!
 //! ## Implementing the trait by hand
 //!
@@ -103,12 +103,12 @@
 //! ## Per-worker state with `Arc`
 //!
 //! When several workers share a value but each should keep its *own* instance - a per-core cache, a
-//! pool you do not want contended across cores - wrap it in the strategy-partitioned
-//! [`Arc<T, PerThread>`](crate::Arc). Relocation materializes a separate `T` for the destination
-//! worker (lazily, on first use there), so the sharing is per-worker instead of process-wide. Reach
-//! for [`Arc<T, PerProcess>`](crate::Arc), which behaves as a vanilla `Arc`, when one shared
-//! instance is what you want, and [`Arc<T, PerNumaNode>`](crate::Arc) for one instance per NUMA
-//! node. This is also the usual bridge to a type that does not implement `ThreadAware` itself.
+//! pool you do not want contended across cores - wrap it in the strategy-partitioned `Arc<T, S>`
+//! that the crate's `std` feature provides (`thread_aware::Arc`). With the `PerThread` strategy,
+//! relocation materializes a separate `T` for the destination worker (lazily, on first use there),
+//! so the sharing is per-worker instead of process-wide. Use `PerProcess`, which behaves as a
+//! vanilla `Arc`, when one shared instance is what you want, and `PerNumaNode` for one instance per
+//! NUMA node. This is also the usual bridge to a type that does not implement `ThreadAware` itself.
 //!
 //! # Choosing an implementation
 //!
@@ -117,8 +117,8 @@
 //! | A compound of thread-aware fields | `#[derive(ThreadAware)]` | Forwards relocation to each field. |
 //! | A field with genuine per-core behavior | a hand-written impl | Only you know what "rebind" means. |
 //! | A foreign type that carries no affinity | [`Unaware<T>`](crate::Unaware) | A `MoveAsIs<T>`: implements the trait as a no-op. |
-//! | Shared state that should differ per worker | [`Arc<T, PerThread>`](crate::Arc) | Materializes a separate `T` per destination. |
-//! | Shared state that is the same everywhere | [`Arc<T, PerProcess>`](crate::Arc) | Behaves as a vanilla `Arc`. |
+//! | Shared state that should differ per worker | `Arc<T, PerThread>` (`std`) | Materializes a separate `T` per destination. |
+//! | Shared state that is the same everywhere | `Arc<T, PerProcess>` (`std`) | Behaves as a vanilla `Arc`. |
 //!
 //! [`Unaware`](crate::Unaware) wraps a value and satisfies `ThreadAware` without reacting to
 //! relocation - use it for inert, foreign, or allocation-free values that legitimately do not care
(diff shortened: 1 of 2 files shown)


crates/thread_aware/src/lib.rs
```diff
@@ -176,6 +176,10 @@ extern crate std;
 mod cell;
 mod wrappers;
 
+/// A guide to authoring thread-aware types: how to implement, test, and debug them, and the
+/// anti-patterns to avoid. See [the guide](_documentation).
+pub mod _documentation;
```
martintmk, line 139: - this should go into thread_aware_core
- rename to just `documentation` (also could you apply the same rename across all our crates)
- protect with `#[cfg(any(doc, test))]` so this module doesn't become part of our public API

psandana: Thanks! I've gated it with `#[cfg(any(doc, test))]` (705d1a97) so the module is doc/test-only and out of the normal public API, matching the existing `_documentation` modules in `recoverable` and `fetch`.

On the other two points:

- **Move to `thread_aware_core`:** this guide is written around the `thread_aware` *facade* — `#[derive(ThreadAware)]`, `Unaware`, the strategy-partitioned `Arc`, `ThreadBuilder` — none of which live in `thread_aware_core`. Placing it there would need `thread_aware` as a dev-dependency of `core` (which risks the cyclic-deps gate) and would break the intra-doc links to those facade types. Keeping it where those APIs live seems most useful to readers. If you'd instead like a separate, trait-contract-focused guide in `core`, I'm happy to split it — just let me know.
- **Rename `_documentation` → `documentation` across all crates:** glad to, but since every crate currently uses the `_documentation` name it's a repo-wide convention change; I'd rather land it as its own sweep so it isn't tangled with this guide. I'll open a follow-up unless you'd prefer it here.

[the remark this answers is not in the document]
psandana: Gated with #[cfg(any(doc, test))] in 705d1a97 to match the 
ecoverable/etch pattern — thanks for the catch.

crates/thread_aware/src/_documentation/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+//! correct if `relocate` is called at the wrong moment, called with the wrong threads, or never
+//! called at all - see [Performance vs. Correctness](crate#performance-vs-correctness). That
+//! single fact drives most of the guidance below: because nothing enforces relocation, a type that
+//! *silently* fails to relocate is the failure mode to design against.
+//!
+//! # Authoring a thread-aware type
+//!
+//! ## Prefer the derive
+//!
+//! In almost all cases, implement [`ThreadAware`](crate::ThreadAware) with the derive macro. It
```
martintmk: could we link to actual macro (in external crate)

crates/thread_aware/src/_documentation/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+//! contract it expresses. This guide is the companion *how-to*: how to make your own types
+//! thread-aware correctly, which implementation to reach for, how to test and debug the result,
+//! and the mistakes that compile cleanly yet quietly do nothing.
+//!
+//! It is written for authors who see `T: ThreadAware` in an API and need to satisfy it, and for
+//! reviewers deciding whether a `#[derive(ThreadAware)]` or a `#[thread_aware(skip)]` is the right
+//! call. The lessons in [Anti-patterns](#anti-patterns) are drawn from migrating a large
+//! production service onto an Oxidizer-backed runtime.
+//!
+//! # Why thread-awareness exists
```
martintmk, line 16: this should just link to main lib.rs docs, too much duplication

crates/thread_aware/src/_documentation/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+//!     #[thread_aware(skip)]
+//!     id: u64,
+//! }
+//! ```
+//!
+//! `skip` is a claim that a field genuinely has nothing to rebind. It is not an escape hatch for
+//! "this field does not implement `ThreadAware` yet" - reach for [`Unaware`](crate::Unaware) or
+//! [`Arc`](crate::Arc) for that, so the intent is visible in the type.
+//!
+//! ## What the generated bounds mean
```
martintmk, line 76: this might be too much detail, I myself had trouble grasping what is this traying to say

crates/thread_aware/src/_documentation/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+//! whose type mentions a generic parameter, it emits `where <field type>: ThreadAware` - the exact
+//! obligation the generated body discharges when it relocates that field. So a `Vec<T>` field
+//! yields `where Vec<T>: ThreadAware`, and a `Wrapper<T>` field yields
+//! `where Wrapper<T>: ThreadAware`, governed by that wrapper's own impl rather than by a bound on
+//! `T`. A field whose
+//! type reaches no parameter, and a marker payload behind a function pointer
+//! (`PhantomData<fn(*const T)>`), owe no bound at all. See
+//! [the derive's reference](crate::ThreadAware#generic-bounds) for the full rules.
+//!
+//! ## Implementing the trait by hand
```
martintmk, line 84: I am missing `thread_aware::Arc` guide here, when to implement this. (when we want to maintain separated PerThread instances)

crates/thread_aware/src/_documentation/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+//!
+//! When work crosses into a worker from the outside - an FFI entry, a hand-off from a foreign
+//! thread - relocate the entire long-lived dependency graph **once**, at that boundary, rather than
+//! special-casing each affinity-bearing dependency downstream. Make the graph's root `ThreadAware`
+//! (by
+//! derive) so a single `relocate` at the entry rebinds every affinity-bearing resource beneath it.
+//! Relocating a subtree while its parent was built from a stale clone (see above) is how affinity
+//! goes stale in practice.
+//!
+//! # Testing
```
martintmk, line 181: check thread_aware::Relocator (under "test-util") that could be used for relocation testing. Also too much detail too, make it more concise.

crates/thread_aware/src/_documentation/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+//!     );
+//! }
+//! ```
+//!
+//! Construct the [`Thread`](crate::Thread) values a real test needs with
+//! [`ThreadBuilder`](crate::ThreadBuilder) (available with the default `std` feature). The
+//! `test-utils` feature additionally offers a [`Relocator`](crate::Relocator) helper for driving
+//! relocations in tests.
+//!
+//! # Debugging and telemetry
```
martintmk: we don't have any answer for this right now, I would just omit this section

psandana: Removed the Debugging and telemetry section (53ca92f2).

crates/thread_aware/src/_documentation/mod.rs
psandana: Agreed - cut it down to a short pointer at the derive's Generic Bounds reference (53ca92f2).

crates/thread_aware/src/_documentation/mod.rs
psandana: Trimmed to a pointer at the crate-level Theory of Operation (53ca92f2); dropped the duplication.

crates/thread_aware/src/_documentation/mod.rs
psandana: Made Testing more concise and now lead with the `test-utils` `Relocator` helper for driving relocations (53ca92f2).

crates/thread_aware/src/_documentation/mod.rs
psandana: Added a "Per-worker state with `Arc`" section (53ca92f2): when to reach for `Arc<T, PerThread>` to keep separate per-worker instances, vs `PerProcess` / `PerNumaNode`.

crates/thread_aware/src/_documentation/mod.rs
psandana: Done in 53ca92f2 - now links to `macro@crate::ThreadAware`.

crates/thread_aware/src/lib.rs
psandana: Gated with `#[cfg(any(doc, test))]` (705d1a97). Opened **AB#7857350** for the repo-wide `_documentation` -> `documentation` rename so it can land as its own sweep. On the core move: this guide leans on the `thread_aware` facade - the derive, `Unaware`, strategy `Arc`, `Relocator` - none of which live in `thread_aware_core`, so moving it there breaks those intra-doc links (and would need `thread_aware` as a dev-dependency of `core`, risking the cyclic-deps gate). Happy to instead split out a separate, trait-contract-focused guide for `core` if you'd prefer.

crates/thread_aware/src/_documentation/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+//! [Performance vs. Correctness](crate#performance-vs-correctness)). Nothing enforces it, so the
+//! failure mode to design against is a type that *silently* fails to relocate.
+//!
+//! # Authoring a thread-aware type
+//!
+//! ## Prefer the derive
+//!
+//! In almost all cases, implement [`ThreadAware`](crate::ThreadAware) with
+//! [the derive macro](macro@crate::ThreadAware). It generates a
+//! [`relocate`](crate::ThreadAware::relocate) that forwards the notification to every field, which
```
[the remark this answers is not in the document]
psandana: Done in 2ce36aaa - switched both derive-macro links (this line and the Generic Bounds link) to the `derive@` disambiguator, matching the `internity` convention.

crates/thread_aware/src/_documentation/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+//!     );
+//!     assert_eq!(
+//!         value.skipped.relocations, 0,
+//!         "skipped fields must not be relocated"
+//!     );
+//! }
+//! ```
+//!
+//! The `test-utils` feature's [`Relocator`](crate::Relocator) drives relocations without hand-built
+//! [`Thread`](crate::Thread) values, which is usually what a real test wants.
```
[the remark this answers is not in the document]
psandana: Fixed in 2ce36aaa - reworded to name the `test-utils` feature and use a plain `Relocator` code span instead of the feature-gated intra-doc link, so rustdoc stays clean without `test-utils`.

crates/thread_aware/src/_documentation/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+//!
+//! ## Per-worker state with `Arc`
+//!
+//! When several workers share a value but each should keep its *own* instance - a per-core cache, a
+//! pool you do not want contended across cores - wrap it in the strategy-partitioned
+//! [`Arc<T, PerThread>`](crate::Arc). Relocation materializes a separate `T` for the destination
+//! worker (lazily, on first use there), so the sharing is per-worker instead of process-wide. Reach
+//! for [`Arc<T, PerProcess>`](crate::Arc), which behaves as a vanilla `Arc`, when one shared
+//! instance is what you want, and [`Arc<T, PerNumaNode>`](crate::Arc) for one instance per NUMA
+//! node. This is also the usual bridge to a type that does not implement `ThreadAware` itself.
```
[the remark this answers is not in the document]
psandana: Fixed in 2ce36aaa. The section now names the `std` feature and drops the `crate::Arc` / `crate::PerThread` / `crate::PerNumaNode` intra-doc links (which broke under `--no-default-features`) in favour of plain code spans. Verified: the guide contributes no broken-intra-doc-link warnings in either the default or `--no-default-features` doc build.

psandana:
Thanks — addressed in c90d9f4b:

- **`crate::Arc` link (line 72):** good catch, I missed one. That last `[Arc](crate::Arc)` in "Skipping a field" is now a plain code span naming the `std` feature, consistent with the rest of the guide. Verified the module contributes **zero** broken-intra-doc-link warnings in the `--no-default-features` doc build.

- **Gating `#[derive(ThreadAware)]` doctests behind `#[cfg(feature = "derive")]`:** I looked into this and don't think it applies here. `docs/feature-gated-doctests.md` scopes that rule to **non-default** features, and CI runs doctests with default features and `--all-features` — `derive` is in `default = ["std", "derive"]`, so it's present in both legs. The crate's own convention matches: the `ThreadAware` derive example in `lib.rs` (`# Example`) is not derive-shimmed, and the `std`-gated example there wraps only the `std`-dependent body (`Arc`/`PerThread`) in `# #[cfg(feature = "std")]`, using `#[derive(ThreadAware)]` freely inside. None of this guide's doctests use `std`-gated items (they import only `Thread`/`ThreadAware`, both unconditional), so no `std` shim is needed either. Adding `derive` shims would diverge from the crate's existing derive examples. Happy to add them if you'd prefer the belt-and-braces version.

martintmk (approved):

crates/thread_aware/src/lib.rs
psandana: Update on the "move to `thread_aware_core`" point: I've switched the guide's references to the facade types (the `ThreadAware` derive, `Unaware`, `Arc`, `Relocator`) to **docs.rs links** (9760e997), matching the pattern `thread_aware_core` already uses for these same types (`[arc]: https://docs.rs/thread_aware/latest/thread_aware/struct.Arc.html`, `[derive]: .../derive.ThreadAware.html`).

That removes the blocker I raised earlier: the guide no longer needs any of those types in scope, so it could move to `thread_aware_core` **without** a dev-dependency on `thread_aware` (and the dependency cycle that would create). It also fixes the feature-gated-link breakage — `Arc` is `std`-gated and `Relocator` is `test-utils`-gated, so intra-doc links to them broke under `--no-default-features`; docs.rs links resolve in every config. The core types it still links by intra-doc (`ThreadAware`, `Thread`, `relocate`) resolve from either crate.

Happy to move the module to `thread_aware_core` in this PR if you'd prefer, or land it here and move it in the rename sweep (AB#7857350). Let me know which you'd like.

crates/thread_aware/src/_documentation/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+//!
+//! In almost all cases, implement [`ThreadAware`](crate::ThreadAware) with
+//! [the derive macro](https://docs.rs/thread_aware/latest/thread_aware/derive.ThreadAware.html). It
+//! generates a [`relocate`](crate::ThreadAware::relocate) that forwards the notification to every
+//! field, which is exactly what a compound type owes its parts:
+//!
+//! ```rust
+//! use thread_aware::{Thread, ThreadAware};
+//!
+//! #[derive(ThreadAware)]
```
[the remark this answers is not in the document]
psandana: Fixed in 180583e6. Rather than shim each doctest, I gated the whole module on the `derive` feature: `#[cfg(all(any(doc, test), feature = "derive"))]`. Every example here is built around `#[derive(ThreadAware)]`, so the guide is present exactly when the derive it documents is available (default and all-feature builds) and simply absent otherwise — which is what `docs/feature-gated-doctests.md` requires. Verified the guide doctests run under default features and are absent (no failure) under `--no-default-features`.

geeknoid (approved):

crates/thread_aware/src/_documentation/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+//!         self.buffer = Vec::new();
+//!     }
+//! }
+//! ```
+//!
+//! ## Per-worker state with `Arc`
+//!
+//! When several workers share a value but each should keep its *own* instance - a per-core cache, a
+//! pool you do not want contended across cores - wrap it in the strategy-partitioned `Arc<T, S>`
+//! ([`thread_aware::Arc`](https://docs.rs/thread_aware/latest/thread_aware/struct.Arc.html), with
```
geeknoid: **The Arc guidance names an API that no longer exists** — *Correctness · Medium · High confidence*

`thread_aware` no longer exports an `Arc` or any strategy types: its current dependencies/exports are the core vocabulary and derive macro. The strategy pointer now lives at [`performables::arc::Arc`](https://github.com/microsoft/oxidizer/blob/9e6f729663aa23ec56bbf9007c02cb6bd78bb3dc/crates/performables/src/arc/mod.rs#L151-L165), whose strategies are [`PerProcess`, `PerThread`, and `PerNuma`](https://github.com/microsoft/oxidizer/blob/9e6f729663aa23ec56bbf9007c02cb6bd78bb3dc/crates/performables/src/arc/mod.rs#L24-L35), not `PerNumaNode`. Because these are external links and prose rather than compiled examples, the default/all-feature doctests do not catch this. A reader following this section gets a missing page and unresolved imports.

**Direction:** Either point this section and table explicitly at `performables::arc::{Arc, PerThread, PerProcess, PerNuma}` (including the dependency/feature requirements), or remove the strategy recommendation if this guide should cover only `thread_aware`'s public surface. Add one compile-checked example for the chosen path.

**Done when:** Every recommended Arc/strategy path resolves at this head, and a doctest imports and constructs the documented strategy names.

crates/thread_aware/src/_documentation/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+//!     #[thread_aware(skip)]
+//!     skipped: Tracker,
+//! }
+//!
+//! fn assert_reaches_the_right_fields(from: Option<&Thread>, to: &Thread) {
+//!     let mut value = UnderTest {
+//!         tracked: Tracker::default(),
+//!         skipped: Tracker::default(),
+//!     };
+//!     value.relocate(from, to);
```
geeknoid: **The testing example never executes its assertions** — *Testing · Medium · High confidence*

This doctest only defines `assert_reaches_the_right_fields`; nothing calls it. The block therefore succeeds after compilation without running `value.relocate` or either assertion. For example, deleting the `value.relocate(from, to)` statement—or generating a derive that skips `tracked`—would leave this doctest green, even though this section presents it as the way to prove relocation reaches the right fields.

**Direction:** Make the documented scenario execute in a real test context (for example, drive it with the feature-gated `Relocator` in a test that CI runs), while keeping the visible snippet aligned with that test. A compile-only helper can remain for readability, but it should not be the sole guard for the behavior being taught.

**Done when:** Removing the relocation call or changing either expected count makes a committed test fail, and the guide points to or contains that executable scenario.

geeknoid (commented):
Static-only review found two user-facing defects in the new guide: its Arc decision path names APIs that no longer exist in `thread_aware`, and its relocation-testing doctest defines assertions but never invokes them.

The remaining claims I traced hold: the guide is gated consistently with the `derive` feature; the other examples are within the default/all-feature doctest configurations; `skip`, generated bounds, `Unaware`, and `Relocator` match the current implementations. No executable security boundary or hot path changes in this documentation-only diff.

All five taxonomies were assessed statically (150 security, 162 correctness, 182 testing, 183 performance, 143 conformance). Nothing was executed. Existing review points were deduplicated, including the already-raised NUMA-allocation wording at line 101.

PR #743, opened 2026-09-08
Pin GitHub Actions to full-length commit SHAs

danfiedler-msft:
## Summary

This PR pins GitHub Actions to full-length commit SHAs for improved security and reproducibility and adds a 7 day cooldown to Dependabot configuration for GitHub Actions. This work is described in more detail at https://aka.ms/action-pinning.

## Why?

Pinning actions to commit SHAs prevents supply-chain attacks where a tag could be moved to point to malicious code. This is a recommended security best practice per the [GitHub Actions security hardening guide](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions).

This change mitigates the risk of tag retargeting to malicious code as seen in incidents like the [tj-actions/changed-files action compromise](https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised) or [codfish/semantic-release-action compromise](https://www.stepsecurity.io/blog/supply-chain-compromise-codfish-semantic-release-action) and improves the integrity and reproducibility of the CI/CD pipeline.

## What changed?

**Action pinning:** Third-party action references in `.github/workflows/` that used mutable tag-based references (e.g., `actions/checkout@v4`) have been updated to full-length commit SHAs with a version comment (e.g., `actions/checkout@<sha> # v4`) using the [pinact](https://github.com/suzuki-shunsuke/pinact) tool. References that were already pinned to a SHA, or that used immutable release tags, were left unchanged.

**Dependabot configuration:** `.github/dependabot.yml` has been updated to ensure a `github-actions` package-ecosystem section is present with a `cooldown` configuration (`default-days: 7`). If the file did not exist, it was created. If a `github-actions` section already existed, only the `cooldown` block was added or its `default-days` value was increased to 7 if it was lower. The 7-day cooldown provides a window for the community to detect and report compromised releases before they are automatically proposed as updates, reducing exposure to supply-chain attacks via newly published malicious versions.

## Is this safe to merge?

Yes. The pinned SHAs correspond to the same commits that the existing tags pointed to. No behavioral changes in action execution are introduced. You can verify the pinned SHA value using the GitHub REST API (e.g., the commit hash for `actions/checkout@v7` can be found in the `sha` property in the JSON response for `GET https://api.github.com/repos/actions/checkout/commits/v7`).

## Additional Information

For more information, please see https://aka.ms/action-pinning

### DIFF
--- a/.github/actions/anvil-setup/action.yml
+++ b/.github/actions/anvil-setup/action.yml
@@ -109,7 +109,7 @@ runs:
 
     # cargo-binstall keeps the cold bootstrap path fast.
     - name: Install cargo-binstall
-      uses: cargo-bins/cargo-binstall@v1.21.0 # pinned by tag: this release is an immutable release (GitHub locks the tag to a commit)
+      uses: cargo-bins/cargo-binstall@ead08b90bd7b2e6d81963fb9cf0b7239f66d5db4 # v1.21.0
 
     - name: Install just
       shell: bash
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -155,7 +155,7 @@ runs:
       # "immutable": true on GET /repos/taiki-e/install-action/releases/tags/v2.81.8). The tag
       # cannot be retargeted, so this reference is intentionally left unpinned to a commit SHA.
       # See https://github.com/taiki-e/install-action/releases/tag/v2.81.8
-      uses: taiki-e/install-action@v2.81.8
+      uses: taiki-e/install-action@0631aa6515c7d545823c67cfae7ef4fc7f490154 # v2.81.8
       with:
         tool: ${{ steps.expand.outputs.cargo_tools }}
 
--- a/.github/workflows/anvil-pr-impl.yml
+++ b/.github/workflows/anvil-pr-impl.yml
@@ -154,13 +154,13 @@ jobs:
       # write tokens).
       - name: Upsert anvil-semver advisory
         if: always() && github.event_name == 'pull_request' && matrix.os == 'linux' && github.event.pull_request.head.repo.full_name == github.repository && hashFiles('target/anvil/comments/semver.md') != ''
-        uses: marocchino/sticky-pull-request-comment@v3.0.5 # pinned by tag: this release is an immutable release (GitHub locks the tag to a commit)
+        uses: marocchino/sticky-pull-request-comment@5770ad5eb8f42dd2c4f34da00c94c5381e49af88 # v3.0.5
         with:
           header: anvil-semver
           path: target/anvil/comments/semver.md
       - name: Clear anvil-semver advisory
         if: always() && github.event_name == 'pull_request' && matrix.os == 'linux' && github.event.pull_request.head.repo.full_name == github.repository && hashFiles('target/anvil/comments/semver.md') == ''
-        uses: marocchino/sticky-pull-request-comment@v3.0.5 # pinned by tag: this release is an immutable release (GitHub locks the tag to a commit)
+        uses: marocchino/sticky-pull-request-comment@5770ad5eb8f42dd2c4f34da00c94c5381e49af88 # v3.0.5
         with:
           header: anvil-semver
           delete: true
@@ -216,7 +216,7 @@ jobs:
         # fails. The separate hashFiles predicates require both feature
         # configurations; one multi-pattern call would accept a partial pair.
         if: always() && matrix.os != 'windows-arm' && hashFiles('target/coverage/lcov-all-features.info') != '' && hashFiles('target/coverage/lcov-no-default.info') != ''
-        uses: codecov/codecov-action@v7.0.0 # pinned by tag: this release is an immutable release (GitHub locks the tag to a commit)
+        uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f # v7.0.0
         with:
           files: target/coverage/lcov-all-features.info,target/coverage/lcov-no-default.info
           flags: ${{ matrix.os }}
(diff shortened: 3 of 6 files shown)


.github/actions/setup/action.yml
```diff
@@ -155,7 +155,7 @@ runs:
       # "immutable": true on GET /repos/taiki-e/install-action/releases/tags/v2.81.8). The tag
       # cannot be retargeted, so this reference is intentionally left unpinned to a commit SHA.
       # See https://github.com/taiki-e/install-action/releases/tag/v2.81.8
-      uses: taiki-e/install-action@v2.81.8
+      uses: taiki-e/install-action@0631aa6515c7d545823c67cfae7ef4fc7f490154 # v2.81.8
```
Vaiz, line 158: this was kept as a tag because it is an immutable release. Is there a reason to pin it by SHA?
https://github.com/taiki-e/install-action/releases/tag/v2.81.8

Vaiz (changes requested):

.github/actions/anvil-setup/action.yml
```diff
@@ -109,7 +109,7 @@ runs:
 
     # cargo-binstall keeps the cold bootstrap path fast.
     - name: Install cargo-binstall
-      uses: cargo-bins/cargo-binstall@v1.21.0 # pinned by tag: this release is an immutable release (GitHub locks the tag to a commit)
+      uses: cargo-bins/cargo-binstall@ead08b90bd7b2e6d81963fb9cf0b7239f66d5db4 # v1.21.0
```
st-dev-gh, line 112: This release tag is already immutable, so it cannot be retargeted. What additional protection or requirement motivates replacing it with a SHA?

If the replacement is justified, please make it in `cargo-anvil` as this file is generated by `cargo-anvil` and not to be edited directly. Update the corresponding template, then regenerate this repository with the updated generator, following the [installation and generation instructions](https://github.com/microsoft/ox-tools/blob/main/crates/cargo-anvil/docs/design/README.md#5-user-experience).

.github/workflows/anvil-pr-impl.yml
```diff
@@ -154,13 +154,13 @@ jobs:
       # write tokens).
       - name: Upsert anvil-semver advisory
         if: always() && github.event_name == 'pull_request' && matrix.os == 'linux' && github.event.pull_request.head.repo.full_name == github.repository && hashFiles('target/anvil/comments/semver.md') != ''
-        uses: marocchino/sticky-pull-request-comment@v3.0.5 # pinned by tag: this release is an immutable release (GitHub locks the tag to a commit)
+        uses: marocchino/sticky-pull-request-comment@5770ad5eb8f42dd2c4f34da00c94c5381e49af88 # v3.0.5
```
st-dev-gh, line 157: This release tag is already immutable, so it cannot be retargeted. What additional protection or requirement motivates replacing it with a SHA?

If the replacement is justified, please make it in `cargo-anvil` as this file is generated by `cargo-anvil` and not to be edited directly. Update the corresponding template, then regenerate this repository with the updated generator, following the [installation and generation instructions](https://github.com/microsoft/ox-tools/blob/main/crates/cargo-anvil/docs/design/README.md#5-user-experience).

.github/workflows/anvil-scheduled-impl.yml
```diff
@@ -73,7 +73,7 @@ jobs:
         # the Codecov UI can distinguish PR-tier uploads from scheduled
         # uploads while still tracking each platform separately.
         if: always() && matrix.os != 'windows-arm' && hashFiles('target/coverage/lcov-all-features.info') != '' && hashFiles('target/coverage/lcov-no-default.info') != ''
-        uses: codecov/codecov-action@v7.0.0 # pinned by tag: this release is an immutable release (GitHub locks the tag to a commit)
+        uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f # v7.0.0
```
st-dev-gh, line 76: This release tag is already immutable, so it cannot be retargeted. What additional protection or requirement motivates replacing it with a SHA?

If the replacement is justified, please make it in `cargo-anvil` as this file is generated by `cargo-anvil` and not to be edited directly. Update the corresponding template, then regenerate this repository with the updated generator, following the [installation and generation instructions](https://github.com/microsoft/ox-tools/blob/main/crates/cargo-anvil/docs/design/README.md#5-user-experience).

.github/actions/anvil-setup/action.yml
geeknoid: **These pins modify generated outputs instead of their source** — *Correctness · Medium · High*

This file declares itself `GENERATED BY cargo-anvil. DO NOT EDIT DIRECTLY` and says to update cargo-anvil and regenerate. The same applies to `.github/workflows/anvil-pr-impl.yml` and `.github/workflows/anvil-scheduled-impl.yml`, but this PR edits all three generated outputs without changing the catalog that produces them. The overwrite is not hypothetical: the later cargo-anvil adoption in [`70bd5aaf`](https://github.com/microsoft/oxidizer/commit/70bd5aafadb9b11fc45d8977bd808066e574b8de) regenerates the same catalog outputs and emits the tag forms. Consequently, three of the new pins will disappear on the next catalog update, so the stated pinning policy is not durable.

**Direction:** change the cargo-anvil source/catalog, consume that version here, and regenerate; add a static policy check over generated output so future catalog updates cannot reintroduce non-SHA external `uses:` references.

**Done when:** a clean cargo-anvil regeneration retains these full-SHA references, and the repository check fails if an external action reference is generated in a non-SHA form.

geeknoid (commented):
I found one cross-pass durability issue: three of the six edited files are cargo-anvil generated outputs, so their pins are overwritten by regeneration rather than becoming the source of truth. The inline comment covers the affected files and a regression guard.

The resulting head tree is otherwise internally consistent for the stated pinning policy: all 55 external `uses:` references under `.github` use 40-hex commit IDs, changed references retain version comments, and the permission/secret contexts are unchanged. I did not repeat the seven existing comments about immutable tags and now-stale explanatory comments.

The Dependabot seven-day cooldown described as a change is already present identically at the merge base, so this PR does not introduce that part of the description. Also, the local inputs cannot prove that each external SHA is the same commit formerly named by its tag; live API/network verification was prohibited, so that claim remains unassessed rather than assumed.

All five taxonomies were assessed (150 security, 162 correctness, 182 testing, 183 performance, 143 conformance). This was static-only: no posting, project execution, tests, builds, benchmarks, profiles, Miri, coverage, mutation, fuzzing, or live calls.

PR #750, opened 2026-09-13
feat(metabench): add Intel VTune support alongside Linux perf

geeknoid:
## Summary

Adds a `Mode::Vtune` probe mode integrated the same way as the existing Linux `perf` mode: it measures exactly the annotated Criterion workload for every discovered benchmark case via a new `vtune.rs` control module.

Unlike `perf`'s FIFO-based enable/disable protocol, VTune's CLI exposes external pause/resume as its own synchronous subcommands (`vtune -command resume|pause -r <result-dir>`), so the worker is launched with `--start-paused` and the annotated function's `begin()`/`Drop` guard resumes/pauses the collection directly instead of polling a FIFO. Results are extracted with `vtune -report hw-events -format csv` and parsed into the same `NativeResult` shape `perf` produces, so they flow through the existing report/baseline/regression machinery unchanged.

VTune is available on both Linux and Windows (gated on the `vtune` binary itself rather than `target_os`), and is opt-in only via `--vtune` / `BENCH_ENGINE=vtune`, mirroring `perf`'s exclusion from the default engine set.

## Testing

- Argument-validation and CSV-parsing unit tests, including short-alias rejection (`-C`/`-R`/`-c`) and thousands-separated hardware-event counts.
- A unit test covering `finish_worker`'s "workload not measured" failure path.
- Fake-`vtune`-script integration tests exercising the full worker/control/report pipeline end-to-end, including a control-protocol failure path that surfaces `Error::VtuneControl`.

Verified with `cargo test -p metabench` (lib + integration), `just clippy`, `just format`, `just spellcheck`, and `just readme`.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>
Copilot-Session: 5c9ea652-e17f-4e3c-87dc-e65c8ca8b5ad


crates/metabench/src/artifact.rs
```diff
@@ -366,6 +374,83 @@ fn perf_counter_value(value: &serde_json::Value) -> Option<f64> {
...
+    for (index, line) in contents.lines().enumerate().skip(1) {
+        let line = line.trim();
+        if line.is_empty() {
+            continue;
+        }
+        let mut fields = line.split(',');
+        let Some(name) = fields.next() else {
+            return Err(Error::ArtifactFormat {
+                path: path.to_owned(),
+                message: format!("line {} does not have an event name and a count", index + 1),
```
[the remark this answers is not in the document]
geeknoid: Fixed: replaced the field-rejoining hack with a proper quote-aware CSV field splitter (`split_csv_fields`) that respects `"..."` quoting (including doubled `""` escapes). A quoted grouped count like `INST_RETIRED.ANY,"1,234,567"` now parses correctly as a single field with its thousands separators stripped before the numeric parse, while an unquoted extra field is rejected as a genuine format error instead of being silently merged. Added `split_csv_fields_honors_quoting_and_escaped_quotes`, a quoted-count regression test, and a test asserting the unquoted-grouping case is now an error.

crates/metabench/src/vtune.rs
```diff
@@ -0,0 +1,130 @@
...
+    } else {
+        Err(Error::VtuneControl(std::io::Error::other(format!(
+            "vtune -command {command} exited with status {status}"
+        ))))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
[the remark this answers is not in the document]
geeknoid: Fixed: added `#[cfg_attr(coverage_nightly, coverage(off))]` immediately above `#[cfg(test)] mod tests` in vtune.rs, matching the pattern in artifact.rs and metric_display.rs.

crates/metabench/src/artifact.rs
```diff
@@ -366,6 +374,113 @@ fn perf_counter_value(value: &serde_json::Value) -> Option<f64> {
...
+    for (index, line) in contents.lines().enumerate().skip(1) {
+        let line = line.trim();
+        if line.is_empty() {
+            continue;
+        }
+        let fields = split_csv_fields(line);
+        let [name, value_text] = fields.as_slice() else {
+            return Err(Error::ArtifactFormat {
+                path: path.to_owned(),
+                message: format!("line {} does not have an event name and a count", index + 1),
```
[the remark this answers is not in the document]
geeknoid: Fixed. `parse_vtune` now parses the header row first and locates the `Hardware Event Type` and `Hardware Event Count:Self` columns by name (new `VTUNE_EVENT_TYPE_COLUMN`/`VTUNE_EVENT_COUNT_COLUMN` constants), rather than assuming an exact 2-field row shape. Each data row's field count is validated against the header's field count so malformed/inconsistent rows are still rejected, but extra columns (sample count, events-per-sample, precision, etc.) are now tolerated and ignored by index. Also updated `fake_vtune`'s `run_report` to emit a realistic 5-column report matching this shape, so the integration tests exercise the header-driven lookup end-to-end, and added unit tests `parses_vtune_csv_report_with_additional_columns_by_header_name`, `vtune_rejects_a_header_missing_the_required_columns`, and `vtune_rejects_a_row_whose_field_count_does_not_match_the_header`. Pushed in 88088b15.

crates/metabench/examples/fake_vtune.rs
```diff
@@ -0,0 +1,87 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! A minimal stand-in for the real `vtune` CLI, used only by
+//! `tests/spawned_benchmark.rs` to exercise `runner::launch_vtune_worker` and
+//! `vtune`'s control protocol on every platform (a real `VTune` install is
+//! Linux/Windows-only and not available in this repository's test
+//! environment). Copied to `vtune`/`vtune.exe` on a temporary `PATH` entry so
+//! it stands in for the real binary.
```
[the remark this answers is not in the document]
geeknoid: Good catch. Rather than adding `/examples/**` to the packaging allowlist (examples are intentionally dev-only per the workspace's packaging policy), added a `fake_vtune_fixture_available()` check that the four vtune integration tests call first: if `examples/fake_vtune.rs` isn't present (as when running from a published tarball where it was never packaged), the tests skip with an explanatory message instead of failing on a fixture that was never meant to ship. This keeps the packaging policy intact while making `cargo test` from a published crate succeed. Pushed in 88088b15.

crates/metabench/src/runner.rs
```diff
@@ -729,6 +749,174 @@ fn validate_perf_arguments(arguments: &[OsString]) -> Result<(), Error> {
...
+/// This mirrors `launch_perf_worker`'s shape, but `VTune`'s own CLI takes the
+/// place of the FIFO-based `perf --control` protocol: the collection starts
+/// paused (`--start-paused`) and is resumed/paused around the annotated
+/// workload by [`crate::vtune::begin`]/[`crate::vtune::Guard::drop`] via
+/// `vtune -command resume|pause -r <result-dir>`, run from inside the worker
+/// process. The exact `vtune` invocation below (analysis type, knobs, and
+/// report format) is our best documented understanding of the `VTune`
+/// command-line interface and has not been verified against a live
+/// installation in this repository's environment; adjust it here if a real
+/// `VTune` version's CLI differs. See [`crate::vtune`] for the control-protocol
```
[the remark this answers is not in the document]
geeknoid: Acknowledged as a real limitation, and it's already called out explicitly in `launch_vtune_worker`'s doc comment: 'has not been verified against a live installation in this repository's environment; adjust it here if a real VTune version's CLI differs.' I don't have access to a real VTune install in this environment to validate the exact command/report contract, so I can't close this gap directly in this PR. In the meantime I strengthened the fake fixture (`fake_vtune::run_report`) to validate the *entire* `-report hw-events -r <dir> -format csv -report-output <path>` invocation shape (previously it only checked for `-report-output`), and to emit a report with the fuller, realistic column set (sample count, events-per-sample, precision, etc.) that a real VTune report includes, so at least the harness now protects the parts of the contract it can express without a real installation. Happy to add an explicit 'experimental'/unverified note to the public docs if you'd like it more prominent than the existing internal comment - let me know.

crates/metabench/src/artifact.rs
```diff
@@ -366,6 +374,113 @@ fn perf_counter_value(value: &serde_json::Value) -> Option<f64> {
...
+/// how it is produced). This is our best documented understanding of that
+/// report's shape and has not been verified against a live `VTune` install in
+/// this repository's environment; adjust it here if a real report's columns
+/// differ. Fields are split with [`split_csv_fields`] so a count that groups
+/// digits with `,` thousands separators (this too remains unconfirmed
+/// against a live install) parses correctly as long as it is quoted per CSV
+/// convention, e.g. `INST_RETIRED.ANY,"1,234,567"`; an unquoted third field
+/// is treated as a genuine format error rather than silently merged in.
+pub(crate) fn parse_vtune(path: &Path, artifact_root: &Path, identity: String, native_identity: String) -> Result<NativeResult, Error> {
+    validate_json_size(path, MAX_ARTIFACT_JSON_BYTES)?;
```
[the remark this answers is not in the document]
geeknoid: Fixed. Refactored `validate_json_size` into a thin wrapper around a new `validate_artifact_size(path, limit, format: &str)` that takes a format label for the error message. `parse_vtune` now calls `validate_artifact_size(path, MAX_ARTIFACT_JSON_BYTES, "CSV")`, so an oversized VTune report now reports a CSV-specific size error instead of the misleading 'JSON file...' message. Pushed in 88088b15.

crates/metabench/src/vtune.rs
```diff
@@ -0,0 +1,131 @@
...
+    result_dir: Option<std::path::PathBuf>,
+}
+
+impl Drop for Guard {
+    fn drop(&mut self) {
+        if let Some(result_dir) = self.result_dir.take()
+            && let Err(error) = run_command(&result_dir, "pause")
+        {
+            record_error(error);
+        }
```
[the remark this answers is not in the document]
geeknoid: Added `vtune_pause_failure_surfaces_vtune_control_error`, which sets `FAKE_VTUNE_FAIL_COMMAND=pause` (so `resume` succeeds and a `Guard` is created, but the `-command pause` call `Guard::drop` makes at workload end fails) and asserts the run fails with 'failed to control the VTune collection', distinct from the existing resume-failure test. Verified locally (passes alongside the other three vtune integration tests). Pushed in 88088b15.

crates/metabench/src/artifact.rs
```diff
@@ -366,6 +374,113 @@ fn perf_counter_value(value: &serde_json::Value) -> Option<f64> {
...
+    let contents = fs::read_to_string(path).map_err(|source| Error::ArtifactIo {
+        path: path.to_owned(),
+        source,
+    })?;
```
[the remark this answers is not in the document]
geeknoid: Same fix as the duplicate comment on this line's sibling thread: introduced `validate_artifact_size(path, limit, format)` and have `parse_vtune` pass "CSV" as the format label, so the size-limit error now reads correctly for the VTune artifact instead of saying 'JSON file'. Pushed in 88088b15.

crates/metabench/src/vtune.rs
```diff
@@ -0,0 +1,131 @@
...
+}
+
+#[inline]
+#[doc(hidden)]
+pub fn begin() -> Option<Guard> {
+    if !*ACTIVE || MEASURED.swap(true, Ordering::Relaxed) {
+        return None;
+    }
+    let result_dir = env::var_os(RESULT_DIR_ENV)?;
+    let result_dir = std::path::PathBuf::from(result_dir);
```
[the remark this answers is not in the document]
geeknoid: Fixed. `begin()` now fetches and validates `RESULT_DIR_ENV` first; if it's missing, it records an `Error::VtuneControl` and returns `None` without ever setting `MEASURED`. `MEASURED.swap(true, ...)` only happens after the env var is confirmed present, so `finish_worker()` can no longer incorrectly report success in that edge case. Pushed in 88088b15.

crates/metabench_macros_impl/src/benchmark.rs
```diff
@@ -397,8 +399,8 @@ fn expand_benchmark(arguments: BenchmarkArguments, mut function: ItemFn, metaben
 /// generated adapter cannot correctly invoke those adapter shapes or uphold
 /// an unsafe function's safety contract. Ordinary generic, const, and
 /// explicitly ABI-qualified functions are supported. Const functions run
-/// under Criterion and Gungraun but do not participate in runtime allocation
-/// or Linux perf tracking.
+/// under Criterion and Gungraun but do not participate in runtime allocation,
```
martin-kolinek, line 402: 🤖: Non-blocking — Three public engine descriptions still omit VTune: `crates/metabench_macros/src/lib.rs:184` says runtime instrumentation is allocation/perf only, `crates/metabench/src/identity.rs:6` lists four engines, and `README.md:51` retains the old crate blurb. **Please update these mirrors alongside this corrected internal macro documentation so users see the same supported-engine set.**

crates/metabench/src/vtune.rs
```diff
@@ -0,0 +1,144 @@
...
+
+#[derive(Debug)]
+pub struct Guard {
+    result_dir: Option<std::path::PathBuf>,
+}
+
+impl Drop for Guard {
+    fn drop(&mut self) {
+        if let Some(result_dir) = self.result_dir.take()
+            && let Err(error) = run_command(&result_dir, "pause")
```
martin-kolinek, line 57: 🤖: This backend cannot produce the exact per-invocation hardware-event counts it advertises. Intel's `collect-with` documentation defines `runsa` as custom hardware event-based sampling, but `MEASURED.swap` admits only the first Criterion invocation; a short workload can produce no sample, while a longer one produces an events-per-sample estimate rather than a direct count. The collection also remains active while `Guard::drop` spawns and waits for `vtune -command pause`, so process-control work extends the supposedly workload-only window.

**Please either collect enough repetitions under an explicit sampling contract, or use an in-process/counting mechanism that isolates the annotated workload before reporting these values as exact one-invocation counters.**

crates/metabench/examples/fake_vtune.rs
```diff
@@ -0,0 +1,112 @@
...
+        "1,INST_RETIRED.ANY,1234.0,1234,Yes\n",
+        "1,CPU_CLK_UNHALTED.THREAD,567.0,567,Yes\n",
+    );
+    if std::fs::write(output, report).is_err() {
+        return ExitCode::FAILURE;
+    }
+    ExitCode::SUCCESS
+}
+
+fn run_wrapped_workload(arguments: &[String]) -> ExitCode {
```
martin-kolinek, line 102: 🤖: Non-blocking — The fake CLI accepts any initial invocation containing `--`, so the end-to-end suite stays green if `-collect-with runsa`, `--start-paused`, `-result-dir`, or forwarded `--vtune-arg` values disappear. **Please validate the collection command shape here and make one success test pass a forwarded VTune argument.** The report branch already validates its contract; this would cover the other external invocation that the fake is meant to protect.

crates/metabench/tests/spawned_benchmark.rs
```diff
@@ -186,12 +187,286 @@ exit "$status"
...
+
+/// Prepends [`fake_vtune_directory`] to the current `PATH` so a spawned
+/// `metabench` worker resolves `vtune` to the fixture.
+/// The workspace's packaging allowlist deliberately keeps `/examples/**` out
+/// of the published `.crate` (examples are dev-only, per the top-level
+/// `Cargo.toml` policy), while `/tests/**` (this file) is packaged. A
+/// `cargo test` run from a published tarball therefore cannot build the
+/// `fake_vtune` fixture this file's vtune tests depend on. Skip them in that
+/// case instead of failing on a missing fixture that was never meant to ship.
+fn fake_vtune_fixture_available() -> bool {
```
martin-kolinek, line 259: 🤖: Non-blocking — The fixture-existence guard makes source-checkout CI silently pass all four VTune integration tests if `examples/fake_vtune.rs` is accidentally moved or renamed. **Please distinguish the intentional published-tarball case from a repository checkout, and assert that the fixture exists in the latter.**

martin-kolinek (approved):
🤖: Approved.

crates/metabench_macros_impl/src/benchmark.rs
geeknoid: Fixed: added Intel VTune to the engine descriptions in metabench_macros/src/lib.rs, identity.rs, and the root README so all four locations now list the same set (Criterion, Gungraun, allocation tracking, Linux perf, and Intel VTune).

crates/metabench/examples/fake_vtune.rs
geeknoid: Fixed: fake_vtune's run_wrapped_workload now validates the full collection-launch argv shape (-collect-with runsa [forwarded --vtune-arg values] --start-paused -result-dir <dir>) before the -- separator, failing with a diagnostic on any mismatch instead of accepting any invocation that merely contains --. Also added --vtune-arg -knob to vtune_measures_exact_workload_and_writes_metrics so a forwarded VTune argument is exercised through this new validation path.

crates/metabench/tests/spawned_benchmark.rs
geeknoid: Fixed: added is_full_repository_checkout() (checks for the workspace-root Cargo.toml two levels above CARGO_MANIFEST_DIR, which a published tarball never includes) and require_fake_vtune_fixture_or_skip(), which panics with a descriptive message if the fixture is missing inside a full repo checkout (signalling an accidental move/rename) but skips gracefully with an eprintln otherwise (the intentional published-tarball case). All four vtune integration tests now use this helper instead of the previous unconditional skip.

crates/metabench/src/vtune.rs
geeknoid: This is a real, inherent limitation of driving VTune's sampling profiler across a single Criterion invocation, and I've documented it explicitly rather than attempting a redesign in this PR: added a "Sampling limitations" section to vtune.rs's module doc explaining that (1) only the first Criterion invocation is measured since MEASURED admits a single resume per process, (2) VTune's counts are sampling estimates rather than perf's exact per-invocation counts and can read zero for short workloads, and (3) Guard::drop's synchronous pause round-trip extends the measured window slightly beyond the workload itself, plus a pointer to perf for callers who need exact counts. A full fix (multi-repetition aggregation or an in-process counting mechanism) would be a substantial redesign of the VTune integration's control protocol; I think that's better scoped as follow-up work once we have real-world usage data on how much this matters in practice, rather than blocking this PR, but happy to discuss further if you feel otherwise.

crates/metabench/src/lib.rs
```diff
@@ -96,10 +96,16 @@
 //!   annotated workload. Use repeated `--perf-arg` options for additional
 //!   `perf stat` arguments.
 //!
+//! - **Intel `VTune`** reports hardware event counts for one invocation of every
+//!   discovered Criterion case, using the same annotated-workload
+//!   measurement as Linux `perf`. Use repeated `--vtune-arg` options for
```
martin-kolinek: 🤖: Non-blocking — The sampling caveat added in this push is invisible to users because `vtune` is a private module, while this public description still says VTune uses the same annotated-workload measurement as Linux `perf`. Short workloads can report zero sampled events, and the pause round-trip extends the measured window. **Please state here (and in the generated README mirror) that VTune reports sampling estimates, and direct callers who need exact per-invocation counts to `perf`.** This is the public-doc follow-up to the sampling-contract thread on `vtune.rs:57`.

crates/metabench/src/artifact.rs
```diff
@@ -366,6 +374,152 @@ fn perf_counter_value(value: &serde_json::Value) -> Option<f64> {
...
+        // Real `vtune -report hw-events` output can repeat the same event
+        // name across multiple rows (for example, one row per core type or
+        // per hardware thread group); the report command does not constrain
+        // the grouping to a single row per event. Sum repeated rows into one
+        // metric instead of treating a second occurrence as malformed input.
+        metrics
+            .entry(name.clone())
+            .and_modify(|metric: &mut Metric| {
+                let existing = metric.value.as_f64();
+                *metric = Metric::float("vtune", &name, existing + value, MetricDirection::LowerIsBetter);
```
martin-kolinek: 🤖: Non-blocking — Summing two individually finite repeated counts can overflow to infinity, bypassing the finite-value check above and putting a non-finite metric into the report. **Please reject a non-finite aggregate as `ArtifactFormat`.**

crates/metabench/examples/fake_vtune.rs
martin-kolinek: 🤖: Non-blocking — The new check still accepts arbitrary arguments after the `-result-dir` value, so it does not enforce the full launch shape described above and the integration test can miss a trailing-argument regression. **Please require the result directory to be the final argument before `--`.**

geeknoid: Fixed: run_wrapped_workload now also requires -result-dir <dir> to be the final two arguments before the -- separator, so a trailing argument after the result directory now fails the fixture instead of being silently accepted.

crates/metabench/src/lib.rs
geeknoid: Fixed: added the sampling-estimate/single-invocation caveat to the public VTune engine description in lib.rs (and regenerated the README mirror via just readme), pointing callers who need exact per-invocation counts at perf.

crates/metabench/src/artifact.rs
geeknoid: Fixed: the aggregation now checks the summed value with is_finite() and returns ArtifactFormat if two individually finite counts sum to infinity, instead of storing the non-finite result. Added a regression test using two f64::MAX rows.

crates/metabench/tests/spawned_benchmark.rs
```diff
@@ -186,12 +187,316 @@ exit "$status"
...
+        }
+        directory
+    })
+}
+
+/// The workspace's packaging allowlist deliberately keeps `/examples/**` out
+/// of the published `.crate` (examples are dev-only, per the top-level
+/// `Cargo.toml` policy), while `/tests/**` (this file) is packaged. A
+/// `cargo test` run from a published tarball therefore cannot build the
+/// `fake_vtune` fixture this file's vtune tests depend on.
```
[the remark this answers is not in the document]
geeknoid: Good catch on the scope of the claim. This guard was only ever meant to harden the new vtune tests specifically, not to make the whole file tarball-safe: run()/run_target()'s --example basic/--example parameterized invocations have unconditionally depended on unpackaged examples since before this fixture existed, so the whole-suite packaging gap is pre-existing and out of scope for this PR. I've reworded fake_vtune_fixture_available's doc comment to say so explicitly, so it no longer reads as a broader compatibility claim than it is.

ralfbiedert (approved):

martin-kolinek (approved):

PR #756, opened 2026-09-14
fix(rallocator): bound bump allocations that cannot fit a chunk segment

martintmk:
## Summary

`can_fit_tracked_in_chunk_segment` admitted a bump layout whenever `size + size_of::<TrackingHeader>() + (align - 1)` fitted `BUMP_SEGMENT_SIZE`, ignoring the allocator-owned prefix that begins every chunk segment. `allocate_tracked` places the user pointer *after* that prefix, so requests landing in the gap were admitted, never fitted, and drove the retry loop to map a fresh 64 KiB chunk on every attempt until the process exhausted memory.

This is reachable from safe code with the **default** options, because `max_allocation_bytes` defaults to `BUMP_SEGMENT_SIZE`:

```rust
let heap = Heap::bump(bump::Options::new());
with_hint(&heap, || {
    let mut values: Vec<u8> = Vec::new();
    values.resize(32_713, 0); // never returns; maps chunks until memory runs out
});
```

The test-only `allocate` already used `can_fit_in_chunk_segment`, which *does* account for `size_of::<BumpChunk>()`, so only the tracked production variant was affected.

## Defect window

Each size was probed in its own process with a fresh bump heap, under an external memory/time supervisor, with the original predicate in place:

| Size (align 16) | Result | Why |
| --- | --- | --- |
| ≤ 32_704 | ok | served by a chunk's roomier second segment |
| **32_705..=32_713** | **runaway** | admitted, but fits neither segment |
| ≥ 32_714 | ok | rejected by the old predicate, fell back to the general heap |

That window is nine bytes wide, which is why uniform sampling had not reached it.

## Fix

Model the placement `allocate_tracked` actually performs, keyed to a fresh normal chunk's first segment. Because `advance_chunk` yields either a chunk's second segment (a 16-byte prefix) or a new/recycled normal chunk reset through `chunk_start` (a 32-byte aligned prefix), the first segment is the tightest placement the retry loop can be forced onto. An admitted layout therefore always fits within a bounded number of advances, and the largest requests fall back to the general heap. The measurable cost is at most 16 bytes of bump capability at the very top of the range.

A `debug_assert!` states the invariant directly rather than silently capping retries, so future arithmetic drift fails loudly instead of hanging. Restoring the defective predicate makes that assertion fire in 0.00 s, which was verified by fault injection.

## Hardening coverage

Adds 78 tests: payload, alignment, reallocation and cross-thread invariants; arena lifecycles including attachment-cache eviction and escaped allocations; private region, retirement and failure-injection paths; real container workloads through the installed global allocator; and a bounded operation-trace model shared by a seeded soak and a Bolero fuzz target.

Test design follows practices taken from mimalloc (cookie-stamped payloads verified at free), jemalloc (packing and accounting-based retention rather than RSS), snmalloc (live-set uniqueness, dirty-then-zero `calloc`), tcmalloc (realloc pattern preservation across size-class boundaries) and bumpalo (operation model with overlap and containment invariants). Boundary sizes around the chunk segment capacity are enumerated explicitly, since a nine-byte window is not reachable by sampling.

## Verification

| Check | Result |
| --- | --- |
| `cargo test -p rallocator --all-features --locked` | 175 passed, 0 failed (141 before) |
| `cargo clippy --all-targets --all-features -- -D warnings` | clean |
| `cargo fmt --check` | clean |
| Miri (lib, `hardening`, `arena_hardening`, `global_hardening`) | 94 passed, no undefined behaviour |
| libFuzzer + AddressSanitizer, 20 min | 23 752 runs, 0 crashes |
| 8 192-seed soak | 65 s, 34 MB peak RSS |

The soak returns `live_bytes` to baseline on every seed and `mapped_bytes` plateaus at 4 784 128 bytes from roughly seed 3 840, against a derived ceiling of 105 218 048 bytes. Seed 74, which originally consumed 741 691 392 bytes of private commit, now peaks at 286 720 bytes.

`seeded_soak` is `#[ignore]`d and opt-in through `RALLOCATOR_STRESS_SEEDS` and `RALLOCATOR_STRESS_START_SEED`; any scenario panic reports the exact seed to replay.

## Limitations

- The invariant assertion is debug-only, so release-build drift still needs an external timeout or memory supervisor.
- The soak bound covers allocator-reported committed backing only; it cannot bound `System` allocations, recorder buffers or virtual address reservations.
- Concurrency coverage is schedule-dependent and does not establish race freedom under all interleavings.
- No finite campaign can prove the absence of memory corruption; these checks bound specific failure modes and raise detection probability.


crates/rallocator/src/heap/bump/state.rs
```diff
@@ -486,13 +500,22 @@ pub(crate) unsafe fn release_handle(state: *mut BumpState) {
...
+/// matches aligning the addresses [`chunk_start`] produces.
 fn can_fit_tracked_in_chunk_segment(layout: Layout) -> bool {
-    layout
-        .size()
-        .max(1)
-        .checked_add(size_of::<TrackingHeader>())
-        .and_then(|size| size.checked_add(layout.align().max(align_of::<TrackingHeader>()) - 1))
-        .is_some_and(|size| size <= BUMP_SEGMENT_SIZE)
+    let alignment = layout.align().max(align_of::<TrackingHeader>());
+    align_up(size_of::<BumpChunk>(), 16)
```
martin-kolinek, line 514: 🤖: Non-blocking — The segment-prefix offset remains independently encoded in admission and placement, so a later change to `chunk_start` can again admit a layout that never fits and repeatedly maps chunks.

**Please share the prefix-alignment constant and first-segment offset calculation between `chunk_start`, both admission predicates, and the test oracle.** Today `can_fit_tracked_in_chunk_segment`, `can_fit_in_chunk_segment`, and `tracked_first_segment_capacity` each repeat `align_up(size_of::<BumpChunk>(), 16)`, while `chunk_start` computes the real pointer separately. The `32_688` assertion pins today's copied model, but it would not fail if placement alone changed; a shared helper would make the admission/placement invariant mechanical and leave the tests as a backstop.

martin-kolinek (approved):
🤖: Approved.

crates/rallocator/tests/hardening.rs
```diff
@@ -0,0 +1,369 @@
...
+mod common;
+mod scenarios;
+mod support;
+
+// SAFETY: This is the only rallocator configuration used in this test binary.
+// System remains the global allocator so test bookkeeping does not contaminate
+// the rallocator counters or inadvertently consume active allocation hints.
+static ALLOCATOR: rallocator::Rallocator = unsafe { rallocator::Rallocator::new() };
+static TEST_LOCK: Mutex<()> = Mutex::new(());
+const MAX_SCENARIO_LIVE_BLOCKS: usize = if cfg!(miri) { 4 } else { 16 };
```
martintmk, line 26: **Posted by an AI agent · Non-blocking**

**The scenario live-set limit has two names and definitions**

**Problem**
`MAX_SCENARIO_LIVE_BLOCKS` duplicates `scenarios::MAX_LIVE`, including their `cfg!(miri)` values.

**Why this matters**
Changing one definition without the other silently invalidates `scenario_backing_retention_bound`, weakening the soak retention check.

**Suggested fix**
Expose one `scenarios::MAX_LIVE_BLOCKS` constant and use it in both the scenario runner and bound calculation.

crates/rallocator/tests/scenarios/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+use allocation_hints::heaps::{Heap, bump, general};
+use allocation_hints::with_hint;
+
+use crate::common::{Block, check_disjoint};
+
+pub(crate) const MAX_OPERATIONS: usize = if cfg!(miri) { 12 } else { 96 };
+pub(crate) const INPUT_BYTES: usize = 1 + 6 * MAX_OPERATIONS;
+const MAX_LIVE: usize = if cfg!(miri) { 4 } else { 16 };
+
+pub(crate) struct Recording;
```
martintmk, line 15: **Posted by an AI agent · Nit**

**The recording scope diverges from the crate’s guard naming**

**Problem**
`Recording` changes global recorder configuration and restores it through `Drop`, while the sibling RAII type is named `RecordingGuard`.

**Why this matters**
The current name reads like captured data and obscures the type's cleanup behavior.

**Suggested fix**
Rename it to `RecordingGuard` and update its three consumers.

martintmk (commented):
**Posted by an AI agent**

Complete Review Lens coverage found two non-blocking naming and maintainability issues. Existing unresolved discussion and failing exact-head CodeQL and Codecov checks mean this is not a clean report.

Coverage included API design, correctness, tests, performance, naming, telemetry, resilience, consistency, public API, and public documentation across all 11 changed files. The paired public API and rustdoc comparisons found no public-surface changes.

Existing unresolved items remain around independently encoded segment-prefix arithmetic, two inaccurate safety/lifecycle comments, 14 open high-severity CodeQL invalid-pointer alerts, and two uncovered changed lines reported by Codecov.

Material limitations: Miri was unavailable locally (the exact-head runtime-analysis CI passed); rustdoc coverage used default features on the host target; concurrent benchmark runs were not a paired Criterion comparison.

crates/rallocator/src/heap/bump/state.rs
```diff
@@ -979,4 +1010,76 @@ mod tests {
...
+    #[test]
+    fn tracked_admission_stops_at_the_first_segment_capacity() {
+        // Pin the empirically verified align-16 boundary (capacity 32,688 with a
+        // 40-byte tracking header). A future change to the header size or the
+        // prefix arithmetic that moved this boundary is exactly the class of
+        // defect that previously admitted a layout no segment could hold.
+        assert_eq!(size_of::<TrackingHeader>(), 40);
+        assert_eq!(tracked_first_segment_capacity(16), 32_688);
+
+        for &align in &[1_usize, 16, 64, 4096] {
```
martintmk, line 1035: **Posted by an AI agent · Non-blocking**

**Tracked admission tests omit supported high-alignment boundaries**

**Problem**
This matrix stops at 4 KiB, while `with_max_alignment` accepts 8, 16, and 32 KiB and the first-segment capacity changes with alignment. The integration tests do not distinguish bump allocation from fallback to the general heap.

**Why this matters**
A regression in high-alignment admission could remain hidden because payload correctness still holds after fallback.

**Suggested fix**
Cover 8, 16, and 32 KiB at the exact first-segment capacity and one byte beyond it, and include those admitted layouts in the bounded chunk-growth test.

martintmk (commented):
**Posted by an AI agent**

Reviewed the complete pinned allocator-hardening change at `0ed0a2172322c04f269d9e72acca8d2a74f7b558` across all ten required areas. I found one new non-blocking regression-coverage gap: tracked bump admission is tested only through 4 KiB alignment even though the internal option supports alignments through 32 KiB. Existing discussions about duplicated segment-prefix arithmetic, scenario live-set constants, guard naming, the PR test-count claim, and arena reallocation wording remain applicable and were not duplicated.

The exact-head default-feature `rallocator` suite and targeted hardening tests passed. Default-feature host public API captures were byte-identical and public rustdoc had no changes. CodeQL remains red with 14 invalid-pointer annotations concentrated in allocator hardening/test paths, and Codecov reports two uncovered lines in `allocator/hardening.rs`; the required CI owners still need to disposition those checks.

martintmk (commented):
**Posted by an AI agent**

A fresh complete Review Lens pass at the pinned head found no additional findings. Existing unresolved discussion remains around shared segment-prefix arithmetic, scenario live-set constants, guard naming, high-alignment admission coverage, and two inaccurate test comments; those were not duplicated.

All ten required areas completed. The paired public API and rustdoc comparisons found no public-surface changes, and `cargo check -p rallocator` passed. Exact-head CodeQL still reports 14 invalid-pointer annotations, while Codecov reports two uncovered changed lines.

psandana (approved):

geeknoid (commented):
## Review summary

No new actionable findings. The production fix makes tracked-bump admission model the actual placement order—chunk prefix, tracking header, requested alignment, and non-zero payload—and the affected retry now terminates after at most one fresh-segment advance. I also traced root versus normal chunks, second-segment fallback, checked arithmetic, allocation failure, cleanup, and the added unsafe test scaffolding without finding a new memory-safety or state-consistency defect.

The added hardening suite exercises exact-fit/one-byte-over boundaries, bump/arena/global behavior, injected map/commit/reallocation failures, reuse, purge, and deterministic stress scenarios. Existing review discussion already covers the remaining notable concerns (shared placement arithmetic, maximum-alignment coverage, duplicated test limits, naming, inaccurate comments/claims, and CodeQL annotations), so I did not repeat them.

All five static passes were completed: security 150/150, correctness 162/162, testing 182/182, performance 183/183, and conformance 143/143 (820/820 total). This was a static-only review: no project code, tests, builds, benchmarks, profiles, Miri, coverage, mutation testing, fuzzing, or live probes were run. Generated `Cargo.lock` content was excluded per repository guidance; configuration changes were reviewed.

PR #757, opened 2026-09-14
fix: strengthen Rust safety invariants

martintmk:
## Summary

- document and harden unsafe UTF-8, executor lifetime, and thread-safety invariants
- replace an internal HTTP-body panic with the existing typed error channel
- add regression coverage for multibyte formatting across the 128-byte stack-buffer boundary
- preserve all public names, signatures, visibility, trait surfaces, and future output types

## Cargo evaluate

Ran `cargo evaluate` across all 80 workspace crates. The final report contains 1,323 findings: 23 deny, 1,245 warn, and 55 suggestion. The remaining denials require prohibited public API changes, cover pre-existing test-fixture or unsafe-abstraction audits, or apply M365-only Tower/URI policy to this open-source compatibility surface. No cargo-evaluate report, JSON evidence, HTML output, or loop journal is committed.

## Validation

The repository README, formatting, Clippy, unit tests, doctests with default and all features, build, rustdoc, Cargo sorting, spelling, licenses, cargo-deny, and SemVer checks completed successfully. Compile-based workspace checks excluded only the unchanged `msvc_spectre_libs` package because this machine does not have the Visual Studio Spectre-mitigated libraries installed.

### DIFF
--- a/crates/anyspawn/src/handle.rs
+++ b/crates/anyspawn/src/handle.rs
@@ -14,10 +14,12 @@ use performables::sync::channel::OneshotReceiver;
 /// This is returned by [`Spawner::spawn`](crate::Spawner::spawn) and implements
 /// [`Future`] to allow awaiting the task's completion.
 ///
+/// A child panic, runtime cancellation, or early result-channel close is a hard failure.
+///
 /// # Panics
 ///
-/// Awaiting a `JoinHandle` will panic if the spawned task panicked or its
-/// runtime stopped before delivering the result.
+/// Panics if the spawned task panics, its runtime cancels it, or its result channel closes
+/// before sending a result.
 pub struct JoinHandle<T>(pub(crate) JoinHandleInner<T>);
 
 pub(crate) enum JoinHandleInner<T> {
@@ -35,7 +37,7 @@ impl<T> Future for JoinHandle<T> {
             JoinHandleInner::Tokio(jh) => Pin::new(jh).poll(cx).map(unwrap_tokio_result),
             JoinHandleInner::Custom(rx) => Pin::new(rx)
                 .poll(cx)
-                .map(|res| res.expect("spawned task did not produce a result because its channel closed")),
+                .map(|res| res.expect("spawned task result channel closed early")),
         }
     }
 }
@@ -45,7 +47,7 @@ impl<T> Future for JoinHandle<T> {
 fn unwrap_tokio_result<T>(result: Result<T, tokio::task::JoinError>) -> T {
     match result {
         Ok(value) => value,
-        Err(error) => panic!("spawned task did not complete: {error}"),
+        Err(error) => panic!("spawned task did not complete: {error} (panic or cancellation)"),
     }
 }
 
--- a/crates/arty_executor/src/wake.rs
+++ b/crates/arty_executor/src/wake.rs
@@ -197,7 +197,10 @@ impl PinnedDrop for WakeSignal {
     #[cfg_attr(test, mutants::skip)] // Only used for assertions, effect-free.
     fn drop(self: Pin<&mut Self>) {
         // This is too common to do a release-mode assert.
-        debug_assert!(self.is_inert());
+        debug_assert!(
+            self.is_inert(),
+            "WakeSignal dropped while wakers are still outstanding (waker_count > 0)"
+        );
     }
 }
 
--- a/crates/arty_executor/src/builder.rs
+++ b/crates/arty_executor/src/builder.rs
@@ -68,7 +68,9 @@ impl ExecutorBuilder {
     /// # Safety
     ///
     /// The returned object must not be dropped until a call to
-    /// [`execute_cycle()`][1] returns [`CycleOutcome::Shutdown`][2].
+    /// [`execute_cycle()`][1] returns [`CycleOutcome::Shutdown`][2]. Dropping it earlier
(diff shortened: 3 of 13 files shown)


geeknoid (approved):

crates/http_extensions/src/body/mod.rs
```diff
@@ -200,9 +200,10 @@ impl HttpBody {
     /// # }
     /// ```
     pub fn into_bytes(self) -> impl Future<Output = Result<BytesView>> + Send {
-        self.into_buffered().map_ok(|body| {
+        self.into_buffered().map(|result| {
+            let body = result?;
             body.into_bytes_no_buffering()
-                .unwrap_or_else(|| unreachable!("once body is buffered, it must be a view over a byte sequence"))
+                .ok_or_else(|| HttpError::validation("buffered body did not contain an in-memory byte sequence"))
```
martintmk, line 206: **Posted by an AI agent · Non-blocking**

**`HttpBody::into_bytes` classifies an internal invariant failure as validation**

**Problem**
The new fallback uses `HttpError::validation`, even though failure after a successful `into_buffered()` means the library violated its own representation invariant; the caller supplied no invalid input.

**Why this matters**
The standard validation label feeds metrics and logging, so an implementation regression would be attributed to caller input and routed to the wrong owner.

**Suggested fix**
Return an internal/other error with a stable dedicated label and `RecoveryInfo::never()` instead of the validation classification.

crates/anyspawn/src/handle.rs
```diff
@@ -14,10 +14,12 @@ use performables::sync::channel::OneshotReceiver;
 /// This is returned by [`Spawner::spawn`](crate::Spawner::spawn) and implements
 /// [`Future`] to allow awaiting the task's completion.
 ///
+/// A child panic, runtime cancellation, or early result-channel close is a hard failure.
```
martintmk, line 17: **Posted by an AI agent · Non-blocking**

**Spawner methods under-document the expanded `JoinHandle` failure contract**

**Problem**
This type now explicitly documents runtime cancellation and early result-channel closure as hard failures, but `Spawner::spawn` and `spawn_blocking` still mention only child panic, while `spawn_anywhere` has no matching panic section.

**Why this matters**
Consumers commonly learn the returned handle's behavior from its construction method and may incorrectly assume cancellation or channel closure is distinguishable or non-panicking.

**Suggested fix**
Synchronize all three spawner methods with this complete panic contract, preferably by linking back to `JoinHandle` to prevent future drift.

crates/arty_executor/src/wake.rs
```diff
@@ -197,7 +197,10 @@ impl PinnedDrop for WakeSignal {
     #[cfg_attr(test, mutants::skip)] // Only used for assertions, effect-free.
     fn drop(self: Pin<&mut Self>) {
         // This is too common to do a release-mode assert.
-        debug_assert!(self.is_inert());
+        debug_assert!(
+            self.is_inert(),
+            "WakeSignal dropped while wakers are still outstanding (waker_count > 0)"
```
martintmk, line 202: **Posted by an AI agent · Nit**

**`WakeSignal` assertion message violates the lowercase convention**

**Problem**
The new diagnostic begins with uppercase `WakeSignal`, while this repository requires panic and assertion messages to start lowercase.

**Why this matters**
Keeping diagnostics consistent makes panic output and message review predictable across the workspace.

**Suggested fix**
Use `wake signal dropped while wakers are still outstanding (waker_count > 0)`.

martintmk (commented):
**Posted by an AI agent**

The review found one unsafe-contract defect, two non-blocking contract inconsistencies, and one diagnostic-message nit. All ten Review Lens areas completed at the pinned head; paired public API captures for all seven affected crates were unchanged, and `cargo check` passed for every changed crate.

Because this is the author's own PR, this review is COMMENT-only and casts no vote.

---

**Posted by an AI agent**

**`WakeSignal::waker` reverses the unsafe drop condition**

**Problem**
`is_inert()` returns `true` only when no wakers remain and the signal is safe to drop, but this safety contract says the owner must keep the signal alive until it returns `false`. The actual owner correctly waits for `true`.

**Why this matters**
A caller following this unsafe API documentation could drop the storage while raw wakers still reference it, allowing use-after-free and undefined behavior.

**Suggested fix**
Change `false` to `true` so the documented precondition matches `is_inert()` and the existing call sites.

geeknoid (commented):
Early feedback on this draft (static review only; nothing was executed).

I found no additional inline findings beyond the four points already recorded on the PR: the inverted `WakeSignal::waker` safety condition, the HTTP-body internal error being classified as validation, incomplete `Spawner` panic contracts, and the assertion-message casing nit. I did not duplicate those threads.

The remaining stated claims held under review: the UTF-8 fast path is sound because a successful `write_fmt` completes whole `str` segments, and the new cases cover exact 128-byte and overflow boundaries for two- and four-byte scalars; the revised `SyncHolder` and `SpanBuilder` `Sync` rationales match their shared-reference APIs and underlying thread-safe ownership contracts; and the diff preserves public names, signatures, visibility, trait surfaces, and future output types.

All five complete taxonomies were assessed over the diff and blast radius (820 classes total), including an exhaustive static pass over the touched unsafe abstractions. No tests, builds, benchmarks, profilers, Miri, coverage, mutation, fuzzing, or live probes were run.

PR #761, opened 2026-09-15
fix: update rustls for RUSTSEC-2026-0285

martin-kolinek:
🤖 Updates the locked `rustls` version from 0.23.44 to 0.23.45, which contains the fix for RUSTSEC-2026-0285. This restores cargo-deny without suppressing the advisory.

## Seismograph coverage unblock

The lockfile is an Anvil impact-analysis trip wire, so this PR runs full-workspace coverage. Linux x64 repeatedly reported `crates/seismograph_protocol/src/lib.rs:159` as uncovered even though `frame_size_boundaries_and_chunking_are_exact` executed that chunk-reading path successfully in both feature configurations.

The failure was reproduced locally under the exact Linux x64 Anvil package selection and `target/cov` layout. With the same 7,618 passing tests, LLVM's LCOV export reported the line count as 0, 1, or 2 as the coverage build/object layout changed. The line is present in both the crate's `cfg(test)` unit-test build and the normal library build used by downstream crates; only the unit-test build originally exercised it.

The suspected cause is LLVM/cargo-llvm-cov reconciliation of multiple coverage mappings for the same source line from those separately compiled crate variants. The experiments establish that the exported result varies despite equivalent test execution, but they do not establish the precise internal ordering or selection algorithm.

To unblock this PR, a public-API integration test now round-trips a response larger than the 8 KiB frame chunk through the normal library build. This makes the line covered in both compilation variants, so the exported result remains covered regardless of how the mappings are reconciled. This is a targeted workaround, not a general requirement to duplicate unit tests as integration tests; the coverage pipeline should ultimately union `cfg(test)` and normal-library coverage reliably.

Validation:
- `just anvil-deny`
- `cargo +1.96.1 clippy --package seismograph_protocol --all-targets --all-features --locked -- -D warnings`
- exact Linux x64 Anvil package selection and `target/cov` layout:
  - all features: 7,619 tests passed; line 159 covered
  - no default features: 6,306 tests passed; line 159 covered


psandana:
🤖 Pato's Pull Request Agent: The rustls 0.23.44 → 0.23.45 bump for RUSTSEC-2026-0285 looks correct and minimal. Two things worth confirming before merge:

1. Several required checks are currently red — `Tests and Coverage (linux)`, `static-analysis (ubuntu-latest / ubuntu-24.04-arm)`, `external-type-exposure`, `coverage (ubuntu-latest)` and `Anvil / llvm-cov [pr-test]`. A lockfile-only patch bump shouldn't cause these, so please confirm they're flaky/infra and unrelated rather than a regression pulled in by the new rustls (e.g. `external-type-exposure` could react to a changed re-exported rustls type).
2. The diff also downgrades `getrandom` (0.4.3 → 0.3.4) in the `tempfile` subtree as a side effect of re-resolving. That's benign resolver output, but since the stated intent is a single-crate security bump, worth a quick sanity check that no other advisory-relevant crate shifted.

Vaiz (approved):

crates/seismograph_protocol/tests/frame_chunking.rs
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Integration coverage for frame payloads that cross a read-chunk boundary.
+
+use seismograph_protocol::message::Response;
+use seismograph_protocol::{read_response, write_response};
+
+#[test]
+fn large_response_round_trip_covers_the_normal_library_chunking_path() {
```
AdomasBekeras, line 10: this does not seem related to the PR?

AdomasBekeras (approved):

crates/seismograph_protocol/tests/frame_chunking.rs
martin-kolinek: Yeah, it's a stabilization of coverage - the current main has unstable coverage where some runs don't get to 100%. Added a test that fixes that here.

PR #764, opened 2026-09-16
perf(rallocator): consolidate v1 and Seismograph improvements

ralfbiedert:
## Summary

Consolidates the Rallocator v1 performance work and Seismograph snapshot/monitor improvements: medium-span caching and reclamation, eligible in-place reallocation, remote-accounting improvements, independently owned snapshot storage, and an offline snapshot viewer.

## Public API changes

- Exposes `rallocator::tuning_telemetry` behind the existing `tuning-telemetry` feature, with explicit session control and allocation-free counter observations.
- Adds `PeakLiveBytesScope`, `Stats::peak_live_bytes_scope`, and optional lifetime/sample peak accessors. Rallocator reports aggregate-query sample peaks, not lifetime high-water marks.
- Requires allocator statistics section version 2. Legacy version 1 statistics are rejected; backward compatibility is intentionally not retained.
- Adds the documentation-hidden integration hook `seismograph::snapshot::snapshot_collection_active()`; no `snapshot_arena_active()` alias is retained.
- Adds `seismograph view <SNAPSHOT-FILE>` for read-only interactive inspection of saved captures.
- Propagates Cachet's Seismograph feature to BytesBuf when that optional dependency is present.

Existing allocator entry points and configuration remain unchanged. Eligible allocations can resize in place through the existing `GlobalAlloc::realloc` interface.

## Validation so far

- Windows: 8,211 all-feature tests passed (4 skipped), excluding only the unchanged `msvc_spectre_libs` package because this machine lacks Visual Studio Spectre-mitigated CRT libraries.
- Windows: all-workspace no-default-feature tests passed (6,904 passed, 4 skipped).
- Workspace doctests passed: 1,262 with all features (same Spectre-package exclusion), 1,057 with default features.
- Rust 1.95 tests for all 11 changed crates passed: 1,314 all-feature and 1,261 default-feature tests.
- All 36 each-feature builds passed; Rallocator all-target/all-feature debug and release builds passed.
- Workspace Clippy passed with the same Spectre-package exclusion. Formatting, spelling, README, license, dependency-policy and branch-crate unused-dependency checks passed.

Draft while validation continues. The complete local `just anvil-pr` gate stops at the missing Spectre libraries; coverage, runtime-analysis, mutation, Linux and ARM results are not yet all confirmed. Hosted CI is the next source of actionable failures.


crates/seismograph/src/monitor.rs
```diff
@@ -337,33 +445,42 @@ fn handle_client(mut stream: TcpStream, descriptor: &MonitorDescriptor, stop: &A
...
             Ok(request) => request,
-            Err(ClientError::Stopped | ClientError::Disconnected) => return Ok(()),
+            Err(ClientError::Stopped | ClientError::Disconnected | ClientError::IdleTimedOut) => return Ok(()),
             Err(error) => return Err(error),
         };
         let response = authenticated_response(&request);
-        seismograph_protocol::write_response(&mut stream, request_id, &response).map_err(ClientError::Protocol)?;
+        let snapshot_response = matches!(response, Response::Snapshot(_));
+        if snapshot_response {
+            stream.set_write_timeout(None).map_err(ClientError::Io)?;
```
geeknoid: **Snapshot writes can hold every monitor slot indefinitely** — *Correctness · Medium · High confidence*

For snapshot responses this removes the only write deadline, while `SnapshotRequestGuard` is dropped when `authenticated_response` returns—before `write_response` starts. An authenticated client can request a capture and then stop reading; that worker retains both its `ActiveClient` registration and the full snapshot buffer. Repeating this sequentially can occupy all 16 slots, after which `ActiveClients::register` rejects every new client. The idle timeout only guards the next request read, so it cannot release these blocked writers.

The guard scope is visible at https://github.com/microsoft/oxidizer/blob/82cb0f66b325ed4fe0e171ff6adb67a179b97588/crates/seismograph/src/monitor.rs#L480-L490 and the slot limit at https://github.com/microsoft/oxidizer/blob/82cb0f66b325ed4fe0e171ff6adb67a179b97588/crates/seismograph/src/monitor.rs#L247-L252.

**Direction:** keep snapshot transmission subject to a bounded progress/deadline policy (or move it behind bounded streaming), and keep the snapshot admission permit through transmission if concurrent retained responses must also be limited.

**Done when:** a client that authenticates, requests a snapshot, and stops reading is disconnected within a documented bound, releases its snapshot and slot, and cannot prevent a subsequent client from connecting.

crates/seismograph_protocol/src/lib.rs
```diff
@@ -152,15 +160,14 @@ fn read_frame(reader: &mut impl Read, maximum: usize) -> Result<Frame, Error> {
         return Err(Error::MessageTooLarge);
     }
     let mut payload = Vec::new();
+    payload.try_reserve_exact(len).map_err(|_error| Error::MessageTooLarge)?;
```
geeknoid: **A frame header can force a 4 GiB reservation before any payload arrives** — *Correctness · Medium · High confidence*

`read_response` permits lengths through `u32::MAX`, and this new eager reservation runs immediately after the 20-byte header. A truncated or malformed peer therefore makes the client reserve the entire declared snapshot before supplying even one payload byte; on constrained processes this can fail an otherwise recoverable read or consume a large virtual-address reservation. The previous chunked path grew only as bytes were actually received, so this is a behavioral regression rather than merely an implementation choice.

**Direction:** preserve incremental allocation while reading, or introduce a substantially smaller explicit protocol limit and reserve only a bounded initial chunk.

**Done when:** a maximum-length header followed by EOF/timeout does not attempt the full declared allocation, while valid large snapshot frames still decode correctly.

crates/seismograph_cli/src/commands/monitor/snapshot.rs
```diff
@@ -0,0 +1,306 @@
...
+}
+
+pub(super) fn prepare_with_progress(
+    mut decoded: seismograph::snapshot::DecodedSnapshot,
+    progress: &mut impl FnMut(Phase),
+) -> Result<Box<CapturedSnapshot>, String> {
+    let mut allocator = None;
+    let mut runtime_source = None;
+    for source in std::mem::take(&mut decoded.sources) {
+        if source.id == seismograph_rallocator::source::ID {
```
geeknoid: **Duplicate source IDs are silently resolved by last-one-wins** — *Correctness · Medium · High confidence*

Snapshot capture rejects duplicate `SourceId`s, but decoding accepts them and this loop overwrites `allocator` or `runtime_source` each time the same known ID appears. The new offline viewer therefore treats a structurally conflicting file as valid and displays whichever section occurs last, which can produce internally plausible but misleading diagnostics instead of reporting a malformed capture. The decoder currently appends every section without a uniqueness check: https://github.com/microsoft/oxidizer/blob/82cb0f66b325ed4fe0e171ff6adb67a179b97588/crates/seismograph/src/snapshot.rs#L367-L386.

**Direction:** enforce `SourceId` uniqueness while decoding the container (preferred, matching capture-time validation), or reject duplicates before assigning known sources here.

**Done when:** a file containing two allocator or two runtime sections is rejected with a clear malformed/duplicate-source error, while distinct unknown source IDs remain loadable.

geeknoid (commented):
Early static feedback on this draft found three correctness issues worth addressing before broader validation: snapshot responses can occupy all monitor client slots indefinitely, response headers can trigger an eager 4 GiB reservation, and offline files silently accept duplicate source identities with last-one-wins semantics.

The allocator unsafe/UB pass traced the changed allocation, reallocation, medium-span, remote-accounting, snapshot-storage, and platform-memory paths; I did not find a proven soundness or UB defect that met the reporting bar, and I did not duplicate the existing CodeQL pointer comments. The public API/versioning changes are consistent with the description: v1 allocator statistics are intentionally rejected, aggregate-query peaks are explicitly scoped, and existing allocator entry points remain unchanged.

All five static taxonomies were assessed (820 classes total). Nothing was executed: no builds, tests, benchmarks, profilers, Miri, coverage, mutation, fuzzing, or live probes. Generated `Cargo.lock` and generated README prose were excluded per repository instructions. Performance candidates without a defensible hot-path impact and pre-existing issues not worsened by this PR were omitted.

PR #765, opened 2026-09-16
ci(anvil): limit mutation concurrency

martin-kolinek:
## Purpose

Diagnostic change for the Linux and Linux-ARM mutation runner shutdowns observed on microsoft/oxidizer#764.

Both jobs successfully found 1,025 mutants and passed their unmutated baseline, then timed out on the same allocator mutations before the runner service received SIGTERM (exit 143). The generated Anvil recipe currently passes `--jobs 0`, which enables automatic CPU-based parallelism; the former Oxidizer mutation runner used one mutation job at a time.

## Change

Run PR diff mutation testing with `--jobs 1` instead of `--jobs 0`.

This intentionally modifies a generated Anvil recipe only for diagnosis. If the mutation matrix completes reliably, the equivalent change should be made in the upstream Cargo Anvil catalog and regenerated here rather than merging this generated-file override as the permanent fix.

## Evidence

- Failing Linux job: https://github.com/microsoft/oxidizer/actions/runs/35124397692/job/104890414339
- Failing Linux-ARM job: https://github.com/microsoft/oxidizer/actions/runs/35124397692/job/104890414405
- Both jobs passed the baseline and then ended with runner shutdown / exit 143.

## Validation

- `just --dry-run anvil-mutants-diff`
- `git diff --check`


martin-kolinek:
🤖 Diagnostic result: the runner shutdown is a genuine OOM caused by one deterministic mutant, not mutation parallelism.

Exact mutant:

```text
crates/rallocator/src/allocator.rs:1955:23:
replace &= with |= in telemetry_used_slice_indices
```

This changes `remaining &= remaining - 1`, which clears one bitmap bit per iteration, into `remaining |= remaining - 1`. The iterator therefore stops making progress and supplies an unbounded stream to snapshot collection.

Linux x64 and ARM64 both reproduced the failure with `--jobs 1` and only this single mutant selected. Streamed telemetry showed system memory rising to roughly 15.8-15.9 GiB of 15.9 GiB, available memory falling to 0-167 MiB, and `rallocator-*` processes reaching roughly 8.4-8.6 GiB plus 6.8-7.1 GiB RSS before the runner received SIGTERM/exit 143.

Earlier isolated mutations at allocator.rs:1259:57 and the following eight `allocate_medium` arithmetic mutants were all caught normally and stayed near 2.4 GiB used memory.

This means `--jobs 1` should not be upstreamed as the fix. The repository should skip or otherwise contain this runaway iterator mutant; a general Anvil improvement would be per-mutant memory containment rather than lower concurrency.

PR #767, opened 2026-09-17
perf(miri): reduce runtime analysis duration

martin-kolinek:
🤖 This narrows Miri to checks that provide memory-safety value and scales native stress/property breadth under `cfg(miri)` while preserving normal test workloads.

## Summary

- excludes safe build/integration crates and low-value test modules from their own Miri runs
- scales deep-path, sampling, allocation, synchronization, and pseudo-fuzz workloads only under Miri
- retains direct unsafe allocator, arena, recorder, callback, resolver, and concurrency coverage
- updates the generated cargo-anvil catalog to 0.11.0, including affected-package-scoped LCOV exports
- documents the runtime-analysis validation boundary and exclusion criteria

## Runtime impact

- original full runtime-analysis baseline: 5,637s (1h 33m 57s)
- best comparable optimized run: 1,772s (29m 32s), a 68.6% reduction
- latest cold run: 2,012s (33m 32s), showing expected compile/scheduling variance
- targeted package reductions include Internity 251s → 111s, HTTP Extensions 232s → 189s, Performables 217s → 41s, and WinHTTP Impl 194s → 161s

## Validation

- `just anvil-pr-fast` with the cargo-anvil 0.11.0 generated catalog
- `ANVIL_IMPACT=off just anvil-pr-runtime-analysis`
- package-scoped `just anvil-miri --package ...` checks for every changed runtime-sensitive crate
- targeted native test suites
- `just anvil-fmt`
- `just anvil-clippy`
- `just anvil-cargo-sort`
- `just anvil-readme-check`
- `just anvil-spellcheck`


justfiles/anvil/checks/llvm-cov.just
```diff
@@ -102,14 +97,17 @@ anvil-llvm-cov: anvil-llvm-cov-validate-prereqs anvil-impact
     # choose and must exist before --output-path runs.
     [System.IO.Directory]::CreateDirectory('target/coverage') | Out-Null
     function Invoke-AnvilLcovReport {
-        param([Parameter(Mandatory)][string] $OutputPath)
+        param(
+            [Parameter(Mandatory)][string] $OutputPath,
+            [Parameter(Mandatory)][string[]] $PackageArgs
```
martin-kolinek: 🤖: `justfiles/anvil/checks/llvm-cov.just` no longer matches its `.anvil.lock` checksum, so cargo-anvil treats this generated recipe as locally modified and stops updating it automatically.

**Move these `llvm-cov` fixes into the cargo-anvil catalog and regenerate this file before merging.** The file's header explicitly forbids direct edits; the lock records `sha256:662c85b7...`, while the checked-in content hashes to `sha256:5c4d7454...`. Leaving that drift in place freezes this recipe out of future catalog updates.

deny.toml
```diff
@@ -11,26 +11,30 @@ unmaintained = "all"
...
-# licenses" on https://docs.opensource.microsoft.com/legal/resources/oss-licenses-by-type/.
 allow = [
     "MIT",
     "Apache-2.0",
-    "ISC",
+    "Apache-2.0 WITH LLVM-exception",
     "BSD-2-Clause",
     "BSD-3-Clause",
+    "ISC",
+    "MPL-2.0",
```
martin-kolinek: 🤖: This catalog bump silently widens the workspace license gate to permit `MPL-2.0`.

**Keep the existing permissive-only allowlist in this PR, or split this policy widening into an explicit, independently reviewed change.** The same hunk removes the repository's rule limiting additions to Microsoft's permissive-license list, while `unused-allowed-license = "allow"` means the new weak-copyleft allowance remains latent until a future dependency uses it and then passes the gate without notice.

crates/fetch_winhttp_impl/src/request.rs
```diff
@@ -858,11 +858,17 @@ mod tests {
...
+        for (index, (uri, expected_host, expected_port)) in [
             ("https://example.com:8443/", "example.com", 8443),
             ("https://[::1]:8443/", "[::1]", 8443),
             ("https://[::1]/", "[::1]", 443),
-        ] {
+        ]
+        .into_iter()
+        .enumerate()
+        {
+            if cfg!(miri) && index == 0 {
```
martin-kolinek: 🤖: Non-blocking — These Miri reductions select cases by array index, so inserting or reordering a row silently changes which lifecycle semantics Miri exercises.

**Select the reduced cases by semantic value, or define explicit `#[cfg(miri)]` case tables that name the transitions being retained.** The six new predicates at lines 869, 898, 953, 999, 1060, and 1351 all still compile after a table edit, but can then test a different port, protocol, TLS, or upload case without any visible change to the Miri policy.

docs/runtime-analysis.md
```diff
@@ -0,0 +1,137 @@
...
+- `cachet` and `ohno` contain no runtime `unsafe` code. Their behavioral suites
+  run natively and under `cargo careful`; their dependencies retain their own
+  Miri coverage.
+- `seismograph_cli` and `seismograph_runtime` are safe integration layers over
+  `seismograph`. The unsafe recorder and snapshot implementations remain in the
+  Miri-selected `seismograph` package.
+- `rest_over_grpc` keeps its runtime transcoding tests under Miri, but omits
+  build-time descriptor and code-generation unit tests. The build half is safe
+  Rust and remains covered by native tests and `cargo careful`.
+- `rallocator` keeps its library and allocator integration suites under Miri,
```
martin-kolinek, line 70: 🤖: Non-blocking — The new Miri scope inventory is already incomplete for two crates changed by this PR.

**Document that `rallocator/tests/performables_telemetry.rs` is also skipped under Miri, and add the missing `seismograph` reductions.** The current `rallocator` entry names only `tests/telemetry.rs`, while the sole test in `performables_telemetry.rs` is now ignored; `seismograph` also reduces its sampling population, snapshot count, and default test ring capacity but has no inventory entry.

crates/rallocator/src/telemetry/core.rs
```diff
@@ -2051,12 +2068,17 @@ mod tests {
...
-            ..Default::default()
+            ..test_recorder_configuration()
         });
 
-        let selected = (0..10_000).filter(|_| begin_allocation().is_some()).count();
+        let attempts = if cfg!(miri) { 1_000 } else { 10_000 };
+        let selected = (0..attempts).filter(|_| begin_allocation().is_some()).count();
 
-        assert!((70..=130).contains(&selected), "selected {selected} allocations");
+        let expected_range = if cfg!(miri) { 5..=15 } else { 70..=130 };
```
martin-kolinek: 🤖: Non-blocking — The reduced Miri range makes this sampling test depend too strongly on the process-global allocation-ID offset, so valid sampling windows can fail the assertion.

**Widen the Miri range or retain enough attempts for the expected band to be stable across starting IDs.** `begin_allocation()` samples 1,000 consecutive values from `NEXT_ALLOCATION_ID`, whose starting point depends on earlier tests. Evaluating the actual hash over the first million possible 1,000-ID windows puts about 7.47% outside `5..=15`, with valid counts ranging from 1 to 23.

crates/rest_over_grpc/src/build/service_definition.rs
```diff
@@ -1035,7 +1035,7 @@ pub(crate) fn to_snake_case(name: &str) -> String {
     heck::AsSnakeCase(name).to_string()
 }
 
-#[cfg(test)]
+#[cfg(all(test, not(miri)))]
```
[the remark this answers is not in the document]
martin-kolinek: 🤖 No change needed: `to_snake_case` remains unconditionally available at line 1034 and all production callers still compile. This diff only changes the test-only `ServiceDefinition::generate` helper and the unit-test module from `#[cfg(test)]` to `#[cfg(all(test, not(miri)))]`.

deny.toml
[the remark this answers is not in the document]
martin-kolinek: 🤖 Accepted as the cargo-anvil 0.12 catalog standard per maintainer decision. The PR description now explicitly calls out the generated catalog migration.

.anvil.lock
```diff
@@ -1,11 +1,11 @@
 version = 1
 tool = "anvil"
-tool_version = "0.9.0"
-catalog_checksum = "sha256:baa2a5e479fe8f466e8f54e71ef0c498e07ac225b59a61a76ee99d46d5498650"
+tool_version = "0.12.0"
+catalog_checksum = "sha256:b097d6ce3927a9bf4acb84ec0d17f18d32a4fc0fd46ea999ea14feab96634dd9"
```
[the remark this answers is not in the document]
martin-kolinek: 🤖 Fixed in 5b6c38a6 and the updated PR description: this PR intentionally upgrades the generated cargo-anvil catalog from 0.9.0 to 0.12.0. Regeneration is clean (`cargo anvil --dry-run`: 164 unchanged), and `llvm-cov.just` is catalog-owned.

geeknoid (commented):
I found no additional reportable findings after the existing review threads. The revised selection scheme now uses explicit per-case Miri markers, documents every package/test reduction, and leaves native workloads unchanged.

The core safety-coverage claims held under static tracing: routerama keeps vector-boundary cases around each unsafe load width; multitude preserves allocation thresholds by lowering configurable boundaries rather than removing them; seismograph keeps minimum-capacity recorder lifecycles while capacity-specific tests remain explicit; and performables retains every synchronization role plus repeated transitions. Package exclusions are limited to safe integration/code-generation layers while the unsafe owner crates remain selected.

I did not repeat the 11 existing inline threads, including the catalog/license discussion and earlier Miri inventory, selection, and sampling concerns. The reported runtime figures were not independently reproduced; the review only verified that the code and documentation support the claimed reduction mechanism.

This was a static-only review across all five complete taxonomies; no project code, tests, builds, benchmarks, profilers, Miri, coverage, mutation, fuzzing, network calls, or live probes were executed.

crates/ohno/Cargo.toml
```diff
@@ -17,6 +17,11 @@ homepage = { workspace = true }
 include = { workspace = true }
 repository = "https://github.com/microsoft/oxidizer/tree/main/crates/ohno"
 
+# Ohno and its generated API contain no runtime unsafe code. Native tests and
```
Vaiz: instead of having a comment we should use `[forbid(unsafe_code)]` in `lib.rs`. That way it's a guarantee instead of a promise

crates/cachet/Cargo.toml
```diff
@@ -17,6 +17,12 @@ homepage.workspace = true
 include.workspace = true
 repository = "https://github.com/microsoft/oxidizer/tree/main/crates/cachet"
 
+# Cachet contains no runtime unsafe code. Its suite exercises cache behavior and
+# async orchestration, which remain covered by native tests and cargo-careful;
+# unsafe dependencies retain their own Miri suites.
+[package.metadata.anvil.miri]
```
psandana: 🤖 Pato's Pull Request Agent: The Miri `exclude = true` for `cachet`, `seismograph_cli`, `seismograph_runtime`, and `routerama_build` all rest on a hand-written comment asserting "no runtime unsafe code". The workspace only sets `unsafe_op_in_unsafe_fn = "warn"`, so nothing enforces that claim — a future `unsafe` block would silently drop out of Miri coverage while the comment still reads clean.

As already suggested for `ohno` (discussion_r4094669727), turn the promise into a guarantee by adding `#![forbid(unsafe_code)]` to each of these crates' `lib.rs` (or `main.rs` for `seismograph_cli`). Then the exclusion is self-documenting and compiler-enforced. Non-blocking.

crates/ohno/Cargo.toml
martin-kolinek: 🤖 Added `#![forbid(unsafe_code)]` to `ohno` in 80cf479b, so the package-level Miri exclusion is now backed by a compiler-enforced invariant.

crates/cachet/Cargo.toml
martin-kolinek: 🤖 Addressed in 80cf479b for all five package-level exclusions. `ohno`, `routerama_build`, `seismograph_cli`, and `seismograph_runtime` now use unconditional `#![forbid(unsafe_code)]`. `cachet` uses `#![cfg_attr(not(test), forbid(unsafe_code))]`: its unit-test-only `testing_aids::init_tracing!()` expands through `ctor`, whose generated code carries `allow(unsafe_code)` and cannot coexist with an unconditional forbid. The shipped Cachet library is still compiler-prohibited from introducing unsafe code. The policy and this test-harness exception are documented in `docs/runtime-analysis.md`.

st-dev-gh (approved):
[copilot speaking] Non-blocking, two documentation/consistency nits:

1. `docs/runtime-analysis.md:105-108` says `rest_over_grpc`'s runtime tests provide Miri coverage for the transcoding buffer path "over `bytes` and `hyper`". The crate has no `hyper` dependency in either `[dependencies]` or `[dev-dependencies]`, so Miri never interprets hyper code here; the covered path is `bytes` plus `http-body`/`http-body-util`. Suggest naming those instead.

2. `docs/runtime-analysis.md:40-41` states recorder tests use the minimum supported event-ring capacity unless ring sizing or wraparound is under test. The new `recorder::test_configuration()` helper is applied at four sites (`lib.rs:244`, `recorder.rs:1253`, `recorder.rs:1343`, `snapshot.rs:1224`), but around ten other recorder setups that neither test capacity nor set it explicitly still use `..Default::default()` — for example `recorder.rs:1277`, `:1321`, `:1374`, `:2011`, and `:2163`. Either apply the helper there or scope the wording to the cases you intend.

Minor: the `:57` lead-in calls the list "package-level exclusions", but it also covers test-level ignores and workload reductions.

Vaiz (approved):
approved, though that feels like a lot of tinkering, maybe we need a better solution

PR #769, opened 2026-09-18
chore: align lint configuration with cargo-anvil

martin-kolinek:
🤖 Aligns Oxidizer's lint configuration with cargo-anvil's namespace-specific managed-region layout from microsoft/ox-tools#191.

The Rust, rustdoc, and Clippy catalogs now live in separate explicit subtables, so `cargo sort --grouped` treats them as stable sorting units. Oxidizer's repository-specific `missing_docs`, `unreachable_pub`, and `clippy::panic` rules remain as bare keys immediately after their namespace's closing sentinel. `doc-valid-idents` remains repository-owned Clippy behavior configuration.

This also includes the upstream `empty_structs_with_brackets` warning and `literal_string_with_formatting_args` allow override. The lockfile records the forthcoming region IDs and checksums. Until cargo-anvil 0.12.0 from microsoft/ox-tools#191 is merged and adopted, the currently installed 0.9.0 catalog cannot reconcile these new regions; that temporary ordering gap is intentional.


Cargo.toml
```diff
@@ -396,39 +423,5 @@ clippy.should_panic_without_expect = "allow"
...
-
-# Profile used by `cargo bolero` (it passes `--profile fuzz` by default).
-# Optimized build with single codegen unit, as recommended by cargo-bolero
-# to avoid known compilation bugs in the libfuzzer instrumentation.
-[profile.fuzz]
-inherits = "dev"
-opt-level = 3
-incremental = false
-codegen-units = 1
+# <<< anvil-managed: anvil-workspace-lints
```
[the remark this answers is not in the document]
martin-kolinek: 🤖 No lockfile update is needed. The managed bodies now exactly match the existing cargo-anvil 0.9 catalog render, so their recorded checksums remain valid; this change only moves repository-owned settings outside the sentinels and removes duplicated catalog text. `cargo anvil --dry-run` reports both regions unchanged with no manifest update required.

st-dev-gh (approved):

Cargo.toml
```diff
@@ -317,102 +317,101 @@ zerocopy = { version = "0.8.26", default-features = false }
...
+multiple_unsafe_ops_per_block = "warn"
+redundant_type_annotations = "warn"
+renamed_function_params = "warn"
+semicolon_outside_block = "warn"
+undocumented_unsafe_blocks = "warn"
+unnecessary_safety_comment = "warn"
+unnecessary_safety_doc = "warn"
+unneeded_field_pattern = "warn"
+unused_result_ok = "warn"
+unwrap_used = "warn"
```
psandana: 🤖 Pato's Pull Request Agent: This managed clippy block drops `too_long_first_doc_paragraph = "warn"`, which was present before this change (it only appears as a deletion in the diff and is absent from the head file). The PR description lists what 0.12.0 *adds* (`empty_structs_with_brackets`, `literal_string_with_formatting_args`) but not this removal, so it reads as an unintended loss of lint coverage. Please confirm the 0.12.0 catalog deliberately dropped `too_long_first_doc_paragraph` — if the repo still wants it, it can be re-added as a repository-owned bare key after the managed sentinel (like `panic`).

Vaiz: we need to include this into default anvil setup

Vaiz (approved):

wukchung (approved):

Cargo.toml
martin-kolinek: The lint is part of the nursery group so it was extraneous.

PR #770, opened 2026-09-18
feat(fakeable): add fakeable macros

martin-kolinek:
🤖 ## Summary

- import the `fakeable` and `fakeable_impl` procedural macro crates at version 0.2.0
- add a private end-to-end fixture covering real, manual fake, generic, async, and Mockall behavior
- preserve generic implementation arguments and async method constraints in generated code
- reject Mockall method shapes that cannot match the generated wrapper API
- document the generated wrapper design and explicitly mark the crate and Mockall integration as experimental
- integrate workspace dependencies, generated READMEs, packaging metadata, and spelling

## Validation

- `just anvil-clippy`
- `just anvil-fmt --fix`
- `just anvil-readme --fix`
- `just anvil-spellcheck`
- `just anvil-mutants-diff`
- `cargo +nightly-2026-05-30 udeps --package fakeable_impl --all-targets --all-features`
- `cargo +nightly-2026-05-30 llvm-cov --package fakeable_impl --all-features --tests --summary-only --fail-under-lines 100`
- `cargo +nightly-2026-05-30 test --no-run --package fakeable_test --no-default-features --locked`


crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,747 @@
...
+        Err(err) => return err.into_compile_error(),
+    };
+
+    // Generate internal type names without prefix, to be placed in dedicated module
+    let enum_name = syn::Ident::new("Enum", proc_macro2::Span::call_site());
+    let helper_module_name = quote::format_ident!("__fakeable__{struct_name}");
+    let fakes_attribute = &args.fakes_feature;
+
+    let mut real_impl = item_impl.clone();
+    *real_impl.self_ty = parse_quote!(#helper_module_name::#struct_name);
```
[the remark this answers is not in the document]
martin-kolinek: 🤖 Fixed in 53d59356. The hidden real implementation now preserves the original self-type path segment, including generic arguments, and the end-to-end fixture plus snapshot coverage verifies `impl<T> GenericService<T>`.

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1022 @@
...
+            }
+
+            let delegation_method = generate_delegation_method(method, fakes_attribute, enum_name, struct_name, helper_module_name)?;
+            delegation_methods.push(syn::ImplItem::Fn(delegation_method));
+        }
+    }
+
+    let mut fake_impl = original_impl.clone();
+    fake_impl.items = delegation_methods;
+    Ok(quote! { #fake_impl })
```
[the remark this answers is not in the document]
martin-kolinek: 🤖 Fixed in 72519bee. Wrapper trait impls now preserve non-function associated items, with snapshot coverage for an associated type and const; inherent associated items remain hidden with the real implementation.

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1027 @@
...
+        method_name,
+        param_names,
+        is_async,
+    );
+
+    quote! {
+        match #receiver .inner {
+            #helper_module_name::#enum_name::Real(real) => #real_call,
+            #fakes_cfg
+            #helper_module_name::#enum_name::Fake(fake) => #fake_call,
```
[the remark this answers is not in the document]
martin-kolinek: 🤖 Fixed in 4516aa2b. Typed receivers now produce a targeted diagnostic listing the supported `self`, `&self`, and `&mut self` forms, with contract, coverage, and mutation tests.

crates/fakeable/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+/// #[fakeable::fakeable(
+///     fake_impl = mocks::MockMyService,
+/// )]
+/// pub struct MyService {
+///     // fields...
+/// }
+///
+/// #[fakeable::fakeable(generate_mockall_fake = true, mockall_fake_module = "mocks")]
+/// impl MyService {
+///     // methods...
```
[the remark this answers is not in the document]
martin-kolinek: 🤖 Fixed in 77484fd9. `mockall` is now a doctest development dependency, ignored only by the unused-dependency check, and the docs explicitly require consumers to declare their own direct `mockall` dependency. Both all-feature and no-default-feature doctests pass.

crates/fakeable/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+///
+///     pub async fn save_user(&self, name: &str) -> Result<u64, String> {
+///         // Real implementation would save to database
+///         Ok(42)
+///     }
+/// }
+///
+/// // Manual fake implementation
+/// #[cfg(any(feature = "test-util", test))]
+/// pub mod fakes {
```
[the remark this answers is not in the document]
martin-kolinek: 🤖 Fixed in 77484fd9. The compiling doctest now covers the production path, while fake construction is documented separately as consumer-side usage that requires the consuming crate to declare and enable `test-util`.

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1034 @@
...
+        {
+            return syn::Error::new_spanned(item_impl, "generate_mockall_fake requires the fakeable `mockall` feature")
+                .into_compile_error();
+        }
+
+        #[cfg(feature = "mockall")]
+        {
+            let fake_name = struct_name.to_string();
+            let fake_module = args.mockall_fake_module.as_deref().unwrap_or("fakes");
+            match generate_mockall_fake(item_impl, &fake_name, fakes_attribute, fake_module) {
```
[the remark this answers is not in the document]
martin-kolinek: 🤖 Fixed in be797b38. Mockall generation now rejects generic impl blocks with a targeted diagnostic and directs users to manual fakes; generic manual-fake implementations remain supported and tested.

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1042 @@
...
+    let fake_ident = quote::format_ident!("{}", fake_name);
+
+    // Extract methods from the impl block
+    let mut mock_methods = Vec::new();
+
+    for item in &item_impl.items {
+        if let syn::ImplItem::Fn(method) = item {
+            // Skip constructors (methods that don't take self and return Self)
+            let has_self = method.sig.inputs.iter().any(|input| matches!(input, syn::FnArg::Receiver(_)));
+            let is_delegated = has_self && matches!(method.vis, syn::Visibility::Public(_) | syn::Visibility::Restricted(_));
```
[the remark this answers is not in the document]
martin-kolinek: 🤖 Fixed in d5ecfb4c. `generate_mockall_fake` now rejects trait impl blocks with a targeted diagnostic and directs users to a manual fake, avoiding an incomplete generated mock API.

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1042 @@
...
+            // Real struct (copy of the original struct but with a different name)
+            #real_struct
+
+            // Compile-time size guard for the fake implementation.
+            #size_assert
+        }
+
+        // Wrapper struct
+        #(#transformed_struct_attrs)*
+        #struct_vis struct #struct_name #struct_generics {
```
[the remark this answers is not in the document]
martin-kolinek: 🤖 Fixed in d5ecfb4c. Direct `#[cfg(...)]` attributes now gate the helper module, wrapper, and generated constructor impl together. A `cfg_attr` that conditionally applies `cfg` is rejected with a targeted diagnostic, with nested and malformed metadata coverage.

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1109 @@
...
+    quote! {
+        // Helper module containing internal types
+        #(#cfg_attrs)*
+        #[allow(non_snake_case)]
+        mod #helper_module_name {
+            use super::*;
+
+            // Internal enum - needs derive attributes for wrapper struct to work
+            #(#derive_attrs)*
+            pub(super) enum #enum_name #struct_generics {
```
martin-kolinek: 🤖: Bounded generic structs expand to invalid Rust because declaration generics are reused in type positions. For `Service<T: Clone> where T: Send`, the generated references include `Service<T: Clone>` and `Enum<T: Clone>`, and the wrapper `impl` does not place the `where` clause correctly; the existing unbounded `GenericService<T>` fixture does not exercise either failure.

**Please split the generics with `split_for_impl()` and use declaration, type, impl, and `where` forms in their respective positions, then cover a struct with both inline and `where` bounds.**

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1109 @@
...
+
+                let delegation_method = generate_delegation_method(method, fakes_attribute, enum_name, struct_name, helper_module_name)?;
+                delegation_methods.push(syn::ImplItem::Fn(delegation_method));
+            }
+            _ if is_trait_impl => delegation_methods.push(item.clone()),
+            _ => {}
+        }
+    }
+
+    let mut fake_impl = original_impl.clone();
```
martin-kolinek, line 351: 🤖: Copying an `unsafe impl` onto the wrapper makes its safety promise apply to an unconstrained fake variant. For example, a sound `unsafe impl Send for Service` on the real representation also marks `Service::fake(...)` as `Send` even when the fake contains `Rc`; safe code can then move that value across threads and trigger undefined behavior through `Rc`'s non-atomic reference count.

**Reject unsafe impl blocks unless the macro can establish the unsafe trait's invariant for the fake representation as well.**

crates/fakeable_impl/src/lib.rs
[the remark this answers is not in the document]
martin-kolinek: 🤖: The typed-receiver rejection still leaves other accepted receiver and `Self` forms misgenerated. `fn consume(mut self)` becomes `match mut self.inner`, `fn merge(self, other: Self) -> Self` passes the wrapper where each branch expects its own concrete type, and `fn maybe(&self) -> Option<Self>` declares `Option<Wrapper>` while returning `Option<HiddenReal>` or `Option<Fake>`.

**Please derive inner access from the receiver kind and either translate every input/output occurrence of `Self` across the wrapper boundary or reject signatures the macro cannot translate.**

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1109 @@
...
+    );
+    let fake_call = generate_method_call(
+        &proc_macro2::Ident::new("fake", proc_macro2::Span::call_site()),
+        method_name,
+        param_names,
+        is_async,
+    );
+
+    quote! {
+        match #receiver .inner {
```
martin-kolinek, line 531: 🤖: Trait wrappers can execute an inherent method instead of the trait method they wrap. The real branch calls `real.method(...)` using ordinary method lookup, so if the hidden real type has an inherent method with the same name, `Trait::method(&wrapper)` returns the inherent method's behavior rather than the annotated trait implementation's behavior. Receiver-less trait functions have the same ambiguity.

**Use fully qualified trait calls for the real branch, including receiver-less trait functions, so the wrapper delegates to the impl it is replacing.**

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1109 @@
...
+    let mut sig = sig.clone();
+
+    // Check if the signature already has explicit lifetime parameters
+    let has_explicit_lifetimes = sig
+        .generics
+        .params
+        .iter()
+        .any(|param| matches!(param, syn::GenericParam::Lifetime(_)));
+
+    // If there are already explicit lifetimes, return as-is
```
martin-kolinek: 🤖: Non-blocking — Mockall generation leaves nested elided lifetimes in signatures that this macro accepts, causing the emitted `mock!` block to fail compilation. `fn mixed<'a>(&self, direct: &'a str, nested: Option<&str>)` bypasses normalization because any explicit lifetime returns early, while `fn nested(&self, value: &Option<&str>)` is missed because the detector does not recurse through the outer reference. Mockall 0.14 rejects these remaining nested elisions.

**Normalize each nested elided reference with a fresh non-colliding lifetime, including beneath direct references and alongside existing lifetime parameters, or reject these signatures before emitting the mock.**

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1109 @@
...
+    } else {
+        quote! {}
+    };
+
+    quote! {
+        #real_impl
+
+        #wrapper_impl
+
+        #mockall_fake
```
martin-kolinek, line 263: 🤖: A conditional impl can leave its generated Mockall module enabled after the impl itself is removed. `real_impl` and `wrapper_impl` retain `#[cfg(...)]` because they clone the source impl, but `mockall_fake` is emitted without that cfg; if a copied method signature uses a type available only under the condition, the opposite configuration fails to compile in the unconditional mock module.

**Gate the Mockall emission with the impl's direct `cfg` attributes, and reject cfg-disabling `cfg_attr` forms as the struct path does, so all generated items are enabled as a unit.**

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1109 @@
...
+
+    quote! {
+        // Helper module containing internal types
+        #(#cfg_attrs)*
+        #[allow(non_snake_case)]
+        mod #helper_module_name {
+            use super::*;
+
+            // Internal enum - needs derive attributes for wrapper struct to work
+            #(#derive_attrs)*
```
martin-kolinek, line 147: 🤖: Non-blocking — Copying every derive onto both the enum and wrapper creates surprising extra requirements and can reject otherwise valid services. A `#[derive(Clone)]` service requires the fake to be `Clone` (the primary documentation's `FakeUserService` is not), while `#[derive(Default)]` fails because the generated enum has no default variant; shape-sensitive derives such as `Debug` also describe wrapper internals rather than the original struct.

**Please make derive propagation an explicit supported contract: forward only behavior that can be preserved or reject unsupported derives clearly, and document the bounds imposed on fake types. At minimum, fix the primary example and list these limitations.**

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1109 @@
...
+    let fake_impl_path = match args.fake_impl.as_ref() {
+        Some(path) => path.clone(),
+        None => {
+            return syn::Error::new_spanned(struct_name, "fake_impl must be specified for struct definitions").into_compile_error();
+        }
+    };
+
+    // Extract the fake constructor name (default to "fake" if not provided)
+    let fake_constructor_name = args.fake_constructor.as_deref().unwrap_or("fake").to_string();
+    let fake_constructor_ident = quote::format_ident!("{}", fake_constructor_name);
```
martin-kolinek, line 83: 🤖: Non-blocking — Invalid string-valued names make the procedural macro panic instead of returning a source diagnostic. Values such as `fake_constructor = "create fake"` or `mockall_fake_module = "my-mocks"` reach `format_ident!`, which panics without pointing at the offending literal, unlike the targeted `syn::Error`s used for other unsupported input.

**Validate both values as Rust identifiers while parsing the arguments and return a `syn::Error` spanned on the literal when validation fails.**

crates/fakeable/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+///
+/// # Optional mockall Mock Generation
+///
+/// This integration is experimental and should be used carefully. Generated
+/// Mockall types are not a stable compatibility boundary: upgrading either
+/// `fakeable` or `mockall` can change the generated API or behavior.
+///
+/// By specifying `generate_mockall_fake = true` in the attribute for the impl block, this macro
+/// will generate a mock implementation using the `mockall` crate. The generated mock will be placed
+/// in the specified module (default: "mocks"). This option requires the `fakeable` `mockall` Cargo
```
martin-kolinek: 🤖: Non-blocking — The documented default Mockall module name is wrong: omitting `mockall_fake_module` generates `fakes`, not `mocks`, so code following this paragraph will look for the generated type in the wrong module.

**Please change the documented default to `fakes`.**

crates/fakeable_impl/src/lib.rs
martin-kolinek: 🤖 Fixed in 569683ca. Nested elided references are now found through outer references and normalized even when explicit lifetimes already exist, using a fresh non-colliding lifetime inserted before type parameters.

crates/fakeable_impl/src/lib.rs
martin-kolinek: 🤖 Fixed in 569683ca. Generated Mockall output now carries the impl block direct `cfg` attributes, and cfg-disabling `cfg_attr` forms are rejected consistently with struct expansion; snapshot coverage verifies the gating.

crates/fakeable_impl/src/lib.rs
martin-kolinek: 🤖 Fixed in 569683ca. Real trait branches and trait constructors now use fully qualified calls. A compiled fixture with colliding inherent and trait method names verifies the trait implementation is selected for real and fake wrappers.

crates/fakeable/src/lib.rs
martin-kolinek: 🤖 Fixed in 569683ca. The documented default Mockall module is now `fakes`, matching the generator.

crates/fakeable_impl/src/lib.rs
martin-kolinek: 🤖 Fixed in 569683ca. Struct expansion now uses `split_for_impl()` forms in type and impl positions and emits the where clause on declarations; the compiled generic fixture has both inline and where-clause bounds.

crates/fakeable_impl/src/lib.rs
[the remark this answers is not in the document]
martin-kolinek: 🤖 Fixed in 569683ca. `mut self` is now rejected, bare `Self` parameters and nested `Self` returns are rejected, and direct `Self` returns remain translated and covered for sync/async methods.

crates/fakeable_impl/src/lib.rs
martin-kolinek: 🤖 Addressed in 569683ca. The primary manual-fake example now derives `Clone`, and the documented experimental limitations explicitly state that derives are copied to the enum/wrapper, impose bounds on the fake, and may be unsuitable for shape/default-sensitive derives.

crates/fakeable_impl/src/lib.rs
martin-kolinek: 🤖 Fixed in 569683ca. `fake_constructor` and non-dot `mockall_fake_module` values are validated as Rust identifiers during parsing and return literal-spanned diagnostics instead of reaching `format_ident!`.

crates/fakeable_impl/src/lib.rs
martin-kolinek: 🤖 Fixed in 569683ca. Unsafe impl blocks are rejected before wrapper generation because their invariants cannot be established for an arbitrary fake representation; the diagnostic is contract-tested.

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1283 @@
...
+    }
+
+    if receiver.is_some_and(|receiver| receiver.mutability.is_some() && matches!(receiver.kind, syn::ReceiverKind::Value)) {
+        return Err(syn::Error::new_spanned(
+            method_sig,
+            "mut self receivers are not supported; use self, &self, or &mut self",
+        ));
+    }
+
+    for input in &method_sig.inputs {
```
martin-kolinek, line 385: 🤖: The `Self` validation still misses method generic bounds and `where` predicates. For `fn consume<T: Marker<Self>>(&self, value: T)`, the wrapper requires `T: Marker<Wrapper>` but delegates to the hidden method, which requires `T: Marker<HiddenReal>`; the accepted expansion therefore cannot type-check.

**Inspect the complete method signature and either translate or reject bare `Self` in generic bounds and `where` predicates as well.**

crates/fakeable_impl/tests/output_verification.rs
```diff
@@ -0,0 +1,614 @@
...
+    let input = quote! {
+        impl<T> MyService<T>
+        where
+            T: Clone,
+        {
+            pub fn new(value: T) -> Self {
+                Self { value }
+            }
+
+            #[test]
```
martin-kolinek, line 377: 🤖: Non-blocking — The bounded-generic test is nested inside another test's `quote!` input, so it never runs and its code was accidentally blessed as generated output in the surrounding snapshot. The compiled fixture still exercises bounded generics, but this snapshot no longer tests what its name claims.

**Move `fakeable_on_bounded_generic_struct_preserves_generic_forms` to module scope with its own snapshot, and remove the embedded function from the generic-impl snapshot.**

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1283 @@
...
+        } else {
+            false
+        }
+    });
+
+    if !needs_lifetime {
+        return sig;
+    }
+
+    let mut existing_lifetimes = std::collections::HashSet::new();
```
martin-kolinek, line 712: 🤖: Non-blocking — The synthesized Mockall lifetime ignores higher-ranked lifetimes bound inside parameter types. For `fn f(&self, cb: for<'mock> fn(Option<&str>))`, this pass chooses a new method-level `'mock` and then rewrites the nested reference beneath the existing `for<'mock>` binder, producing illegal lifetime shadowing instead of a usable mock declaration.

**Make lifetime-name collection and rewriting binder-aware: choose a name fresh across nested binders and do not rewrite references across a higher-ranked binding boundary.**

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1283 @@
...
+    is_async: bool,
+    fakes_cfg: &proc_macro2::TokenStream,
+    receiver: &syn::Receiver,
+    trait_path: Option<&syn::Path>,
+) -> proc_macro2::TokenStream {
+    let real = proc_macro2::Ident::new("real", proc_macro2::Span::call_site());
+    let await_suffix = is_async.then(|| quote! { .await });
+    let real_call = if let Some(trait_path) = trait_path {
+        quote! {
+            <#helper_module_name::#real_struct_name as #trait_path>::#method_name(
```
martin-kolinek, line 515: 🤖: Generic trait impls now generate fully qualified calls against the hidden type without its generic arguments. For `impl<T> Trait for Service<T>`, the wrapper branch emits `<__fakeable__Service::Service as Trait>::method(...)`, but the hidden `Service` requires `<T>`, so an otherwise supported generic trait impl fails with E0107.

**Pass the full extracted self-type segment, including its generic arguments, into the constructor, ordinary-method, and `Self`-returning trait call generators.**

docs/design/fakeable.md
```diff
@@ -0,0 +1,92 @@
...
+use identifier patterns. `mut self`, parameters containing bare `Self`, and
+nested `Self` return types are rejected because their values cannot be
+translated across the wrapper boundary.
+
+Trait delegation uses fully qualified calls on the hidden real type so an
+inherent method with the same name cannot intercept a trait method call.
+Unsafe impl blocks are rejected because the macro cannot prove the unsafe
+trait's invariant for an arbitrary fake representation.
+
+Direct `#[cfg(...)]` attributes on a struct gate the helper module, wrapper, and
```
martin-kolinek, line 50: 🤖: Non-blocking — The documented `cfg` contract still says it applies only to structs, although this push now propagates direct `#[cfg(...)]` and rejects cfg-applying `cfg_attr` on impl blocks too. Readers therefore cannot discover the new impl-level behavior from either this design page or the mirrored limitation in `crates/fakeable/src/lib.rs`.

**Update both limitation statements to cover structs and impl blocks.**

crates/fakeable_impl/tests/output_verification.rs
martin-kolinek: 🤖 Fixed in 91581251. The bounded-generic test is now a module-level test with its own snapshot; the accidental embedded function was removed from the generic-impl input.

crates/fakeable_impl/src/lib.rs
martin-kolinek: 🤖 Fixed in 91581251. Mockall generation now detects a higher-ranked lifetime binder combined with nested elision and rejects the signature before rewriting across that boundary; the test now runs at module scope.

crates/fakeable_impl/src/lib.rs
martin-kolinek: 🤖 Fixed in 91581251. Bare `Self` is now detected and rejected across method generic bounds and where predicates, with separate contract tests for both forms.

docs/design/fakeable.md
martin-kolinek: 🤖 Fixed in 91581251. Both the public limitations and design document now describe direct cfg propagation and cfg-disabling cfg_attr rejection for structs and impl blocks.

crates/fakeable_impl/src/lib.rs
martin-kolinek: 🤖 Fixed in 91581251. The full self-type path segment, including generic arguments, now flows into constructor, ordinary, and `Self`-returning trait-qualified real calls. A generic trait impl snapshot covers the result.

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1334 @@
...
+
+            if is_delegated && matches!(method.vis, syn::Visibility::Restricted(_)) {
+                return Err(syn::Error::new_spanned(
+                    &method.vis,
+                    "generate_mockall_fake does not support restricted method visibility; use `pub` or a manual fake implementation",
+                ));
+            }
+
+            if is_delegated {
+                if method.sig.inputs.iter().any(|input| {
```
martin-kolinek: 🤖: Non-blocking — The new higher-ranked-lifetime guard still allows lifetime shadowing when the nested elision and binder are in different parameters. For `fn f(&self, value: Option<&str>, callback: for<'mock> fn(&'mock str))`, no single parameter satisfies both checks, so `add_explicit_lifetimes` introduces a method-level `'mock` that conflicts with the callback's binder and the expansion fails with E0496.

**Make the check signature-wide, or include nested binder names when selecting the synthesized lifetime.** A cross-parameter test should cover this case.

This follows up on the earlier higher-ranked-lifetime contract thread at `lib.rs:733`; the new guard handles the original same-parameter example but misses this split form.

crates/fakeable/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+/// - Mockall generation rejects mutable methods and methods with restricted visibility because
+///   it cannot generate a fake that matches the wrapper's delegated method set.
+/// - Mockall generation rejects generic impl blocks; use a manual fake for generic services.
+/// - Mockall generation rejects trait impl blocks; use a manual fake for trait implementations.
+/// - Unsafe impl blocks are rejected because the macro cannot establish their safety invariants for
+///   the fake representation.
+/// - Receiver-less methods must return `Self`; other associated functions cannot select a real or
+///   fake implementation to delegate to.
+/// - `mut self`, `Self` parameters, and nested `Self` return types such as `Option<Self>` are
+///   rejected because they cannot be translated across the wrapper boundary.
```
martin-kolinek: 🤖: Non-blocking — The limitations still present `Self` restrictions as applying only to parameters and return types, but this push also rejects `Self` in method generic bounds and `where` predicates. A consumer following the documented contract can therefore write `fn consume<T: Marker<Self>>(...)` and encounter an undocumented hard error.

**Add generic bounds and `where` predicates to this limitation and the matching section in `docs/design/fakeable.md`.**

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1334 @@
...
+
+struct BareSelfVisitor {
+    found: bool,
+}
+
+impl<'ast> syn::visit::Visit<'ast> for BareSelfVisitor {
+    fn visit_type_path(&mut self, i: &'ast syn::TypePath) {
+        if i.qself.is_none() && i.path.segments.len() == 1 && i.path.segments[0].ident == "Self" {
+            self.found = true;
+            return;
```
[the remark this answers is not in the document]
martin-kolinek: 🤖 Fixed in 4860fc67. The Self visitor now rejects paths beginning with `Self`, including projected forms such as `Self::Assoc`, across parameters, outputs, and generic predicates.

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1334 @@
...
+    let fake_ident = quote::format_ident!("{}", fake_name);
+
+    // Extract methods from the impl block
+    let mut mock_methods = Vec::new();
+
+    for item in &item_impl.items {
+        if let syn::ImplItem::Fn(method) = item {
+            // Skip constructors (methods that don't take self and return Self)
+            let has_self = method.sig.inputs.iter().any(|input| matches!(input, syn::FnArg::Receiver(_)));
+            let is_delegated = has_self && matches!(method.vis, syn::Visibility::Public(_) | syn::Visibility::Restricted(_));
```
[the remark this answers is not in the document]
martin-kolinek: 🤖 Fixed in 4860fc67. Generated Mockall fakes now reject consuming `self` receivers and direct users to manual fakes, avoiding incompatible Arc-backed mock storage.

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1334 @@
...
+
+    // Parse the fakes attribute
+    let fakes_attribute = &args.fakes_feature;
+    let fakes_cfg = fakes_cfg_attr(fakes_attribute);
+
+    // Extract derive attributes to apply to internal types
+    let (derive_attrs, _other_attrs): (Vec<_>, Vec<_>) = struct_attrs
+        .iter()
+        .cloned()
+        .partition(|attr| attr.path().get_ident().is_some_and(|ident| ident == "derive"));
```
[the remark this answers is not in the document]
martin-kolinek: 🤖 Fixed in 4860fc67. A `cfg_attr` that conditionally applies `derive` is now rejected with a targeted diagnostic; direct derives remain the supported contract and are copied consistently.

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1334 @@
...
+            #real_struct
+
+            // Compile-time size guard for the fake implementation.
+            #size_assert
+        }
+
+        // Wrapper struct
+        #(#transformed_struct_attrs)*
+        #struct_vis struct #struct_name #struct_generics #where_clause {
+            inner: #helper_module_name::#enum_name #type_generics,
```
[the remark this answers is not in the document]
martin-kolinek: 🤖 Fixed in 4860fc67. Struct layout `repr` attributes are rejected before expansion because the wrapper cannot preserve the original field layout or ABI; the restriction is documented and tested.

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1334 @@
...
+
+/// Helper function to generate a method call with optional await
+fn generate_method_call(
+    target: &proc_macro2::Ident,
+    method_name: &proc_macro2::Ident,
+    param_names: &[proc_macro2::Ident],
+    is_async: bool,
+) -> proc_macro2::TokenStream {
+    let await_suffix = is_async.then(|| quote! { .await });
+    quote! { #target.#method_name(#(#param_names),*) #await_suffix }
```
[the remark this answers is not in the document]
martin-kolinek: 🤖 Fixed in 4860fc67. Delegated constructor and receiver calls now forward method type and const generic arguments through turbofish syntax; lifetime arguments remain inferred. Snapshot coverage includes non-inferable type/const methods.

crates/fakeable_impl/src/lib.rs
martin-kolinek: 🤖 Fixed in 4860fc67. Higher-ranked-lifetime detection is now signature-wide: nested elision in any parameter combined with a binder in any parameter is rejected before synthesis. A cross-parameter regression test covers the reported case.

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1334 @@
...
+                // Add explicit lifetimes to avoid mockall compilation errors
+                let sig_with_lifetimes = add_explicit_lifetimes(&method.sig);
+
+                // Convert async functions to use impl Future return type for better mockall compatibility
+                let method_signature = if method.sig.asyncness.is_some() {
+                    convert_async_to_impl_future(&sig_with_lifetimes)
+                } else {
+                    quote! { #sig_with_lifetimes }
+                };
+                mock_methods.push(quote! { pub #method_signature; });
```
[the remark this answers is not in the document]
martin-kolinek: 🤖 Fixed in 4860fc67. Const methods are now rejected for generated Mockall fakes with a targeted diagnostic; manual fakes remain available for const APIs.

crates/fakeable/src/lib.rs
martin-kolinek: 🤖 Fixed in 4860fc67. The public limitations and design documentation now include rejected `Self` usage in method generic bounds and where predicates.

crates/fakeable_impl/src/lib.rs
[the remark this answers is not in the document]
martin-kolinek: 🤖: Non-blocking — The new guard can still be bypassed with `#[cfg_attr(feature = "ffi", repr(C))]`. `cfg_attr_applies_attribute` already recursively detects conditional attributes, but `process_struct` checks only direct `repr` attributes; when the condition is true, the generated wrapper keeps `repr(C)` despite having a different field layout and ABI.

**Reject `cfg_attr` forms that apply `repr` before generating the wrapper, and add a conditional-layout test.**

geeknoid (approved):

geeknoid (commented):
Static review found three new code-generation contract defects and one systemic test-oracle gap. The highest-impact issue is namespace relocation: qualified impl targets and `self::`/`super::` fake paths can resolve in the wrong generated module. Parameter forwarding also accepts binding/configuration forms it cannot preserve, and Mockall's generated dependency path is capture-prone.

I also rechecked the PR's stated claims around bounded generics, fully qualified trait delegation, unsafe-impl rejection, direct cfg propagation, projected `Self`, and higher-ranked lifetime rejection; the current head implements those mechanisms. Existing discussions for `impl Trait` returns, generated arm-name collisions, conditional `repr`, Mockall method-level cfg propagation, and async `+ Send` were not duplicated.

Security found no attacker-controlled source-to-sensitive-sink path in this source-to-source proc macro. Performance found no actionable hot-path opportunity without a representative workload or measurement objective. Generated READMEs and the lockfile were treated as generated evidence, not authored review targets.

All five taxonomies were assessed statically (150 security, 162 correctness, 182 testing, 183 performance, 143 conformance). Nothing was built, tested, benchmarked, profiled, fuzzed, mutation-tested, covered, run under Miri, or posted.

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1433 @@
...
+    false
+}
+
+/// Extracts the final type path segment from an impl block.
+fn extract_struct_segment(item_impl: &ItemImpl) -> Result<syn::PathSegment, syn::Error> {
+    match &*item_impl.self_ty {
+        syn::Type::Path(type_path) => Ok(type_path
+            .path
+            .segments
+            .last()
```
geeknoid: **Relocating paths changes which module they resolve in** — *Correctness · Medium · High confidence*

This keeps only the final segment of an impl target. A valid impl written outside the declaration module, such as `impl services::MyService`, is therefore retargeted to `__fakeable__MyService::MyService` in the impl's module instead of `services::__fakeable__MyService::MyService`. The same scope shift affects the configured fake path: [`Fake(#fake_impl_path)`](https://github.com/microsoft/oxidizer/blob/4860fc671f375de8245a0e20bba4b48297aa27fa/crates/fakeable_impl/src/lib.rs#L175) is emitted inside the helper module, so `fake_impl = self::fakes::Fake` now resolves `self` relative to that generated module. Both accepted `syn::Path` forms produce unresolved generated code.

**Direction:** preserve and rewrite the full parent path for hidden impl targets, and bind/rebase fake paths at the attribute's original scope before using them in the helper module.

**Done when:** compile-pass consumer fixtures cover an out-of-module qualified impl plus `self::` and `super::` fake paths.

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1433 @@
...
+
+    for input in &sig.inputs {
+        match input {
+            syn::FnArg::Receiver(_) => {
+                has_self = true;
+            }
+            syn::FnArg::Typed(syn::PatType { pat, .. }) => {
+                if let syn::Pat::Ident(pat_ident) = &**pat {
+                    // Extract just the identifier name, ignoring mut keyword
+                    param_names.push(pat_ident.ident.clone());
```
geeknoid: **Flattening every `PatIdent` to its name loses binding and cfg semantics** — *Correctness · Medium · High confidence*

`Pat::Ident` includes `ref`/`ref mut` bindings, while `PatType` may carry conditional attributes. For `pub fn set(&self, ref value: String)`, the copied wrapper signature binds `value` as `&String`, but the generated call forwards it to the hidden method's `String` parameter. Likewise, a `#[cfg(...)] value` parameter can disappear from the copied signature while its name remains unconditionally present in the generated call. These are valid Rust signatures that the macro accepts and then turns into type or name-resolution errors.

**Direction:** validate the complete parameter pattern and attributes, rejecting unsupported by-reference or conditionally present parameters with targeted diagnostics, or preserve their semantics in the forwarding expression. Apply the same rule to conditionally present generic parameters.

**Done when:** compile-fail tests reject `ref`/`ref mut` and conditional signature components, or compile-pass tests prove forwarding works in every cfg mode.

crates/fakeable_impl/src/lib.rs
```diff
@@ -0,0 +1,1433 @@
...
+        }
+    }
+
+    let fakes_cfg = fakes_cfg_attr(fakes_attribute);
+
+    if fake_module == "." {
+        // Generate in current module
+        Ok(quote! {
+            #fakes_cfg
+            mockall::mock! {
```
geeknoid: **The generated Mockall dependency path is shadowable** — *Conformance · Medium · High confidence*

The expansion resolves `mockall::mock!` in the consumer's module. A consumer that already defines `mod mockall` shadows the extern-prelude dependency, so enabling this integration can fail even though the required direct dependency is present. This is the hygiene problem covered by [M-MACRO-HELPERS](https://microsoft.github.io/rust-guidelines/guidelines/macros/#M-MACRO-HELPERS): generated references to third-party items need a controlled path. The current-module branch and nested-module branch both use the same relative path.

**Direction:** emit an absolute dependency path such as `::mockall::mock!`, or route it through a hygienic helper path that cannot be captured by consumer names.

**Done when:** a consumer fixture that declares its own `mod mockall` still expands the generated mock successfully.

crates/fakeable_impl/tests/output_verification.rs
```diff
@@ -0,0 +1,797 @@
...
+            other_value: i32,
+        }
+    };
+
+    let args = quote! {
+        fake_impl = fakes::FakeMyService
+    };
+
+    let result = fakeable_impl::fakeable_impl(args, input);
+    let result_file = syn::parse_file(&result.to_string()).unwrap();
```
geeknoid: **Syntax snapshots cannot detect type-invalid macro expansions** — *Testing · Medium · High confidence*

These snapshot cases parse the generated text with `syn::parse_file`, which proves only that it is syntactically valid. Unresolved helper paths, a `ref` binding forwarded with the wrong type, conditional parameters referenced after removal, and shadowed dependency paths all remain parseable and can be blessed by these snapshots. The private end-to-end fixture covers several happy paths, but not the signature and namespace equivalence classes exercised only here.

**Direction:** retain snapshots for token-shape review, but add compile-pass/compile-fail consumer fixtures (for example, a `trybuild` matrix) for every supported/rejected path, pattern, cfg, receiver, generic, and return-type class.

**Done when:** removing a parent path segment or accepting the `ref`/conditional-parameter cases above makes a compile-contract test fail before any snapshot update.

